PR #15, opened 2018-12-10
Refine Pixel increment in file information

### LINKED ISSUE #11: [file info] pixel increment
The entry "Pixel increment" in the file info now adopts the unit defined in the header. To have a better UX, please transform it to user friendly units, like milliarcsec, arcsec, arcmin, or degree, depending on the value itself. If it is velocity, please use km/s.

duidae:
fix #11

### DIFF
--- a/FileInfoLoader.cc
+++ b/FileInfoLoader.cc
@@ -424,6 +424,9 @@ bool FileInfoLoader::fillHdf5ExtFileInfo(FileInfoExtended* extendedInfo, string&
         double crval2 = (getDoubleAttribute(val, attributes, "CRVAL2") ? val : 0.0);
         double cdelt1 = (getDoubleAttribute(val, attributes, "CDELT1") ? val : 0.0);
         double cdelt2 = (getDoubleAttribute(val, attributes, "CDELT2") ? val : 0.0);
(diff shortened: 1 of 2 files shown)


FileInfoLoader.cc
```diff
@@ -443,11 +446,13 @@ bool FileInfoLoader::fillHdf5ExtFileInfo(FileInfoExtended* extendedInfo, string&
         crDegStr = fmt::format("[{}, {}]", cr1, cr2);
         if (!(cdelt1 == 0.0 && cdelt2 == 0.0))
             axisInc = fmt::format("{}, {}", unitConversion(cdelt1, cunit1), unitConversion(cdelt2, cunit2));
+        if (!(bmaj == 0.0 && bmin == 0.0 && bpa == 0.0))
```
veggiesaurus: if either bmaj or bmin are zero, we don't really have beam info, surely?

kswang1029: none of them should be zero, otherwise a beam cannot be defined.

veggiesaurus: bpa can be zero, right? 

the current check only guards against all three of them being zero.

kswang1029: Oops, right. None of bmaj and bmin should be zero. bpa can be zero.

Issue #37, opened 2018-12-20
partial update of cube histogram should show current histogram


veggiesaurus:
Currently, the cube histogram is blank until `progress == 1`. The partial updates should include the current state of the histogram (once the max/min have been determined). That way, the user can get a rough estimate of the cube histogram as it is calculated.

pford:
Added this in branch pam/cubeHistogramUpdate.  Shall I proceed with pull request for version 1.0, or wait until after release?  It is really a minor change, added the Histogram submessage in each progress message.

veggiesaurus:
@pford Partial updates work nicely, thanks. You can merge it in for V1.0

pford:
Fixed in PR #43

PR #43, opened 2018-12-21
Update histogram bins in progress message

pford:
Noting here that the commit in this branch was released in version 1.0.


pford:
Fixes issue #37

Issue #50, opened 2019-01-02
[user feedback] wrong coordinate grid with casa simobserve/simanalyze generated images


kswang1029:
This issue is reported by a user who wanted to display a casa image generated by using CASA's simobserve/simanalyze tasks. The displayed coordinate grid is incorrect (left: CARTA v1.0, incorrect. right: casaViewer, correct) 

<img width="958" alt="carta_image" src="https://user-images.githubusercontent.com/20819712/50582226-be7b9280-0e9b-11e9-91ce-4dc12cbf791a.png">

By using CASA's exportfits to generate a FITS image and loading it with CARTA v1.0, the result is now consistent with the casaViewer.
![screen shot 2019-01-02 at 2 38 09 pm](https://user-images.githubusercontent.com/20819712/50582265-13b7a400-0e9c-11e9-9089-8e0d1522d0b6.png)

I tracked down the issue and found that in the file info and header, celestial frame definition and projection type are missing. Both do show up in its FITS format. Interestingly with CASA's imhead task, these entries do exist. **The missing entries will be filled by some AST defaults to render the coordinate grid and that's why the end result is not correct.** 

Test images are attached. 
[demo_images.zip](https://github.com/CARTAvis/nrao-carta-backend/files/2720034/demo_images.zip)
 
Update: confirmed with other images generated with simobserve/simanalyze. Same symptom.

kswang1029:
@pford Tested with the images provided from the user feedback. Now the coordinate grid is correctly generated. However, the beam size of casa image is not displayed correctly. FITS one is correct.

![screen shot 2019-01-14 at 10 52 19 am](https://user-images.githubusercontent.com/20819712/51094798-fe6f3d80-17ea-11e9-876e-b5a579d3fd64.png)
![screen shot 2019-01-14 at 10 52 31 am](https://user-images.githubusercontent.com/20819712/51094801-04fdb500-17eb-11e9-8279-66ead0a02d38.png)

The root cause may be the implicit unit for bmaj and bmin in the header. 
For casa image,
```BMAJ = 9.52158e-06
BMIN = 8.8016e-06
BPA = 85.5142```
and for FITS image,
```BMAJ = 0.000545547
BMIN = 0.000504294
BPA = 85.5142```

The frontend may also need to be modified to adopt the implicit unit (@veggiesaurus)

My suggestion is to have unified unit for beam definition (i.e. degree) regardless the image types.

kswang1029:
[...]

PR #56, opened 2019-01-11
Add equinox and radesys to ext file info header and computed entries

pford:
also convert rad to deg for reference coordinates

Issue #50


pford:
The failure looks like a Travis test script error (no $HOME/package/bin), but build and package stages worked and showed a $PACKAGE/bin directory.  I cannot restart the PR build; since it is the weekend at ASIAA, will merge this PR and if master build fails either back out the PR or sort it out next week.

Issue #62, opened 2019-02-01
Backend memory usage when loading image


pford:
When loading the image channel cache, the data is loaded into a casacore::Array, copied to a std::vector of the same size, then the Array is destroyed.  Instead, the data should be loaded incrementally and added to the std::vector so that it is not in memory twice.

PR #70, opened 2019-02-19
Avoid the copy of channelCache if it is not necessary

markccchiang:
The original code in `Frame.cc` line 813: `data = channelCache`, which copies the channelCache to a new `std::vector<float> data`. It makes the memory doubled and spends a significant time if channelCache is very large. It can be avoided if the copy is not necessary. 

Another part makes the memory doubled is in the function `Frame::setChannelCache(size_t channel, size_t stokes)`. It is the issue about how to convert `casacore::Array` to `std::vector` in a proper way. I will deal with it in another branch.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -812,6 +812,8 @@ bool Frame::fillRegionHistogramData(int regionId, CARTA::RegionHistogramData* hi
                     region->fillHistogram(newHistogram, data, configChannel, currStokes, configNumBins, minval, maxval);
                 } else { // requested channel (current or specified)
                     if (configChannel == currentChannel()) {  // use channel cache
+                        bool writeLock(false);
+                        tbb::queuing_rw_mutex::scoped_lock cacheLock(cacheMutex, writeLock);
                         region->getMinMax(minval, maxval, channelCache);
                         region->fillHistogram(newHistogram, channelCache, configChannel, currStokes, configNumBins, minval, maxval);
                     } else {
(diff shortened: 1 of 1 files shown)


veggiesaurus:
You've removed the read lock from this, doesn't this mean that another thread could update the channel cache while this thread is calculating the histogram? Wasn't this the reason for the additional copy to `data` in the first place?

markccchiang:
Thanks for the comment! Function `fillHistogram(...)` takes the channel cache as a constant variable, it would not change the cache. One concern is that multi-threads read a cache at the same time. I have added a cache lock back. Please correct me if I am wrong. I found copying a whole channel cache is quite expensive, which is no less than converting the casacore::Array to std::vector. It has to be handled before we taking care of the casacore::Array issue.

veggiesaurus:
Nothing wrong with multiple threads reading the cache simultaneously, but the main issue is that another thread might be writing to the cache if we're trying to handle many things at once. I agree that copying the channel cache is expensive, so this is a good optimisation.

If we keep trying to handle multiple animation frames simultaneously, this might end up being an issue down the line, because we will be filling histograms of one frame while the next frame is being read into a channel cache. We need to approach all of this with caution.

veggiesaurus:
@pford this PR avoids an additional `malloc` and `memcpy` (via the copy constructor of the vector) and reduces required memory significantly, but probably holds the mutex for a longer period of time (two manual iterations through the data, compared to one highly optimised memcpy), depending on how effective the memory allocation is. I think it's an improvement over the current situation, and once #62 is fixed, this should mean our memory requirements are roughly equal to the size of a single channel, rather than 2x or 3x.

Issue #71, opened 2019-02-20
loading a file without specifying an HDU fails


veggiesaurus:
The frontend has a secondary method of loading files, via URL parameters. In this case, the HDU field will be left blank. However, this results in the file not loading, because the `casacore::string::fromString` call has the `chk` flag set to `true`, resulting in an exception being thrown. 

The other instance of `fromString` being used (in `getFITShdu`) explicitly guards against empty strings. This should be done in `openFile` as well.

veggiesaurus:
@ajm-asiaa can you report your findings here?

ajm-asiaa:
I'm not sure this is working. When we open an image using the file browser, we see the following in the browser console log:
![screenshot 2019-02-27 at 09 32 33](https://user-images.githubusercontent.com/20802551/53470702-57461d00-3a9d-11e9-9c66-1c1d4e307d45.png)

When we try to open the same image using the URL, we get the following:
![screenshot 2019-02-27 at 14 07 55](https://user-images.githubusercontent.com/20802551/53470760-8a88ac00-3a9d-11e9-939e-f4b8af13e3a9.png)

The only difference between the two cases seems to be at the "hdu" line where a 0 is missing in the URL case. The error in the URL case, "fromString failure for string", may be related to this.

pford:
This worked with the HDF5 file I tested, but didn't try FITS.  I modified one of the ICD tests to open each image type with hdu="" and all worked with the new fix in pam/no_hdu branch.

pford:
Branches merged, ready to re-test in master branch.

ajm-asiaa:
I confirm that images can now be opened directly using the URL method, but the "hdu" value still does not appear in the web browser's console log.

This is the first part of the web browser console log when opening the image through the file browser:
> e {directory: "home/acdc/CARTA/Images", file: "aJ.fits", hdu: "0", fileId: 0, renderMode: 0}
> directory: "home/acdc/CARTA/Images"
> file: "aJ.fits"
> fileId: 0
> hdu: "0"
> renderMode: 0

This is the first part of the console log when opening the image with the URL method:
> e {directory: "home/acdc/CARTA/Images", file: "aJ.fits", hdu: "", fileId: 0, renderMode: 0}
> directory: "home/acdc/CARTA/Images"
> file: "aJ.fits"
> fileId: 0
> hdu: ""
> renderMode: 0

veggiesaurus:
[...]

PR #74, opened 2019-02-26
Handle missing hdu and naxis entries

pford:
In extended file info and open file (Issue #71)


veggiesaurus:
@ajm-asiaa can you report your findings here?

Issue #65, opened 2019-02-01
Backend should save cube min/max and histograms


pford:
Cube region and image region hold histograms separately.  Cube histogram should reuse or create channel histograms in image region, and save cube histogram in cube region.

pford:
Changed description, since per-channel histograms use per-channel min/max but cube histograms use cube min/max.  This changes the bin sizes and therefore the histograms.  Will save cube min/max and histogram for each channel/stokes for reuse.

PR #75, opened 2019-02-27
Use RO_LatticeIterator to load data cache while opening a new file or channel

markccchiang:
This PR is addressed to solve the issue #62. Combined with PR #70, the memory usage is expected to be no greater than the size of a channel + a tile size mapped on the channel. The original function to get the data `getSlice(...)` is replaced by a new function `getData(...)`, where `RO_LatticeIterator` and tile algorithms are applied on it. Note that the `getSlice(...)` method is not replaced for Cube-histogram calculations, because I found it is faster (nearly x2) to dump data using `getSlice(...)` than using the new function `getData(...)`. Users who choose to calculate the Cube-histogram are supposed to hope it is done in a short time without considering the usage of memory. I will write a testing code to test its performance in detail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ jobs:
         - echo "Building carta-backend" 
         - cd $TRAVIS_BUILD_DIR
         - git submodule init && git submodule update
-        - sed -i '' 's/target_link_libraries(carta_backend/target_link_libraries(carta_backend uv/g' CMakeLists.txt # fix to find uv
+#        - sed -i '' 's/target_link_libraries(carta_backend/target_link_libraries(carta_backend uv/g' CMakeLists.txt # fix to find uv
         - mkdir build && cd build
         - cmake -DCMAKE_CXX_FLAGS="-I /usr/local/opt/openssl/include" -DCMAKE_CXX_STANDARD_LIBRARIES="-L /usr/local/Cellar/fmt/5.1.0/lib -L /usr/local/Cellar/hdf5/1.10.4/lib -L /usr/local/lib -L /usr/local/opt/openssl/lib" ..
         - make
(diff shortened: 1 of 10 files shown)


veggiesaurus:
I'm not sure that creating a number of TBB tasks to do a simple `memcpy` will really get us any speed gains, since that's almost certainly not the bottleneck. It might be slower, due to the overhead of scheduling tasks. 
I've asked @markccchiang to do some profiling to see if the parallel method provides any benefit, and whether the 2x slowdown is applicable to FITS files as well as CASA files.

pford:
Slicing the Lattice for an entire cube in order to calculate a cube histogram should be removed, since the Session calculates the cube histogram one channel at a time.  I am refactoring the backend for issue #65 and will remove calculating the cube histogram from Frame. So this case does not need to be profiled.

markccchiang:
Below plot shows time spending to dump raw data with respect to image files (with 10k, 20k, and 40k Fits/CASA formats). Solid lines are for new function `getData(...)` and dash lines are for original function `getSlice(...). Each of the points is the average of three tests under the same condition.
time-vs-filesize

veggiesaurus:
@markccchiang are you clearing disk cache between each test? If you're oping a 6GB file on a 16GB machine, is it possible that the old function is running into memory issues, hence the non-linear behaviour? Could you test a few more image sizes?

markccchiang:
Thanks for the comments! I have re-done the measurement, used more image sizes and cleared the disk cache every time. Following plot shows the results: 
time-vs-filesize

veggiesaurus:
thanks @markccchiang. Can you post the data in table form as well? The graph is a bit cluttered in the bottom left.

markccchiang:
As shown in the table below. Note that each of the points on the plot is the average of three tests under the same condition.

pford:
@markccchiang The performance tests should be re-run with the latest changes using shared casacore::Arrays, then please make a recommendation whether you think this PR should be merged.  Resolving conflicts in this branch is not a good use of your time.

markccchiang:
This PR doesn't need to be merged since the performance of using shared casacore::Arrays is much better than using LatticeIterator. For the thread safety concern (using shared casacore::Array) I will try to do more tests on this issue. And this PR is just a backup if we encounter an unexpected problem in the future.

PR #83, opened 2019-03-04
automatically calculate min and max channels if they are < 0

### LINKED ISSUE #82: SET_REGION rejects messages with min or max channels < 0
As per the [ICD](https://docs.google.com/document/d/1V4KzJaV3zQYJoA_ims28iBTit_aixo_xYcAfPiq8hic/edit?ts=5c6124cc#heading=h.h8gjsouxzb4y), `SET_REGION` should accept regions where the min and max channels are < 0. In this case, it implies that the region spans the entire channel range, and the min/max channels can be calculated automatically. Will add a PR shortly to resolve this.

veggiesaurus:
fixes #82

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -565,6 +565,13 @@ bool Frame::setRegion(int regionId, std::string name, CARTA::RegionType type, in
     // Create or update Region
     // check channel bounds and stokes
     int nchan(nchannels());
+    if (minchan < 0) {
+        minchan = 0;
+    }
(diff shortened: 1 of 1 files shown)


pford:
Waiting for 1.0.1 patch to be released before merging any new pull requests, but this is good to go.

kswang1029:
@pford I am updating the user manual to v1.0.1 so now I am the bottleneck.... Will get it done asap.

PR #93, opened 2019-03-06
Refactor histogram for reuse

pford:
Issue #65 .  Separated min/max calculation from histogram calculation, useful for cube histogram.  Both are saved in RegionStats by channel and stokes; previously only saved current stokes and started over when stokes changed.  Created functions in Frame (calls Region calls RegionStats) to _get_ min/max (retrieve if stored) and _calc_ min/max (calculate and store), as well as to _get_ histogram (retrieve if stored) and _calc_ histogram (calculate and store).

This may speed up the histograms for the cube or animation (whichever is requested second) as the min/max for each channel are stored in the image region and can be retrieved.  When the final cube histograms are calculated, the functions setMinMax and setHistogram are used to store them for retrieval later.  Since the finalhistogram is the only one needed for later retrieval, the interim per-channel histograms are destroyed.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -848,10 +848,8 @@ bool Frame::fillRegionHistogramData(int regionId, CARTA::RegionHistogramData* hi
             // get histogram requirements for this index
             CARTA::SetHistogramRequirements_HistogramConfig config = region->getHistogramConfig(i);
             int configChannel(config.channel()), configNumBins(config.num_bins());
-            bool useCurrentChannel(false);
             if (configChannel == CURRENT_CHANNEL) {
                 configChannel = currentChannel();
-                useCurrentChannel = true;
             }
             auto newHistogram = histogramData->add_histograms();
             newHistogram->set_channel(configChannel);
--- a/Frame.cc
+++ b/Frame.cc
@@ -148,9 +148,7 @@ std::vector<float> Frame::getImageData(CARTA::ImageBounds& bounds, int mip, bool
 std::vector<float> Frame::getImageChanData(size_t chan) {
     // get data for given chan, current stokes
     std::vector<float> data;
-    casacore::Matrix<float> chanMatrix;
-    getChannelMatrix(chanMatrix, chan, currentStokes());
-    data = chanMatrix.tovector();
+    getChannelMatrix(data, chan, currentStokes());
     return data;
 }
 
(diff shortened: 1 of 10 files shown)


Frame.cc
```diff
@@ -484,14 +482,14 @@ void Frame::setChannelCache(size_t channel, size_t stokes) {
...
-void Frame::getChannelMatrix(casacore::Matrix<float>& chanMatrix, size_t channel, size_t stokes) {
+void Frame::getChannelMatrix(std::vector<float>& chanMatrix, size_t channel, size_t stokes) {
     // fill matrix for given channel and stokes
     casacore::Slicer section = getChannelMatrixSlicer(channel, stokes);
     casacore::Array<float> tmp;
     // slice image data
     std::unique_lock<std::mutex> guard(latticeMutex);
     loader->loadData(FileInfo::Data::XYZW).getSlice(tmp, section, true);
-    chanMatrix.reference(tmp);
+    chanMatrix = tmp.tovector();
```
confluence: This causes a new vector to be allocated in `.tovector`, which is then copied to `chanMatrix`. If you change this to `tmp.tovector(chanMatrix)`, that will reuse `chanMatrix` directly and avoid the copy.

Region/RegionStats.cc
```diff
@@ -36,51 +37,81 @@ CARTA::SetHistogramRequirements_HistogramConfig RegionStats::getHistogramConfig(
...
+        haveMinMax = true;
+    } catch (std::out_of_range) {
+        // not stored
+    }
+    return haveMinMax;
+}
+
+void RegionStats::setMinMax(int channel, int stokes, minmax_t minmaxVals) {
+    // Save min, max for given channel and stokes
+    if (channel == -2) { // all channels (cube); don't save intermediate channel min/max
```
confluence: It would be good to use the `ALL_CHANNELS` constant for this consistently and avoid hardcoding the numeric value -- perhaps the constants should be defined in a different file which can be imported in multiple places. But it's not essential to fix this in this PR.

Session.cc
```diff
@@ -744,33 +757,36 @@ void Session::sendCubeHistogramData(const CARTA::SetHistogramRequirements& messa
...
+                        sendFileEvent(fileId, "REGION_HISTOGRAM_DATA", requestId, histogramMessage);
 
+                        // get histogram bins for each channel and accumulate bin counts
+                        std::vector<int> cubeBins;
+                        CARTA::Histogram chanHistogram;  // histogram for each channel
+                        for (size_t chan=0; chan < nchan; ++chan) { 
+                            frames.at(fileId)->calcRegionHistogram(regionId, chan, stokes, numbins, cubemin,
+                                cubemax, chanHistogram);
                             // add channel bins to cube bins
                             std::vector<int> channelBins = {chanHistogram.bins().begin(), chanHistogram.bins().end()};
```
confluence: Is this faster than accessing `chanHistogram.bins()` directly below? We should perhaps profile both options to compare them.

confluence: I will look into profiling this; I think it's OK the way it is for now.

Frame.cc
pford: Done.

Session.cc
pford: Will look into this.

veggiesaurus: I've dug into the generated protobuf code for getting  and setting  elements of  a repeated field and it's basically inlining an array access, so I'd imagine using it directly should be the fastest method.

pford: Changed to using iterators for addition, tested old/new on 2 datasets [1920,1920,478] and [4224,1824,222].  Using transform() reduced the total time to calculate cube histogram bins by ~250 ms (total time of ~104 s and ~112 s).

std::transform(chanHistogram.bins().begin(), chanHistogram.bins().end(), cubeBins.begin(),
                                    cubeBins.begin(), std::plus<int>());

PR #97, opened 2019-03-07
Fix reading of fixed size null-terminated string attributes

### LINKED ISSUE #86: Null-terminated string attributes are not read correctly
Part of #45. As described in https://docs.google.com/document/d/1T1T85zvGHy7IsePwztN7SjrMPJ4hQJv8X5ihZKf5XA8/edit#heading=h.2fonicmwctib -- this has been verified to be triggered by null-terminated string attributes. We don't require a specific string format, and one of our converters produces null-terminated strings. The issue is possibly in line 60 of `HDF5Attributes.cc`.

confluence:
Fixes #86.

One of our FITS-to-HDF5 converters (the C implementation) writes string attributes as 256-character fields which are null-terminated (unlike the Python implementation, which writes them as zero-padded fields which are truncated to different sizes). Our schema does not specify a particular storage mechanism for fixed strings, so both these options should be legal and supported.

Overwriting the internal c-string buffer of a `casacore::String` object does not change the size property of the string, and the subsequent resize which uses the size obtained from the attribute doesn't help, because that's the fixed maximum size of the field. So for these null-terminated fields we currently produce 256-character strings with embedded nulls followed by (as far as I can make out) a repeated padding character (which does not trigger any protocol buffer errors). Firefox seems to treat the nulls in these characters as terminators, and displays the strings truncated, but Chromium/Chrome and Edge display the full strings including padding characters, which causes the long lines of unicode boxes documented in the bug.

We can calculate the correct length of the string (in both cases) by calling `strlen` on the modified c-string buffer, and resizing the string object to this length solves the problem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,5 @@ fabric.properties
 
 # End of https://www.gitignore.io/api/clion
 permissions.txt
+
+build
--- a/.gitignore
+++ b/.gitignore
@@ -65,5 +65,3 @@ fabric.properties
 
 # End of https://www.gitignore.io/api/clion
 permissions.txt
-
-build
--- a/ImageData/HDF5Attributes.cc
+++ b/ImageData/HDF5Attributes.cc
@@ -54,11 +54,20 @@ void HDF5Attributes::readScalar (hid_t attrId, hid_t dtid, const casacore::Strin
             }
             break;
         case H5T_STRING: {
-            casacore::String value;
-            value.resize(sz+1);
-            casacore::HDF5DataType dtype(value);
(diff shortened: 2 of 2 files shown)


veggiesaurus:
Just to be safe, perhaps replace `strlen` with `strnlen` with `n=256`, in case the string is 256 characters in length and doesn't get null terminated.

confluence:
The 256 is not significant. After some investigation I think that the current usage of `strlen` is safe: we always start off with a string resized to one more character than the reported size, and `resize()` fills the space with nulls. So there will always be a null at the end of the string.

PR #99, opened 2019-03-08
reuse vector memory when creating large temporary casacore::array containers.

veggiesaurus:
This PR changes the construction of three large temporary `casacore::array` containers to re-use memory already defined by a `std::vector`, thus solving the 2x memory usage issue and speeding up reading a bit by avoiding unnecessary memory allocations and copies.

Going forward, we should probably consider the following:
- Switching back to `casacore::Matrix` and `casacore::Vector` containers, as they allow us to specify a no-initialisation policy. This will require some careful checking of the thread safety issues that are apparently fixed
- Whether we can directly utilise memory in the protocol buffer objects. At the moment, we're allocating some memory, getting a profile, then copy to a protocol buffer object. Probably a minor optimisation, but might help in future for large allocations of a fixed size.

Those two are less urgent than the 2x memory issue though.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -476,11 +476,11 @@ void Frame::setChannelCache(size_t channel, size_t stokes) {
     if (channel != currentChannel() || stokes != currentStokes()) {
         bool writeLock(true);
         tbb::queuing_rw_mutex::scoped_lock cacheLock(cacheMutex, writeLock);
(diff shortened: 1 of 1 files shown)


veggiesaurus:
note: there are some places where I've left `.tovector` as is, because the vector size is small.

PR #101, opened 2019-03-09
optimisations to down-sampling implementation

veggiesaurus:
While implementing similar downsampling code in another project, I realised that we should be accessing the channel cache along the contiguous axis (x) in the inner loop, so that memory reads are more efficient. At the same time, I've switched from using a combination of `isnan` and `isinf` to simply using `std::isfinite` (introduced in c++11). 

These may seem like micro-optimisations, but I've tested and found a significant performance improvement when viewing large files. See below for two tests. Each point is the average of three test runs on the same file with the same zoom actions to trigger them. 

Cropping and downsampling 10K x 10K image by various factors
Cropping and downsampling 48K x 48K image by various factors

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -107,21 +107,23 @@ std::vector<float> Frame::getImageData(CARTA::ImageBounds& bounds, int mip, bool
     if (meanFilter) {
         // Perform down-sampling by calculating the mean for each MIPxMIP block
         auto range = tbb::blocked_range2d<size_t>(0, numRowsRegion, 0, rowLengthRegion);
-        auto loop = [&](const tbb::blocked_range2d<size_t> &r) {
-            for(size_t j = r.rows().begin(); j != r.rows().end(); ++j) {
-                for(size_t i = r.cols().begin(); i != r.cols().end(); ++i) {
+        auto loop = [&](const tbb::blocked_range2d<size_t>& r) {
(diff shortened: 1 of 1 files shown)


Frame.cc
```diff
@@ -107,21 +107,23 @@ std::vector<float> Frame::getImageData(CARTA::ImageBounds& bounds, int mip, bool
     if (meanFilter) {
         // Perform down-sampling by calculating the mean for each MIPxMIP block
         auto range = tbb::blocked_range2d<size_t>(0, numRowsRegion, 0, rowLengthRegion);
-        auto loop = [&](const tbb::blocked_range2d<size_t> &r) {
-            for(size_t j = r.rows().begin(); j != r.rows().end(); ++j) {
-                for(size_t i = r.cols().begin(); i != r.cols().end(); ++i) {
+        auto loop = [&](const tbb::blocked_range2d<size_t>& r) {
```
confluence: This looks like an odd spacing change -- deliberate, or typo?

veggiesaurus: intentional, since the `&` is put before the space in most cases in the rest of the file. However, we should probably employ a proper linting tool like `clang-tidy` to make sure we're consistent with a particular style.

confluence: I'm happy to defer all pedantic style issues until we can clean up everything at once and start using a commit hook.

Issue #80, opened 2019-03-01
Implement basic regions for single image


pford:
- basic ROI for single image including point, rectangle, ellipse, and polygon
- import and export ROI file
- ROI statistics
- ROI spectral profile
- ROI histogram

https://docs.google.com/document/d/1IdEh2x7lrsul4kSJ3lzGCiEUyI9gGTX7yVp8BtCKUyw

pford:
Opened PR 107 to merge regions into master, failed on package upload because no space left on device (prior to running tests).

pford:
This will be ready to test after branch _pam/region_updates_ completes build/test and PR.

pford:
Pull request merged, commit  2d44acc on master building now.  Also fixed rectangle pixels and SET_REGION for cursor region id in this PR.

kswang1029:
A quick test suggests that the major and minor axes of ellipse are swapped. Not sure if it is from  frontend or from backend. Below is the case: the bright pixel is 100 and the backend is all 1.
<img width="856" alt="Screen Shot 2019-03-28 at 11 52 34" src="https://user-images.githubusercontent.com/20819712/55128956-53ff8900-5150-11e9-8d9e-251224cc4973.png">
<img width="855" alt="Screen Shot 2019-03-28 at 11 52 19" src="https://user-images.githubusercontent.com/20819712/55128957-53ff8900-5150-11e9-9025-05d303f809f0.png">

veggiesaurus:
For a rotation of zero, the major axis is north-south, correct? @kswang1029 does this look right?
![image](https://user-images.githubusercontent.com/592504/55139311-1eea3b00-513e-11e9-9e87-e551485d09ab.png)

Can confirm that the backend is receiving and correctly parsing `bmaj=1` and `bmin=0.5` and passing them into `LCEllipsoid` constructor as major and minor axes. I think `LCEllipsoid` assumes a different axis orientation though. The comment above the constructor says:
```
// Construct a two dimensional ellipse with theta being the angle from
    // the x-axis to the major axis of the ellipse in radians.
    LCEllipsoid (
    	const Float xcenter, const Float ycenter,
    	const Float majorAxis, const Float minorAxis,
    	const Float theta, const IPosition& latticeShape
    );
```

kswang1029:
@veggiesaurus yes, major axis should be north-south in astronomy world...

pford:
Ellipse angle PR merged in commit d3f992f, ready to test after builds complete.

[...]

PR #103, opened 2019-03-14
Refactor Session, Frame

pford:
Moved code from Session to Frame to prepare for ROI (issue #80). Region parameters validated and stored in Region class. Next step will be to create lattice region(s) in Region from these parameters, to use in Frame to create a SubLattice from the Lattice for the entire image.

All tests pass in this branch.  Also updated version number to 1.1.

### DIFF
--- a/FileInfoLoader.cc
+++ b/FileInfoLoader.cc
@@ -79,14 +79,18 @@ CARTA::FileType FileInfoLoader::convertFileType(int ccImageType) {
 
 bool FileInfoLoader::getHduList(CARTA::FileInfo* fileInfo, const std::string& filename) {
     // fill FileInfo hdu list
-    bool hduOK(true);
+    bool hduOK(false);
     if (fileInfo->type()==CARTA::FileType::HDF5) {
         casacore::HDF5File hdfFile(filename);
         std::vector<casacore::String> hdus(casacore::HDF5Group::linkNames(hdfFile));
-        for (auto groupName : hdus) {
-            fileInfo->add_hdu_list(groupName);
+        if (hdus.empty()) {
+            fileInfo->add_hdu_list("");
+            hduOK = true;
+        } else {
+            for (auto groupName : hdus)
+                fileInfo->add_hdu_list(groupName);
+            hduOK = true;
         }
-        hduOK = (fileInfo->hdu_list_size() > 0);
     } else if (fileInfo->type()==CARTA::FITS) {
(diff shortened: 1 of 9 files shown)


veggiesaurus:
Just browsing through the code, surely there's nothing wrong with defining a region that has its center off the image, such as the first image below? Currently, your checks would pass with the region in the second image below, but fail with the region in the first image if I'm understanding the code correctly. 
image

image

Also not sure that we should prevent regions that have larger widths than the image itself (e.g. below). This is important if the user wants to see a spectral profile of the entire file. 
image

@kswang1029 thoughts?

kswang1029:
Ultimately with WCS group, images within the same group may have different FOV. So we do need the flexibility to define regions anywhere on the celestial sphere with any size.

pford:
Modified so that if the region is completely outside the image range (region xrange < 0 or > image xmax, or region yrange < 0 or > image ymax) the region is invalid and will not be processed.

pford:
There may be further refinements when I am able to test regions.  This is just a start.

veggiesaurus:
@pford after discussing with @kswang1029 in more detail, the consensus is that regions should only be considered invalid and be rejected by `SET_REGION` if they contain malformed or invalid configurations, such as
- negative width or height
- NaN values in any of the control points
- The incorrect number of control points for the required region type

Thus regions that are entirely outside of the image range should not be flagged as invalid

veggiesaurus:
@confluence and/or @markccchiang can you review this?

pford:
OK, will apply criteria to POINT region points as well (except width/height).  Does the channel min/max need to be in range?  Of course applying such a region to the Lattice will fail, so what does SET_REGION_ACK success=true mean exactly?

veggiesaurus:
I'd say for 1.1, yes, the channels should be in range.

 `success = true` means the region was defined correctly and assigned a region ID, and further changes to the region requirements and control points can be made. Any actual analysis product for that region would be empty / full of NaNs, but that's ok. 

As a reference, here's a region defined in DS9 that's entirely off the image: 
image
The resulting statistics blurb looks like this: 

```
center=-79 76
image

reg	sum	error		area		surf_bri		surf_err
				(pix**2)		(sum/pix**2)		(sum/pix**2)
---	---	-----		--------		------------		------------
1	0	0		0		-nan		-nan

reg	sum	npix	mean	median	min	max	var	stddev	rms	
---	---	----	----	------	---	---	---	------	---	
1	0	0	0	0	1.7976931e+308	-1.7976931e+308	0	0	0	

```

I'd imagine we'd want something similar, (although with NaNs in place of obviously incorrect values)

Issue #17, opened 2018-12-11
[image viewer] backend crashes when multiple sessions access same dataset


pford:
Image locking for disk access is within Session, not across Sessions

veggiesaurus:
@kswang1029 can you remember how to reproduce this?

kswang1029:
@veggiesaurus with the help from @ajm-asiaa for setting up the demo server, I did the following tests with one backend and two sessions from two different computers:
1) requesting a file list from the same folder ==> crash
2) requesting file info of the same file ==> appears fine
3) opening same big image ==> appears fine
4) playing animation at low and high fps ==> appears fine.

PR #106, opened 2019-03-20
Lock on the file list request handler

markccchiang:
This PR is for solving issue #17. Since the current lock is within Session, not cross Sessions. I extract the file list handler (`onFileListRequest`) from Session as an independent object, which handles all users' file list requests. Applying `tbb::mutex::scoped_lock` on file list handler so that it only process for one user a time. It could avoid the crash. There is no merge conflict with Pam's refactor PR.


pford:
Please review the includes in FileListHandler.[cc, h].  Many are not needed.

markccchiang:
Thanks for the comment! I have updated the branch including removed unused headers.

pford:
Please fix centos6 and ubuntu failures in new FileListHandler code then will merge this.

markccchiang:
Sorry for the late reply! I will fix it later today.

PR #108, opened 2019-03-25
updated to latest protocol buffer definitions (minor compatibility changes)

veggiesaurus:
have merged latest changes into master. Only minor adjustments required. 

protobuf commit here: https://github.com/CARTAvis/carta-protobuf/commit/cc044a61b3132697e7bc19c898044d4b3948ded1


pford:
OK for an immediate fix, preferable to remove channel range from region settings rather than set to -1.  Will change one of the region branches for this.

veggiesaurus:
Figured it's best to make this change minimal and leave the refactoring to you :)

PR #112, opened 2019-03-26
Enable loading 2D and 3D statistics from HDF5 files

confluence:
Existing code for loading statistics from a file has been refactored as follows:

* 3D statistics were added. Various bugs in the original 2D code were fixed.
* The objects used to store statistics were moved from frame to loader, together with the definition of the statistics struct.
* The function used to load statistics was moved from frame to loader and split into multiple functions for loading individual types of statistics datasets. The top-level functions which are independent of HDF5-specific code are in the top-level loader, so that this functionality can in theory easily be extended to other loaders if the appropriate format-specific functions are implemented (as was the case before).
* `ndims`, `nchannels` and `nstokes` properties were added to the loader -- they are frequently used in the statistics functions.
* These and other properties are now calculated in a loader function called by the frame constructor, not directly in the frame constructor.
* The frame now has an `nstokes` helper function and corresponding private property, analogous to `nchannels`. 
* Statistics datasets are no longer loaded with the `loadData` function. This was a poor fit, since many statistics datasets contain integer data and this function can currently only load floating point data. If it later becomes possible to load datasets of any type with this function, it may be possible to combine it with the more generic function for loading statistics. Currently `loadImage` always returns the main dataset, as is the case in other loaders. In future we can also use it to load the swizzled dataset.
* For a similar reason, statistics datasets could not be stored in the internal dataset map of the HDF5 loader. Since it's unnecessary to store the original arrays at all after the data has been loaded into the structs, I decided to eliminate this map entirely and store the main dataset in an `image` property, as is done in the other loaders. In future when we add the swizzled dataset we can store it in a separate property.
* Also for consistency with the other loaders, the main dataset is loaded in `openFile`, and no initial `loadData` call is needed in the frame constructor. We assume in other functions that this dataset exists. This, together with the removal of the map, makes it simpler to reuse its reference to the parent group when loading other datasets.
* The `hasData` function is still used for statistics datasets. It now checks the dimensions of the main dataset, and otherwise falls back to checking for the presence of other datasets in the file.
* To load a statistics dataset we construct an `HDF5DataSet` object directly, reusing the reference to the parent group from the main dataset, instead of opening the file again by creating an `HDF5Lattice`.
* Because the data type passed to the dataset constructor has to match the native datatype exactly, and we can't guarantee the exact type used by each file for the statistics, we autodetect the native type of the dataset and use it to construct the object. But because the HDF5 interface as well as the  `HDF5DataSet` object allow data to be read into a buffer of any type of the same class, we coerce all floating point datasets to `casacore::Float` and all integer datasets to `casacore::Int64` by creating casacore arrays of those types to read in the data from the dataset objects. This allows us to cast the arrays to a known type inside the statistics loading functions. Some of the specifics of this implementation could be simplified if it were possible to use the destination type when constructing the dataset object -- we may suggest a patch to casacore to make this possible.
* Neither `HDF5Lattice` nor `HDF5DataSet` can handle scalar datasets correctly -- the object can be created, but it's impossible to read data from it. As a temporary workaround we are using the HDF5 interface directly to read these datasets and wrapping the data in 1D arrays. This could possibly also be patched in casacore.
* Because of this issue, we cannot use the shape of the array to check the correctness of the stats dimensions for the given image dimensions (both scalar and 1D datasets become 1D arrays). This is why we have a separate function for retrieving the original shape.
* The lowest-level functions for reading the HDF5 data are templated to avoid a lot of code repetition, but we don't use templated functions directly from the stats loading functions, to make it possible for all the generic functions (and stubs of the data loading functions) to be defined in the parent loader (templated functions can't be virtual). It's possible to eliminate templating entirely from the implementation of the shape function by using the HDF5 interface directly and bypassing the typed dataset object, but it's unclear whether this would be superior to the current templated solution.


Frame.cc
```diff
@@ -91,253 +85,6 @@ bool Frame::checkStokesIndex(int stokes) {
     return ((stokes >= 0) && (stokes < nstokes));
```
pford: nstokes can use nstokes( ) instead of determining from stokesAxis

confluence:
I just noticed that the past few commits have failed the checks -- if I have to fix something anyway, I will fix the extra `nstokes` too.

pford:
Looks like they did not fail, they were canceled probably due to new commits on the same branch.

PR #115, opened 2019-03-27
Rwjs/reduce maps

rwjsimmonds:
Solves a number of problems in the current master branch. Most importantly:-
  Solves seg fault if a client disconnects unexpectedly while tasks are still trying to write.
  Prevents a single session from blocking all threads when set channels messages are queued.

This started out to eliminate the multiple maps that were holding different object types in main.cc. This was done by placing the objects associated with a session into the Session class. It has also eliminated most of the message queues that were not doing anything useful. Also the message types are now handled in a map to unsigned ints in the first step to eliminate the use of long strings for message types in the ICD. Also started to specialise the OnMessageTask types to eliminate multiple lookups for any message that arrives and make the execute functions easier to understand. Given that the code was crashing on unexpected client disconnects, simple reference counting has also been added to Session objects so that they don't get deleted until all tasks operating on them have completed.


rwjsimmonds:
Hi Pam:

The cancel is just moving the functionality from the cube histogram method to the main thread. If this is logically incorrect for other functionally I can easily remove this, but it seemed odd to have to create and schedule a task to  just set the value of one parameter. If it is a problem for now remove it and we can think about this more with the update to the ICD.

Rob

pford:
Re: histogram requirements, only the cube histogram can cancel in the popup window, sending a message with empty histograms[ ], and only the cube histogram implementation looks for this cancellation.  So we don't have to worry about the region_id and this should be safe.  Mark is working on cancelling tasks in general, e.g. when an image or region is closed.

Proceeding with merge since "pr" tests (merge branch into master) passed even though "push" (branch) tests failed.

PR #122, opened 2019-03-28
Fix ellipse angle

pford:
Fix for test feedback in Issue #80 .
Region rotation field is measured from north-south axis but LCEllipsoid theta is from east (x-axis)

### DIFF
--- a/Region/Region.cc
+++ b/Region/Region.cc
@@ -287,10 +287,10 @@ casacore::LCRegion* Region::makeEllipseRegion(const std::vector<CARTA::Point>& p
     if (points.size()==2) {
         float cx(points[0].x()), cy(points[0].y());
         float bmaj(points[1].x()), bmin(points[1].y());
-	// rotation is in degrees from y-axis
+        // rotation is in degrees from y-axis
         // ellipse rotation angle is in radians from x-axis
-	casacore::Quantity theta((rotation+90.0), "deg");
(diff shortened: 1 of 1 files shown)


pford:
circleci and push failures are from animator test (known issue), pr pending...

Region/Region.cc
```diff
@@ -286,7 +287,11 @@ casacore::LCRegion* Region::makeEllipseRegion(const std::vector<CARTA::Point>& p
     if (points.size()==2) {
         float cx(points[0].x()), cy(points[0].y());
         float bmaj(points[1].x()), bmin(points[1].y());
-        ellipse = new casacore::LCEllipsoid(cx, cy, bmaj, bmin, rotation, m_latticeShape.keepAxes(m_xyAxes));
+	// rotation is in degrees from y-axis
+        // ellipse rotation angle is in radians from x-axis
+	casacore::Quantity theta((rotation+90.0), "deg");
```
veggiesaurus: wouldn't it be easier to just do (rotation+90) * PI / 180.0f? (Similar to line 239)

pford: I was thinking ahead to when it could be deg or rad imported from a CRTF file (or is the frontend going to parse and convert the file settings to the ICD?).  Also thought it would normalize deg > 360 => rad > 2*pi.  But I will simplify for now.

PR #124, opened 2019-03-29
Fix compiler warnings

confluence:
This is a small set of fixes to eliminate compiler warnings in the portions of the code I recently worked on. The biggest change is in the Frame constructor -- now only loader calls are wrapped in try blocks. Statistics can fail to load without invalidating the whole frame. Because the stats are intertwined, I saw no reason to continue loading stats datasets if some of them failed -- the loading function will stop after the first failure. The stats struct now has a `valid` flag, so that we don't try to use the loaded stats if they only partially loaded.

I haven't made any fixes outside the loaders and the related frame code -- I don't want to cause unnecessary merge issues. Please fix your own sections as you go.


confluence:
I can see the test failure -- it probably has something to do with the frame constructor changes. I will try to fix it.

Issue #123, opened 2019-03-29
[region] pixel index in region definition


kswang1029:
Thanks @zarda to highlight this issue in the ICD test spec doc.

With CASA, if blc=[850, 1200, 0, 0] and trc=[1150, 1400, 0, 0], the npts = 60501 which means the width is 301 px and the height is 201.

With CARTA, if we define a region centered at (1000px, 1300px) and a width of 300 px and a height of 200 px (so blc=[850,1200], trc=[1150,1400]), statistics gives npts = 60000. 
If I set (width, height) = (301, 201), then I will get exact statistics (npts, mean, sum, etc) as from CASA.

I suspect the difference may be due to 
- WCS treatment in CASA
- how pixel is counted in a region
- index rounding

To simply the above case (which is actually an ICD test), see the following 
![Screen Shot 2019-03-29 at 11 49 25 AM](https://user-images.githubusercontent.com/20819712/55208308-c045bf80-5218-11e9-9879-c7e4be1445f9.png)

This problem is quite tricky in this special case (which does happen):
If I define a rectangle centered at (1,1) and width=height=2px, how many pixels are covered exactly? With CASA, I think the answer is 9 (all pixels in). And in CARTA it is 4 (maybe just (0,0), (0,1), (1,0), and (1,1)). Naively if width=height=2px, the area is 4 px^2. I think CASA's treatment is correct and intuitive.

So what's your insight on this issue? @veggiesaurus @pford We may have to consider to be consistent with CASA.  

UPDATE:
in carta if we define a square then rotate by 90 degrees, npts may change (and consistent with CASA)! To be consistent with CASA when rot=0, maybe we need to treat it as a polygon as well.

veggiesaurus:
![image](https://user-images.githubusercontent.com/592504/55210752-a1800280-51f1-11e9-834f-67665a0a8fbc.png)
for the record, DS9 says 4 pixels. However, when making an ellipse with the same dimensions, it shows 5 pixels (where a consistent rounding behaviour between shapes would be 3 pixels)

![image](https://user-images.githubusercontent.com/592504/55210932-73e78900-51f2-11e9-97ab-2667eaf93af8.png)


This is a tricky issue, and opinion might be split depending on user's existing software preference. But at the very least we need to be consistent between the various shapes. 

![image](https://user-images.githubusercontent.com/592504/55211153-87dfba80-51f3-11e9-9cf2-6ea70ebafe52.png)

[...]

PR #126, opened 2019-03-30
Use LCPolygon for correct pixel inclusion in rectangle regions

pford:
For issue #123 

Have not run region ICD tests yet.


pford:
Region tests also passed.  Added NumPixels to REGION_STATISTICS test, expected value 60501 passed.

PR #128, opened 2019-04-01
Stop old spectral profile process if cursor's position renewed

markccchiang:
Let the old spectral profile process stoppable can save a lot of time if an image file has many channels. By default, the profile process will check per 100 channels whether if the region parameter is changed. If so, stop the old process and let the new process start early. The performances to get the complete profile data between this and original methods are roughly similar, as shown in the below plot:
time-vs-case
where I use SgrB2-N.spw0.line.fits with 7680 channels to do the test.

### DIFF
--- a/Region/Region.cc
+++ b/Region/Region.cc
@@ -384,12 +384,11 @@ bool Region::getData(std::vector<float>& data, casacore::SubLattice<float>& subl
             }
             size_t begin = 0; // the begin index of profile vector in each copy
             size_t upperBound = (profileSize%checkPerChannels == 0 ? profileSize/checkPerChannels : profileSize/checkPerChannels + 1);
-            // get regional parameters setting as this moment
+            // get regional parameters setting at this moment
             std::vector<CARTA::Point> tmp_ctrlpoints = m_ctrlpoints;
             // get profile data section by section with a specific length (i.e., checkPerChannels)
             for (size_t i=0; i<upperBound; ++i) {
                 if (pointsChanged(tmp_ctrlpoints)) { // check if regional parameters setting changed during this loop
-                    data.clear();
                     return false; // stop profile process and don't send the data
                 }
                 // modify the start position for slicer
@@ -399,9 +398,8 @@ bool Region::getData(std::vector<float>& data, casacore::SubLattice<float>& subl
                 casacore::Slicer slicer(start, count);
                 casacore::Array<float> buffer;
                 sublattice.doGetSlice(buffer, slicer);
-                std::vector<float> tmp = buffer.tovector();
-                memcpy(&data[begin], &tmp[0], tmp.size()*sizeof(float));
-                begin += tmp.size();
+                memcpy(&data[begin], buffer.data(), count(profileAxis)*sizeof(float));
+                begin += count(profileAxis);
             }
             dataOK = true;
         } catch (casacore::AipsError& err) {
--- a/Region/Region.cc
+++ b/Region/Region.cc
@@ -364,16 +364,54 @@ casacore::IPosition Region::xyShape() {
 // ***********************************
 // Region data
 
-bool Region::getData(std::vector<float>& data, casacore::SubLattice<float>& sublattice) {
+bool Region::getData(std::vector<float>& data, casacore::SubLattice<float>& sublattice, int checkPerChannels) {
     // fill data vector using region SubLattice
     bool dataOK(false);
     casacore::IPosition sublattShape = sublattice.shape();
     data.resize(sublattShape.product());
-    casacore::Array<float> tmp(sublattShape, data.data(), casacore::StorageInitPolicy::SHARE);
-    try {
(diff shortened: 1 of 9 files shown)


Region/Region.cc
```diff
@@ -364,16 +364,55 @@ casacore::IPosition Region::xyShape() {
...
+            for (int i=0; i<sublattShape.size(); ++i) {
+                if (count(i)>1) {
+                    profileAxis = i;
+                    profileSize = count(i);
+                }
+            }
+            size_t begin = 0; // the begin index of profile vector in each copy
+            size_t upperBound = (profileSize%checkPerChannels == 0 ? profileSize/checkPerChannels : profileSize/checkPerChannels + 1);
+            // get regional parameters setting as this moment
+            std::vector<CARTA::Point> tmp_ctrlpoints = m_ctrlpoints;
```
veggiesaurus: do we need a mutex around this? What if one thread is busy updating a region, while the other thread is calculating a spectral profile?

Region/Region.cc
```diff
@@ -364,16 +364,55 @@ casacore::IPosition Region::xyShape() {
...
+                }
+                // modify the start position for slicer
+                start(profileAxis) = i*checkPerChannels;
+                 // modify the count for slicer
+                count(profileAxis) = (checkPerChannels*(i+1) < profileSize ? checkPerChannels : profileSize - i*checkPerChannels);
+                casacore::Slicer slicer(start, count);
+                casacore::Array<float> buffer;
+                sublattice.doGetSlice(buffer, slicer);
+                std::vector<float> tmp = buffer.tovector();
+                memcpy(&data[begin], &tmp[0], tmp.size()*sizeof(float));
```
veggiesaurus: you can copy directly from the `casacore::Array` using `casacore::Array::data()` without creating a temporary vector. In fact, you could even reuse the `data` vector, and just create a `casacore::Array` that uses the correct section of the `data` vector.

Region/Region.cc
markccchiang: There is a mutex start from https://github.com/CARTAvis/carta-backend/blob/master/Frame.cc#L738. So I think this block is also been locked.

veggiesaurus: Doesn't that mean that there's no way for `m_ctrlpoints` to ever be changed during `Region::getData`?

Region/Region.cc
markccchiang: Thanks for the suggestion! I have changed the code and use `casacore::Array::data()` to copy data.

Region/Region.cc
veggiesaurus: just reading through the code, that lock is the `latticeMutex` which the frame does not attempt to lock when updating region parameters. So it's still possible to update `m_ctrlpoints` while `getData` is running, which makes sense. However, my concern is that there's a very small chance that you could be copying `m_ctrlpoints` to `tmp_ctrlpoints` _while_ `m_ctrlpoints` is being updated by `setPoints`

Unrelated, note that changing region rotation will also invalidate the current region calculations

markccchiang: I see. I am thinking a good way to prevent this case happens.

veggiesaurus: it's a very unlikely situation, but I guess it _could_ happen?

pford: This could happen, as Session::onSetRegion sends the spatial profile before initiating the spectral profile.  The calls for data in onSetRegion could include its region points/rotation and used to check for change (not only for profiles, but also stats and histograms).  Or perhaps SubLattice::positionInParent would tell us these xy coordinates?

Region/Region.cc
```diff
@@ -509,7 +546,7 @@ void Region::fillSpectralProfileData(CARTA::SpectralProfileData& profileData, in
             newProfile->set_coordinate(profileCoord);
             newProfile->set_stats_type(CARTA::StatsType::None);
             std::vector<float> profile;
-            if (getData(profile, sublattice)) {
+            if (getData(profile, sublattice, 100)) {
```
pford: Noting here that if getData returns false before the profile is complete, the SPECTRAL_PROFILE_DATA message will still be sent with empty values[ ] array.

veggiesaurus: What we really want is a solution that checks for cancellation, _and_ sends partial data as well, so users can be reassured that progress is happening if a region takes a long time to calculate. Cancellation checks can happen at a higher cadence than the updates. 
Below is a demo of this happening on the old CyberSKA viewer (insanely slow performance due to server issues)

region_updates

markccchiang:
I have updated the code based on the new master branch. The stoppable spectral profile logic is moved to the Frame. By comparing the latest cursor's position with the cursor's position from `SubLattice::positionInParent`, we can determine whether the spectral profile process is outdated and can be stopped.

pford:
Proceeding with pull request, needed for other development.  Partial messages are requested in issue #135.

Issue #137, opened 2019-04-02
Incorrect HDF5 data streams


pford:
Regions which include NaN values as well as finite values return stats NaN values and spectral profile NaN values.  Region histograms are always identical to image histogram (as reported in issue #129) even though a new message is sent when the region changes.

pford:
Frame::fillRegionHistogramData is used for *all* regions.  The loader is asked for image stats for the current chan/stokes regardless of the region id, and if the stats are valid, the loaded histogram is used.  But this is for the image region so none is calculated for the specified region and the image histogram is always returned.

pford:
This should be fixed in PR #138 and PR #139 , then will be ready to test.

veggiesaurus:
struggling to test due to the other HDF5 issue, which still seems to occur when loading an un-tiled HDF5  image

kswang1029:
Tried some ALMA images converted to HDF5, region stats/spectral profile/histogram works well with different sizes (including/excluding nan pixels). 🥇

pford:
Just pushed branch pam/improve_pixel_mask to use LatticeIterator for the pixel mask rather than reading the entire region data at once.  Also only gets the sublattice mask for statistics (non-point-region spectral profile and region stats data), mask not needed for histogram calculations.

veggiesaurus:
HDF5 regions now work similarly to FITS, closing

PR #162, opened 2019-04-04
Use lattice iterator to generate pixel mask

pford:
Use casacore::LatticeIterator for pixel mask rather than reading entire SubLattice for better memory usage.  Set SubLattice pixel mask for statistics only (region spectral profile and stats data).  If SubLattice has no tile shape, make one with TiledFileAccess for iterator cursor shape; after casacore PR 892 is merged, this will be done in HDF5Lattice automatically and this code will not be used (should be removed later).

Response to issue #137

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -2,6 +2,7 @@
 #include "compression.h"
 #include "util.h"
 
+#include <casacore/tables/DataMan/TiledFileAccess.h>
 #include <tbb/blocked_range2d.h>
 #include <tbb/parallel_for.h>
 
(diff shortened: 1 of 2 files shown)


veggiesaurus:
> after casacore PR 892 is merged, this will be done in HDF5Lattice automatically and this code will not be used (should be removed later).

So we can use the master branch of `casacore` for now with this fix, until https://github.com/casacore/casacore/pull/892 is merged?

pford:
Yes, this code compiles with the current casacore master branch.

Issue #113, opened 2019-03-26
Cache region statistics


pford:
casacore LatticeStatistics calculates all stats types, then the requested values are retrieved into an array to add to the spectral profile.  If all were cached, stats changes/additions could be retrieved from the cache rather than recalculated.

pford:
If the region/channel/stokes doesn't change, do not need to send previous stats, only new ones.

PR #164, opened 2019-04-05
Cache region stats data

pford:
Issue #113 , store per-channel, per-stokes region stats.  Will help performance during animation when region stats are requested.  When the region is changed (control points and rotation), the stats and histograms are invalidated.  When they are recalculated for the new region, the old stats are cleared, the new stats are stored, and the valid flag is set again.

Not required for release 1.1, wait to merge for 1.2--?

### DIFF
--- a/Region/RegionStats.cc
+++ b/Region/RegionStats.cc
@@ -19,22 +19,36 @@
 using namespace carta;
 using namespace std;
 
+RegionStats::RegionStats() {
+    clearStats();
+}
+
+RegionStats::~RegionStats() {
+}
+
+// ***** Cache *****
+
+void RegionStats::clearStats() {
+    m_histogramsValid = false;
+    m_statsValid = false;    
+}
+
 // ***** Histograms *****
 
 // config
 bool RegionStats::setHistogramRequirements(const std::vector<CARTA::SetHistogramRequirements_HistogramConfig>& histogramReqs) {
-    m_configs = histogramReqs;
+    m_histogramReqs = histogramReqs;
     return true;
 }
 
 size_t RegionStats::numHistogramConfigs() {
-    return m_configs.size();
+    return m_histogramReqs.size();
 }
 
 CARTA::SetHistogramRequirements_HistogramConfig RegionStats::getHistogramConfig(int histogramIndex) {
     CARTA::SetHistogramRequirements_HistogramConfig config;
-    if (histogramIndex < m_configs.size())
-        config = m_configs[histogramIndex];
+    if (histogramIndex < m_histogramReqs.size())
+        config = m_histogramReqs[histogramIndex];
     return config;
 }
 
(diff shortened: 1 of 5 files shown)


pford:
Noting here I will be out of town the next four days, another reason to delay the merge.

Region/RegionStats.cc
```diff
@@ -127,51 +143,78 @@ void RegionStats::calcHistogram(int channel, int stokes, int nBins, float minVal
...
+        if (m_statsValid && (m_statsData.count(stokes)) && (m_statsData.at(stokes).count(channel))) {
+            // used stored stats
+            try {
+                std::vector<double> storedStats(m_statsData.at(stokes).at(channel));
+                for (size_t i=0; i<m_statsReqs.size(); ++i) {
+                    // add StatisticsValue to message
+                    auto statsValue = statsData.add_statistics();
+                    auto cartaStatType = static_cast<CARTA::StatsType>(m_statsReqs[i]);
+                    statsValue->set_stats_type(cartaStatType);
+                    statsValue->set_value(storedStats[cartaStatType-13]);
```
confluence: Why 13? Could we add a named constant for this?

Region/RegionStats.cc
```diff
@@ -127,51 +143,78 @@ void RegionStats::calcHistogram(int channel, int stokes, int nBins, float minVal
...
+                    }
                 } else {
-                    statsValue->set_value(std::numeric_limits<double>::quiet_NaN());
+                    value = results[i][0];
+                }
+                statsValue->set_value(value);
+
+                // cache stats values
+                if (m_statsData[stokes][channel].empty()) { // resize vector, set to NaN
+                    m_statsData[stokes][channel].resize(CARTA::StatsType_MAX-13, // 13 is the first stat
```
confluence: See other comment about the magic number.

Region/RegionStats.cc
veggiesaurus: The enum I got from @markccchiang from CARTA 0.9 had weird numbering scheme...so I'm to blame for this. We can fix this magic number by fixing the ICD.

confluence: Ah. I have no objection to the *value* of the magic number; I'd just like it not to be magic. :)

pford: [cartaStatType - CARTA::NumPixels] is also odd.  It would be clearest in code if the enum started with NumPixels and ended with None, then the index into the vector would be the stat's enum value itself.  A bit uneasy about this whole thing, maybe yet another map (!) would be better?

Region/RegionStats.cc
pford: Again, carta-protobuf has StatsType_MIN = None (0), StatsType_MAX = PluginDefined (31), and StatsType_ARRAYSIZE = StatsType_MAX + 1, which are totally useless unless we have an always-half-empty vector of size 32.

Region/RegionStats.cc
veggiesaurus: @pford that sounds like a good idea, we can start with NumPixels. Do we even need a value for `None`? We could just default to using `Sum` for single-pixel profiles. That way, if we extend the list of stats, we don't have to move `None` around, or start adding additional types _after_ `None`.

pford: The backend does not need type None, does not read the stats type for a point region.  If spectral requirements are set for a Point, the backend just gets the spectral profile data, sets the type in the return message to None, and sends it.  We could use another type (Sum) for both messages.  StatsType::None is not used to turn off profiles; an empty SpectralConfig is sent instead.

PR #166, opened 2019-04-06
Use 3D histogram loaded from HDF5 file, if it exists

### LINKED ISSUE #114: 3D histogram loaded from HDF5 file is not actually used
The cube histogram is requested in a completely different way to per-channel histograms, so the code which would access the data loaded from a HDF5 file is not currently triggered. This needs refactoring in Frame and Session.

confluence:
I found a simple way to fix this with minimal refactoring, by moving the code for loading the image histogram into its own function. That leaves some duplication between the `Session` and `Frame` code which could probably be reduced, but that can be done later.

This fixes #114.


Session.cc
```diff
@@ -506,13 +506,21 @@ bool Session::sendCubeHistogramData(const CARTA::SetHistogramRequirements& messa
                 if (frames.at(fileId)->getRegionHistogram(regionId, channel, stokes, numbins, *histogram)) {
                     // use stored cube histogram
                     sendFileEvent(fileId, "REGION_HISTOGRAM_DATA", requestId, histogramMessage);
+                    dataSent = true;
```
confluence: This was missing from the original code -- was this intentional, or a bug?

PR #167, opened 2019-04-06
Update CMakeLists.txt

zarda:
Add -O2 optimization flag to enable vectorization. Speed up to 4x.


veggiesaurus:
Optimisation flags should only be applied when building in release mode. I assume all official builds and performance tests occur with release mode compilation (i.e. calling cmake -DCMAKE_BUILD_TYPE=Release)

Cmake defaults to either O2 or O3 for release mode, so these flags are not needed.

zarda:
https://github.com/Kitware/CMake/blob/2f01090d81c086cffd6d4dc048a635343d7a4528/Modules/Compiler/GNU.cmake#L53

PR #178, opened 2019-04-11
Move two functions from main.cc to util

markccchiang:
To let `main.cc` looks a little bit clean. I move two functions `readPermissions(...)` and `checkRootBaseFolders(...)` from `main.cc` to `util.h/.cc`.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -1,15 +1,11 @@
-#include <casacore/casa/OS/File.h>
 #include <casacore/casa/OS/HostInfo.h>
-#include <casacore/casa/Inputs/Input.h>
 #include <fmt/format.h>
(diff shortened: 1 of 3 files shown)


main.cc
```diff
@@ -1,15 +1,11 @@
-#include <casacore/casa/OS/File.h>
 #include <casacore/casa/OS/HostInfo.h>
-#include <casacore/casa/Inputs/Input.h>
```
pford: casacore::Input is used in main, not util

markccchiang: Thanks for reminding! I put `#include <casacore/casa/Inputs/Input.h>` back to `main.cc`.

Issue #179, opened 2019-04-12
session with unfinished jobs is not killed when reload CARTA


kswang1029:
To reproduce the issue:
1) open set_QA/orion_12co_hera.fits and enable spectral profiler
2) move cursor around to see profiles, should update quickly as the image is small
3) reload CARTA, then the backend console will show Num sessions: 1 
4) open set_QA/OrionKL_sci.spw19.cube.I.pbcor.fits and move cursor around to see spectral profiles 
5) don't wait the profile update and reload CARTA, now the backend console shows Num sessions: 2. This suggests the session associated with step 4 is still alive but CARTA has been reloaded.

markccchiang:
Currently, the backend cannot kill the spectral profile instantly. It only set the checkpoint per 100 channels and stop it early if the cursor's position changed. It is possible that the closed file's cursor position does not change anymore and leads to its spectral profile process continuous. I will check it. Does this case crash the backend?

kswang1029:
if the first image is large enough, likely yes.

veggiesaurus:
@markccchiang we need to wait until the task is finished before destroying the objects. @robsimmonds any ideas? Maybe in the destructor for `Region` or `Frame`, wait for the task to exit?

markccchiang:
For region spectral profile data, backend sends four messages `Session::sendSpatialProfileData()`, `Session::sendSpectralProfileData()`, `Session::sendRegionHistogramData()` and `Session::sendRegionStatsData()` a time when we update a region status. That means Session will call the same Frame four times for calculations. Destroy the Frame during this process (for example one a new image file) will carsh the backend. This is the casa which @kswang1029 has experienced.

markccchiang:
I think this issue is already solved. @kswang1029 could you please confirm it? Thank you!

kswang1029:
@markccchiang with the help from @ajm-asiaa I think the issue is fixed. 👍 Close the issue.

PR #183, opened 2019-04-15
Patches to fix z profile bugs

markccchiang:
* Last time in order to fix the z profile problem for a single channel image. I set the `sublattShape > 1` as a necessary condition for executing the stoppable process. However, it is not correct, which leads to all z profile processes for the multi-channel images are non-stoppable. Now I correct it.
* Add a private bool variable `connected` in the `Frame`. If Session is closed or the image file under Session is closed, Session fires a signal to Frame and set `connected=false`. So the z profile process belongs to such Frame will be stopped.
* If the stopped process is called, the Frame will not continue the next process `region->fillSpectralProfileData(...)` and will also not send z profile data to the frontend. This could avoid the crash.

This is for issue #179.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -244,6 +244,7 @@ bool Frame::setImageView(const CARTA::ImageBounds& newBounds, int newMip,
 }
 
 bool Frame::setImageChannels(int newChannel, int newStokes, std::string& message) {
+    increase_job_count_();
     bool updated(false);
     if (!valid || (regions.count(IMAGE_REGION_ID)==0)) {
         message = "No file loaded";
@@ -263,6 +264,7 @@ bool Frame::setImageChannels(int newChannel, int newStokes, std::string& message
             }
         }
     }
+    decrease_job_count_();
     return updated;
 }
 
@@ -454,6 +456,7 @@ bool Frame::setRegionStatsRequirements(int regionId, const std::vector<int> stat
 
 bool Frame::fillRasterImageData(CARTA::RasterImageData& rasterImageData, std::string& message) {
     // fill data message with compressed channel cache data
+    increase_job_count_();
     bool rasterDataOK(false);
     std::vector<float> imageData;
     if (getImageData(imageData)) {
@@ -512,6 +515,7 @@ bool Frame::fillRasterImageData(CARTA::RasterImageData& rasterImageData, std::st
     } else {
         message = "Raster image data failed to load";
     }
+    decrease_job_count_();
     return rasterDataOK;
 }
 
--- a/Frame.cc
+++ b/Frame.cc
@@ -13,6 +13,7 @@ Frame::Frame(const string& uuidString, const string& filename, const string& hdu
     : uuid(uuidString),
       valid(true),
       connected_(true),
+      job_count_(0),
       cursorSet(false),
       filename(filename),
       loader(FileLoader::getLoader(filename)),
@@ -1018,6 +1019,7 @@ bool Frame::getSublatticeXY(casacore::SubLattice<float>& sublattice, std::pair<i
 }
 
 bool Frame::getSpectralData(std::vector<float>& data, casacore::SubLattice<float>& sublattice, int checkPerChannels) {
+increase_job_count_();
     bool dataOK(false);
     casacore::IPosition sublattShape = sublattice.shape();
     data.resize(sublattShape.product());
@@ -1033,8 +1035,10 @@ bool Frame::getSpectralData(std::vector<float>& data, casacore::SubLattice<float
             // get profile data section by section with a specific length (i.e., checkPerChannels)
             for (size_t i=0; i<upperBound; ++i) {
                 // check if cursor's position changed during this loop, if so, stop the profile process
-                if (tmpXY != cursorXY || !connected_)
+                if (tmpXY != cursorXY || !connected_) {
+decrease_job_count_();
                     return false;
+                }
                 // modify the start position for slicer
                 start(spectralAxis) = i*checkPerChannels;
                 // modify the count for slicer
@@ -1055,5 +1059,6 @@ bool Frame::getSpectralData(std::vector<float>& data, casacore::SubLattice<float
         } catch (casacore::AipsError& err) {
         }
     }
+decrease_job_count_();
     return dataOK;
 }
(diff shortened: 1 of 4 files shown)


Frame.cc
```diff
@@ -1012,20 +1018,20 @@ bool Frame::getSublatticeXY(casacore::SubLattice<float>& sublattice, std::pair<i
 bool Frame::getSpectralData(std::vector<float>& data, casacore::SubLattice<float>& sublattice, int checkPerChannels) {
     bool dataOK(false);
     casacore::IPosition sublattShape = sublattice.shape();
+    casacore::IPosition count(sublattShape);
+    int profileAxis = -1; // profile axis index
```
pford: (spectral) profileAxis is already determined and saved as local variable spectralAxis, and number of channels (profileSize) is local variable nchan (function nchannels())

Session.cc
```diff
@@ -229,10 +236,12 @@ void Session::onCloseFile(const CARTA::CloseFile& message, uint32_t requestId) {
     std::unique_lock<std::mutex> lock(frameMutex);
     if (fileId == ALL_FILES) {
         for (auto& frame : frames) {
+            frame.second->disconnect_called();
             frame.second.reset();  // delete Frame
```
pford: Does this give the Frame time to react to being disconnected (stop spectral profile) before deleting it?  I think it may still seg fault on Frame destruction.

pford: Same with main OnDisconnect -> Session::disconnect_called() -> Frame::disconnect_called(): Session is deleted after Session::disconnect_called(), and ~Session deletes Frame

veggiesaurus: unless `disconnet_called` is doing some form of busy wait to ensure that the task has been concluded (either because it finished properly, or because it was interrupted correctly), then you'll still get a seg fault.

Frame.cc
```diff
@@ -804,10 +809,11 @@ bool Frame::fillSpectralProfileData(int regionId, CARTA::SpectralProfileData& pr
                     auto cursorPos = region->getControlPoints()[0];
                     // try use the loader's optimized cursor profile reader first
                     if (!loader->getCursorSpectralData(spectralData, profileStokes, cursorPos.x(), cursorPos.y())) {
-                        getSpectralData(spectralData, sublattice, 100);
+                        // only send spectral profile data to frontend while it is complete
+                        if (getSpectralData(spectralData, sublattice, 100))
+                            region->fillSpectralProfileData(profileData, i, spectralData);
                     }
                     guard.unlock();
```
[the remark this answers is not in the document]
markccchiang: I have changed the code that unlocks latticeMutex after reading data into spectralData and before filling SpectralProfileData message.

Frame.cc
markccchiang: Done.

Session.cc
markccchiang: It is true. Now I add a private variable `tbb::atomic<int> job_count_` to count the number of jobs in the Frame and use a while loop to wait for Frame's jobs finished before deleting it.

markccchiang: Using the job count and wait for job count becomes zero before deleting an object looks work. However, the downside is that we have to carefully arrange the position of `increase_job_count_()` and `decrease_job_count_()` to make sure the job count goes to zero when all processes are finished. Also, the job count has to be counted immediately when the message comes in.

Frame.cc
```diff
@@ -804,10 +810,13 @@ bool Frame::fillSpectralProfileData(int regionId, CARTA::SpectralProfileData& pr
                     auto cursorPos = region->getControlPoints()[0];
                     // try use the loader's optimized cursor profile reader first
                     if (!loader->getCursorSpectralData(spectralData, profileStokes, cursorPos.x(), cursorPos.y())) {
-                        getSpectralData(spectralData, sublattice, 100);
+                        // get spectral profile data
+                        bool complete = getSpectralData(spectralData, sublattice, 100);
+                        guard.unlock();
+                        // only send spectral profile data to frontend while it is complete
+                        if (complete)
+                            region->fillSpectralProfileData(profileData, i, spectralData);
```
pford: The job count is decremented at the end of getSpectralData(), Frame could be deleted while its Region is filling the spectral profile message.

Session.cc
robsimmonds: I have not tested it but I think it looks logically correct. Normally I don't like busy waits but the way it is used here means that it will never wait long. The style is not consistent and job_count is not very descriptive. Also there is now a busy wait in the for frames in the Session destructor. This is OK, but since the Session destructor should never be called until all tasks processing on this Session are complete, it is confusing to have a busy wait here. If job count is not zero here it would indicate that a Session object is being destroyed due to a code bug.

markccchiang: Thanks for the comments! I moved the wait for finish step into `Frame::DisconnectCalled()`. When `Session::disconnect_called()` is called, it will trigger `Frame::DisconnectCalled()` to execute stop and wait for finish step in Frames. So there is no busy wait in the Session destructor now. The code style I use is google's style, since Angus wants a unified style. But I can change them back to the consistent style before the decision is made. Also, I may change the name `job_counter` to `spectral_profile_counter`(?) I am wondering should I apply the "wait for finish step" on the other Frame's functions which are expected to have a long process time. Because they will also cause a crash like the spectral profile process. The coding is not finished yet and they need more checks. Any comments or suggestions are welcome.

Frame.cc
markccchiang: Moving the count increment/decrement to the function `Frame::fillSpectralProfileData()` to make sure the whole spectral profile process is done before deleting the Frame.

Frame.cc
```diff
@@ -1010,41 +1019,34 @@ bool Frame::getSublatticeXY(casacore::SubLattice<float>& sublattice, std::pair<i
 }
 
 bool Frame::getSpectralData(std::vector<float>& data, casacore::SubLattice<float>& sublattice, int checkPerChannels) {
+increase_job_count_();
```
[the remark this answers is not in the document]
markccchiang: Ok.

markccchiang:
The stop procedure for spectral profile process is that: 
1. Before destroying a Frame, execute the function `Frame->disconnectCalled()`
2. Set a flag `connected = false` in the Frame, so the spectral profile process can stop early
3. Use a while loop and wait for the whole spectral profile process is done, i.e., wait for `zProfileCount` becomes zero
4. Destroy the Frame at last

PR #190, opened 2019-04-18
Remove conditional behaviour of function which populates histogram messages

### LINKED ISSUE #189: hdf5 cube histogram hangs
from a general test I found that requesting a cube histogram from a hdf5 image hangs. In my test case, I have enabled statistics widget, spectral profiler widget, and histogram widget. A region is created and all the three widgets are set to that region. Then when I request the cube histogram, the dialog hangs. It is supposed to be instantly since the cube histogram should have been calculated and saved in the hdf5 image.

confluence:
The `createCubeHistogramMessage` function is used to populate cube histogram messages. It previously failed silently if the global histogram progress variable in the session was set to `HISTOGRAM_CANCEL`. This was causing a hang in the frontend if a cube histogram was requested after a region histogram had been requested. The check is unnecessary within the cube histogram calculation, because it is already being performed in the calling function wherever this function is used.

This fixes #189.

### DIFF
--- a/Session.cc
+++ b/Session.cc
@@ -647,14 +647,12 @@ bool Session::sendCubeHistogramData(const CARTA::SetHistogramRequirements& messa
 }
 
 void Session::createCubeHistogramMessage(CARTA::RegionHistogramData& message, int fileId, int stokes, float progress) {
-    // check for cancel then update progress and make new message
-    if (histogramProgress != HISTOGRAM_CANCEL) {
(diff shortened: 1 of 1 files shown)


veggiesaurus:
looks fine to me. @kswang1029 can you test on mix 'n match with this branch? I haven't spotted any regression bugs in my brief testing.

kswang1029:
Yes, it works well now. 👍

PR #192, opened 2019-04-20
Rwjs/animation start

robsimmonds:
Added support for the StartAnimation message. Note that this not get called until someone writes a client to support it.

Changed the names of the queue and the functions previously with an "ani" in them to reflect that they really are to do with setting the image channels.

Change the SetImageChannelsTask::execute() so that tasks are recycled after each execution so that this does not hold on to a TBB thread.

### DIFF
--- a/Session.cc
+++ b/Session.cc
@@ -880,97 +880,99 @@ Session::execute_animation_frame( void )
 		 this );
     exit(1);
   }
-
   if( _ani_obj->_stop_called ) {
-
-    // Note not checking for the stop frame yet.
-    
+    // Note: not checking for the stop frame yet, since need clarity on logic
+    // ... so just stop now!
     return false;
   }
-  
-  ::CARTA::AnimationFrame curr_frame= _ani_obj->_curr_frame;
-  ::CARTA::AnimationFrame delta_frame= _ani_obj->_delta_frame; 
+
   bool recycle_task= true;
-  
-  auto fileId(_ani_obj->_file_id);
-  if (frames.count(fileId)) {
-    try {
-      std::string errMessage;
-      auto channel = curr_frame.channel();
-      auto stokes = curr_frame.stokes();
-      bool channelChanged(channel != frames.at(fileId)->currentChannel());
-      bool stokesChanged(stokes != frames.at(fileId)->currentStokes());
-      if (frames.at(fileId)->setImageChannels(channel, stokes, errMessage)) {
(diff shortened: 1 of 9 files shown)


veggiesaurus:
@ajm-asiaa can you re-trigger this test? Failed due to DNS failure

Session.cc
```diff
@@ -52,7 +55,10 @@ Session::~Session() {
     }
     frames.clear();
     outgoing->close();
-    
+    if( _ani_obj ) {
```
veggiesaurus: Can we make _ani_obj a unique_ptr to avoid this?

Session.cc
```diff
@@ -830,3 +839,159 @@ void Session::sendLogEvent(std::string message, std::vector<std::string> tags, C
...
+	_ani_obj->_curr_frame= tmpF;
+      }
+    }
+    _ani_obj->_tLast= std::chrono::high_resolution_clock::now();
+  }
+  return recycle_task;
+}
+
+
+void
```
veggiesaurus: can you modify this to conform to the google c++ formatting styles, in preparation for our code formatting cleanup? 
https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions
(should be on one line. If the line is too long, wrap the arguments, not the return type

Session.h
```diff
@@ -39,7 +40,7 @@ class Session {
  public:
     std::string uuid;
     carta::FileSettings fsettings;
-    tbb::concurrent_queue<std::pair<CARTA::SetImageChannels,uint32_t>> aniq;
+    tbb::concurrent_queue<std::pair<CARTA::SetImageChannels,uint32_t>> setchanq;
```
veggiesaurus: I presume once we have animation messages working properly, we can limit this queue to a size of one, since _only_ the latest SET_CHANNELS message will be valid

ajm-asiaa:
@veggiesaurus OK. It worked this time. The IAA website must have been down.

Session.cc
robsimmonds: I have done that.

Session.h
robsimmonds: Yes, this could just become a pointer.

Session.cc
robsimmonds: done.

Session.h
robsimmonds: Yes could just be a pointer once it is not used for animation any more.

PR #197, opened 2019-04-24
Rwjs/minmax opt

robsimmonds:
Tiny optimisation to minmax function. Since this appears to dominate the performance of
the histogram function, may as well squeeze all we can from this.

### DIFF
--- a/Region/MinMax.h
+++ b/Region/MinMax.h
@@ -45,9 +45,9 @@ void MinMax<T>::operator()(const tbb::blocked_range<size_t> &r) {
         T val = data[i];
 	if( std::isfinite(val) ) {
 	    if( val < tmin ) {
-	        tmin= val;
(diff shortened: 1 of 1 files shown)


Region/MinMax.h
```diff
@@ -43,10 +43,10 @@ void MinMax<T>::operator()(const tbb::blocked_range<size_t> &r) {
     T tmax = maxval;
     for(size_t i = r.begin(); i != r.end(); ++i) {
         T val = data[i];
-	if(std::isfinite(val)) {
-	    if(val < tmin) {
+	if( std::isfinite(val) ) {
```
veggiesaurus: I'm pretty sure she meant `if (std::isfinite(val)) {`, not `if( std::isfinite(val) ) {`

Issue #175, opened 2019-04-10
Implement HDF5Image (casacore ImageInterface)


pford:
As first step to supporting FluxDensity (#174 ) and CASA imageanalysis (region import/export, fitting, moments, pv generation, etc.).

Current casacore::HDF5Image assumes a casacore::Record interface with coordinate system info, image info, and optional misc. info.  This information will have to be assembled from the HDF5 header attributes.

pford:
Tested images in set_HDF5 using CARTA mix 'n match with pam/hdf5_image backend branch.  All opened without crashing.

pford:
Closing this issue although more may be needed with Lattice->Image conversion and WCS region import/export.  Works with current Lattice implementation.

PR #200, opened 2019-04-25
carta HDF5Image

pford:
Issue #175 - casacore::HDF5Image is incompatible with IDIA's HDF5 schema.  The new class carta::HDF5Image is derived from casacore::ImageInterface.  Works with current casacore::Lattice-based implementation.

Swizzled data in HDF5Loader remains an HDF5Lattice, coordinate system not needed.  Some hdf5 attributes-related code was moved from FileInfoLoader to HDF5Attributes.  Also fixed the cursor z-profile message with hdf5 swizzled data from the loader.

### DIFF
--- a/ImageData/HDF5Image.h
+++ b/ImageData/HDF5Image.h
@@ -1,0 +1,58 @@
+//# HDF5Image.h : CARTA HDF5Image derived from casacore::HDF5Image
+
+#ifndef CARTA_HDF5IMAGE_H
+#define CARTA_HDF5IMAGE_H
+
+#include <casacore/images/Images/HDF5Image.h>
+#include <casacore/images/Images/MaskSpecifier.h>
+#include <casacore/casa/Containers/Record.h>
+
+namespace carta {
+
+class HDF5Image : public casacore::ImageInterface<float> {
+    public:
+      // Construct an image from a pre-existing file.
+      HDF5Image(const std::string& filename, const std::string& array_name,
+          const std::string& hdu, casacore::MaskSpecifier = casacore::MaskSpecifier());
+      // Copy constructor
+      HDF5Image(const carta::HDF5Image& other);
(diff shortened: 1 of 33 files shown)


veggiesaurus:
Wonder if we should call it something less vague than `carta::HDF5Image` to avoid potential clashes or confusion, despite them existing in separate namespaces. `IDIAHDF5Image`? (that's not a great name either, not sure what other options are)

confluence:
How does this image class differ from `casacore::HDF5Image`? Is it only a matter of changing the hardcoded paths to datasets? Could we submit a patch to casacore to make their class more flexible (e.g. by adding optional function parameters to override their defaults)? That doesn't mean we shouldn't proceed with this change in the meantime, though.

veggiesaurus:
@confluence I think the attributes for the cassacore image are stored very differently, IIRC

ImageData/HDF5Image.h
```diff
@@ -0,0 +1,58 @@
+//# HDF5Image.h : CARTA HDF5Image derived from casacore::HDF5Image
```
confluence: Should this not be `casacore::ImageInterface`?

pford: Yes, change of plans.  Will fix.

pford:
Looking at the casacore code, it expects a casacore::HDF5Image to be created in a certain way with a casacore::Record containing CoordinateSystem info, ImageInfo, and ObsInfo.  So opening a casacore::HDF5Image means reading in this Record and supplying the subrecord to the appropriate objects.  The name of the data attribute is only one small part.

confluence:
I see now that this looks much simpler than `casacore::HDF5Image`, and also uses a lot of our other custom code. Given how specific this is to our schema, I think this approach makes sense.

PR #202, opened 2019-04-30
Icd improvements 1.2

robsimmonds:
Major part of the update of the ICD; the header. There are still a few updates to the message bodies that will need to be done, but best to get everyone using this code as soon as possible.


OnMessageTask.h
```diff
@@ -31,75 +31,82 @@ class OnMessageTask : public tbb::task {
...
   tbb::task* execute();
 public:
- SetImageChannelsTask(Session *session_,
-		      std::pair<CARTA::SetImageChannels,uint32_t> req_ )
-       : OnMessageTask( session_ ) {
-    _req= req_;
+ SetImageChannelsTask(Session *session,
+		      std::pair<CARTA::SetImageChannels,uint32_t> req )
+       : OnMessageTask( session ) {
+    req= req;
```
confluence: Missed a space, but we're about to reformat everything anyway so it's not super important.

PR #205, opened 2019-05-02
Confluence/clang format

veggiesaurus:
This PR adjusts include guard style, include ordering and reformats using clang-format.

Renaming variables to match the code style will be performed in a separate PR


confluence:
Note: there are some awkward edge cases, e.g. wrapping of very very long trailing comments. I suggest that we fix these afterwards on a case-by-case basis, e.g. by moving long trailing comments to their own lines.

Issue #201, opened 2019-04-25
Convert Lattice-based objects to Image-based objects.


pford:
Making this a separate issue since #171  and #174 depend on it.  This issue depends on #175 which is done.

Issue #174, opened 2019-04-10
Support stats type FluxDensity


pford:
Requires image info (beam) not in Lattices. First step is to convert Lattice-based objects to Image-based objects (#201 ) then use the ImageStatsCalculator in casacode imageanalysis (I think) since LatticeStatistics cannot compute this stat.

pford:
Noting here that FluxDensity is computed only if the image's CoordinateSystem has a DirectionCoordinate and either the unit is "K" or the image has a beam and the unit is per beam (unit contains "/beam")

PR #209, opened 2019-05-08
Implement casacore ImageInterface types and ImageStatistics

pford:
Issues #201 and #174 
 
As opposed to the previous casacore Lattice, LCRegion (Lattice Coordinate), and SubLattice implementation, the new types have a casacore::CoordinateSystem which will be needed for future development, including region import/export and WCS groups.

Although the backend architecture has not been changed, the objects and associated variable names have been updated. The FileLoader returns a casacore Image type (base class casacore::ImageInterface). The regions are casacore::WCRegion (World Coordinate) types returned as a casacore::ImageRegion, which can be applied to the image to create a casacore::SubImage for profiles and statistics.

A new CartaHdf5Image class, derived from casacore::ImageInterface, creates a casacore::HDF5Lattice for data, casacore::CoordinateSystem for coordinate conversions, casacore::ImageInfo for restoring beam, and pixel mask for NaN values. The hdf5 header keyword/value pairs are used to create a casacore::CoordinateSystem using casacore::ImageFITSConverter.

The SubImage created for the image plane or region is passed to RegionStats as an ImageInterface so that ImageStatistics can compute FluxDensity if possible.


ImageData/HDF5Attributes.h
```diff
@@ -2,22 +2,25 @@
...
+#pragma once
+
 #include <casacore/casa/Containers/Record.h>
 #include <casacore/casa/HDF5/HDF5HidMeta.h>
 
 class HDF5Attributes {
 public:
     // HDF5Record::doReadRecord modified to not iterate through links
     // (links get HDF5Error "Could not open group XXX in parent")
-    static casacore::Record doReadAttributes(hid_t groupHid);
```
confluence: Is the "do" in this function name necessary? It's already a verb. :)

pford: Not necessary, started with casacore "doReadRecord" function.  Feel free to rename during code styling name changes.

PR #210, opened 2019-05-09
Angus/more icd updates

veggiesaurus:
updates the backend for some of the ICD refinements discussed in the CARTA 1.2 and 1.3 ICD discussion document: 
- Makes use of double precision values when sending spectral profiles from regions
- Removed some deprecated fields and changed the stats enum to start from zero
- Added support for feature flags for server, client, and individual files
- Moved the EventType enum into the protocol buffer definitions to ensure consistency between frontend and backend. 
- Added a flow control message for animation (unused for now)

Companion PR of [carta-frontend/#326](https://github.com/CARTAvis/carta-frontend/pull/326)


veggiesaurus:
I'm quite surprised the checks passed, given that it's an ICD change.

Region/RegionStats.cc
```diff
@@ -199,12 +199,11 @@ void RegionStats::fillStatsData(CARTA::RegionStatsData& statsData, const casacor
                 statsValue->set_value(value);
 
                 // cache stats values
-                if (m_statsData[stokes][channel].empty()) {                        // resize vector, set to NaN
-                    m_statsData[stokes][channel].resize(CARTA::StatsType_MAX - 13, // 13 is the first stat
-                        std::numeric_limits<double>::quiet_NaN());
+                if (m_statsData[stokes][channel].empty()) { // resize vector, set to NaN
+                    m_statsData[stokes][channel].resize(CARTA::StatsType_MAX, std::numeric_limits<double>::quiet_NaN());
                 }
                 // first type enum is 13; make 0-based
```
pford: Not critical but this comment should be removed

PR #214, opened 2019-05-10
Angus/variable renaming

veggiesaurus:
Giant PR bringing file, class, function and variable names in line with the style described in `CODESTYLE.md`. A few minor code cleanup changes as well, and a number of `TODO` comments added where things need changing or inspecting by the original author.


FileInfoLoader.h
```diff
@@ -15,59 +15,58 @@
...
-        const casacore::String& axisType3, const casacore::String& axisType4, int& chanAxis, int& stokesAxis);
+    void AddComputedEntries(CARTA::FileInfoExtended* ext_info, const std::string& xy_coords, const std::string& cr_pixels,
+        const std::string& cr_coords, const std::string& cr_ra_dec, const std::string& rade_sys, const std::string& spec_sys,
+        const std::string& bunit, const std::string& axis_inc, const std::string& rs_beam);
+    void AddShapeEntries(CARTA::FileInfoExtended* extended_info, const casacore::IPosition& shape, int chan_axis, int stokes_axis);
+    void FindChanStokesAxis(const casacore::IPosition& data_shape, const casacore::String& axis_type_1, const casacore::String& axis_type_2,
+        const casacore::String& axis_type_3, const casacore::String& axis_type_4, int& chan_axis, int& stokes_axis);
 
     // casacore::Record attributes
+    // TODO: what are these for?
```
pford: Some HDF5 images have attributes that should be numeric but are strings instead.  These return the values as int or double.

Region/Region.h
```diff
@@ -25,116 +25,116 @@ class Region {
...
-    bool getSpectralConfigStokes(int& stokes, int profileIndex);
-    std::string getSpectralCoordinate(int profileIndex);
-    bool getSpectralConfig(CARTA::SetSpectralRequirements_SpectralConfig& config, int profileIndex);
-    void fillSpectralProfileData(CARTA::SpectralProfileData& profileData, int profileIndex, std::vector<float>& spectralData);
-    void fillSpectralProfileData(CARTA::SpectralProfileData& profileData, int profileIndex, casacore::ImageInterface<float>& image);
+    bool SetSpectralRequirements(const std::vector<CARTA::SetSpectralRequirements_SpectralConfig>& profiles, const int num_stokes);
+    size_t NumSpectralProfiles();
+    bool GetSpectralConfigStokes(int& stokes, int profile_index);
+    std::string GetSpectralCoordinate(int profile_index);
+    // TODO: this isn't implemented
```
pford: Yikes!  It used to be!  How does this even work?

pford: Implemented GetSpectralConfig, misspelled in commit message

FileInfoLoader.h
veggiesaurus: `FileInfoLoader::getDoubleAttribute` isn't implemented though, and not used anywhere (instead, it seems like Hdf5Attributes::GetDoubleAttribute` is used.

pford: This was moved from FileInfoLoader to Hdf5Attributes but mistakenly not removed from the .h file.  I fixed this and removed the unused _region local variable from CartaHdf5Image.

pford:
Should we go ahead and rearrange Frame.h to put the public section first?

veggiesaurus:
> Should we go ahead and rearrange Frame.h to put the public section first?

Good idea

pford:
Working on making *Loader::LoadData uniform

veggiesaurus:
There are still a number of outstanding issues picked up by `clang-tidy`. Should we handle these in a separate PR? Lots based on brace use and using `else if` after `return` or `continue`, so quite minor. But we may want to tune our `.clang-tidy` config in future as well

pford:
I made sure clang-tidy is in my $PATH then ran ./scripts/checkstyle.sh and got no output -- ?

veggiesaurus:
> I made sure clang-tidy is in my $PATH then ran ./scripts/checkstyle.sh and got no output -- ?

I had to install the python YAML package as well, not sure if that's the issue? it's looking for the protobuf files in `build`. If you use a different folder for your cmake build, you may need to create a symlink. @confluence any ideas?

confluence:
The script runs `run-clang-tidy`, which is a wrapper script that comes with `clang-tidy`. So I'd guess that needs to be in your path as well. If it is, then maybe run the `find` command without piping first to make sure you're finding the files. But you may also be able to integrate clang-tidy into your IDE, as Angus suggests.

Issue #212, opened 2019-05-10
backend crash when opening set_QA/supermosaic.10.hdf5 


kswang1029:
this happens with the dev backend. 

h5dump -H log is attached.

[https://app.zenhub.com/files/160788905/f66a726b-2c5d-4e43-a475-1547c79fcb62/download](https://app.zenhub.com/files/160788905/f66a726b-2c5d-4e43-a475-1547c79fcb62/download)

update with testing with images in set_HDF5
h_m51_b_s05_drz_sci.hdf5 --> crash
orion_12co_hera.hdf5          --> crash

Noticed that when opening the hdf5 images in set_HDF5, it takes few to several seconds to open images  which are not that big in xy dimension. The same images in FITS are opened much faster.

pford:
It works with the hdf5 images I have.  I cannot scp this image from asiaa (permission denied), sent message to Anthony.

The backend is reading the header attributes, creating a casacore::Record, then feeding that to a utility to create a casacore::CoordinateSystem.  Will see how I can reuse the headers in FileInfoExtended for hdf5.

pford:
The issue was that it was using OBSFREQ for RESTFRQ, but I neglected to convert the OBSFREQ string value to a float.  Also converted FileInfoExtended into casacore::Record for coordinate system rather than reading headers from file again.

kswang1029:
@pford Thanks. supermosaic.10.hdf5 works now. However, the issue remains for 
set_HDF5/h_m51_b_s05_drz_sci.hdf5
set_HDF5/orion_12co_hera.hdf5

Guessing again it is header issue.

pford:
The orion dataset works in my developer build on the dev branch.  The m51 dataset seg faults in wcslib.

pford:
The problem is that the hdf5 converter (incorrectly) makes all header items in the hdf5 image a string value.  CARTA has to figure out which ones to convert from string to bool, int, or float so that the keyword/value pairs can be used to make a coordinate system.  I am fixing this for the test datasets but it may trip up on other hdf5 header keywords in the future.

veggiesaurus:
This is done to maintain round trip compatibility with FITS, where the header entries are all strings.

pford:
Not according to the papers cited by wcslib http://www.atnf.csiro.au/people/mcalabre/WCS/
e.g. table 8 on page 14 of "Representations of world coordinates in FITS"
[...]

PR #215, opened 2019-05-13
CartaHdf5Image header issue

pford:
Issue #212 

Did not convert OBSFREQ header from a string to a float.  Also reuse CARTA::FileInfoExtended header entries for creating coordinate system to improve performance.


Frame.cc
```diff
@@ -11,7 +11,7 @@
 using namespace carta;
 using namespace std;
```
confluence: Should we eliminate this if we're using `std::` explicitly? Are there other bare keywords from this namespace elsewhere in the file?

ImageData/CartaHdf5Image.cc
```diff
@@ -136,36 +137,79 @@ casacore::Bool CartaHdf5Image::doGetMaskSlice(casacore::Array<bool>& buffer, con
...
+            default:
+                header_record.define(header_entry.name(), header_entry.value());
+        }
+    }
+    return header_record;
+}
+
+bool CartaHdf5Image::ShouldBeFloat(const std::string& entry_name) {
+    casacore::String name(entry_name);
+    return ((name == "EQUINOX") || (name == "EPOCH") || (name == "LONPOLE") || (name == "LATPOLE") || (name == "RESTFRQ") ||
```
confluence: Should we maybe have a vector of exact names and a vector of starting substrings and iterate over them? Would that be significantly slower?

pford: Will redo this

PR #231, opened 2019-05-26
Rwjs/task cancel

robsimmonds:
This adds additional TBB contexts for more fine grain task cancellation.
Also moved more event processing to main thread, mainly to reduce the delay before cancellation starts.
Added a few more checks for when animation should be cancelled.
Added -exit_after arg that if set causes the backend to exist after the number of seconds set when there are no sessions left. This is needed for the apache2 based AuthN/AuthZ scheme I have been working on.

### DIFF
--- a/Main.cc
+++ b/Main.cc
@@ -141,10 +141,21 @@ void OnMessage(uWS::WebSocket<uWS::SERVER>* ws, char* raw_message, size_t length
                     }
                     break;
                 }
+                case CARTA::EventType::CLOSE_FILE: {
+                    CARTA::CloseFile message;
+                    if (message.ParseFromArray(event_buf, event_length)) {
+                        session->CheckCancelAnimationOnFileClose(message.file_id());
+                        session->_file_settings.ClearSettings(message.file_id());
+                        session->OnCloseFile(message);
+                    }
+                    break;
+                }
                 case CARTA::EventType::START_ANIMATION: {
                     CARTA::StartAnimation message;
                     message.ParseFromArray(event_buf, event_length);
-                    tsk = new (tbb::task::allocate_root(session->AnimationContext())) AnimationTask(session, head.request_id, message);
+                    session->cancelExistingAnimation();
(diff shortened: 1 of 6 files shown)


Session.h
```diff
@@ -70,13 +70,23 @@ class Session {
         OnSetImageChannels(request.first);
     }
     void CancelSetHistRequirements() {
-        _histogram_progress.fetch_and_store(HISTOGRAM_CANCEL);
+        _histo_context.cancel_group_execution();
+    }
+    tbb::task_group_context& HistoContext() {
```
confluence: This seems like a weird abbreviation. I'd suggest either `Hist` (which is already used elsewhere) or writing `Histogram` out in full.

Session.h
```diff
@@ -120,9 +130,13 @@ class Session {
     tbb::task* getWaitingTask_ptr() {
         return _animation_object->_waiting_task;
     }
-    bool waiting_flow_event() {
+    bool waitingFlowEvent() {
```
confluence: Should this not be CapitalCamelCase?

Main.cc
```diff
@@ -136,15 +137,26 @@ void OnMessage(uWS::WebSocket<uWS::SERVER>* ws, char* raw_message, size_t length
                     if (message.histograms_size() == 0) {
                         session->CancelSetHistRequirements();
                     } else {
-                        tsk = new (tbb::task::allocate_root(session->context()))
+                        tsk = new (tbb::task::allocate_root(session->HistoContext()))
```
markccchiang: It seems have to reset the `_histo_context` like adding `session->HistoContext().reset();` before this line. Otherwise, the new Per-Cube process cannot be started once have set `_histo_context.cancel_group_execution()`.

kswang1029:
@robsimmonds Short question: is this PR the generic solution of cancelling tasks? Or it is just related to animator? I would like to know more a bit in order to perform tests.

PR #238, opened 2019-05-30
fixed histogram cancellation

### LINKED ISSUE #234: per cube histogram is not functional
using the dev branch as of May 30, 2019, the per cube histogram is not functional.

When the progress bar reaches 50%, the calculation ends immediately with a partial histogram shown in the render config widget.

veggiesaurus:
fixes #234


veggiesaurus:
@kswang1029 can you test this on mix 'n match to confirm it solves the cube histogram issue?

kswang1029:
Yes, 75bd013 works!

PR #239, opened 2019-05-30
Swizzled dataset support for region spectral profiles

### LINKED ISSUE #157: Proper HDF5 support for regions using rotated dataset

confluence:
This fixes #157. Region spectral profile stats are calculated on the HDF5 loader using the swizzled dataset, and recalculated whenever the region is resized or moved. They are not recalculated when a different statistic is requested for an unchanged region.

Edit: diff includes some minor automatic formatting fixes in unrelated code.


veggiesaurus:
One thing we need to implement (possibly in another issue, but definitely before 1.2) is a check to see whether the swizzled dataset should be used, based on some guesstimate of how many reads are required.

confluence:
Good idea, and I think we can add it to this PR. We just need a formula.

veggiesaurus:
> Good idea, and I think we can add it to this PR. We just need a formula.

As a first guess, we could assume that casacore reads in one row at a time, so for a region of size _X_ * _Y_ in an image of size _W_ * _H_ * _D_, that would be _Y_ * _D_ reads of size _X_ * 4 bytes. For the swizzled set, it would be _X_ reads of size _Y_ * _D_ * 4 bytes.  

However, I’m pretty sure that casacore would probably be more efficient than that under some circumstances, particularly in situations where _X_ is close to _W_, in which case a single sequential read might be used for a number of rows. Can you dig into the code to work out what it’s actually being read, specifically for HDF5 files? (Might just want to debug print out when it calls `H5Dread` To get a better idea)

veggiesaurus:
@confluence for now, should we simply use the above approach as an initial method for determining swizzled dataset use, until we have some more concrete numbers using casacore? I'd like to get this PR merged soon

confluence:
I'm happy to do that. Let's make a separate issue for the optimisation.

ImageData/Hdf5Loader.h
```diff
@@ -293,13 +296,138 @@ bool Hdf5Loader::GetCursorSpectralData(std::vector<float>& data, int stokes, int
...
+                return nullptr;
+            }
+
+            for (size_t y = 0; y < num_y; y++) {
+                // skip all Z values for masked pixels
+                if (!mask->getAt(IPos(2, x, y))) {
+                    continue;
+                }
+                for (size_t z = 0; z < num_z; z++) {
+                    float& v = slice_data[y * num_z + z];
```
veggiesaurus: no need to use a reference for a primitive type. In fact, since all the stats are done using doubles, it would be better to just use `double v`

ImageData/Hdf5Loader.h
```diff
@@ -293,13 +296,138 @@ bool Hdf5Loader::GetCursorSpectralData(std::vector<float>& data, int stokes, int
...
+
+        float mean_sq;
+
+        // Calculate final stats
+        for (size_t z = 0; z < num_z; z++) {
+            if (num_pixels[z]) {
+                mean[z] = sum[z] / num_pixels[z];
+
+                mean_sq = sum_sq[z] / num_pixels[z];
+                rms[z] = sqrtf(mean_sq);
```
veggiesaurus: stats are all done using `double` precision, so this should be `sqrt` in order to give the same results as casacore

style.out
```diff
@@ -0,0 +1,860 @@
+Enabled checks:
```
veggiesaurus: should this file be here?

confluence: Oh, no. My bad.

confluence:
I misunderstood your comment about the initial approach to take for deciding to use the swizzled dataset or not -- I thought we were leaving it out of the PR entirely. Should I just do a comparison between the predicted number of reads, regardless of read size?

PR #241, opened 2019-05-31
Task cancellation for regional spectral statistics

markccchiang:
This cancellation method is the same as the cursor's z profile cancellation. During the regional spectral process, it checks every 100 channels whether the region states changed or the image file is closed. If there are no other better ways to stop running tasks, this can be a backup solution.


veggiesaurus:
I think that, in general, this is the right way to handle currently executing region calculations, and to use the task context cancellations to prevent pending region calculations from running. There are a few things to point out though:

1) This only interrupts calculations if the _region_ changes. If the region requirements change (e.g. user selects a different region in a spectral profile), this will not be handled. There may be a simpler way to check this and the region changes, by assigning a timestamp using an int (atomic by nature), and then updating the timestamp whenever the region changes or the region requirements change. Possibly this is an over-simplification...
2) This PR conflicts with @confluence's work on using swizzled datasets for regions (#239). When using swizzled data, the data is read in slice by slice aligned to the YZ-plane, iterating over X (whereas the conventional method reads aligned to the XY-plane, iterating over Z). So this will need to be taken into acount when handling swizzled datasets. It's important to make sure that this is done correctly, so as not to reverse the speedup from using swizzled datasets (can be up to 50x faster, depending on image and region size)
3) A fixed step size of 100 channels is not appropriate, as performance will depend entirely on (a) image size, (b) region size and (c) I/O performance. I would suggest that we rather define some target time interval (50 - 100 ms perhaps?) and start from a small step size, increasing or decreasing it automatically, depending on whether the last iteration took longer or shorter than the target time interval.

markccchiang:
Thanks for the comments! I will improve it according to the suggestions and solve the merged conflicts with the new updated dev branch.

markccchiang:
The improvement of the cancellation for region spectral statistics is done.  It includes a) allow interrupting the calculation if region requirements changed, including change(switch) the region ID. b) let the step size of channels for each calculation can be adjusted automatically and set the time interval between two steps is about 75 ms. Note that this task cancellation is not available for swizzled datasets since it applies a different way to access the data and calculate statistics. I will try to find a good way to handle its cancellation.

Frame.cc
```diff
@@ -810,18 +812,19 @@ bool Frame::FillSpectralProfileData(int region_id, CARTA::SpectralProfileData& p
                     if (stats_values) {
                         region->FillSpectralProfileData(profile_data, i, *stats_values);
                     } else {
-                        casacore::SubImage<float> sub_image;
+                        std::vector<std::vector<double>> stats_values;
                         std::unique_lock<std::mutex> guard(_image_mutex);
-                        GetRegionSubImage(region_id, sub_image, profile_stokes);
-                        region->FillSpectralProfileData(profile_data, i, sub_image);
```
confluence: I believe that this `Region` method is now unused, and should be removed. This will leave two very similar methods for filling stats from data which has been passed in (one which accepts a map and one which accepts a vector), and these should probably be merged into one -- but that is not a trivial change and can be done in a separate PR.

Frame.cc
```diff
@@ -1057,3 +1063,73 @@ bool Frame::GetSpectralData(std::vector<float>& data, casacore::SubImage<float>&
     }
     return data_ok;
 }
+
+bool Frame::GetRegionalSpectralData(std::vector<std::vector<double>>& stats_values, int region_id, int profile_index,
```
confluence: For consistency I would call this `GetRegionSpectralData`.

Frame.cc
```diff
@@ -635,7 +641,7 @@ bool Frame::FillRegionHistogramData(int region_id, CARTA::RegionHistogramData* h
                     } else {
                         std::unique_lock<std::mutex> guard(_image_mutex);
                         casacore::SubImage<float> sub_image;
-                        GetRegionSubImage(region_id, sub_image, curr_stokes, config_channel);
+                        GetRegionSubImage(region_id, sub_image, curr_stokes, {config_channel, config_channel});
```
confluence: Should we not use the constructor with a single from/to value to avoid this repetition? We frequently use this kind of range.

Frame.h
```diff
@@ -118,24 +125,24 @@ class Frame {
     // get lattice slicer for profiles: get full axis if set to -1, else single value for that axis
     void GetImageSlicer(casacore::Slicer& image_slicer, int x, int y, int channel, int stokes);
     // make Lattice sublattice from Region given channel and stokes
-    bool GetRegionSubImage(int region_id, casacore::SubImage<float>& sub_image, int stokes, int channel = ALL_CHANNELS);
+    bool GetRegionSubImage(int region_id, casacore::SubImage<float>& sub_image, int stokes, ChannelRange = {0, ALL_CHANNELS});
```
confluence: Similarly, should we not use the default range constructor here?

Region/Region.h
```diff
@@ -112,8 +117,8 @@ class Region {
     casacore::WCRegion* MakePolygonRegion(const std::vector<CARTA::Point>& points);
 
     // Extend xy region to make LCRegion
-    bool MakeExtensionBox(casacore::WCBox& extend_box, int stokes, int channel = ALL_CHANNELS); // for extended region
-    casacore::WCRegion* MakeExtendedRegion(int stokes, int channel = ALL_CHANNELS);             // x/y region extended chan/stokes
+    bool MakeExtensionBox(casacore::WCBox& extend_box, int stokes, ChannelRange channel_range = {0, ALL_CHANNELS}); // for extended region
+    casacore::WCRegion* MakeExtendedRegion(int stokes, ChannelRange chanel_range = {0, ALL_CHANNELS});              // x/y region extended chan/stokes
```
confluence: Typo: `chanel_range`

markccchiang:
@confluence I have pushed the commits regarding your comments. Please review them. Thank you!

Frame.cc
```diff
@@ -633,9 +639,10 @@ bool Frame::FillRegionHistogramData(int region_id, CARTA::RegionHistogramData* h
                             region->CalcHistogram(config_channel, curr_stokes, num_bins, min_val, max_val, data, *new_histogram);
                         }
                     } else {
+                        ChannelRange single_channel(config_channel);
```
confluence: It seems unnecessarily verbose to create named variables for these -- I was suggesting anonymous objects, so `ChannelRange(some_channel_here)` or just `ChannelRange()`. But I don't feel that strongly about it, and it's only a few places in the code, so if you prefer this style I don't mind if you keep it.

markccchiang: Agree. It looks cleaner using anonymous objects. I will change them and add a note on how to use ChannelRange structure.

Frame.cc
```diff
@@ -1061,3 +1075,73 @@ bool Frame::GetSpectralData(std::vector<float>& data, casacore::SubImage<float>&
...
+    RegionState region_state = region->GetRegionState();
+    // get statistical profile data
+    int start = 0;
+    int count, end;
+    casacore::SubImage<float> sub_image;
+    while (start < profile_size) {
+        // start the timer
+        auto tStart = std::chrono::high_resolution_clock::now();
+        // check if frontend queries changed, if so, terminate this loop process
+        if (!_connected) {
```
confluence: To make it easier to reuse this sequence of checks from the swizzled calculation function on the HDF5 loader, maybe refactor it into its own public function and pass the frame object as a parameter to the loader function so that you can call the function on the frame from inside the loader.

Inside the loader function I guess you would check this before reading each X slice from the swizzled dataset.

Frame.cc
```diff
@@ -1153,3 +1186,188 @@ bool Frame::GetSpectralData(std::vector<float>& data, casacore::SubImage<float>&
...
+        }
+        if (delta_channels > profile_size) {
+            delta_channels = profile_size;
+        }
+    }
+    stats_values = std::move(results);
+    return true;
+}
+
+bool Frame::CalcRegionSpectralData(std::map<CARTA::StatsType, std::vector<double>>& stats_values, int stokes, int region_id,
```
confluence: I strongly feel that this code should remain inside the HDF5 loader -- we have consistently factored format-specific code out of the frame object. The frame object should not be aware of HDF5-specific implementation details -- it should only call loader functions to check if the loader has a particular feature or can produce some dataset.

veggiesaurus: there might be a need for some of the logic and final stats combining to be placed in the Frame, especially since we want to return _partial_ results to the frontend. But I don't think just moving the entire function to `Frame` is the right solution.

veggiesaurus: @markccchiang please keep in mind that the end goal is to also send partial results to the frontend at regular intervals, whether those partial results are from the original dataset or the swizzled data.

markccchiang: I see. I will move it back to the HDF5 loader. Then the loader will be responsible for determining whether to interrupt the stats calculation.

veggiesaurus: @confluence can you give some thought as to how we can return partial results from the loader while it is still in progress? The messages to the frontend will get sent from `Session`. Not sure if passing a callback into the frame and loader functions would be the most elegant solution, but it might work

confluence: I think a callback is the most sensible solution -- it seems like a good place to start.

veggiesaurus: I wonder if a callback could also be used to allow the loader to check whether the region is still valid

markccchiang:
Note that the latest commission is a temporary version that applies interruption checks for swizzled data calculation. I will move the stats calculation back to the HDF5 loader (Frame should not see swizzled data) and the loader will be responsible for interruption checks according to @confluence's comment.

Frame.cc
markccchiang: BTW, sending partial results from the original dataset is no problem. But it looks unlikely for swizzled data?

confluence: Well, they'd be partial in a different direction. :) You'd have to recalculate the mean, RMS, etc., after every pass, and they'd get more accurate as you read in more slices, until eventually you'd have the exact values.

ImageData/Hdf5Loader.h
```diff
@@ -43,6 +47,9 @@ class Hdf5Loader : public FileLoader {
     casacore::ArrayBase* GetStatsData(FileInfo::Data ds) override;
 
     casacore::Lattice<float>* LoadSwizzledData(FileInfo::Data ds);
+    // current region states
+    std::unordered_map<int, RegionState> _region_states;
```
confluence: I would prefer it if you passed in references to the original properties on the frame object instead of duplicating them here (and changing the state in two different places from the frame). The tidiest way to do this may be to pass in a reference to the entire frame object, and calling a (new) public getter on the frame to do this check. You could call the same getter from the non-swizzled calculation. Rob has some concerns that values may be cached if the compiler uses certain optimisations, so you may want to test this with different build types and possibly use the `volatile` keyword for the properties.

ImageData/Hdf5Loader.h
```diff
@@ -362,21 +370,21 @@ std::map<CARTA::StatsType, std::vector<double>>* Hdf5Loader::GetRegionSpectralDa
...
+        RegionState region_state = _region_states[region_id];
+
         // Load each X slice of the swizzled region bounding box and update Z stats incrementally
         for (size_t x = 0; x < num_x; x++) {
-            casacore::Slicer slicer;
-            if (_num_dims == 4) {
-                slicer = casacore::Slicer(IPos(4, 0, y_min, x + x_min, stokes), IPos(4, num_z, num_y, 1, 1));
-            } else if (_num_dims == 3) {
-                slicer = casacore::Slicer(IPos(3, 0, y_min, x + x_min), IPos(3, num_z, num_y, 1));
+            if (!_connected) {
```
confluence: See comment next to these properties. This could be a function on the frame object which uses the frame's original state variables.

Frame.cc
confluence: You could call the same function to check state from here and from the loader function.

Frame.cc
```diff
@@ -1153,3 +1204,95 @@ bool Frame::GetSpectralData(std::vector<float>& data, casacore::SubImage<float>&
...
+            delta_channels = profile_size;
+        }
+        // send partial result by the callback function
+        cb(results, progress);
+    }
+    stats_values = std::move(results);
+    return true;
+}
+
+bool Frame::CanUseSiwzzledData(const casacore::ArrayLattice<casacore::Bool>* mask) {
```
confluence: Typo in the name. I would also rename this to something like `UseImageRegionStats` -- this function specifically checks if the swizzled dataset can and should be used for this specific application, not if it exists in general.

confluence:
Sorry, Mark, I made my review comments before I saw your latest changes. I will look at them now. :)

ImageData/FileLoader.h
```diff
@@ -136,10 +136,16 @@ class FileLoader {
     // Return a casacore image type representing the data stored in the
     // specified HDU/group/table/etc.
     virtual ImageRef LoadData(FileInfo::Data ds) = 0;
-    virtual bool GetCursorSpectralData(std::vector<float>& data, int stokes, int cursor_x, int cursor_y);
-    virtual std::map<CARTA::StatsType, std::vector<double>>* GetRegionSpectralData(
-        int stokes, int region_id, const casacore::ArrayLattice<casacore::Bool>* mask, IPos origin);
+    virtual bool GetCursorSpectralData(std::vector<float>& data, int stokes, int cursor_x, int count_x, int cursor_y, int count_y);
+    // check if one can apply swizzled data under such image format and region condition
+    virtual bool CanUseSiwzzledData(const casacore::ArrayLattice<casacore::Bool>* mask);
```
confluence: Updated comment: typo is still a typo; but if this is on the loader now I would suggest a different name -- `HasRegionSpectralData`.

confluence: Actually, `UseRegionSpectralData` is probably better -- it may *have* the data, but this function determines whether to *use* it or not.

ImageData/Hdf5Loader.h
markccchiang: Now I move the region and cursor states variables from the Frame to FileLoader. The states check functions are also moved to the FileLoader. So they can be called from the Frame or the FileLoader itself. The states variables are set as  `volatile` now, including `RegionState`, `RegionConfig`, `CursorXy`, and `_connected`.

confluence: It doesn't make any sense for these variables to be on the loader. They have nothing to do with loading images. I strongly suggest that you leave them on the frame object, where they were before, and pass some kind of reference into the loader calculation function so that they can be queried from that function. I still believe that the cleanest way to do this is to add a public getter to the frame object and to pass in a reference to the frame object to the calculation function so that that getter can be called.

confluence:
Apart from the location of the state variables and the functions used to access them (see the reply in the comment thread), I think this all looks fine.

ImageData/Hdf5Loader.h
```diff
@@ -327,6 +343,8 @@ std::map<CARTA::StatsType, std::vector<double>>* Hdf5Loader::GetRegionSpectralDa
             std::piecewise_construct, std::forward_as_tuple(region_id, stokes), std::forward_as_tuple(origin, mask->shape(), num_z));
         recalculate = true;
     } else if (!_region_stats[region_stats_id].IsValid(origin, mask->shape())) { // region stats expired
```
confluence: One more thing: because we are now updating these stats in the loader state incrementally, and may abort the calculation before it's complete, we need to add a flag to the state which indicates that the calculation completed -- so that if we try to calculate stats for the same region later we don't think that the cached stats are valid if they were not fully calculated.

confluence: If we want to be really clever, we can save the actual position to which we previously calculated, use a separate check, and resume from there instead of recalculating from scratch.

confluence:
The region state looks fine -- now we only need the check that the cached loader calculations were completed.

markccchiang:
I think that is all. @confluence could you please check again? Thank you!

veggiesaurus:
@markccchiang this is mostly looking really good. I managed to get it to compile on gcc by just removing the `volatile` keywords on the structs (not sure if they're strictly needed if passing by reference, but please check this, or see if there's another way around).

Two remaining issues:
- The partial updates are working nicely for both FITS and HDF5, but they're being sent every 50 ms. It's ok to check for cancellation every 50 ms, but we shouldn't send partial updates every 50 ms. I'd suggest sending every 1000 ms, or even every 2000 ms. @kswang1029 is that fast enough?
- Second not really an issue, but a question: should we be passing `std::function` arguments by value or as a `const` reference (to avoid the copy constructor)? I suspect we should pass by reference.

kswang1029:
Updating profiles at 1000ms or 2000 ms should be good. I prefer 1000ms.

veggiesaurus:
@markccchiang have sent a PR (into this branch) for you to review, which updates the partial profiles at 1000 ms.

kswang1029:
with b89eae6 and hdf5 cube, region spectral profiles may be broken if there are NaN pixels. Example is shown below:
Screen Shot 2019-07-15 at 19 40 57

kswang1029:
with b89eae6 and fits cube, if a region contains all NaN, requesting region spectral profile will cause the backend crash.

kswang1029:
with 83c1a47, the backend hangs if we define a region which is completely outside the image. To reproduce:
1) make a region inside the image
2) request region spectral profile
3) seeing it updating the profile
4) before it completes, move the region to be completely outside the image
5) we should see the profile stops updating
6) move the region back inside the image, now it is not responding.
7) changing channels is not working, neither. The backend hangs.

markccchiang:
It seems the hanging problem for the region spectral profile is caused by the casacore. If you move the region completely outside the image, somewhere in the casacore would be stuck. I can reproduce it on the dev branch too. Looking for the way to avoid this phenomenon.

kswang1029:
@markccchiang yes I confirmed the hanging backend issue is also in dev. Lets have a separate issue for that.

markccchiang:
The broken spectral profile for the HDF5 cube can also be reproduced on the dev branch. It occurs while the region contains most of NaN pixels. Like the following plot:
螢幕快照 2019-07-16 下午5 11 46
I suspect there is a problem in the stats calculation while dealing with the NaN pixels. Will look into the code in detail.

PR #242, opened 2019-05-31
Removed use of _waiting_task which was causing crash

robsimmonds:
Fixed some flow control logic.

### DIFF
--- a/AnimationObject.h
+++ b/AnimationObject.h
@@ -60,7 +60,7 @@ public:
         _file_open = true;
         _waiting_flow_event = false;
         _waiting_task = nullptr;
-	_last_flow_frame = start_frame;
+        _last_flow_frame = start_frame;
     }
     tbb::task* waiting_task_ptr() {
         return _waiting_task;
--- a/AnimationObject.h
+++ b/AnimationObject.h
@@ -3,6 +3,7 @@
 
 #include <tbb/task.h>
 #include <chrono>
+#include <iostream>
 
 #include <carta-protobuf/animation.pb.h>
 #include <carta-protobuf/set_image_channels.pb.h>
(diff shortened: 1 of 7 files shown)


kswang1029:
tested with the mix page and the commit 1d61da9:

With supermosaic.10.fits at desired 15fps, at home with slower internet, the actual fps is smaller.
When I click stop, the animation does not stop immediately but keep showing some/several channels then jump back to the channel which I clicked stop. It feels like it is showing buffered channels until running out.
animator

kswang1029:
another issue is when animation is ongoing, if I clicked "play" again, the channel will jump back to a certain former channel then keep going.
animator2

veggiesaurus:
> tested with the mix page and the commit [1d61da9](https://github.com/CARTAvis/carta-backend/commit/1d61da90affa82982b83d58f23dae32ba34d26d7):
> 
> With supermosaic.10.fits at desired 15fps, at home with slower internet, the actual fps is smaller.
> When I click stop, the animation does not stop immediately but keep showing some/several channels then jump back to the channel which I clicked stop. It feels like it is showing buffer channels until running out.
> animator

I think this will be fixed with tiled rendering. We want it to stay on whichever channel you click stop on. At the moment there's no way for the front-end to know which channels are from an animation that has been stopped.

veggiesaurus:
> another issue is when animation is ongoing, if I clicked "play" again, the channel will jump back to a certain former channel then keep going.

Front-end bug. I'll fix that

robsimmonds:
@kswang1029 have discussed the too many frames issue with Angus. We understand this issue. It is due to the way we are calculating the flow control window not doing what we want on a slow network. There is a simple fix for it, but we will not apply that now since Angus is travelling. We can get that done later this week. It needs a more sophisticated fixed based on the sliding-window idea from TCP, but that will need more work and a lot of testing. So for now we have agreed to merge it into dev as is and accept that there is some strange behaviour when using very high frame rates with slow networks.

kswang1029:
ok. please let me know if I can help on testing.

PR #249, opened 2019-06-26
Take stokes into account when caching spectral profile stats on the HDF5 loader.

### LINKED ISSUE #247: Spectral profile test with various stokes fails: Hdf5Loader returns same stats
https://github.com/CARTAvis/carta-backend/blob/3253c1b94fc4e7c0e6252ee220420b838d78cb58/Session.cc#L788
About updating the profile data, the backend will check if the frontend has already set a cursor. In case of region operation, is this a must to check? At current ICD backend test, the tester needs to set both region and cursor so that the profile data updates.

confluence:
This fixes #247 by adding the stokes value to the key used in the map of cached stats on the HDF5 loader. It also fixes an unrelated bug: when the stats were recalculated after a change in the region, the region properties were not updated on the stats object, so we kept recalculating unnecessarily.

### DIFF
--- a/ImageData/FileLoader.h
+++ b/ImageData/FileLoader.h
@@ -24,6 +24,19 @@ struct ImageStats {
     bool valid;
 };
 
+struct RegionStatsId {
+    int region_id;
+    int stokes;
+
+    RegionStatsId() {}
+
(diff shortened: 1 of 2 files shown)


pford:
Verified this branch passes the REGION_SPECTRAL_PROFILE_STOKES.test.ts test.

PR #250, opened 2019-06-27
Angus/tiled rendering

veggiesaurus:
Note: Companion PR of https://github.com/CARTAvis/carta-frontend/pull/346. This should only be merged once both PRs are approved.

Initial implementation of tiled rendering. Tiled rendering allows us to render much larger images more efficiently (see [here](https://docs.google.com/document/d/1w5TC7hZk-0Nrsg8BarAUF71MYtid_-g6/edit#heading=h.gjdgxs) for description). It also simplifies issues around `SET_IMAGE_VIEW` considerably. 

Tiled rendering is used when not animating channels / stokes. See [here](https://docs.google.com/document/d/1ryrqjW6f9VFtLpSHB6UiIHNxXb18XCzJX90HzZQY5gc/edit) for details.

There are some known issues, which can be fixed separately after the merge. Those relating to the backend are:
- The ICD documentation needs to be updated
- Currently, all requested tiles are processed without checking to see if they are still required. This can be improved with a proper queuing system.

Additional issues are noted on the frontend PR


veggiesaurus:
note CI tests will fail as there are significant changes to the ICD

Frame.cc
```diff
@@ -538,20 +549,20 @@ bool Frame::GetRasterData(std::vector<float>& image_data, CARTA::ImageBounds& bo
     bool write_lock(false);
     tbb::queuing_rw_mutex::scoped_lock lock(_cache_mutex, write_lock);
 
-    if (mean_filter) {
+    if (mean_filter && mip > 1) {
         // Perform down-sampling by calculating the mean for each MIPxMIP block
-        auto range = tbb::blocked_range2d<size_t>(0, num_rows_region, 0, row_length_region);
```
veggiesaurus: this is not strictly related to tiled rendering, but splitting the down-sampling along one axis only helps to avoid cache misses and speeds up down-sampling around 10%.

veggiesaurus:
> Clearly you have run the style/format checks, which resulted in changes unrelated to this PR. We need everyone to do this before submitting a PR.

sorry, was wondering whether to keep those as a separate PR, but I suppose as long as we're consistent in running those checks, it shouldn't be too much of an issue.

Issue #257, opened 2019-07-10
inconsistent full-frame statistics


kswang1029:
with the current dev+dev branches as of July 10

While testing region statistics, I spotted a weird issue in statistics widget. In some cases, the full image statistics are not consistent with the region statistics if the region (ellipse or rectangle) covers the entire image. Examples are shown below:

consistent:
![Screen Shot 2019-07-10 at 9 19 35 AM](https://user-images.githubusercontent.com/20819712/60933539-3a481300-a2f5-11e9-9c48-459d1daeb8f4.png)
![Screen Shot 2019-07-10 at 9 31 04 AM](https://user-images.githubusercontent.com/20819712/60933577-79766400-a2f5-11e9-9ba1-078e97ebf738.png)

inconsistent:
![Screen Shot 2019-07-10 at 9 25 28 AM](https://user-images.githubusercontent.com/20819712/60933585-81ce9f00-a2f5-11e9-86c6-42f835959503.png)
![Screen Shot 2019-07-10 at 9 24 19 AM](https://user-images.githubusercontent.com/20819712/60933600-8f842480-a2f5-11e9-884e-78bd9e938f73.png)

Same h_m51 image with v1.1 does not have this issue:
![Screen Shot 2019-07-10 at 9 33 03 AM](https://user-images.githubusercontent.com/20819712/60933654-cd814880-a2f5-11e9-85e8-2a68bd6662fa.png)

confluence:
It looks like the number of pixels is incorrect in one of the calculations. I think all the values that differ depend on this value (number of pixels, mean, rms, std dev).

confluence:
I have encountered the same discrepancy in my branch, where I am adding the use of cached HDF5 stats for full channels. The HDF5 stats disagree with the calculated image stats, but agree with the region stats (except for the standard deviation, which may be a precision issue). If you look at `miniCube_5px_5px_IQU_4ch.fits`, it's clear that the image calculation results in the incorrect number of pixels (5x5 with no NaNs = 25, not 21).

I suspect that some kind of rounding error is happening in the creation of the image region in `Frame::SetImageRegion` (this may explain why it affects some images and not others; it would depend on the X and Y dimensions), but I can't see how this would result in the number of pixels being 21 and not 20.

confluence:
Here's the boolean mask of the minicube full image region: 

```
1 1 1 1 1 
0 1 1 1 1 
0 1 1 1 1 
0 1 1 1 1 
[...]

PR #261, opened 2019-07-15
Use cached HDF5 image statistics

### LINKED ISSUE #253: HDF5 schema update: use cached SUM and SUM_SQ stats instead of MEAN
We have changed the HDF5 schema to store the sum and sum of squares instead of the mean, as these are more useful intermediate components which can be used to calculate other statistics (https://github.com/idia-astro/cpp_hdf5converter/issues/3). We should switch to using these.

Because the change should be simple, I intend only to deprecate reading the mean for now, so that we don't have to worry about updating the backend and regenerating all HDF5 files at exactly the same time. We can entirely remove support for the mean at a later stage, once all files have been converted. To simplify this and future schema changes it may be a good idea to add functionality to the converter to rewrite the secondary datasets in an existing file (https://github.com/idia-astro/cpp_hdf5converter/issues/8).

confluence:
This fixes #253 and also has an unrelated fix for #257.


veggiesaurus:
I'll have a closer look at the code tomorrow, but here are some initial performance numbers that should make @arusstaylor happy:

* playing back FITS file with stats widget set to "Image" region: 1.5 fps
* playing back HDF5 file (**without** this implementation) with stats widget set to "Image" region: 4.2 fps (HDF5 file using cached channel histograms)
* playing back HDF5 file (**with** this implementation) with stats widget set to "Image" region: 11 fps (HDF5 file using cached channel histograms _and_ cached channel `SUM`, `SUMSQ`)

The CPU usage is also reduced from 200% to 175%, even though the frame rate more than doubled. 

(all tests with supermosaic.10 and requested frame rate set to 15)

Frame.cc
```diff
@@ -1045,9 +1059,12 @@ bool Frame::GetImageHistogram(int channel, int stokes, int num_bins, CARTA::Hist
         int image_num_bins(current_stats.histogram_bins.size());
 
         if ((num_bins == AUTO_BIN_SIZE) || (num_bins == image_num_bins)) {
+            double min(current_stats.basic_stats[CARTA::StatsType::Min]);
```
veggiesaurus: I am a little uncomfortable with using `min` and `max` as variable names, considering so many other namespaces will do so. All it takes is for us to include a header file that adds `using namespace std` for this code to break. Could we call them `min_val` and `max_val`?

ImageData/Hdf5Loader.h
```diff
@@ -399,6 +399,36 @@ bool Hdf5Loader::GetRegionSpectralData(int stokes, int region_id, const casacore
         auto t_start = std::chrono::high_resolution_clock::now();
         auto t_latest = t_start;
 
+        // Lambda to calculate mean, sigma and RMS
+        auto calculate_means = [&]() {
```
veggiesaurus: not sure `calculate_means` is an accurate label, since it's doing mean, rms and sigma. Perhaps `calculate_stats` or something else indicating that we're doing more than calculating the mean?

Frame.cc
confluence: Done.

ImageData/Hdf5Loader.h
confluence: Also done.

Issue #265, opened 2019-07-16
Assertion warning in Region.cc


pford:
After PR #241 merge:

Region.cc:425:65: warning: comparison of
      constant 0 with expression of type 'bool' is always true
`assert(max_chan >= min_chan && all_channels >= max_chan >= 0 && all_channels >= min_chan >= 0);`

I also suggest ( ) around each comparison for clarity and readability, e.g. `((max_chan >= min_chan) &&` ...

pford:
Fixed.

PR #267, opened 2019-07-17
Modify channel range assertion

markccchiang:
This code modification is according to @pford's suggestion (issue #265).

### DIFF
--- a/Region/Region.cc
+++ b/Region/Region.cc
@@ -416,7 +416,10 @@ bool Region::MakeExtensionBox(casacore::WCBox& extend_box, int stokes, ChannelRa
     try {
         double min_chan(channel_range.from), max_chan(channel_range.to);
         double all_channels = _image_shape(_spectral_axis);
-        if (channel_range.to == ALL_CHANNELS) { // extend to nchan
+        if (channel_range.from == ALL_CHANNELS) {
+            min_chan = 0;
+        }
+        if (channel_range.to == ALL_CHANNELS) {
             max_chan = all_channels - 1;
         }
         assert((max_chan >= min_chan) && (all_channels > max_chan));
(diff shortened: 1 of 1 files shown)


Region/Region.cc
```diff
@@ -422,7 +422,11 @@ bool Region::MakeExtensionBox(casacore::WCBox& extend_box, int stokes, ChannelRa
         if (channel_range.to == ALL_CHANNELS) { // extend to nchan
             max_chan = all_channels;
         }
-        assert(max_chan >= min_chan && all_channels >= max_chan >= 0 && all_channels >= min_chan >= 0);
+        assert((max_chan >= min_chan) && (all_channels >= max_chan));
```
pford: I am confused by the code block before this assertion.  If the image has 64 channels (all_channels=64) and the channel_range is (ALL_CHANNELS, ALL_CHANNELS), shouldn't min_chan=0 and max_chan=all_channels-1?  It is setting the range to (64, 64) which would be invalid.  But it looks like the ALL_CHANNELS range is never used, Frame sets the ChannelRange argument to GetRegion to a single channel, a range of channels, or ChannelRange() (which does the right thing).

pford: therefore (all_channels > max_chan), not >=

markccchiang:
I think I should also correct `max_chan = all_channels` as `max_chan = all_channels - 1`.

Region/Region.cc
markccchiang: Sorry for the confusion! For `ChannelRange()` sets the channel range (0, ALL_CHANNELS - 1). For ChannelRange(ALL_CHANNELS, ALL_CHANNELS) also sets the channel range (0, ALL_CHANNELS - 1). Is it looks reasonable?

PR #268, opened 2019-07-17
Fix a bug for HDF5 region definition

markccchiang:
I think this is the reason why @kswang1029 saw the broken spectral profile for a range in the HDF5 image.


confluence:
This fix has been merged into `dev` in my most recent PR.

PR #271, opened 2019-07-17
Confluence/remove ordered maps

confluence:
We don't need any of these maps to be ordered. Using unordered maps should give us slightly better performance.


confluence:
We've profiled this and concluded that it actually makes performance slightly worse -- so I'm closing this PR and deleting the branch. :)

Issue #252, opened 2019-07-08
"Current" Stokes profiles aren't updated when changing channels if there are also fixed requirements


veggiesaurus:
See GIF below. When the only spectral requirements is the current Stokes profile, the profile is updated every time the channel is sent. However, if another spectral profile widget exists with a _fixed_ requirement, then no new profiles are sent by the backend. I've confirmed that the requirements are being sent correctly by the frontend (`z` for the current profile, and `Iz` for the fixed Stokes I profile), and that no profile data is sent to the frontend when switching channels. 

![stokes_profile_bug](https://user-images.githubusercontent.com/592504/60802688-c6cdc680-a179-11e9-8132-602c2fa1910e.gif)

veggiesaurus:
@pford could you look into this?

PR #272, opened 2019-07-18
Update data streams correctly

pford:
Issue #252

When requirements, a region, channel, or stokes change, the backend sends new data streams in response.  This PR ensures that the correct data streams are updated.  In certain cases, data is not sent: (1) when only the channel changes, a cube histogram is not sent and spectral profiles are not sent; (2) when the stokes changes, profiles with a fixed stokes are not sent.

Also sends the spatial profile message with no profiles when not in requirements to update the cursor value.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -673,8 +673,9 @@ bool Frame::GetRasterTileData(std::vector<float>& tile_data, const Tile& tile, i
 // ****************************************************
 // Region histograms, profiles, stats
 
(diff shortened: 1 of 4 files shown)


pford:
I neglected to test 2 regions, will do tomorrow first thing.

veggiesaurus:
> I neglected to test 2 regions, will do tomorrow first thing.

seems to work fine with 2 regions, but please do test as well

PR #278, opened 2019-07-22
Modify the cursor spectral profile response

markccchiang:
The backend will send the partial spectral profile of cursor every 500 ms.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -1189,7 +1189,8 @@ bool Frame::GetSubImageXy(casacore::SubImage<float>& sub_image, CursorXy& cursor
 }
 
 bool Frame::GetCursorSpectralData(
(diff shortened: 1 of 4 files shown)


Frame.cc
```diff
@@ -1231,9 +1235,11 @@ bool Frame::GetCursorSpectralData(
                 if (delta_channels > profile_size) {
                     delta_channels = profile_size;
                 }
-                progress = (float)start(_spectral_axis) / profile_size;
-                // send partial result by the callback function
-                cb(data, progress);
+                if (dt_partial_profile > TARGET_PARTIAL_TIME / 2 || progress >= 1.0f) {
```
veggiesaurus: can we rather define a separate constant for this?

Issue #273, opened 2019-07-18
unexpected profile update with multiple spectral profile widgets and multiple regions


kswang1029:
with dev+dev July 18, 2019

Spectral profiles are updated unexpectedly when switching to different regions.

To reproduce:
1. create three regions on the image
2. create three spectral profile widgets and 
     2.1 assign widget 1 to region 1 with "mean"
     2.2 assign widget 2 to region 2 with "StdDev"
     2.3 assign widget 3 to region 3 with "min"
3. once all profiles are delivered, switch to region 2 in widget 1
==> now the profiles in widget 1 and widget 2 are updated simultaneously (this is probably okay)
4. once all profiles are delivered, switch to region 1 in widget 1
==> now widget 1 updates first, then widget 2 updates (widget 2 seems updates unnecessarily) 

also see the demo
![regionSpecProUpdate](https://user-images.githubusercontent.com/20819712/61425990-6265f080-a94b-11e9-8be3-e56a2f979862.gif)

veggiesaurus:
@pford can you investigate? A diff on the server side between existing and incoming requirements seems to be needed

pford:
Currently, the statistics in the REGION_STATS_DATA are cached, but not the spectral profiles.  When a region's requirements are set, no diffs are made and all profile data is sent.  Will fix this.

veggiesaurus:
if they're cached, then it should be sufficient to just return all, instead of a diff, I guess (if the diff process is too messy)

pford:
I started with the spatial profiles and have a problem, but I think it is in the frontend.  If you have one profile displayed (e.g. "x") then you add the other profile "y", backend only sends y-profile (the diff in requirements) but then x-profile goes blank.  I pushed the branch for issue 273 (spatial profiles only, so far).

veggiesaurus:
Yes, currently the frontend still expects you to send all the spatial profiles instead of the diff (because they were previously all being sent). I'll sort that out tomorrow. For now, you might have to use the dev console to ensure that profiles are being delivered properly

veggiesaurus:
@kswang1029 can you verify that this is fixed in the latest `dev`? (`398fc62`)

pford:
This broke the partial spectral profiles (backend thinks it is sent after first update), fixing in branch for issue #263.

kswang1029:
Issue solved! Nice. 👍

Issue #263, opened 2019-07-16
Point region get no z-profile


zarda:
Confirmed on frontend [branch](https://github.com/CARTAvis/carta-frontend/tree/hj/%23239_Support_point_region) and ICD test [branch](https://github.com/CARTAvis/carta-backend-ICD-test/tree/hj/issue%2315_region_spectral_point).

Reproduce : After setting up a point region 2 and any region 1, changing from another region 1 to point region 2 should SetSpectralRequirements to let region 1 be null, SetSpectralRequirements to call point region 2, and then backend send SpectralProfileData but with empty z-profile. The ICD test [REGION_SPECTRAL_PROFILE.test.ts](https://github.com/CARTAvis/carta-backend-ICD-test/blob/hj/issue%2315_region_spectral_point/src/test/REGION_SPECTRAL_PROFILE.test.ts) asserts this issue.

veggiesaurus:
@pford can you look into this? We can't test the frontend point regions feature until this is fixed

pford:
I added a note in the weekly meeting page: backend dev, backend master, and ICD-test are all on different protobuf commits.  I cannot get **any** tests to pass in REGION_SPECTRAL_PROFILE.test.ts.

pford:
Tested with frontend hj/#239_Support_point_region branch.  The backend was comparing the Point region's subimage xy point with the cursor position instead of its own region control point.  Since they differed, it exited the spectral profile.

pford:
Side note in issue #273, backend was marking profile as sent after first partial profile, so it did not send profile updates.  Fixed in this branch pam/point_region, profile sent = false until progress is complete.

pford:
There are still issues with interrupting spectral profiles but they will be dealt with separately.  Ready to test that a point region gets a spectral profile.

zarda:
Tested well on both frontend and testing.

PR #279, opened 2019-07-22
Fix point region (non-cursor)  and partial spectral profiles

pford:
Issue #263 and fix bug introduced for issue #273.

Modified cursor spectral profile code for point region, so that it compares subimage cursor with Point region control points instead of Frame's _cursor_xy.  Also Region does not set profile sent = true until progress is complete.


kswang1029:
Partial cursor profile is still being delivered when region spectral profile has been requested. We will have to wait until cursor profile is completed then start to see region spectral profile updates progressively.

Frame.cc
```diff
@@ -1234,17 +1236,30 @@ bool Frame::GetCursorSpectralData(
...
                 // start the timer
                 auto t_start = std::chrono::high_resolution_clock::now();
-                // check if frontend's requirements changed
-                if (Interrupt(cursor_xy)) {
-                    return false;
+                // check if region point changed from subimage point
+                if ((region_id == CURSOR_REGION_ID) && (Interrupt(_cursor_xy, subimage_cursor))) {
+                    return false; // cursor moved
                 }
+                if (region_id > CURSOR_REGION_ID) {
```
markccchiang: It seems the `region_id` is always equal to 0. Because we have already set to execute the function `GetPointSpectralData()` only when it is a point region (on line 923 `if (region->IsPoint()) {..}`).

pford: A point region can be a cursor (region id 0) or a fixed point region (SET_REGION with type Point and id -1 which gets set to the next available region id).  Point regions are being implemented for release 1.2.

pford:
I thought we already had code to cancel a spectral profile if the requirements changed.  I will look into it.

Frame.cc
markccchiang: I see. So @kswang1029's problem is due to the lack of interruption condition that users switch between different regions or cursor. I will do it and send another PR. You can merge this PR first 
since it is irrelevant to the subject.

pford: @kswang1029's problem is that the cursor spectral profile doesn't know that its requirements have been changed (to profiles=[], which should cancel).  In addition the cursor/point position, the number of profiles should be checked, and the profile coordinate (stokes) should be checked (for example if user changes I -> Q you do not need to send stokes I any more).  For non-point regions, the region position, stokes, and requested stats should be checked.

pford:
We may have to send profiles for multiple regions (cursor and point) if the user opens multiple widgets, so we can't just check region_id.  The cursor profile should be cancelled because the requirements have been set to empty profiles.

Issue #280, opened 2019-07-23
backend crash when requesting region spectral profile from an HDF5 cube


kswang1029:
to reproduce:
1) define a region which includes zero NaN pixel
2) request to see region spectral profile
3) while the profile is updating, move the region to be totally outside the image
4) then move the region to cover non-NaN pixels, NaN pixels and empty space.
5) possibly the backend will crash, if not, repeat 2) to 4)

The following is just to indicate the three different types of regions. To reproduce this issue, we just need one region and move it around.
![Screen Shot 2019-07-23 at 10 33 46 AM](https://user-images.githubusercontent.com/20819712/61678555-bbf86180-ad35-11e9-8c6c-1e25eff8aaf2.png)

FITS cube does not have this issue.

markccchiang:
It is possible that we have to add a lock while calling the region mask `region->XyMask()`. I will fix it.

kswang1029:
with dev 36d6a07, i think the issue is gone. Shall we close this issue?

markccchiang:
Ok.

PR #281, opened 2019-07-23
Add a lock on region mask to avoid the crash for HDF5

markccchiang:
Add a lock on `region->XyMask()`, since it would cause the crash for HDF5 image (issue #280).

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -962,13 +962,12 @@ bool Frame::FillSpectralProfileData(
                         return profile_ok;
                     }
                     bool use_swizzled_data(false);
+                    std::unique_lock<std::mutex> guard(_image_mutex);
                     try {
                         // check is the region mask valid (outside the lattice or not)
                         region->XyMask();
                         // if region mask is valid, then check is swizzled data available
-                        std::unique_lock<std::mutex> guard(_image_mutex);
                         use_swizzled_data = _loader->UseRegionSpectralData(region->XyMask());
-                        guard.unlock();
                     } catch (casacore::AipsError& err) {
                         std::cerr << err.getMesg() << std::endl;
                         CARTA::SpectralProfileData profile_data;
@@ -980,6 +979,7 @@ bool Frame::FillSpectralProfileData(
                         profile_ok = true;
                         return profile_ok;
                     }
+                    guard.unlock();
                     if (use_swizzled_data) {
                         std::unique_lock<std::mutex> guard(_image_mutex);
                         _loader->GetRegionSpectralData(profile_stokes, region_id, region->XyMask(), region->XyOrigin(),
--- a/Frame.cc
+++ b/Frame.cc
@@ -962,14 +962,18 @@ bool Frame::FillSpectralProfileData(
                         return profile_ok;
                     }
                     bool use_swizzled_data(false);
+                    const casacore::ArrayLattice<casacore::Bool>* mask = nullptr;
                     std::unique_lock<std::mutex> guard(_image_mutex);
                     try {
                         // check is the region mask valid (outside the lattice or not)
-                        region->XyMask();
+                        mask = region->XyMask();
+                    } catch (...) { }
+                    guard.unlock();
+                    if (mask) {
                         // if region mask is valid, then check is swizzled data available
-                        use_swizzled_data = _loader->UseRegionSpectralData(region->XyMask());
-                    } catch (casacore::AipsError& err) {
-                        std::cerr << err.getMesg() << std::endl;
+                        use_swizzled_data = _loader->UseRegionSpectralData(mask);
+                    } else {
+                        // if region mask not valid, send a NaN to the frontend
                         CARTA::SpectralProfileData profile_data;
                         profile_data.set_stokes(curr_stokes);
                         profile_data.set_progress(1.0);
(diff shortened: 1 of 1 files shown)


Frame.cc
```diff
@@ -961,16 +961,20 @@ bool Frame::FillSpectralProfileData(
                         profile_ok = true;
                         return profile_ok;
                     }
+                    bool region_mask_ok(false);
                     bool use_swizzled_data(false);
+                    std::unique_lock<std::mutex> guard(_image_mutex);
                     try {
                         // check is the region mask valid (outside the lattice or not)
                         region->XyMask();
```
veggiesaurus: why are we calling `XyMask()` twice? @confluence, did you put this in? Getting the mask might be expensive, since it involves per-pixel testing of polygons. Its performance is likely to be _O(A)_, where _A_ is the area of the bounding box of the region

markccchiang: Actually is me. Since `XyMask()` may throw an exception, then it will cause the hanging issue if a region is outside the lattice.

Frame.cc
```diff
@@ -961,16 +961,20 @@ bool Frame::FillSpectralProfileData(
...
+                        region_mask_ok = true;
                         // if region mask is valid, then check is swizzled data available
-                        std::unique_lock<std::mutex> guard(_image_mutex);
                         use_swizzled_data = _loader->UseRegionSpectralData(region->XyMask());
-                        guard.unlock();
```
veggiesaurus: ideally, we should only lock while getting the Xy mask, and then unlock, rather than locking while doing the entire maks check

Frame.cc
veggiesaurus: Actually, given that it seems to just return a reference to the mask lattice, I don't think it's a problem.

veggiesaurus: might make more sense to reuse the first result of `region->XyMask()`, otherwise it's quite hard to understand what's going on.

veggiesaurus: Another option would be to check the region bounding box first, perhaps?

markccchiang: I think we can call `region->XyMask()` only once. Have submitted a new commit.

Frame.cc
```diff
@@ -961,16 +961,12 @@ bool Frame::FillSpectralProfileData(
...
-                        // if region mask is valid, then check is swizzled data available
-                        std::unique_lock<std::mutex> guard(_image_mutex);
-                        use_swizzled_data = _loader->UseRegionSpectralData(region->XyMask());
-                        guard.unlock();
-                    } catch (casacore::AipsError& err) {
-                        std::cerr << err.getMesg() << std::endl;
+                    bool region_mask_ok;
+                    bool use_swizzled_data;
+                    std::unique_lock<std::mutex> guard(_image_mutex);
+                    use_swizzled_data = _loader->UseRegionSpectralData(region->XyMask(), region_mask_ok);
```
veggiesaurus: this seems rather involved, adding an additional parameter that just gets set to false if the mask is a null pointer. Why not just keep that in this function, by using something like 
```
const casacore::ArrayLattice<casacore::Bool>* mask = region->XyMask();
guard.unlock();
if (mask) {
    use_swizled_data = _loader->UseRegionSpectralData(mask);
}
if (!mask) { // if region mask not valid, send a NaN to the frontend    
    profile_data.set_stokes(curr_stokes);
    ...
}
```

Region/Region.cc
```diff
@@ -520,10 +520,22 @@ casacore::IPosition Region::XyOrigin() {
 const casacore::ArrayLattice<casacore::Bool>* Region::XyMask() {
     // returns boolean mask of xy region
     casacore::ArrayLattice<casacore::Bool>* mask;
+    bool extended_region_ok(false);
 
     if (_xy_region != nullptr) {
         // get extended region
-        auto extended_region = static_cast<casacore::LCExtension*>(_xy_region->toLCRegion(_coord_sys, _image_shape));
+        casacore::LCExtension* extended_region;
```
veggiesaurus: no problem with this approach, I think it makes sense to handle the exceptions in here.

Region/Region.cc
```diff
@@ -520,10 +520,22 @@ casacore::IPosition Region::XyOrigin() {
...
+        try {
+            extended_region = static_cast<casacore::LCExtension*>(_xy_region->toLCRegion(_coord_sys, _image_shape));
+            extended_region_ok = true;
+        } catch (casacore::AipsError& err) {
+            std::cerr << err.getMesg() << std::endl;
```
veggiesaurus: we probably don't need to output the error if it's simply a region outside an image. it's not an "error" really, it's just a region that contains no data.

Util.h
```diff
@@ -142,7 +142,8 @@ struct RegionRequest {
         config = config_;
     }
     bool IsAmong(int profile_index, std::vector<int> other_stats) {
-        if (config.size() > profile_index) {
+        // check dose the region spectral requirement with respect to the profile index exist
+        if ((config.size() > 0) && (profile_index >= 0) && (config.size() - 1 >= profile_index)) {
```
pford: The profile index can change.  If a user sets region 1 profile 0 = "z"/Sum, then adds a second zprofile widget for region 1 with "Qz"/Mean, the profiles for region 1 are now profile 0="Qz"/Mean and profile 1="z"/Sum.

Frame.cc
pford: Called again instead of reusing mask:
`_loader->GetRegionSpectralData(profile_stokes, region_id, region->XyMask(), region->XyOrigin(),`

Util.h
markccchiang: I'm refactoring the spectral profile cancellation according to your suggestions. The struct `RegionRequest` will be removed. Using the spectral profile configs which already stored in the `Region` to do the interruption checks.

pford: I would suggest fixing the hdf5 mask issue in this PR and refactor the interrupts separately.

markccchiang:
I will merge this PR and solve the task cancellation issue in PR #288.

PR #288, opened 2019-07-24
Change the task cancellation codes for spectral profiles

markccchiang:
Bellow in the order of commitments, I did:

1. Remove `_region_states` and `region_requests` from the `Frame`. Use the region states and region configs which are already stored in the `Region`.
2. Add `_region_id` in the `Frame` which represents the current region id for cursor or region spectral profile. And use it as an additional condition for interruption checks.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -999,7 +999,7 @@ bool Frame::FillSpectralProfileData(
                     } else {
                         std::unique_lock<std::mutex> guard(_image_mutex);
                         profile_ok = GetRegionSpectralData(
-                            region_id, i, profile_stokes, [&](std::vector<std::vector<double>> results, float progress) {
+                            region_id, i, profile_stokes, [&](std::map<CARTA::StatsType, std::vector<double>> results, float progress) {
                                 CARTA::SpectralProfileData profile_data;
                                 profile_data.set_stokes(curr_stokes);
                                 profile_data.set_progress(progress);
@@ -1307,7 +1307,7 @@ bool Frame::GetPointSpectralData(std::vector<float>& data, int region_id, casaco
 }
 
 bool Frame::GetRegionSpectralData(int region_id, int profile_index, int profile_stokes,
-    const std::function<void(std::vector<std::vector<double>>, float)>& partial_results_callback) {
+    const std::function<void(std::map<CARTA::StatsType, std::vector<double>>, float)>& partial_results_callback) {
     bool data_ok(false);
     auto& region = _regions[region_id];
 
@@ -1318,7 +1318,7 @@ bool Frame::GetRegionSpectralData(int region_id, int profile_index, int profile_
     }
 
     // initialize the stats data
-    std::vector<std::vector<double>> results;
+    std::map<CARTA::StatsType, std::vector<double>> results;
     size_t profile_size = NumChannels(); // total number of channels
     if (!region->InitStatsData(profile_index, profile_size, results)) {
         return data_ok;
@@ -1331,7 +1331,6 @@ bool Frame::GetRegionSpectralData(int region_id, int profile_index, int profile_
     size_t start = 0;
     size_t count, end;
     float progress = 0;
-    int stats_size = results.size();
     casacore::SubImage<float> sub_image;
     int delta_channels = INIT_DELTA_CHANNEL; // the increment of channels for each step
     int dt_target = TARGET_DELTA_TIME;       // the target time elapse for each step, in the unit of milliseconds
@@ -1352,10 +1351,14 @@ bool Frame::GetRegionSpectralData(int region_id, int profile_index, int profile_
 
         // try to get sub-image and its stats data
         if (GetRegionSubImage(region_id, sub_image, profile_stokes, ChannelRange(start, end))) {
-            std::vector<std::vector<double>> buffer;
-            if (region->GetSpectralProfileData(buffer, profile_index, sub_image)) {
-                for (int j = 0; j < stats_size; ++j) {
-                    memcpy(&results[j][start], &buffer[j][0], buffer[j].size() * sizeof(double));
+            std::map<CARTA::StatsType, std::vector<double>> buffers;
+            if (region->GetSpectralProfileData(buffers, profile_index, sub_image)) {
+                for (const auto& buffer : buffers) {
+                    auto stats_type = buffer.first;
+                    if (results.count(stats_type)) {
+                        const std::vector<double>& stats_data = buffer.second;
+                        memcpy(&results[stats_type][start], &stats_data[0], stats_data.size() * sizeof(double));
+                    }
                 }
             } else {
                 std::cerr << "Can not get zprofile (statistics), region id: " << region_id << ", channel range: [" << start << "," << end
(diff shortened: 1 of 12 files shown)


Frame.cc
```diff
@@ -437,7 +435,9 @@ bool Frame::SetRegionSpectralRequirements(int region_id, const std::vector<CARTA
     if (_regions.count(region_id)) {
         auto& region = _regions[region_id];
         region_ok = region->SetSpectralRequirements(profiles, NumStokes());
-        SetRegionSpectralRequests(region_id, profiles);
+        if (profiles.size() > 0) {
+            SetRegionId(region_id);
```
pford: There can be multiple regions with spectral profiles requested if the user opens additional zprofile widgets, so saving the latest region id and checking it will not work for interrupting a spectral profile.  Stop only if this profile's parameters change (its region changed, stokes changed, requirements changed).

markccchiang:
As Pam has reminded, there could be a condition that users open multiple regions with multiple zprofile widgets. Setting the current region id as an interruption condition would not get the multiple results that users want to see. So I remove this interruption condition.

Frame.cc
```diff
@@ -1370,66 +1379,105 @@ bool Frame::GetRegionSpectralData(std::vector<std::vector<double>>& stats_values
...
-bool Frame::AreSameRegionSpectralRequests(int region_id, int profile_index, const std::vector<int>& requested_stats) {
-    return (_region_requests.count(region_id) && _region_requests[region_id].IsAmong(profile_index, requested_stats));
+bool Frame::IsSameRegionSpectralConfig(
+    int region_id, int num_profiles, int profile_index, int profile_stokes, const std::vector<int>& config_stats) {
+    bool is_same(false);
+    if (_regions.count(region_id)) {
+        auto& region = _regions[region_id];
+        // check is number of profiles changed
+        size_t new_num_profiles(region->NumSpectralProfiles());
+        if ((new_num_profiles == 0) || (new_num_profiles != num_profiles)) {
```
pford: Sorry, I was not clear.  If calculating "z"/RMS and new_num_profiles=0, then always cancel (no profiles requested).  But if new requirements are "Qz"/Mean, "z"/RMS (num profiles=2 now) then the "z"/RMS profile is still needed (note its profile index changed).  If new requirements are "z"/[Mean, RMS] then "z"/RMS is still needed (note config stats changed).  If new requirements are "z"/Sum then "z"/RMS is no longer needed and should be cancelled.  So you have to check the region requirements for both the stokes and stats types to determine if the current profile is still needed.

markccchiang: If we are calculating "z"/RMS and a new requirement "z"/[Mean, RMS] is set. Then the backend will cancel the old "z"/RMS calculation and start new "z"/[Mean, RMS] calculations. However, in this case, it is not necessary to cancel the old "z"/RMS and start a new exactly same calculation. In order to avoid this, the `_spectral_profiles` in `RegionProfiler` and its usage (for calculating stats data, filling the data in protobuf) may need to be changed. Should I finish this work in this PR, or start a another branch to do it?

markccchiang:
According to @kswang1029's suggestion on issue #298, I add the current region id and an interruption condition for spectral profile calculation. This is especially for the interruption of the cursor's spectral profile calculation.

kswang1029:
Please also make sure point region is interruptable too.

markccchiang:
Just have discussed with @kswang1029, there are some items need to improve:
1. The spectral profile calculations with respect to the widgets show on the panel can not be stopped. However, the priority of profile calculation dependent on the current widget users choose.
2. If a region's state changed (change its position, rotation or size), the widgets on the panel corresponding to this region have to be updated.

markccchiang:
If a region id is removed on the zprofile widgets (i.e., `NumSpectralProfiles() == 0`), then the calculation is cancelled (defined in `RegionProfiler::IsValidSpectralConfigStats(..)`). If a stoke index is removed or changed on zprofile widgets, the calculation is also cancelled (in `Frame::IsSameRegionSpectralConfig(..)` and `RegionProfiler::IsValidSpectralConfigStats(..)`). If we require "z"/[RMS] and then "Qz"/[Mean], "z"/[RMS]. Due to the lock of `_image_mutex`, we have to wait for "z"/[RMS] finished and set `profiles_sent = true`. So the next "z"/[RMS] will not calculate again (only calculate "Qz"/[Mean]).

Now the restriction is that if we require "z"/[RMS] and then "z"/[Mean, RMS]. The previous "z"/[RMS] is canceled and new "z"/[Mean, RMS] is started ("z"/[RMS] started twice). Under the same stoke index, if a stats type vector changed, the calculation is canceled. Despite there may be some stats type elements repeated in a vector before and after the change. To improve it I would propose to do it in another branch. Because there would be a big change in current codes. We can consider calculating all stats types a time, save in the cache even it is not finished. We can reuse the cache for the next calculation if a region state is not changed.

veggiesaurus:
> To improve it I would propose to do it in another branch. Because there would be a big change in current codes. We can consider calculating all stats types a time, save in the cache even it is not finished. We can reuse the cache for the next calculation if a region state is not changed.

Yes, we don't want to make more major changes just before a release. We will also have to re-evaluate these changes after fixing the threading approach for region calculations after 1.2

pford:
In hindsight, using `profile_index` was a huge mistake (by me) to access spectral profiles in RegionProfiler because it can change midstream, when checking for cancellation or just sending profiles in a `for` loop.  When profile="z", "z" is index 0, but then profile=["Qz", "z"], "z" is now index 1.  We should either make it a map with the stokes index (int)/coordinate (string) as the key, or append the new requirements to the existing ones so that ["Qz", "z"] is saved in RegionProfiler as ["z", "Qz"] if "z" already exists.  If we change the profile accessing scheme that will simplify many problems.  What do you think?

pford:
I am refactoring this branch to remove accessing profiles by the profile index.

markccchiang:
`profile_index` and `profile_stokes` actually refer to the same thing. It sounds sensible using a map with the stokes index (int)/coordinate (string) as the key. But considering the current structure, appending the new requirements to the existing std::vector may be a quick solution.

pford:
> `profile_index` and `profile_stokes` actually refer to the same thing

Not true, `profile_index` is the index into the std::vector<SpectralProfile> (which can be reassigned then the index is wrong), and `profile_stokes` is the stokes index of the requested profile (CURRENT_STOKES -1 which is assigned to the current stokes, I=0, Q=1, U=2, V=3).

I realized we can't just append new requirements to the vector, we also have to delete requirements which are no longer needed.  So basically make a new vector each time requirements are sent, which is what is being done now.  And all the RegionProfiler methods which return a value based on profile_index are incorrect once the vector changes.

Looking at this yesterday, the Frame gets nprofiles then iterates through them by profile_index 0 to nprofiles-1.  But since the vector can be replaced at any time, this is not going to work.  I think a better strategy is for Frame to get a copy of the vector at the beginning, then iterate through the copy, checking with RegionProfiler during each iteration whether the particular profile is still needed (not removed or already sent).

pford:
Closing this PR, changes included in PR #324 for spectral profile cancellation.

PR #290, opened 2019-07-25
Round the XY coordinates of a point region correctly

### LINKED ISSUE #289: Point profiles are not correct when using HDF5 images
Testing using `dev` branch, with the newly added point regions. The profiles returned from point regions seem to be offset by half a pixel when using HDF5 files. Probably this line in `Frame.cc`:
```
_loader->GetCursorSpectralData(spectral_data, profile_stokes, cursor_point.x(), 1, cursor_point.y(), 1);`
```
Cursor points should be rounded instead of truncated. Example of this in action:

point offset

confluence:
Fixes #289.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -926,7 +926,7 @@ bool Frame::FillSpectralProfileData(
                     // try use the loader's optimized cursor profile reader first
                     std::unique_lock<std::mutex> guard(_image_mutex);
                     bool have_spectral_data =
-                        _loader->GetCursorSpectralData(spectral_data, profile_stokes, cursor_point.x(), 1, cursor_point.y(), 1);
+                        _loader->GetCursorSpectralData(spectral_data, profile_stokes, cursor_point.x() + 0.5, 1, cursor_point.y() + 0.5, 1);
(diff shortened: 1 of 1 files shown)


Frame.cc
```diff
@@ -926,7 +926,7 @@ bool Frame::FillSpectralProfileData(
                     // try use the loader's optimized cursor profile reader first
                     std::unique_lock<std::mutex> guard(_image_mutex);
                     bool have_spectral_data =
-                        _loader->GetCursorSpectralData(spectral_data, profile_stokes, cursor_point.x(), 1, cursor_point.y(), 1);
+                        _loader->GetCursorSpectralData(spectral_data, profile_stokes, cursor_point.x() + 0.5, 1, cursor_point.y() + 0.5, 1);
```
veggiesaurus: see [here](https://stackoverflow.com/questions/47302544/why-do-lots-of-old-programs-use-floor0-5-input-instead-of-roundinput), perhaps we should use `std::round`?

veggiesaurus: never mind, it's too nit-picky :)

Issue #76, opened 2019-02-27
HDF5 file stays open if reading main dataset in file info loader fails


confluence:
If the file info loader fails to load the main dataset into a `casacore::HDF5Lattice` near line 173 of `FileInfoLoader.cc` (for example, because the data is not the expected type), the file remains open, which prevents it from being read by other tools (e.g. `h5ls`) until the backend is shut down.

This seems to be a casacore issue with the lattice object, which needs to be cleaned up somehow.

I will file a separate issue for the specific datatype incompatibility which caused the crash.

PR #291, opened 2019-07-25
Always open HDF5 file as read-only

### LINKED ISSUE #87: HDF5 files should be opened in read mode, not write mode
This is apparently a casacore limitation in the parameters passed to the lattice constructor. We may be able to fix it by first creating a file object first and then using it to create the lattice. We need to check that this is also done in the info loader.

confluence:
I believe that this causes the HDF5 file to be opened as read-only in all three places where we do this, which fixes #87 (files open in CARTA are blocked from being opened by other processes) and #76 (if the info loader fails to open a file, it stays blocked until CARTA is shut down).

But this seems suspiciously easy, so I'd appreciate a review. Is this correct use of `CountedPtr`, and is it OK to remove the explicit closing of the HDF5 group and file in the info loader (as far as I can see, the destructors of the casacore classes clean them up automatically)?

### DIFF
--- a/FileInfoLoader.cc
+++ b/FileInfoLoader.cc
@@ -238,23 +238,21 @@ bool FileInfoLoader::FillFileExtInfo(CARTA::FileInfoExtended* ext_info, std::str
 bool FileInfoLoader::FillHdf5ExtFileInfo(CARTA::FileInfoExtended* ext_info, std::string& hdu, std::string& message) {
     // Add extended info for HDF5 file
     try {
-        casacore::HDF5File hdf_file(_filename);
+        auto hdf_file_ptr = new casacore::HDF5File(_filename);
+
         if (hdu.empty()) { // use first
-            hdu = casacore::HDF5Group::linkNames(hdf_file)[0];
+            hdu = casacore::HDF5Group::linkNames(*hdf_file_ptr)[0];
         }
-        casacore::HDF5Group hdf_group(hdf_file, hdu, true);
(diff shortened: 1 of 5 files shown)


FileInfoLoader.cc
```diff
@@ -263,8 +261,11 @@ bool FileInfoLoader::FillHdf5ExtFileInfo(CARTA::FileInfoExtended* ext_info, std:
         // check dimensions
         casacore::uInt num_dims;
         casacore::IPosition data_shape;
+
+        auto counted_ptr = casacore::CountedPtr<casacore::HDF5File>(hdf_file_ptr);
```
veggiesaurus: can you use a `CountedPtr` in the usual way? Wouldn't it be better to construct this straight away, instead of constructing `hdf_file_ptr`?

veggiesaurus:
This looks like the right way to handle it, but I don't know enough about `CountedPtr` use to say for sure if this is cleaning up properly.

FileInfoLoader.cc
confluence: It seems that you can. I also removed the intermediate variables in other places; they seemed unnecessarily verbose.

confluence:
> This looks like the right way to handle it, but I don't know enough about `CountedPtr` use to say for sure if this is cleaning up properly.

`CountedPtr` uses `std::shared_ptr` internally, so I would expect it to behave in a similar way.

PR #296, opened 2019-07-29
Extend code for using mask creation to check if the region is out of bounds to point regions.

### LINKED ISSUE #292: backend crash if requesting spectral profile from a point region totally outside the image
this happens to all image format.

confluence:
Should fix #292.

* The code in `Frame::FillSpectralProfileData` which attempts to create a mask (which will fail if the region is entirely outside the image) is now applied to all regions.
* `Region::XyMask` now handles point regions correctly and no longer segfaults. It produces empty masks for point regions, which is not terribly useful, but it fails in a consistent way if the point region is outside the image (which allows us to check all regions in the same way).

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -889,12 +889,34 @@ bool Frame::FillSpectralProfileData(
         if (!region->IsValid()) {
             return false;
         }
+
+        int curr_stokes(CurrentStokes());
+
+        const casacore::ArrayLattice<casacore::Bool>* mask = nullptr;
+        std::unique_lock<std::mutex> guard(_image_mutex);
+        try {
+            // check is the region mask valid (outside the lattice or not)
+            mask = region->XyMask();
+        } catch (casacore::AipsError& err) {
+        }
+        guard.unlock();
+        if (!mask) {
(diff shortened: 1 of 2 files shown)


Frame.cc
```diff
@@ -889,12 +889,34 @@ bool Frame::FillSpectralProfileData(
...
+
+        const casacore::ArrayLattice<casacore::Bool>* mask = nullptr;
+        std::unique_lock<std::mutex> guard(_image_mutex);
+        try {
+            // check is the region mask valid (outside the lattice or not)
+            mask = region->XyMask();
+        } catch (casacore::AipsError& err) {
+        }
+        guard.unlock();
+        if (!mask) {
```
pford: When SET_IMAGE_CHANNELS is used to change the stokes, new spectral profile data is sent for all regions which have requirements set (checked below in line 915).  This is sending a NaN profile before this check, for regions which may not require spectral profiles.

confluence: Good point. I moved the check closer to where it had been before, since there are two other checks which should also be done beforehand.

Issue #299, opened 2019-07-31
backend crash if we delete a region when spectral profile is still being updated wrt the region


kswang1029:
with the dev as of July 31, 2019

To reproduce the issue:
1) open a large fits cube (e.g., set_IRAS_17150-3224/IRAS_17150-3224_sci.spw27.cube.I.manual.image.pbcor.fits)
2) make a big rectangle region
3) launch the spectral profile widget
4) request the profile from the rectangle region
5) when the profile is being updated, delete the rectangle region
6) backend should crash

kswang1029:
fixed via #302 
Close this issue.

PR #302, opened 2019-08-01
Delete a Region after its spectral profile is finished

markccchiang:
This is for issue #299.

While deleting the Region:
1. Set the connection flag is false to interrupt the running job.
2. Wait for the job finished before deleting the Region.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -226,6 +226,7 @@ bool Frame::RegionChanged(int region_id) {
 
 void Frame::RemoveRegion(int region_id) {
     if (_regions.count(region_id)) {
+        _regions[region_id]->DisconnectCalled();
         _regions[region_id].reset();
         _regions.erase(region_id);
     }
@@ -889,6 +890,10 @@ bool Frame::FillSpectralProfileData(
         if (!region->IsValid()) {
             return false;
         }
+
+        // Increase the spectral profile count
+        region->IncreaseZProfileCount();
+
         size_t num_profiles(region->NumSpectralProfiles());
         if (num_profiles == 0) {
             return false; // not requested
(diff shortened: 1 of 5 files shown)


Frame.cc
```diff
@@ -889,6 +890,10 @@ bool Frame::FillSpectralProfileData(
         if (!region->IsValid()) {
             return false;
         }
+
+        // Increase the spectral profile count
+        region->IncreaseZProfileCount();
```
pford: There are cases where this function returns without decreasing the zprofile count (no profiles or region outside image), so the region's count will never reach zero.  Why doesn't the region inc/dec its own count when its zprofile functions are called?

Frame.cc
```diff
@@ -1252,15 +1261,15 @@ bool Frame::GetPointSpectralData(std::vector<float>& data, int region_id, casaco
                 // start the timer
                 auto t_start = std::chrono::high_resolution_clock::now();
                 // check if region point changed from subimage point
-                if ((region_id == CURSOR_REGION_ID) && (Interrupt(_cursor_xy, subimage_cursor))) {
+                if ((region_id == CURSOR_REGION_ID) && (Interrupt(region_id, _cursor_xy, subimage_cursor))) {
                     return false; // cursor moved
```
pford: Return values for GetPointSpectralData and GetRegionSpectralData are not checked, so if the profile is interrupted it is still sent to the frontend.  I think this profile is no longer needed.

pford:
Overall I think the region having a connection status and waiting for the zprofiles to complete will work.  A couple issues in the implementation.

Frame.cc
markccchiang: Not sure is it a good solution. Now I synchronized the zprofile counts for the Frame and the Region objects in the commit 716337aeec2a65e711e07f6543bf34194adc54f8. To make sure the zprofile process is completely done, and then destroy the objects.

PR #303, opened 2019-08-01
Use shared pointers for some attributes to avoid causing crashes through premature deletion

### LINKED ISSUE #301: crash when updating cursor region while spectral profile is still being delivered
possibly the same root cause as #299 or #294 

The backend is crashing when cursor spectral profiles are requested too quickly. To reproduce this bug, load a large file with many channels from a slow storage device, such as a hard drive, (I've tested with `supermosaic.10.fits` and `GNPilot.IQU.fits`) and then open a spectral profile widget linked to the cursor. By moving the cursor around, the backend can be crashed. 

I'm fairly certain the issue is related to `Region::MakeExtendedRegion`. The line `region = new casacore::WCExtension(*_xy_region, ext_box);` derefrerences `_xy_region`, but passes it to `WCExtension` as a const reference. However, another thread can delete `_xy_region` while the `WCExtension` is being constructed.

confluence:
This is an attempt to fix #301 and possibly some related issues. I am using shared pointers for the `_xy_region` and `_xy_mask` attributes, and using local shared pointers to the `_xy_region` object in functions which use it for calculations, so that they don't crash if the region changes and the old object is deleted.

I haven't modified any of the functions which create the region objects -- they all still return raw pointers.

Does this make sense, and does it fix the bug(s)? I have an SSD, so I haven't been able to reproduce the Z-profile crashes locally.

### DIFF
--- a/Region/Region.cc
+++ b/Region/Region.cc
@@ -29,8 +29,6 @@ Region::Region(const std::string& name, const CARTA::RegionType type, const std:
       _spectral_axis(spectral_axis),
       _stokes_axis(stokes_axis),
       _xy_axes(casacore::IPosition(2, 0, 1)),
-      _xy_region(nullptr),
-      _xy_mask(nullptr),
       _coord_sys(coord_sys) {
     // validate and set region parameters
     _num_dims = image_shape.size();
@@ -47,14 +45,7 @@ RegionState Region::GetRegionState() {
 }
 
 Region::~Region() {
-    if (_xy_region) {
-        delete _xy_region;
-        _xy_region = nullptr;
-    }
-    if (_xy_mask) {
-        delete _xy_mask;
-        _xy_mask = nullptr;
-    }
+    // TODO: is it really necessary to do this explicitly?
     _region_stats.reset();
     _region_profiler.reset();
 }
@@ -181,7 +172,7 @@ bool Region::PointsChanged(const std::vector<CARTA::Point>& new_points) {
 bool Region::GetRegion(casacore::ImageRegion& region, int stokes, ChannelRange channel_range) {
     // Return ImageRegion for given stokes and region parameters.
     bool region_ok(false);
-    if (!IsValid() || (_xy_region == nullptr) || (stokes < 0))
+    if (!IsValid() || (!_xy_region) || (stokes < 0))
         return region_ok;
 
     casacore::WCRegion* wc_region = MakeExtendedRegion(stokes, channel_range);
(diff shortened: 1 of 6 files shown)


pford:
Not an expert on shared_ptr but it sounds like we will still have a problem when the xy region is being reassigned to a `new` region:
https://stackoverflow.com/questions/9127816/stdshared-ptr-thread-safety-explained

confluence:
@pford Under what circumstances? When `_xy_value` is reassigned, the old value will not be destroyed until all the local pointer instances go out of scope. The behaviour demonstrated in the top answer in that thread is exactly what we want.

pford:
I looked at the final answer, which says the ref counter is thread-safe but the object pointer (which is changed in Region::SetXyRegion) is not:

std::shared_ptr is not thread safe.

A shared pointer is a pair of two pointers, one to the object and one to a control block (holding the ref counter, links to weak pointers ...).

There can be multiple std::shared_ptr and whenever they access the control block to change the reference counter it's thread-safe but the std::shared_ptr itself is NOT thread-safe or atomic.

If you assign a new object to a std::shared_ptr while another thread uses it, it might end up with the new object pointer but still using a pointer to the control block of the old object => CRASH.

confluence:
@pford Oh, I see. Could we solve this by using a lock when we reassign the main pointer, and also when we create the local pointers?

confluence:
@pford does this look right?

pford:
This is totally beyond my realm of expertise.  It seems to me that if the xy region changes, then any other thread already trying to access the xy region/mask should be signaled to abandon ship somehow, because the new region setting is going to trigger new profile/stats/histogram data anyway.  But maybe that is beyond the scope of this PR.

confluence:
Yes, that is a separate issue -- we may be calculating some unnecessary profiles which are immediately replaced by newer ones. This PR is just to eliminate the server crash caused by threads deleting pointers from under each other.

confluence:
I'm paging Rob, who knows stuff about threads. :)

pford:
Excellent idea.

Region/Region.cc
```diff
@@ -47,14 +45,7 @@ RegionState Region::GetRegionState() {
...
 Region::~Region() {
-    if (_xy_region) {
-        delete _xy_region;
-        _xy_region = nullptr;
-    }
-    if (_xy_mask) {
-        delete _xy_mask;
-        _xy_mask = nullptr;
-    }
+    // TODO: is it really necessary to do this explicitly?
```
veggiesaurus: no, it's not

Region/Region.cc
```diff
@@ -490,9 +480,11 @@ bool Region::MakeExtensionBox(casacore::WCBox& extend_box, int stokes, ChannelRa
 }
 
 casacore::WCRegion* Region::MakeExtendedRegion(int stokes, ChannelRange channel_range) {
```
veggiesaurus: could these not return a `shared_ptr`?

Region/Region.cc
confluence: I've removed it.

Region/Region.cc
confluence: I was initially going to convert all of these to return `shared_ptr` -- but the casacore functions return C pointers anyway, so there has to be a conversion at some point anyway. I was worried about knock-on effects. But I can revisit this. Is there a specific place where we would benefit from using `shared_ptr`s here?

confluence:
@angus do you have debugging output from the segfault? Do you know where in the function it happens?

Issue #307, opened 2019-08-08
incorrect and inconsistent per-frame histogram


kswang1029:
with the dev backend/frontend as of Aug 8, 2019

With the test image set_StokesCube/image_IQU.im, we see inconsistent and incorrect per-frame histogram. A region is created to cover the entire image. In some cases (at different channel), the histogram in the render config widget is correct. But in some other cases, the histogram widget is correct. Moreover, both can be incorrect. 

![Screen Shot 2019-08-08 at 11 42 48 AM](https://user-images.githubusercontent.com/20819712/62673712-15b58880-b9d2-11e9-8aa3-1a3b299bd162.png) 

![Screen Shot 2019-08-08 at 11 43 19 AM](https://user-images.githubusercontent.com/20819712/62673710-151cf200-b9d2-11e9-83c7-7de30efd16a1.png)

![Screen Shot 2019-08-08 at 11 42 57 AM](https://user-images.githubusercontent.com/20819712/62673711-151cf200-b9d2-11e9-80fa-3377f637bdf8.png)

Here I attach another example fits image which is 8x8x100 (https://drive.google.com/open?id=1V5VUX79NWSiUAv82_adpPcQfOR-rC1iY). For each channel, there is a pixel ~1 with all other channel ~0.1. So the histogram would be similar to
![Screen Shot 2019-08-08 at 12 19 29 PM](https://user-images.githubusercontent.com/20819712/62674842-d2a9e400-b9d6-11e9-9d06-e60071894d16.png)
That is, a bump on the left and a one-count bin on the right.
However, by changing channels, sometimes we will see the one-count bin on the right is missing. In other word, there is just a bump. 
![Screen Shot 2019-08-08 at 12 24 14 PM](https://user-images.githubusercontent.com/20819712/62674989-78f5e980-b9d7-11e9-951d-60b0f1e769a2.png)

Could it be memory leakage?

veggiesaurus:
@markccchiang or @confluence can you look into this? @kswang1029 have you found a way to consistently reproduce this?

pford:
Looking at the histogram messages, the histograms for both regions have the same number of bins.  When the histograms match, the histograms for each region have the same bin width; when they differ, the bin widths differ.

kswang1029:
> @markccchiang or @confluence can you look into this? @kswang1029 have you found a way to consistently reproduce this?

@veggiesaurus with set_StokesCube/image_IQU.im we can see the issue quite easily.

[...]

PR #310, opened 2019-08-09
Fix a bug in Min/Max calculation

markccchiang:
This is to solve issue #307.

### DIFF
--- a/Region/MinMax.h
+++ b/Region/MinMax.h
@@ -43,7 +43,8 @@ void MinMax<T>::operator()(const tbb::blocked_range<size_t>& r) {
         if (std::isfinite(val)) {
             if (val < t_min) {
                 t_min = val;
-            } else if (val > t_max) {
+            }
+            if (val > t_max) {
                 t_max = val;
             }
         }


markccchiang:
If the first element comes in the loop and it is the largest value in an array, then it will be assigned to `t_min`, not `t_max`. That is why it has to be two `if`. To avoid using two `if` and using one `if else`, maybe we can initialize `t_min` and `t_max` values as the first element of the finite array.

veggiesaurus:
> If the first element comes in the loop and it is the largest value in an array, then it will be assigned to `t_min`, not `t_max`. That is why it has to be two `if`. To avoid using two `if` and using one `if else`, maybe we can initialize `t_min` and `t_max` values as the first element of the finite array.

You can do that, although we will just need to ensure that we handle the case where the array is empty. Better to have some extra branching code in an _O(1)_ part of the function than in an _O(N)_ part.

markccchiang:
After several trials, I think it is not a good idea to initialize `t_min` and `t_max` as the first element of the finite array. Since it is a parallel calculation, any step to identify the value is finite or NaN and then use it to initialize `t_min` and `t_max` would increase the complexity of this algorithm.

veggiesaurus:
why not just use `-FLOAT_MAX` and `FLOAT_MAX`? (or rather, the c++11 equivalent)  and then in the `join` handle those situations where _t_min or _t_max are still those values.

markccchiang:
> why not just use `-FLOAT_MAX` and `FLOAT_MAX`? (or rather, the c++11 equivalent) and then in the `join` handle those situations where _t_min or _t_max are still those values.

markccchiang:
> > why not just use `-FLOAT_MAX` and `FLOAT_MAX`? (or rather, the c++11 equivalent) and then in the `join` handle those situations where _t_min or _t_max are still those values.

Sorry! I just accidentally close this PR. If my understanding is correct, `t_min` and `t_max` are already initialized as `std::numeric_limits<T>::max()` and `std::numeric_limits<T>::min()`.

markccchiang:
Test the time spending for min/max calculation with large images like `LMCq.fits` or `h_m51_b_s05_drz_sci.fits`. I don't see there is a large difference between the original and newly changed code.

kswang1029:
@markccchiang could you also try with images in set_Gaussian2D? ie 20k, 40k and 80k.

markccchiang:
For 20k image, the average time spent is 2653(original) / 2541(new) ms. For 40k image, the average time spent is 10313(original) / 11794(new) ms.

veggiesaurus:
Ok. I guess that's an acceptable change. I'll approve and merge

PR #315, opened 2019-08-13
Rwjs/mongodb

robsimmonds:
This adds the use of mongodb for the authenticating server deployment of the backend. Note that by default cmake will create a non-authenticating version for desktop deployments and will not link mongoc, bson or jsoncpp. To build the authenticating version you need to run:

        cmake -DAuthServer:BOOL=ON ..


Main.cc
```diff
@@ -310,6 +319,7 @@ void ExitBackend(int s) {
 }
 
 void ReadJSONfile(string fname) {
+#if AuthServer
```
veggiesaurus: shouldn't this be _AUTH_SERVER?

Issue #316, opened 2019-08-13
region import: missing point region


kswang1029:
The CRTF point region
`symbol[[320pix, 200pix], '.']`
is missing when importing.

pford:
In imageanalysis this is a display definition, along with line, vector, and text.  AnnotationBase::isRegion() returns false since AnnSymbol is not derived from AnnRegion.  I thought for carta a point region would be a rectangle with blc=trc.

kswang1029:
If we define a point region with casaviewer can export it as a crtf file, we see it is “symbol”. The same file can be imported to casaviewer and we will see a point region.

PR #319, opened 2019-08-13
import/export point regions

pford:
Issue #316, use casacode AnnSymbol for point regions.

Also added try/catch for annotation regions, for issues with coord sys for some images.  Will try to fix this later.


pford:
Do not merge, further testing showed there is still an issue.

pford:
> Do not merge, further testing showed there is still an issue.

Never mind, this is ready to merge.

pford:
Now I am getting the error again when exporting a point region, the symbol '.' is unknown.

Issue #321, opened 2019-08-14
empty histogram if all pixels are negative in a frame


kswang1029:
test image: set_noise/noise_negative2.miriad which has pixel values of ~-10

veggiesaurus:
@markccchiang it's because `_max_val` is being initialised to `std::numeric_limits<T>::min()` instead of `-std::numeric_limits<T>::max()`

markccchiang:
Yes, may be use `std::numeric_limits<T>::lowest()` instead of `std::numeric_limits<T>::min()` for initializing the `_max_val`.

veggiesaurus:
yes, good idea.

markccchiang:
Will push later, I am trying to build the casacode.

PR #322, opened 2019-08-14
Change the initialization of _max_val for Min/Max calculation

markccchiang:
It is to solve issue #321. Since `std::numeric_limits<T>::min()` is a positive minimum value. Use it to initialize the `_max_val` can not get the correct result if all pixel values are negative. Use `std::numeric_limits<T>::lowest()` instead.

### DIFF
--- a/Region/MinMax.h
+++ b/Region/MinMax.h
@@ -28,11 +28,11 @@ public:
 
 template <typename T>
 MinMax<T>::MinMax(const std::vector<T>& data)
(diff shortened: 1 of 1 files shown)


veggiesaurus:
Looks good. @kswang1029 can you confirm that this fixes #321?

kswang1029:
@markccchiang @veggiesaurus Confirmed! 👍

Issue #311, opened 2019-08-12
cancelable region spectral profiler with hdf5 images?


kswang1029:
With dev as of Aug 12 2019

backend team please confirm the following issue.

It appears that the backend is still busy on region spectral profile calculations even if there is no profile update messages being sent when the frontend no longer needs the profile.

To reproduce the issue:
1) open a large enough hdf5 image
2) freeze cursor and launch one spectral profile widget. we should see a profile now.
3) make a region (region 1)
4) request profile from region 1
5) while profile of region 1 is updating, switch back to cursor and unfreeze cursor and move around, now we should see cursor profile updates along with the cursor movement but we don't. We will start to see it update _once_ the backend completes the region 1 profile calculations.

markccchiang:
It appears the hdf5 image is not included in the cancellation while the region id is removed from the frontend. I will fix this issue.

markccchiang:
Fixed it in the commit 2a42f613db5bb0c9b185b452828976832e3f41b1 (in PR #288).

pford:
Merged into dev

kswang1029:
@pford Yes, issue solved. 👍

Issue #298, opened 2019-07-31
spectral profile from a point region is not stoppable 


kswang1029:
with dev as of July 31, 2019

the backend keeps sending updates of a point region spectral profile even if the user switches to another region.

UPDATE: same for cursor region.

kswang1029:
shall we consider this as critical (from the usability point of view if the image is big)? This means users always have to wait the cursor profile finished then start to see region profile updating. When this happens, the spectral profiler widget appears _blank_ until the region profile starts to update.

markccchiang:
In PR #288, I have considered the current region id which can be used to cancel the cursor or the other region id spectral profile calculation. However, if we open additional zprofile widgets for the other regions, then only the widgets with specific region id can be updated. Is it acceptable for you?

kswang1029:
@markccchiang if you are referring to "origin/mark/refactor_profile_cancellation 296dd45 5027 2019-08-05 15:58:50 CST", cursor profile is still non-stoppable.

markccchiang:
This function was removed according to Pam's suggestion. Maybe you can test 2a9413619e2d5a7ccafe9f938e392d5f4d346ecf.

kswang1029:
@markccchiang sorry with the mix n' match page, 2a94136 is not available anymore. Is it merged?

markccchiang:
It is not merged yet. Should I move the current region id as an interruption condition back?

kswang1029:
The desired feature is the following. If we have three regions: cursor, point, and rectangle, and users request cursor spectral profile, then if users switch to either point or rectangle region while the cursor profile is updating, the cursor profile should stop updating immediately and start the new profile calculations for the new region request.

pford:
If the user changes the profile widget cursor->region X, the frontend sends a new cursor spectral requirement message with no profiles, which triggers cancelling the cursor profile.  Or perhaps you are saying to prefer the other region over cursor when both are required, but that will require some bookkeeping to resume the cursor profile once the other region profile is done.

veggiesaurus:
I think what @kswang1029 is saying is that if the cursor profile is no longer required, then it should be cancelled. If it is still required by another widget, it shouldn't be cancelled. I believe this is the intended behaviour of the current implementation. 

[...]

PR #324, opened 2019-08-14
Remove profile index access for spectral profiles

pford:
For issue #298, building on mark/refactor_profile_cancellation branch.  My brief testing of this branch did not reproduce these issues.

The functionality is basically the same as Mark's, but instead of accessing spectral profiles by an index into the profile vector, the requested profile is found by its stokes index (the int equivalent of the string coordinate: I=0, Q=1, etc., CURRENT_STOKES=-1).  This means that the Frame has to keep track of two stokes index values:  `config_stokes` is the stokes_index in the spectral config, and `profile_stokes` is the stokes index for slicing the image data (equal to `config_stokes` except when CURRENT_STOKES must be replaced by the current stokes index).

This is still not perfect, but I hesitated to make greater changes just before the pre-release.  There are no safeguards to prevent one thread from replacing the vector while another is searching it for the requested profile.  To minimize this, the vector is retrieved before the search when possible to make it more atomic.  Another inefficiency is that the current profile calculations in progress are interrupted if the spectral config for the stokes index changes (even if the change is to add another stats type).  However, if the first stat is complete and sent before the requirements change, then it is not re-sent.  These issues can be improved in the next release.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -999,7 +999,7 @@ bool Frame::FillSpectralProfileData(
                     } else {
                         std::unique_lock<std::mutex> guard(_image_mutex);
                         profile_ok = GetRegionSpectralData(
-                            region_id, i, profile_stokes, [&](std::vector<std::vector<double>> results, float progress) {
+                            region_id, i, profile_stokes, [&](std::map<CARTA::StatsType, std::vector<double>> results, float progress) {
                                 CARTA::SpectralProfileData profile_data;
                                 profile_data.set_stokes(curr_stokes);
                                 profile_data.set_progress(progress);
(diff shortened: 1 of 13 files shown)


pford:
This does not fix issue #311, unless I am misunderstanding the issue.  With a spectral profile for region 1 being calculated, the spectral profile for the cursor is not updating but the spatial profiles are updated.

markccchiang:
> This does not fix issue #311, unless I am misunderstanding the issue. With a spectral profile for region 1 being calculated, the spectral profile for the cursor is not updating but the spatial profiles are updated.

I thought issue #311 means to cancel the spectral profile calculation if such region id or cursor is not shown on the frontend panel (for HDF5 images). @kswang1029 could you please confirm does it fix issue #311?

kswang1029:
@pford @markccchiang with 8238d71, if the profile of region 1 is updating and we switch to cursor, yes, the calculation of profile of region 1 is terminated and cursor profile will be updating. 👍 Before the fix, we will need to wait region 1 profile completes (it's updating in the background and we see the widget is doing nothing) then start to see cursor profile updating along with the cursor movement.

PR #327, opened 2019-08-20
Reject pv image as "not supported"

pford:
Use image coordinate system to detect pv image, then return open file ack message with success = false and error message.  Also check that coord sys number of pixel axes == image dimensions, else this will fail later when creating a region (including image and default cursor) and converting between pixel/world coordinates.  The number of pixel axes can be incorrect when header keywords describing each axis are missing.

Also check open file success before calling method to send data streams.

There may still be an issue in the frontend.

### DIFF
--- a/ImageData/FileLoader.cc
+++ b/ImageData/FileLoader.cc
@@ -82,42 +82,60 @@ void FileLoader::FindCoords(int& spectral_axis, int& stokes_axis) {
     }
 }
 
-bool FileLoader::FindShape(IPos& shape, size_t& num_channels, size_t& num_stokes, int& spectral_axis, int& stokes_axis) {
+bool FileLoader::FindShape(IPos& shape, int& spectral_axis, int& stokes_axis, std::string& message) {
     if (!HasData(FileInfo::Data::Image))
         return false;
 
     shape = LoadData(FileInfo::Data::Image)->shape();
-    size_t num_dims = shape.size();
+    size_t _num_dims = shape.size();
 
-    if (num_dims < 2 || num_dims > 4) {
+    if (_num_dims < 2 || _num_dims > 4) {
+        message = "Image must be 2D, 3D, or 4D.";
         return false;
     }
 
     casacore::CoordinateSystem coord_sys;
     if (!GetCoordinateSystem(coord_sys)) {
+        message = "Image does not have valid coordinate system.";
         return false;
     }
-    if (coord_sys.nPixelAxes() != num_dims) {
+    if (coord_sys.nPixelAxes() != _num_dims) {
+        message = "Image coordinate system does not match data shape.";
         return false;
     }
 
+    // use CoordinateSystem to find coordinate axes
+    casacore::Vector<casacore::Int> linear_axes = coord_sys.linearAxesNumbers();
     spectral_axis = coord_sys.spectralAxisNumber();
-    if (spectral_axis == -1) {
-        num_channels = 1;
-    } else {
-        num_channels = shape(spectral_axis);
-    }
-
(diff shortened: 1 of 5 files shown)


kswang1029:
@pford Thanks for quickly fixing this. After I click the load button, the entire frontend hangs. @veggiesaurus any ideas?

pford:
Which dataset did you test?

kswang1029:
@pford I shared the test images with you via Slack

ImageData/FileLoader.cc
```diff
@@ -82,31 +82,60 @@ void FileLoader::FindCoords(int& spectral_axis, int& stokes_axis) {
...
 }
 
-bool FileLoader::FindShape(IPos& shape, size_t& num_channels, size_t& num_stokes, int& spectral_axis, int& stokes_axis) {
+bool FileLoader::FindShape(IPos& shape, int& spectral_axis, int& stokes_axis, std::string& message) {
     if (!HasData(FileInfo::Data::Image))
         return false;
 
     shape = LoadData(FileInfo::Data::Image)->shape();
-    size_t num_dims = shape.size();
+    size_t _num_dims = shape.size();
```
confluence: Does this not mask the `_num_dims` attribute with a local variable with the same name?

pford: I don't understand the question.  Rather than assign the local num_dims then later assign the _num_dims attribute to the local, why not just assign the attribute?

pford:
Added additional check, axes can be [p, v] or [v, p] in test images.

ImageData/FileLoader.cc
```diff
@@ -82,31 +82,60 @@ void FileLoader::FindCoords(int& spectral_axis, int& stokes_axis) {
...
+
+    // pv images not supported (yet); spectral axis is 0 or 1, and the other is linear
+    if (!linear_axes.empty() && (((spectral_axis == 0) && (linear_axes(0) == 1)) || ((spectral_axis == 1) && (linear_axes(0) == 0)))) {
+        message = "Position-velocity (pv) images not supported.";
+        return false;
+    }
+
+    // spectral axis not defined if CTYPE is not "FREQ" (e.g. "FREQUENC" will fail)
+    if ((spectral_axis == -1) && (_num_dims > 2)) {
+        if ((_num_dims == 3) || (stokes_axis == 3)) {
```
pford: Whoa!  That's not right.

ImageData/FileLoader.cc
confluence: What I mean is, are you not declaring a new local variable here, rather than using the attribute, because of the `size_t`?

pford: Yes, you are right!

confluence: In a future refactoring pass I'd like to make this function no longer return values to the caller *and* set state on the loader, and also try to avoid some of the property duplication between the loader and the frame. But I don't think that needs to happen in this PR.

pford: Agree, DRY (don't repeat yourself)!  So then have loaders for all image types slice the data for spatial profiles and point spectral profiles, rather than doing it in the frame?

confluence: The dividing line between loader and frame is fuzzy; I think there are various tasks that could be described as "loading" and moved into the loader, even if they're not format-specific and end up in the parent class.

Right now there's a bunch of stuff in the frame that probably doesn't need to be there -- I think it would be a good idea to discuss how to divide up responsibility between the different classes at the F2F.

pford: Thanks for the catch.

Issue #186, opened 2019-04-16
disk cache of image stats data


kswang1029:
The main idea of this issue is to use disk cache for image stats of non-hdf5 images (casa, fits, miriad) to speed up image loading process. The persistent cache should store per-frame histogram, per-frame statistics, and per-cube histogram once they are calculated for the first time. If the same image is loaded again, the backend should check if cached data are available and applicable to the image first. If not, then compute and save the results in the cache for re-use.

Kechil:
Closed for now - may arise again in consideration of performance

PR #333, opened 2019-08-25
Add region spectral profile cache for non-HDF5 images

markccchiang:
This mechanism is similar to the stats cache for HDF5 images, which calculates all stats types, including `Sum`, `Mean`, `RMS`, `Sigma`, `SumSq`, `Min` and `Max` a time, and then stores the results in the `Region` object. If there is an interruption during the calculation, say, we change settings of region spectral profile widgets, the unfinished results also stored in the cache and can be reused to continue the calculation next time, if the region's state is not changed. 

The advantage is that no matter how we change the number of spectral profile widgets and their settings, the calculations will not be repeated. Although at the start it will spend more time to calculate all stats types. If the region's state changed, the `Region` object will reset the stats cache by `ResetStatsCache()` function, which will start new spectral profile calculations.

### DIFF
--- a/Frame.cc
+++ b/Frame.cc
@@ -1595,14 +1595,20 @@ bool Frame::GetRegionSpectralData(int region_id, int config_stokes, int profile_
 
         // initialize the spectral data
         std::map<CARTA::StatsType, std::vector<double>> results;
+        size_t start;
         size_t profile_size = NumChannels(); // total number of channels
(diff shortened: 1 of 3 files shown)


kswang1029:
I think #186 refers to a more generic approach to speed up loading casa and fits image with persistent disk cache.

markccchiang:
A more detail description of #186 may be needed.

kswang1029:
@markccchiang yes, I have added some info accordingly.

PR #338, opened 2019-08-28
Fix backend crash from incomplete coordinate system axes

pford:
"Forward-port" fix #337 in release branch.

If header is incomplete or has unexpected values, the FileLoader attempts to use extended file info to determine spectral/stokes axes, then resorts to guessing.


pford:
Hopefully I didn't miss anything this time.

PR #354, opened 2019-09-27
Refactor the OnMessage function

markccchiang:
The idea of refactoring the `OnMessage` function is to let it looks a little bit clear for how messages from the frontend are handled and executed. The messages with respect to their handlers (`Command` functions) get together in the `MessageHandler` class.

In the future for the resuming of a session task that may require a message of multiple commands, like opening a series of images or regions. Using this formate can avoid the repetition of codes. We may only need to edit the protobuf message as an input variable for the `Command` function and reuse it.


pford:
Added @robsimmonds to review this PR since he has worked on message handling.

robsimmonds:
I have started to look at this but I will have to give it a lot more thought. Now we are creating an additional new object for every message received. It is not clear to me what that gains us. Also I have been giving a lot of thought to making our threading code more robust and I will have to think if this new MessageHandler class is going to make that more difficult that it would be with out that having been added.

markccchiang:
This PR can be postponed or discard. I can return to this work after @robsimmonds's works on messages/threading issues are done.

Issue #343, opened 2019-09-11
IMPORT_REGION_ACK.success == True even if region syntax in a file is invalid


kswang1029:
If there is invalid syntax in a region file, such as the following:

#CRTFv0 CASA Region Text Format version 0
symbol [[-123.8pix, **613.1**], .]
symbol [[-133.8pix, 613.1pix], .]

or

#CRTFv0 CASA Region Text Format version 0
symbol [[275.13653085deg, -16.17909524pix], .]
symbol [[275.13653085deg, -16.17909524deg], .]

currently all other valid regions are not loaded which is fine. However, IMPORT_REGION_ACK.success is still "True". 

The frontend also needs to show some error messages to users when IMPORT_REGION_ACK.success is False.

veggiesaurus:
I suggest we stick to the strict definition of `success` that we're currently using, because `success` is a binary field. This makes error handling much simpler: Either a request succeeds (and we utilise the response) or it fails (and we display the error message). 

I think a warning is a more appropriate solution here. We can use the `ERROR_DATA` stream with a severity of `WARNING` to display a toast on the frontend

kswang1029:
good idea. 👍

pford:
@kswang1029 the error for your examples (mixed world and pixel coordinates) throws an exception in the casacode CRTF parser, so none of the regions are imported.  I can catch the exception, set success=false, set the error in the ACK message, and send ERROR_DATA for the Log widget.  The message specifies the CRTF line number where the error occurred.

veggiesaurus:
@pford sorry, I think I misunderstood what @kswang1029 was saying. If the import completely fails, setting success=false and setting the error message is sufficient. I can use the error message in the response. No need for error data in this case. 

Perhaps if some succeed and others fail, we should return success and stream a warning?

pford:
If any definitions in the CRTF file fail, the parser throws an exception; there is no partial success.  If this is unacceptable, I will have to discuss this change with Dave for a casacode pull request for carta release 1.3.  The carta Ds9Parser will behave as you describe; I am working on improving the error messages for failed regions.

For me, the pop-up error messages disappear too quickly to read e.g. `test_error.crtf line# 3: _init: Mixed world and pixel coordinates not supported` to know what and where the issue is.  With error data, the user can go to the log and see what happened.

[...]

Issue #334, opened 2019-08-26
Cannot import/export regions when image has invalid or non-existent WCS headers


veggiesaurus:
Some images don't have WCS information. For example, the "LMCq.fits" file has a minimal header, shown below. As such, it has no WCS information. The backend should still be able to export regions defined for this file in _pixel_ coordinates (although it's totally fine for it to fail to export in WCS). Currently, the backend fails to export regions, and displays `Export carta region type 3 failed: AnnotationBase: Coordinate system has no direction coordinate`. 

LMCq.fits header:
```
NAXIS1 = 10482
CRVAL1 = 85.0539
CRPIX1 = 5241
CRVAL2 = -70.7087
NAXIS2 = 10448
NAXIS = 2
CRPIX2 = 5224
DATE-OBS = 2017-10-21T01:29:00
```

pford:
Merged fixes for seg fault in release 1.2 branch for patch.  Now sends "no direction coordinate" error message for wcs export.  Can export DS9 pixel coordinates, tested with region import in SAO DS9 viewer.

TBD: merge these changes into dev and fix pixel CRTF import/export and DS9 import.  Both use imageanalysis Annotation regions, which currently require a direction coordinate.

Kechil:
Document for now.

PR #356, opened 2019-10-02
Region import failures

pford:
Issues #334 and #343 

- Importing regions with no direction coordinate causes an exception (even for pixel imports, for now).  The image's coordinate system is checked and a warning issued, and the region import fails.
- When a region's syntax or definition fails, the import fails for CRTF region files but succeeds for DS9 region files if any regions are successfully imported; warnings are issued for the failed regions.
- If the region file does not exist, the import fails.

### DIFF
--- a/Ds9Parser.cc
+++ b/Ds9Parser.cc
@@ -667,7 +667,9 @@ void Ds9Parser::PrintBoxRegion(const RegionProperties& properties, std::ostream&
         casacore::Quantity cx(points[0]), cy(points[1]);
         casacore::Quantity width(points[2]), height(points[3]);
         // adjust width by cosine(declination) for correct export
-        width *= cos(cy);
+        if (width.isConform("rad")) {
+            width *= cos(cy);
+        }
         os << std::fixed << std::setprecision(6) << cx.get("deg").getValue() << ",";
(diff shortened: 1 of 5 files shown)


markccchiang:
@pford I am trying to understand the codes. So the remaining problem is that when a region's syntax or definition fails, such a file can not be imported if it is a CRTF region, but can be imported if it is a DS9 region?

pford:
@markccchiang that is correct.  For a CRTF import, IMPORT_REGION_ACK should set success=False with a message indicating the error.  For a DS9 import, IMPORT_REGION_ACK should set success=True, import the good regions, and set a message indicating the error as well as a log message (ERROR_DATA) with a warning indicating the failed region(s).  If the DS9 import should also entirely fail, @kswang1029 please let me know.

markccchiang:
According to @kswang1029, we should import good regions which have the right syntax or definitions. If there are bad regions with poor definitions, we should send a message indicating which ones are wrong. So for the importing of DS9 region file, there is no problem. But for the CRTF region file, maybe we have to modify the casacode class `RegionTextParser`, in order to fit this requirement?

pford:
This PR is to get the DS9 region implementation code into dev (already in release/1.2 branch), then I will contact Dave about modifying casacode for a separate PR.

PR #365, opened 2019-10-24
Angus/contour stream

veggiesaurus:
Note: This is a companion PR of [carta-frontend/#573](https://github.com/CARTAvis/carta-frontend/pull/573). It should not be merged until both PRs are approved.

This PR adds the following:
- Adds support for uncompressed contours when the contour compression level is 0. Contours are sent as floating point values, rather than being decimated, encoded and compressed
- Streams contours in chunks (the number of vertices per chunk is sent from the frontend). This provides better pipelining of contour calculation, compression and data delivery, and lets the user see progress as the contours are being delivered. 
- minor adjustments to Gaussian smoothing parameters to match DS9 behaviour

(note, this bumps the ICD version to 10. I will update the google doc after these two PRs have been approved, but before they have been merged)

### DIFF
--- a/Contouring.cc
+++ b/Contouring.cc
@@ -130,17 +130,21 @@ void TraceLevel(const float* image, int64_t width, int64_t height, double scale,
     vector<bool> visited(num_pixels);
     int64_t i, j;
 
+    auto test_for_chunk_overflow = [&]() {
+        if (vertex_cutoff && vertices.size() > vertex_cutoff) {
+            double progress = std::min(0.99, checked_pixels / double(num_pixels));
+            partial_callback(level, progress, vertices, indices);
+            vertices.clear();
+            indices.clear();
+        }
+    };
+
     // Search TopEdge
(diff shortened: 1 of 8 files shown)


Contouring.cc
```diff
@@ -133,31 +135,59 @@ void TraceLevel(const float* image, int64_t width, int64_t height, double scale,
         if (image[(j)*width + i] < level && level <= image[(j)*width + i + 1]) {
             indices.push_back(vertices.size());
             TraceSegment(image, visited, width, height, scale, offset, level, i, j, Edge::TopEdge, vertices);
+            if (vertex_cutoff && vertices.size() > vertex_cutoff) {
+                double progress = std::min(0.99, checked_pixels / double(num_pixels));
+                partial_callback(level, progress, vertices, indices);
+                vertices.clear();
+                indices.clear();
+            }
```
confluence: This seems like a lot of repetition. Could this not be put in a lambda?

Issue #172, opened 2019-04-08
Resume support for dropped connections


veggiesaurus:
Currently, when the server connection is dropped, the CARTA frontend and backend are in inconsistent states, as the previous session is not resumed, but rather replaced by a new session. 

The following features need to be added to support resumed sessions:
- [ ] Persisting sessions for a small duration (customisable by command line argument) after a "dirty" disconnect
- [x] Correctly detecting resume requests from the frontend
- [ ] Finding existing sessions with the same ID
- [x] Basic handshaking to ensure that states are consistent

This should not handle the much more difficult task of resuming a session after the backend _crashes_, but rather a situation where the client connection drops. 

Alternatively, work could instead be done to resume a session by sending all state stored in the frontend (open files, channels, view parameters, regions, requirements) to the backend, which would then construct a new session with the same state as the previous session. This would require the backend to re-open any files listed in the state and re-create any regions associated with these files.

kswang1029:
Tested. If the backend has been time out and users click the resume button on the GUI, nothing will happen and users see a loading bar forever. Reported to https://github.com/CARTAvis/carta-frontend/pull/617

veggiesaurus:
@kswang1029 is the backend still running?

kswang1029:
no. The case here would be that user did not click the resume button **before** the backend session is removed. The frontend should somehow provide a way to notify users that the session cannot be resumed.

veggiesaurus:
The backend _session_ or the backend process itself?

markccchiang:
When backend time out, the session is closed. The frontend should be aware
of the backend does not exist (green light on the top right of the panel
becomes red)?

Angus Comrie <notifications@github.com> 於 2019年12月11日 週三 下午2:42寫道：

veggiesaurus:
> When backend time out, the session is closed. The frontend should be aware of the backend does not exist (green light on the top right of the panel becomes red)?

@kswang1029 in this case, was the indicator light still green?

kswang1029:
@veggiesaurus @markccchiang allow me to try again to reproduce it and observe more carefully.

kswang1029:
ok, looks likely it was because the backend process was gone due to dev update. I just tried again with a long wait, session resume still works.

PR #375, opened 2019-11-22
Resume the session

markccchiang:
This is for issue #172. When the connection drops (due to the crash of frontend or backend, or the network problem). The session can be resumed by the frontend sending a recovering message to the backend. The message includes recovering image files, their channels, and regions on the backend. It is tricky to persist in a session for a period of time. So far the session is deleted once the `On Disconnect` is called. The timeout setting for uWcbsocket is 15 secs. That means after 15 secs of the network dropping down, the backend session will be deleted.


Main.cc
```diff
@@ -109,6 +107,20 @@ void OnDisconnect(uWS::WebSocket<uWS::SERVER>* ws, int code, char* message, size
...
+    switch ((long)user) {
+        case 3:
+            cerr << "Client emitted error on connection timeout (non-SSL)" << endl;
+            break;
+        case 5:
+            cerr << "Client emitted error on connection timeout (SSL)" << endl;
+            break;
+        default:
+            cerr << "FAILURE: " << user << " should not emit error!" << endl;
+            exit(-1);
```
veggiesaurus: does this mean that an error in one message from one user would crash the backend for all users?

veggiesaurus:
@kswang1029 spotted an issue when sending a resume message while calculating a cube historgram. @markccchiang could you look into this before we merge this branch in?

The code itself looks fine, so if we can just check that issue, then we can merge this in.

kswang1029:
It is “streaming spectral profile” causing the backend crash while resuming a session. For cube histogram, a session can be resumed but cube histogram will not be streamed anymore. Sounds like similar root cause?

Main.cc
markccchiang: I think it does. I will remove the line`exit(-1);` to avoid such a case.

markccchiang:
While the `OnResumeSession` called, the old streaming processes including spectral profile, cube histogram, and animation should be stopped. The old message queue should be cleared. @kswang1029 could you please test the updated branch solves the problem?

kswang1029:
@markccchiang @veggiesaurus 
streaming region spectral profile: session is resumed without crash, but region spectral profiler is not functional anymore. 

streaming animation: session is resumed without crash, but I see a persistent indicator of streaming tiles at the top-right corner. And we cannot make new regions anymore.

streaming per-cube histogram: session is resumed without crash, but once I zoom the image, I see a persistent indicator of streaming tiles at the top-right corner. And we cannot make new regions anymore.

kswang1029:
@markccchiang @veggiesaurus 
with d2944a1, all symptoms are gone! 👍👍👍

veggiesaurus (approved):

PR #390, opened 2019-12-04
Update backend for new casa repo

pford:
Updated CMakeLists.txt, README, and Dockerfile for carta-casacore repo which builds libcasa_imageanalysis.

While I was at it, I updated the version to "1.3" for the next release.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.0
 jobs:
   "ubuntu":
     docker:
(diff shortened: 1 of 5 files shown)


ajm-ska:
I decided to make new Docker containers for the Linux CI, both to keep their overall size small, plus I think when we merge, Github triggers builds of the old and new branches, so it will be able to run and pass the tests by using the old and new containers. Although maybe I'm wrong about that. Anyway, I also updated the .circleci/config.yml to pick up the new Docker containers and the new -lcasa_imageanalysis flag.

PR #397, opened 2019-12-12
Pam/13 fits header fixes

pford:


pford:
This PR has no overlap with #398 so merge should not affect it.

PR #398, opened 2019-12-12
Angus/mean sigma calcs

veggiesaurus:
This PR adjusts the simple min/max calculations done before calculating histograms to include mean and standard deviation. I've renamed the `minmax` classes to `BasicStats` to reflect this change. 

Calculating the mean and std dev adds about 10% to the total histogram time, but the additional utility it provides is probably worth it.

(edit: my version of `clang-tidy` was being overzealous, so I haven't done a reformatting pass before this PR.


Region/BasicStatsCalculator.tcc
```diff
@@ -0,0 +1,100 @@
...
+template<typename T>
+void BasicStats<T>::join(BasicStats<T>& other) {
+    if (other.num_pixels) {
+        sum += other.sum;
+        sumSq += other.sumSq;
+        num_pixels += other.num_pixels;
+        min_val = std::min(min_val, other.min_val);
+        max_val = std::max(max_val, other.max_val);
+        mean = sum / num_pixels;
+        stdDev = sqrt(sumSq / num_pixels - mean * mean);
```
confluence: In the HDF5 loader calculation for the standard deviation I use a slightly different formula for consistency with casacore: `sqrt((sum_sq - (sum * sum / num_pixels)) / (num_pixels - 1))`, rather than `sqrt((sum_sq - (sum * sum / num_pixels)) / num_pixels)`, which is equivalent to your formula. The relevant function in casacore is `LattStatsSpecialize::getSigma`, which calls `LattStatsSpecialize::getVariance`. Elsewhere in casacore your formula is used, but I'm pretty sure that the variant is always used for lattice regions, and that I used it to make the results match casacore's.

Region/BasicStatsCalculator.tcc
```diff
@@ -0,0 +1,100 @@
...
+    : _min_val(std::numeric_limits<T>::max()), _max_val(std::numeric_limits<T>::lowest()), _sum(0), _sum_squares(0), _num_pixels(0), _data(mm._data) {}
+
+template<typename T>
+void BasicStatsCalculator<T>::operator()(const tbb::blocked_range<size_t>& r) {
+    T t_min = _min_val;
+    T t_max = _max_val;
+    for (size_t i = r.begin(); i != r.end(); ++i) {
+        T val = _data[i];
+        if (std::isfinite(val)) {
+            if (val < t_min) {
```
confluence, line 46: This is going to trigger the same bug as the converter: *both* min and max have to be initialised to the first non-nan value found, otherwise under certain contrived circumstances one of them will end up being incorrect.

confluence: (Actually maybe less contrived in a small region -- because there will be fewer values in total.)

Region/BasicStatsCalculator.tcc
```diff
@@ -0,0 +1,100 @@
...
+
+template<typename T>
+BasicStats<T> BasicStatsCalculator<T>::GetStats() const {
+    double mean;
+    double stdDev;
+    double rms;
+
+    if (_num_pixels > 0) {
+        mean = _sum / _num_pixels;
+        stdDev = sqrt(_sum_squares / _num_pixels - mean * mean);
```
confluence: See other comment about the sigma.

confluence (changes requested):
The substitution looks straightforward, but there are two small issues with the calculations.

Region/BasicStatsCalculator.tcc
confluence: Suggestion for a fix: use something similar to the lambda solution from the converter, but to make it threadsafe make the initial lambda the explicit independent min/max check which also sets the lambda to the lazy check. It doesn't matter if multiple threads call it; the result will be the same.

confluence: Hmm, this function is actually called on independent slices of the data, and the stats are aggregated later, right? Then the solution doesn't need to be threadsafe and you can do exactly what the converter does.

veggiesaurus: I can't imagine that the lambda approach would really be faster than two comparisons.

The issue in the converter was because we were doing an `if...else`. Here it's an `if...if`, so we don't have the same problem

Region/BasicStatsCalculator.tcc
veggiesaurus: will change it to the sample variance instead of population variance. We should compare if `_num_pixels > 1` in that case, to avoid division by zero

Region/BasicStatsCalculator.tcc
confluence: Oh, I missed that. Never mind, then. :)

Region/BasicStatsCalculator.tcc
confluence: Casacore checks if it's `> 1.5` -- not sure why.

Region/BasicStatsCalculator.tcc
```diff
@@ -0,0 +1,100 @@
...
+template<typename T>
+void BasicStats<T>::join(BasicStats<T>& other) {
+    if (other.num_pixels) {
+        sum += other.sum;
+        sumSq += other.sumSq;
+        num_pixels += other.num_pixels;
+        min_val = std::min(min_val, other.min_val);
+        max_val = std::max(max_val, other.max_val);
+        mean = sum / num_pixels;
+        stdDev = num_pixels > 1 ? sqrt((sumSq - (sum * sum / num_pixels)) / (num_pixels - 1)): NAN;
```
confluence, line 17: Casacore returns zero if the number of pixels is <= 1.5; not sure if we absolutely have to match that behaviour.

confluence (approved):
This resolves my issues -- but see note about the exact casacore behaviour.

Region/BasicStatsCalculator.tcc
veggiesaurus: that's a good question. What's the standard deviation of a sample of one?

veggiesaurus: if the _population_ is 1, then the standard deviation should be zero, that makes sense. But casacore is using the _sample_ standard deviation (I'm not 100% sure why, or whether we should be using the sample std dev for regions, and population std dev for entire images), in which case the standard deviation of a sample would be undefined.

veggiesaurus: paging @kswang1029 for information on when it makes sense to use sample versus population when it comes to regions containing a subset of the image, and regions containing the entire image.

confluence: Not sure where the 1.5 comes from. `n` is a double, but (surely) is an integer value -- maybe this is avoiding a floating point precision error.

pford (approved):
FYI at the F2F we discussed refactoring Frame to remove all this ugly logic to check the cached values and the loaded values stored in FileLoader, and if not found do the calculations and cache the results.  For now this is fine, mostly name changes.  Adrianna already reviewed the calculations compared to casacore.

PR #400, opened 2019-12-13
Send masked data as NaN

pford:
For CASA and MIRIAD images, masks were being ignored and masked data was being shown.  Now masked data is set to NaN.

Also changed Frame calls which get image ptr then call an image method to FileLoader methods (FileLoader::GetSlice, FileLoader::GetCoordinateSystem).


confluence (changes requested):
The code looks fine, but there's a lot of repetition which could be eliminated if we implemented more functionality in the parent class:

* `GetCoordinateSystem` is identical for all loaders, so that can straightforwardly be put in `FileLoader` and removed from all the child loaders.
* Both `GetSlice` and `GetMaskSlice` could be moved into the parent as well, with the part of `GetSlice` which applies the mask conditional on some kind of flag.
* We could add a custom boolean attribute which is only set to true in the Miriad and Casa loaders, but is there perhaps already a flag on the casacore image object which indicates whether the image supports masks?
* (Our `GetMaskSlice` could then probably also be eliminated, since we already check in `GetSlice` whether the image exists.)

pford:
I renamed `LoadData(FileInfo::Data ds)` to `GetLoaderImage()` since the previous method name was misleading and it was always called with `FileInfo::Data::Image`.  If the image has not been opened with `OpenFile`, this returns a nullptr.

Then I moved `GetImageShape`, `GetCoordinateSystem`, `FindCoordinateAxes` (formerly `FindShape`), and `GetSlice` from child classes to FileLoader.

Frame.cc
```diff
@@ -730,7 +744,7 @@ void Frame::SetImageCache() {
     casacore::Slicer section = GetChannelMatrixSlicer(_channel_index, _stokes_index);
     casacore::Array<float> tmp(section.length(), _image_cache.data(), casacore::StorageInitPolicy::SHARE);
     std::lock_guard<std::mutex> guard(_image_mutex);
-    _loader->LoadData(FileInfo::Data::Image)->getSlice(tmp, section, true);
+    _loader->GetSlice(tmp, section, false);
```
confluence: Has the flag value been changed here deliberately, or is this a C&P error?

Frame.cc
```diff
@@ -740,7 +754,7 @@ void Frame::GetChannelMatrix(std::vector<float>& chan_matrix, size_t channel, si
     casacore::Array<float> tmp(section.length(), chan_matrix.data(), casacore::StorageInitPolicy::SHARE);
     // slice image data
     std::lock_guard<std::mutex> guard(_image_mutex);
-    _loader->LoadData(FileInfo::Data::Image)->getSlice(tmp, section, true);
+    _loader->GetSlice(tmp, section, false);
```
confluence: See above.

ImageData/FileLoader.h
```diff
@@ -128,54 +127,65 @@ class FileLoader {
...
     // check for mirlib (MIRIAD) error; returns true for other image types
     virtual bool CanOpenFile(std::string& error);
+    // Do anything required to open the file at given HDU (set up cache size, Image object)
+    virtual void OpenFile(const std::string& hdu) = 0;
 
-    // get shape and axis information from image data and coordinate system
-    bool FindShape(IPos& shape, int& spectral_axis, int& stokes_axis, std::string& message);
-    void FindCoordinates(int& spectral_axis, int& stokes_axis);
+    // Return the opened casacore image
+    virtual ImageRef GetLoaderImage() = 0;
```
confluence: I think renaming this is definitely a good idea, but I would remove the `Loader` and just call it `GetImage` -- it seems redundant given that it will be called on an object which will probably be called `_loader`, and this is consistent with `GetSlice`, etc..

confluence (changes requested):
Minor comment about the renaming, and a query about a boolean flag.

Frame.cc
pford: Changed deliberately. I was getting a casacore Array conformance error (shapes did not match) when degenerate axes were removed.

ImageData/FileLoader.h
pford: OK

confluence (approved):
I think this is all fine now.

pford:
I missed one GetSlice call.  Removed the boolean argument since it is the default.

PR #407, opened 2019-12-23
Angus/server user prefs

robsimmonds:
Added backend support for saving and restoring layouts and preferences


duidae (approved):

CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ set(LINK_LIBS
         zfp
         zstd
         tbb
+	json-c
```
veggiesaurus: tab/space inconsistency here

DBConnect.cc
```diff
@@ -54,4 +58,310 @@ void ConnectToMongoDB() {
     mongoc_cleanup();
 }
 
+bool SaveLayoutToDB(const std::string& name, const std::string& json_string) {
+    const char* uri_string = "mongodb://localhost:27017";
```
veggiesaurus: do you think we should make the mongo address an optional argument? If servers run multiple MongoDB instances, this port might be unavailable. In addition, if this DB is running on a separate node, we'd need to provide the address as well

Session.cc
```diff
@@ -214,7 +219,21 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
...
     ack_message.set_session_type(type);
-    ack_message.set_server_feature_flags(CARTA::ServerFeatureFlags::SERVER_FEATURE_NONE);
+
+    uint32_t feature_flags = CARTA::ServerFeatureFlags::REGION_WRITE_ACCESS;
+#ifdef _AUTH_SERVER_
+    feature_flags |= CARTA::ServerFeatureFlags::USER_LAYOUTS;
+    feature_flags |= CARTA::ServerFeatureFlags::USER_PREFERENCES;
+
+    if (!GetLayoutsFromDB(&ack_message)) {
+        // Can log failure here
```
veggiesaurus, line 229: should we use the error stream for sending this message to the frontend?

Session.h
```diff
@@ -236,6 +242,9 @@ class Session {
 
     tbb::task_group_context _animation_context;
 
+    // Name of server version users.
+    //    string username;
```
veggiesaurus: can we remove this?

DBConnect.cc
```diff
@@ -54,4 +58,310 @@ void ConnectToMongoDB() {
...
+    bool retval;
+    char* str;
+    char user[50];
+    bool result = true;
+
+    mongoc_init();
+
+    uri = mongoc_uri_new_with_error(uri_string, &error);
+    if (!uri) {
+        fprintf(stderr,
```
confluence: Should we perhaps standardise on using `cout` and `cerr` everywhere rather than `fprintf`?

DBConnect.cc
```diff
@@ -54,4 +58,310 @@ void ConnectToMongoDB() {
...
+
+    cuserid(user);
+
+    insert = bson_new();
+
+    bson_init(&layout);
+    BSON_APPEND_UTF8(&layout, "username", user);
+    BSON_APPEND_UTF8(&layout, "name", name.c_str());
+    
+    if( !strcmp(json_string.c_str(), "") ) {
```
confluence: Could do `json_string.empty()` here.

DBConnect.cc
```diff
@@ -54,4 +58,310 @@ void ConnectToMongoDB() {
...
 
+bool SaveLayoutToDB(const std::string& name, const std::string& json_string) {
+    const char* uri_string = "mongodb://localhost:27017";
+    mongoc_uri_t* uri;
+    mongoc_client_t* client;
+    mongoc_database_t* database;
+    mongoc_collection_t* collection;
+    bson_t *command, reply, *insert, layout, *query;
+    bson_error_t error;
+    bool retval;
```
confluence: I would either eliminate this intermediate variable entirely or give it a more descriptive name (e.g. `command_success`). But I'm leaning towards eliminating it and just doing the function call inside the `if (...)`. (Edited for clarity.)

DBConnect.cc
confluence: Should this not be stored in a configuration file, rather than hardcoded?

DBConnect.cc
```diff
@@ -54,4 +58,310 @@ void ConnectToMongoDB() {
...
+
+    mongoc_init();
+
+    uri = mongoc_uri_new_with_error(uri_string, &error);
+    if (!uri) {
+        fprintf(stderr,
+            "failed to parse URI: %s\n"
+            "error message:       %s\n",
+            uri_string, error.message);
+        return false;
```
confluence: We should consistently use `EXIT_FAILURE`.

DBConnect.cc
```diff
@@ -54,4 +58,310 @@ void ConnectToMongoDB() {
...
+
+    database = mongoc_client_get_database(client, "CARTA");
+    collection = mongoc_client_get_collection(client, "CARTA", "preferences");
+
+    command = BCON_NEW("ping", BCON_INT32(1));
+    retval = mongoc_client_command_simple(client, "admin", command, NULL, &reply, &error);
+    if (!retval) {
+        fprintf(stderr, "%s\n", error.message);
+        return EXIT_FAILURE;
+    }
```
confluence: This is near-identical in multiple functions. Could we not factor it out into another function which we can call with the command?

Main.cc
```diff
@@ -310,6 +310,25 @@ void OnMessage(uWS::WebSocket<uWS::SERVER>* ws, char* raw_message, size_t length
...
+                    if (message.ParseFromArray(event_buf, event_length)) {
+                        session->OnSetUserPreferences(message, head.request_id);
+                    } else {
+                        fmt::print("Bad SET_USER_PREFERENCES message!\n");
+                    }
+                    break;
+                }
+                case CARTA::EventType::SET_USER_LAYOUT: {
+                    CARTA::SetUserLayout message;
+                    std::cerr << "SET_USER_LAYOUT" << std::endl;
```
confluence: Stray debugging message?

confluence (changes requested):

DBConnect.cc
veggiesaurus: We should use fmt::print for consistency, and cout if absolutely necessary.

DBConnect.cc
robsimmonds: The backend is not using a config file currently. We could add one again, though it is something else to manage. I could put this as a command like option. Quite a few server programs have configs managed by things like initd which passed them arguments.

veggiesaurus: I think a command line option is ok

DBConnect.cc
confluence: I'm retracting this; it was a problem when it was an inconsistency, but now that we're using true/false consistently that is also fine.

DBConnect.cc
```diff
@@ -54,4 +86,216 @@ void ConnectToMongoDB() {
...
+bool GetLayoutsFromDB(CARTA::RegisterViewerAck* ack_message_ptr) {
+    mongoc_client_t* client;
+    mongoc_database_t* database;
+    mongoc_collection_t* collection;
+    mongoc_cursor_t* cursor;
+    bson_t *insert, *query;
+    const bson_t* doc;
+    char* str;
+    char* uvalue = NULL;
+    bool retval;
```
confluence: Some of these variables are declared but never used. Not sure if `clang-tidy` will pick these up.

Session.cc
robsimmonds: That would be good, but I am not sure how to do it.

DBConnect.cc
```diff
@@ -1,45 +1,77 @@
...
+    bson_error_t error;
+    std::string uri_string;
+
+    GetMongoURIString(uri_string);
+
+    mongoc_init();
+
+    uri = mongoc_uri_new_with_error(uri_string.c_str(), &error);
+    if (!uri) {
+        fmt::format(
```
confluence: Doesn't `format` just return a string, and `print` print?

veggiesaurus: Yep

DBConnect.cc
confluence: The unused `retval`s are gone, but there are still unused `uvalue`s.

confluence (approved):

PR #409, opened 2019-12-31
Missing data in large image

pford:
For very large image frames, the backend was improperly handling the entire data and mask arrays (probable memory error with very large arrays).  FileLoader::GetSlice was refactored to use a masked lattice iterator to apply the mask to data in chunks.  Also added code to catch bad_alloc when resizing the image cache vector.


confluence (approved):
Everything seems fine here.

Issue #404, opened 2019-12-19
MIRIAD image issues: velocities, masks, etc.


pford:
See notes in #374 and #377, which are being revised to refer to CASA only.  The casacore::MIRIADImage is incomplete, resulting in an incorrect FITS header sent to the frontend and no mask handling.

- [ ] incorrect frequency conversion for spectral axis results in incorrect velocity in render widget heading
- [ ] projection is incorrect
- [ ] masks are ignored

I recommend creating a CartaMiriadImage like the CartaHdf5Image, in which the miriad headers are read with mirlib and used to set up the image, and masks are detected and available.

kswang1029:
@pford 
@veggiesaurus and I had some discussions about the unification of headers and we think we should not perform any conversion for the frequency axis. If a miriad image or a  fits image has CTYPE3 = VELO, we should keep it as VELO. 

If you think we need a separate issue for this, please do.

pford:
@kswang1029:

Re: frequency/velocity conversion:
I opened PR #410 .  However, with the casacore SpectralCoordinate, a VELO axis will result in a "native type" of either VRAD or VOPT.  If this is unacceptable, we will have to go back to reading headers with third-party libraries for each image type (HDF5, mirlib, wcslib, casacore).

Re: MIRIAD projection in issue #377 
There is code in casacore::MIRIADImage which reassigns NCP to SIN with parameters; a log message says "NCP projection is now SIN projection in WCS. miriad readers will not handle this correctly." -- ???  I can put in a workaround to read the header with mirlib and restore the original projection type.

PR #410, opened 2019-12-31
Use native type for spectral axis headers, increase precision

pford:
Response to a comment (not the main issue) in #404.  The file info created by casacore::ImageFITSConverter::ImageHeaderToFITS was previously set to prefer frequency over velocity for the spectral axis headers.  This has been changed to check the native type.  For velocity, there is an extra check for doppler type radio/optical, to work around mirlib which defaults to optical velocity when there is no VELREF header (so a VRAD ctype is returned as native type = VOPT).

Float header string values have been changed to .12E precision except pixel and equinox/epoch values .3f.


confluence (approved):
This looks fine.

PR #414, opened 2020-01-08
carta miriad image

pford:
This image class derived from casacore::MIRIADImage addresses its missing implementation of masks and native spectral type.

The file info checks for native NCP projection conversion to SIN (in all image types).  When the direction coordinate projection in the image coordinate system is SIN, the backend checks whether the projection is equivalent to NCP (projection parameter 1 is 1/tan(dec)) and displays the projection as "SIN / NCP".  This is the best approximation unless we read the actual headers for all image types (e.g. FITS), which we are trying to avoid with the unified file info for all image types.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,6 +84,7 @@ set(SOURCE_FILES
         ImageData/FileLoader.cc
         ImageData/Hdf5Loader.cc
         ImageData/CartaHdf5Image.cc
+        ImageData/CartaMiriadImage.cc
         FileList/FileExtInfoLoader.cc
         FileList/FileInfoLoader.cc
         FileList/FileListHandler.cc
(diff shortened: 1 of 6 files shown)


kswang1029:
@pford we will need to make sure the frontend AST lib can understand "SIN / NCP" since we use this info to generate coordinate grid. Need to double check. If the frontend trim it ("/ NCP" as comment), then we should be fine.

kswang1029:
@pford if we load a miriad cube with mask and we start animation, the backend will crash.

confluence (approved):
The code looks OK.

kswang1029:
> The code looks OK.

even when animating?

confluence:
> > The code looks OK.
> 
> even when animating?

No, sorry, I was just reviewing the structure. I haven't looked into why it's broken -- I have added you as a reviewer to make sure that we resolve this issue before merging.

pford:
@kswang1029 this is ready to review/test

kswang1029:
@pford all good! 👍

Issue #383, opened 2019-11-27
backend support of catalogue overlay


kswang1029:
placeholder of all backend work related to catalogue overlay (frontend #571). Joint efforts from @markccchiang and @zarda, and discussions with @panchyo0 and @veggiesaurus

markccchiang:
We may have a document to specify what data that frontend needs from the backend with respect to their protobuf messages.

veggiesaurus:
@panchyo0 has put together a design document [here](https://docs.google.com/document/d/1DD-_AXg7Nk5fzBAXLdXqkAOqyV-R40Wep7X2K0ymp4A/edit#heading=h.get8b5g447o), which should server as a good starting point. @panchyo0, can you move the document into the CARTA team drive's technical documents folder?

veggiesaurus:
@markccchiang @zarda I think the best starting point is to work out what our options are in terms of loading and parsing VOTables. Given that VOTables are simply a specific schema of XML, we have a few options:
1) use an existing VOTable library for c++ (there are very few, and none are maintained)
2) use an XML library for c++ (there are lots of excellent, well-maintained ones) and handle the "VO" stuff ourself
3) delegate to a CASA library if one exists. I've not investigated that, but it would make me a bit nervous to delegate to CASA unless their VOTable loading/parsing is widely used, well-documented and compatible.

My personal preference is (2), but am open to suggestions. Xerces, libXML2 or tinyXML seem like viable options.

panchyo0:
The document is under CARTA team drive's technical documents folder/ frontend development/catalog overlay

markccchiang:
Looks Xerces has the most support of XML parsing (SAX and DOM), but it is a little bulky. TinyXML is very tiny, not sure does it fit our future requirements. LibXML2 looks promising so far. 

@kswang1029 mentioned two VOTable tools: [VO-India](http://voi.iucaa.in/voi/downloads.htm) and [VOTPP](http://ross.iasfbo.inaf.it/votpp/main.html). For the first link, it contains some java based libraries which are not for us. It also contains some C++ libraries. I do not read them in detail but they look outdated. 

For the second link VOTPP, it is C++ based and the version is 2007 (0.3.2-alpha1). It needs to work with another library [MCS](http://ross.iasfbo.inaf.it/mcs/). So it seems not suitable for us. But I found the code is not so large and it also utilizes LibXML2 which provides a good reference for us. I may try to write the VOTable parser based on its concept.

[...]

PR #443, opened 2020-02-21
Support the catalog overlay

markccchiang:
This is for issue #383. The backend support of opening catalog files, parsing and filtering the VOTable data. A few things to be done in the future:

- Add image bounds and region settings in the data filter
- Find a good way to deal with very large files, for example, loading partial data at the first of opening files and then load more data if necessary
- Benchmark the performance


veggiesaurus (changes requested):
some conflicts with the protobuf changes

Due to complexity of the PR (and the frontend PR), and the build changes required, we should merge them after the code freeze

Main.cc
```diff
@@ -460,11 +498,15 @@ int main(int argc, const char* argv[]) {
 
         session_number = 0;
 
-        websocket_hub.onMessage(&OnMessage);
-        websocket_hub.onConnection(&OnConnect);
-        websocket_hub.onDisconnection(&OnDisconnect);
+        // Create a hub group and pass the PERMESSAGE_DEFLATE option
+        group = websocket_hub.createGroup<uWS::SERVER>(uWS::PERMESSAGE_DEFLATE);
```
veggiesaurus: I think you can do this without creating a group explicitly, by just passing these arguments to the hub constructor directly.

Session.cc
```diff
@@ -1243,7 +1283,7 @@ void Session::SendEvent(CARTA::EventType event_type, uint32_t event_id, google::
     head->icd_version = carta::ICD_VERSION;
     head->request_id = event_id;
     message.SerializeToArray(msg.data() + sizeof(carta::EventHeader), message_length);
-    _out_msgs.push(msg);
+    _out_msgs.push(std::make_pair(msg, compress));
```
veggiesaurus: doesn't this result in an additional copy? you should rather form the pair first, and then serialize to the pair's `msg.data()`

Session.cc
```diff
@@ -1257,10 +1297,14 @@ void Session::SendFileEvent(int32_t file_id, CARTA::EventType event_type, uint32
 void Session::SendPendingMessages() {
     // Do not parallelize: this must be done serially
     // due to the constraints of uWS.
-    std::vector<char> msg;
+    std::pair<std::vector<char>, bool> msg;
     if (_connected) {
         while (_out_msgs.try_pop(msg)) {
-            _socket->send(msg.data(), msg.size(), uWS::BINARY);
+            if (msg.second) { // Compress the message
```
veggiesaurus: this can be simplified to just `_socket->send(msg.first.data(), msg.first.size(), uWS::BINARY, nullptr, nullptr, msg.second);`

markccchiang:
@veggiesaurus I have changed codes by your suggestions. But wondering is it the problem of the frontend's version, the image can not be shown on the viewer on my own test(?)

veggiesaurus:
> @veggiesaurus I have changed codes by your suggestions. But wondering is it the problem of the frontend's version, the image can not be shown on the viewer on my own test(?)

I noticed that earlier on mix 'n match. Odd.

markccchiang:
Thanks! With the new version of Qi's catalog branch, the image is back on the viewer. Probably the backend and frontend use different protobuf version.

PR #444, opened 2020-02-21
Add flux density support to selected files

### LINKED ISSUE #424: flux density of a region spectral profile
now with the frontend branch done by @TienHao 
origin/thchang/add_flux_density_602 009dd6e 2020-01-15 16:22:39 CST
we can request the backend to provide statistic="flux density" of a region spectral profile. However, this works only when using statistics widget. With the spectral profile widget, it returns nothing.

confluence:
This fixes #424.

I have merged Pam's branch `424_spectral_flux` into this branch, and added an implementation of the flux density calculation to the HDF5 stats calculation. I have tested a few files to check that we match the level of support that we currently have for FITS files, but I'm not sure if this simple calculation covers all the cases that we want to cover.

Flux density should now work everywhere in supported files, including the stats widget.

This PR also includes a change to the HDF5 attribute parsing code -- we now use scientific notation when we convert double values back into strings. Using fixed notation was causing a loss in precision which was very noticeable in the flux density calculation.

When testing HDF5 files, please check out the latest converter (dev or master branch) -- there's a bug in the older versions which drops the last entry from every header, which leads to weird behaviour if it's something important like `BUNIT`.

### DIFF
--- a/ImageData/FileLoader.cc
+++ b/ImageData/FileLoader.cc
@@ -525,7 +525,7 @@ void FileLoader::LoadImageStats(bool load_percentiles) {
                         stats[CARTA::StatsType::Mean] = sum / num_pixels;
                         stats[CARTA::StatsType::Sigma] = sqrt((sum_sq - (sum * sum / num_pixels)) / (num_pixels - 1));
                         stats[CARTA::StatsType::RMS] = sqrt(sum_sq / num_pixels);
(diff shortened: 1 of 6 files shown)


pford (approved):

veggiesaurus (approved):
code looks fine to me. Just waiting for @kswang1029 to check results accuracy

kswang1029:
@veggiesaurus @pford @confluence verified with CASA, FITS, MIRIAD, and HDF5 images, all consistent! 👍

PR #456, opened 2020-03-04
Check for coordinate axes failure in file info loader (1.3)

### LINKED ISSUE #450: crash when loading a MIRIAD image
Possibly due to that we switch to casacore for header processing.
[a.tar.gz](https://github.com/CARTAvis/carta-backend/files/4266942/a.tar.gz)

pford:


veggiesaurus:
@kswang1029 please confirm this fixes #450

kswang1029:
@veggiesaurus confirmed. Now an error message is displayed in file info and the backend does not crash.

PR #462, opened 2020-03-06
careful handling of NaN values while contouring

### LINKED ISSUE #461: crazy contours when switching between no smooth and box smooth
Seen in the mac v1.3beta and dev on mix n match

To reproduce:
1) load an image
2) create contours with "no smooth" mode
3) switch to "box" smooth and regenerate contours 

M100_combine_CO_cube image mom0 pbcor-image-2020-3-6-14-48-18

test image attached
[M100_12m_CO_central.image.mom0.fits.zip](https://github.com/CARTAvis/carta-frontend/files/4296898/M100_12m_CO_central.image.mom0.fits.zip)

veggiesaurus:
Fixes #461 

Will  backport to `release/1.3` if its deemed to be a low-risk change by @pford or @confluence. @kswang1029 please double check that it's producing correct contours.

### DIFF
--- a/Contouring.cc
+++ b/Contouring.cc
@@ -26,6 +26,10 @@ void TraceSegment(const float* image, std::vector<bool>& visited, int64_t width,
         double b = image[(j)*width + i + 1];
         double c = image[(j + 1) * width + i + 1];
         double d = image[(j + 1) * width + i];
+        a = isnan(a) ? -std::numeric_limits<float>::max() : a;
+        b = isnan(b) ? -std::numeric_limits<float>::max() : b;
+        c = isnan(c) ? -std::numeric_limits<float>::max() : c;
+        d = isnan(d) ? -std::numeric_limits<float>::max() : d;
 
         double x = 0;
         double y = 0;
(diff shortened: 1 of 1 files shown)


kswang1029:
@veggiesaurus works! weird contours cannot be produced anymore.

Contouring.cc
```diff
@@ -151,7 +160,12 @@ void TraceLevel(const float* image, int64_t width, int64_t height, double scale,
 
     // Search RightEdge
     for (j = 0; j < height - 1; j++) {
-        if (image[(j)*width + i] < level && level <= image[(j + 1) * width + i]) {
+        float pt_a = image[(j)*width + i];
+        float pt_b = image[(j + 1) * width + i];
+        pt_a = isnan(pt_a) ? -std::numeric_limits<float>::max() : pt_a;
```
pford: Why not just test for nan rather than reassign these values?
if (isnan(pt_a) || isnan(pt_b)) continue;
or
if (!isnan(pt_a) && !isnan(pt_b) && (pt_a > level && level >= pt_b)) {...}

pford:
I see no reason why this can't go into 1.3.  Makes sense to ignore nans for contours.

confluence (approved):
This seems like a low-risk change for backporting into the release.

Contouring.cc
veggiesaurus: simply skipping over `NaN` points will result in an infinite loop, because the `visited` array wouldn't get updated. 

However, you're right that I don't need to reassign these values. I've updated the code with the check 

`if ((isnan(pt_a) || pt_a < level) && level <= pt_b)`, since `level <= `NaN` will always be false, and if `pt_a` is a `NaN`, we still need to trace around it to update the `visited` array.

@confluence @pford please re-review, and @kswang1029 please confirm this still works after commit `26dbbf0`

confluence (approved):
Still seems OK.

PR #473, opened 2020-03-19
Calculate fail in large image

### LINKED ISSUE #466: regression: missing per-frame histogram with 51200x51200 image
@zarda found this issue

with set_QA_performance/cube_A/cube_A_51200_z00001.fits, histogram is missing
Screen Shot 2020-03-11 at 9 37 02 PM

with set_QA_performance/cube_A/cube_A_38400_z00001.fits, it works fine
Screen Shot 2020-03-11 at 9 38 36 PM

potentially due to OpenMP replacement? @robsimmonds @veggiesaurus

zarda:
Fixed #466, use size_t instead of int to break the limitation of min/max and histogram calculator. (fixed)

### DIFF
--- a/Region/BasicStatsCalculator.h
+++ b/Region/BasicStatsCalculator.h
@@ -13,7 +13,7 @@ namespace carta {
 
 template <typename T>
 struct BasicStats {
-    size_t num_pixels;
+    uint64_t num_pixels;
     double sum;
     double mean;
     double stdDev;
@@ -22,7 +22,7 @@ struct BasicStats {
     double rms;
     double sumSq;
 
-    BasicStats<T>(size_t num_pixels, double sum, double mean, double stdDev, T min_val, T max_val, double rms, double sumSq);
+    BasicStats<T>(uint64_t num_pixels, double sum, double mean, double stdDev, T min_val, T max_val, double rms, double sumSq);
     BasicStats<T>();
     void join(BasicStats<T>& other);
 };
@@ -31,16 +31,16 @@ template <typename T>
 class BasicStatsCalculator {
     T _min_val, _max_val;
     double _sum, _sum_squares;
-    size_t _num_pixels;
+    uint64_t _num_pixels;
     const std::vector<T>& _data;
 
 public:
     BasicStatsCalculator(const std::vector<T>& data);
     BasicStatsCalculator(BasicStatsCalculator& mm, tbb::split);
 
-    void operator()(const tbb::blocked_range<size_t>& r);
+    void operator()(const tbb::blocked_range<uint64_t>& r);
     void join(BasicStatsCalculator& other); // NOLINT
-    void reduce(const int start, const int end);
+    void reduce(const uint64_t start, const uint64_t end);
 
     BasicStats<T> GetStats() const;
 };
--- a/Region/BasicStatsCalculator.h
+++ b/Region/BasicStatsCalculator.h
@@ -13,7 +13,7 @@ namespace carta {
 
 template <typename T>
 struct BasicStats {
-    uint64_t num_pixels;
+    size_t num_pixels;
     double sum;
     double mean;
     double stdDev;
@@ -22,7 +22,7 @@ struct BasicStats {
     double rms;
     double sumSq;
 
-    BasicStats<T>(uint64_t num_pixels, double sum, double mean, double stdDev, T min_val, T max_val, double rms, double sumSq);
+    BasicStats<T>(size_t num_pixels, double sum, double mean, double stdDev, T min_val, T max_val, double rms, double sumSq);
     BasicStats<T>();
     void join(BasicStats<T>& other);
 };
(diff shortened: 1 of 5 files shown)


Region/BasicStatsCalculator.h
```diff
@@ -31,16 +31,16 @@ template <typename T>
 class BasicStatsCalculator {
     T _min_val, _max_val;
     double _sum, _sum_squares;
-    size_t _num_pixels;
```
veggiesaurus: Nothing wrong with using size_t, since it's just a #define for a 64bit int.

Region/Histogram.cc
```diff
@@ -13,40 +13,47 @@ Histogram::Histogram(int num_bins, float min_value, float max_value, const std::
 
 Histogram::Histogram(Histogram& h, tbb::split) : _bin_width(h._bin_width), _min_val(h._min_val), _hist(h._hist.size(), 0), _data(h._data) {}
 
-void Histogram::operator()(const tbb::blocked_range<size_t>& r) {
-    std::vector<int> tmp(_hist);
+void Histogram::operator()(const tbb::blocked_range<uint64_t>& r) {
+    std::vector<uint64_t> tmp(_hist);
```
veggiesaurus: Do we need to store the bins as 64bit? They get sent as 32bit ints anyway, don't they?

Region/Histogram.cc
```diff
@@ -13,40 +13,47 @@ Histogram::Histogram(int num_bins, float min_value, float max_value, const std::
...
     for (auto i = r.begin(); i != r.end(); ++i) {
         auto v = _data[i];
         if (std::isnan(v) || std::isinf(v))
             continue;
-        int bin = std::max(std::min((int)((v - _min_val) / _bin_width), (int)_hist.size() - 1), 0);
+        // int bin = std::max(std::min((int)((v - _min_val) / _bin_width), (int)_hist.size() - 1), 0);
+        uint64_t _1 = (v - _min_val) / _bin_width;
+        uint64_t _2 = _hist.size() - 1;
+        uint64_t _3 = _1 < _2 ? _1 : _2;
+        uint64_t bin = _3 >= 0 ? _3 : 0;
```
veggiesaurus: These changes aren't needed, are they? The new code is pretty ugly

veggiesaurus (changes requested):
Using size_t is not s problem, so there are lots of needless changes here.

Region/BasicStatsCalculator.h
zarda: got it. max size_t: 18446744073709551615

Region/Histogram.cc
zarda: Fixed

Region/Histogram.cc
zarda: Fixed

kswang1029:
@zarda your new commit breaks the histogram
Screen Shot 2020-03-19 at 3 12 33 PM

zarda:
Fixed the break of histogram. Thanks @kswang1029.

Region/Histogram.cc
```diff
@@ -30,23 +31,23 @@ void Histogram::join(Histogram& h) { // NOLINT
...
-        for (auto i = start; i < end; i++) {
+    auto calc_lambda = [&](size_t start, size_t lstride) {
+        size_t* lbins = new size_t[_hist.size()];
+        size_t end = std::min((size_t)(start + lstride), _data.size());
+        memset(lbins, 0, _hist.size() * sizeof(size_t));
+        for (size_t i = start; i < end; i++) {
             auto v = _data[i];
             if (std::isfinite(v)) {
-                int binN = std::max(std::min((int)((v - _min_val) / _bin_width), (int)_hist.size() - 1), 0);
+                size_t binN = std::max<size_t>(std::min<size_t>((size_t)((v - _min_val) / _bin_width), (size_t)_hist.size() - 1), 0);
```
veggiesaurus: this doesn't match our naming convention. Please use something like `bin_number`

veggiesaurus: (I realise you weren't responsible for the actual name :smile: )

Region/Histogram.cc
```diff
@@ -14,12 +14,13 @@ Histogram::Histogram(int num_bins, float min_value, float max_value, const std::
 Histogram::Histogram(Histogram& h, tbb::split) : _bin_width(h._bin_width), _min_val(h._min_val), _hist(h._hist.size(), 0), _data(h._data) {}
 
 void Histogram::operator()(const tbb::blocked_range<size_t>& r) {
-    std::vector<int> tmp(_hist);
+    std::vector<size_t> tmp(_hist);
```
veggiesaurus: aren't we still using 64-bit bins here, though? The question is whether we definitely need them or not.

zarda: The type size_t may reduce the performance. I suppose there might some extreme case while all bins are zero but one bin collects all counts. May I detect the size of incoming data to switch the type? I mean to assign type by template.

Region/Histogram.cc
zarda: Fixed

Region/Histogram.cc
veggiesaurus: I think trying to detect the size might make it slower (it will introduce a jump for each assigment. Given that we _already_ send the bins as `int32` in the protocol buffers, lets just stick to `int` here. If we find we need to adjust to 64-bit, we will need to do this for the protocol buffers as well (or switch to variable-length integers)

zarda: Got it. Fixed them.

veggiesaurus (approved):

PR #482, opened 2020-03-26
Set region states in the main thread

markccchiang:
It is to solve the stuck problem for region spectral profile calculations. By moving the setting of region states and its statistical requirements in the main thread. The region data streams or spectral profile calculations are done in TBB concurrent threads.

### DIFF
--- a/InterfaceConstants.h
+++ b/InterfaceConstants.h
@@ -6,6 +6,10 @@
 // version
 #define VERSION_ID "1.3"
 
+// thread counts
+#define TBB_THREAD_COUNT 2
+#define OMP_THREAD_COUNT 4
+
 // file ids
 #define ALL_FILES -1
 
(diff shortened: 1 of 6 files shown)


veggiesaurus:
this is going to be tricky to test, but @kswang1029 please try so. Once we're happy with it, we can also merge this into the release branch and deploy on IDIA's servers for some testing

kswang1029:
I tested with three regions and requested their profiles in three different widgets. It looks good. Region can be created while profile is streaming. I also tried moving regions randomly while profile is steaming and it looks good too. If I switch to a new channel during profile streaming, it queues up and will update the image after profile streaming is done. 

As far as I can see so far, it is fine. 

will keep an eye on this in the coming days.

veggiesaurus:
Ok then we can merge this once @robsimmonds and @pford have had a look at the PR

confluence (approved):
If this has been tested, I have no objections.

pford (approved):

PR #488, opened 2020-04-03
Reject spectral req for 2D image

### LINKED ISSUE #487: backend crashes when creating a region on a single channel image while the spectral profiler is launched
v1.3 bug report from https://github.com/CARTAvis/carta/issues/13

to reproduce,
1) load a single channel image
2) enable spectral profiler widget
3) create a region and the backend will crash

Update: if before 3) we select "cursor" in the region dropdown, the backend will survive for cursor but if we switch to "active", the backend will crash. Not sure if this is a frontend issue instead.

Update: if the image is W x H the backend will crash. However, if the image is W x H x 1 or W x H x 1 x 1, the backend will survive. Thanks @veggiesaurus for spotting this.

pford:


veggiesaurus (approved):

veggiesaurus:
looks fine. Can you also submit a PR to `dev`?

PR #492, opened 2020-04-15
Change the type of index in an OMP loop

markccchiang:
The main change is to assign the type `int64_t` for the index in an OMP loop. Using `size_t` would lead to a crash in the backend. The reason can be found in this link https://stackoverflow.com/questions/3988470/why-must-loop-variables-be-signed-in-a-parallel-for according to @kswang1029.

### DIFF
--- a/Region/Histogram.cc
+++ b/Region/Histogram.cc
@@ -36,7 +36,7 @@ void Histogram::join(Histogram& h) { // NOLINT
     tbb::parallel_for(range, loop);
 }
 
-void Histogram::setup_bins(const size_t start, const size_t end) {
+void Histogram::setup_bins() {
     int64_t i, stride, buckets;
     int** bins_bin;
 
--- a/Region/Histogram.cc
+++ b/Region/Histogram.cc
@@ -71,8 +71,8 @@ void Histogram::setup_bins() {
         {
             for (omp_index = 0; omp_index <= (buckets - stride * 2); omp_index += stride * 2) {
 #pragma omp task
-                std::transform(
-                    (bins_bin[omp_index + stride]), &(bins_bin[omp_index + stride][_hist.size()]), (bins_bin[omp_index]), (bins_bin[omp_index]), std::plus<int>());
+                std::transform((bins_bin[omp_index + stride]), &(bins_bin[omp_index + stride][_hist.size()]), (bins_bin[omp_index]),
+                    (bins_bin[omp_index]), std::plus<int>());
             }
             stride *= 2;
         }
--- a/Region/Histogram.cc
+++ b/Region/Histogram.cc
@@ -37,14 +37,14 @@ void Histogram::join(Histogram& h) { // NOLINT
 }
 
 void Histogram::setup_bins() {
-    int64_t i, stride, buckets;
+    int64_t omp_index, stride, buckets;
     int** bins_bin;
 
     auto calc_lambda = [&](size_t start, size_t lstride) {
         int* lbins = new int[_hist.size()];
         size_t end = std::min((size_t)(start + lstride), _data.size());
         memset(lbins, 0, _hist.size() * sizeof(int));
-        for (auto i = start; i < end; i++) {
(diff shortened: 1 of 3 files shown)


kswang1029:
in some other discussion threads, people mentioned it depends on the OMP version which might allow uint as loop index. Which OMP version do you use? @markccchiang @ajm-asiaa

kswang1029:
tested with 80k x 80k image and it works fine.

markccchiang:
@kswang1029 I used the OMP version `libomp: stable 9.0.1 (bottled) LLVM's OpenMP runtime library` on Mac.

ajm-ska:
@kswang1029 I'm not sure how @markccchiang was able to get the libomp version number readout like that. I'm afraid to try too much with Homebrew in case I accidentally update the libomp version. Anyway, the system that created the v1.3 carta-backend on MacOS appears to be using version 8.0.1  `/usr/local/Cellar/libomp/8.0.1/lib/libomp.dylib`

I wonder we could rule out the OMP version being the problem if @markccchiang takes the apparently working libomp.dylib from CARTA v1.3 (e.g. `/Applications/CARTA.app/Contents/Resources/app/carta-backend/libs/libomp.dylib`), puts it in a different directory, does the export LD_LIBRARY_PATH= line to that directory so that it will be searched first, then run his locally built carta-backend. I expect it should then use the older libomp.dylib (assuming libomp.dylib doesn't have any other dependencies).

kswang1029:
Thank you both. 👍

markccchiang:
> @kswang1029 I'm not sure how @markccchiang was able to get the libomp version number readout like that. I'm afraid to try too much with Homebrew in case I accidentally update the libomp version. Anyway, the system that created the v1.3 carta-backend on MacOS appears to be using version 8.0.1 `/usr/local/Cellar/libomp/8.0.1/lib/libomp.dylib`
> 
> I wonder we could rule out the OMP version being the problem if @markccchiang takes the apparently working libomp.dylib from CARTA v1.3 (e.g. `/Applications/CARTA.app/Contents/Resources/app/carta-backend/libs/libomp.dylib`), puts it in a different directory, does the export LD_LIBRARY_PATH= line to that directory so that it will be searched first, then run his locally built carta-backend. I expect it should then use the older libomp.dylib (assuming libomp.dylib doesn't have any other dependencies).

@ajm-asiaa you can type the command `brew info libomp` in the terminal.

ajm-ska:
Thanks @markccchiang . In that case, `brew info libomp` is showing `libomp: stable 9.0.1 (bottled) LLVM's OpenMP runtime library` So that must mean we are using the same version. Unless I have already gone and updated it while trying to check because it was definitely 8.0.1 before.

Region/Histogram.cc
```diff
@@ -36,15 +36,15 @@ void Histogram::join(Histogram& h) { // NOLINT
     tbb::parallel_for(range, loop);
 }
 
-void Histogram::setup_bins(const size_t start, const size_t end) {
-    size_t i, stride, buckets;
+void Histogram::setup_bins() {
+    int64_t i, stride, buckets;
```
veggiesaurus: while we're at it, can we change the name of one of the `i` variables, so that we don't have variable shadowing in the lamda?

Region/Histogram.cc
```diff
@@ -36,15 +36,15 @@ void Histogram::join(Histogram& h) { // NOLINT
...
     int** bins_bin;
 
     auto calc_lambda = [&](size_t start, size_t lstride) {
         int* lbins = new int[_hist.size()];
         size_t end = std::min((size_t)(start + lstride), _data.size());
         memset(lbins, 0, _hist.size() * sizeof(int));
-        for (size_t i = start; i < end; i++) {
+        for (auto i = start; i < end; i++) {
```
veggiesaurus: since `start` and `end` are both `size_t`, won't the auto end up as `size_t` as well? Although I suppose since it's not in a parallel for, it's fine

veggiesaurus (approved):

PR #498, opened 2020-04-17
Add try/catch around statistics for flux density calc

### LINKED ISSUE #486: backend hangs when requesting statistics
v1.3 bug report from https://github.com/CARTAvis/carta/issues/14

To reproduce with the attached example images:
1) load testb.mom0.fits
2) launch statistics widget and we should see statistics but we don't. 
3) append testa.mom0.fits, and we will see the tile streaming indicator blinks forever

[Archive.zip](https://github.com/CARTAvis/carta-backend/files/4425703/Archive.zip)

pford:
Return NaN if ImageStatistics throws an exception.  For release 1.3 patch issue #486 .


pford:
casacore::ImageStatistics was throwing an exception when setting the units in the flux Quantity due to the nonstandard brightness unit in the image.  This exception is caught and the stats value set to NaN.

kswang1029 (approved):
Tested with some images with different units. All are looking good.

PR #504, opened 2020-04-17
Fix to make it C++17 compatible the MacOS compiler

ajm-ska:
Attempt number two at this one line change to the CMakeLists.txt file to make it compatible with the MacOS compiler which needs `CMAKE_CXX_STANDARD 17` explicitly set when the code contains C++17 functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
 project(carta_backend)
 
-set(CMAKE_CXX_STANDARD 14)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
```
veggiesaurus: this should be 3.8, right?

ajm-ska: Thanks! Well spotted! I have just updated it now.

veggiesaurus (approved):

PR #515, opened 2020-04-29
Fix duplication of saved preferences and layouts in the database

confluence:
This PR prevents layout and preference entries from being copied repeatedly rather than updated. Existing duplicates must be cleaned up manually, as this code expects the database to be in a clean state. @veggiesaurus has written a script for this (to be attached below).

### DIFF
--- a/DBConnect.cc
+++ b/DBConnect.cc
@@ -115,7 +115,7 @@ bool SaveLayoutToDB(const std::string& name, const std::string& json_string) {
         bson_t* updated = bson_copy(&existing);
         BSON_APPEND_UTF8(updated, "json_string", json_string.c_str());
         
-        if (!mongoc_collection_update_one(collection, &existing, updated, &opts_upsert, NULL, &error)) {
+        if (!mongoc_collection_replace_one(collection, &existing, updated, &opts_upsert, NULL, &error)) {
             fmt::print("{}", error.message);
             result = false;
         }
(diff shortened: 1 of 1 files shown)


veggiesaurus:
Script for removing existing entry duplicates:

```
const MongoClient = require('mongodb').MongoClient;

const url = 'mongodb://localhost:27017';
const dbName = 'CARTA';
const client = new MongoClient(url, { useUnifiedTopology: true });

client.connect().then(async () => {
    console.log("Connected successfully to server");
    const db = client.db(dbName);
    const srcPreferences = db.collection("preferences");
    const destPreferences = db.collection("preferences");
    const docs = await srcPreferences.find().toArray();
    const prefMap = new Map();
    for (const doc of docs) {
        const username = doc.username;
        // assumes keys are always in order _id, username, <key>
        const key = Object.keys(doc)[2];
        const value = doc[key];

        let userPrefMap = prefMap.get(username);
        if (!userPrefMap) {
            userPrefMap = new Map();
            prefMap.set(username, userPrefMap);
        }
        userPrefMap.set(key, value);
    }

    // remove all old documents
    await destPreferences.deleteMany({});

    for (const userEntry of prefMap) {
        const username = userEntry[0];
        const userPrefMap = userEntry[1];
        process.stdout.write(`Fixing entries for user ${username}...`);        
        for (const entry of userPrefMap) {
            const obj = {username};
            obj[entry[0]] = entry[1];            
            await destPreferences.insertOne(obj);
        }
        console.log("Done");
    }
    client.close();
});
```

veggiesaurus (approved):

PR #516, opened 2020-04-29
Proposed backport: Fix duplication of saved preferences and layouts in the database (#515)

confluence:
This bug isn't obvious to the user, but does cause the database to keep growing. The fix is isolated to the AuthServer code and shouldn't affect anything else. Should this be backported?


kswang1029 (commented):
I think it is a good idea to include this in the 1.3.1 patch. 

I don't have an environment to "test" this.

kswang1029 (approved):

veggiesaurus (approved):

Issue #435, opened 2020-02-05
region analytics with multiple image overlay


kswang1029:
in v1.3 multiple images can be matched spatially and spectrally. Spatially this is achieved by translating, rotating, and scaling of raster tiles. As a consequence, region of interest needs to be enhanced to support multiple images involving the above approach of spatial matching.

Analytics including:
- [ ] spectral profile
- [ ] statistics
- [ ] histogram

pford:
Duplicate of #146 ?

Issue #146, opened 2019-04-04
WCS Group Region of Interest Analysis


Kechil:
Regions of interest need to be defined for a single WCS group, and analysis products from a region of interest (profile, stats, histogram) need to be determined for each required frame in the WCS group.

- [ ] Modify ICD to support WCS groups region of interest
- [ ] Convert region definitions between frames
- [ ] Determine analysis products based on updated region requirements
- [ ] Send analysis products for different files

veggiesaurus:
@pford I think we should arrange a time to chat about this feature. It would be good to start development on this before the rest of the WCS stuff (my side of things) is finished

pford:
Requirements doc: https://docs.google.com/document/d/1wVidLdGQOflxGPl4wb7X7v0FEJJx63IenXlkxkbbB40/edit#heading=h.7zehartallcq

Issue #372, opened 2019-11-04
missing PC header keywords from the backend


kswang1029:
The attached file is an image with rotation (thus the PC matrix is not unity).
[rotated.fits.zip](https://github.com/CARTAvis/carta-backend/files/3802874/rotated.fits.zip)

Its header looks like
SIMPLE  =                    T /Standard FITS                                   
BITPIX  =                  -32 /Floating point (32 bit)                         
NAXIS   =                    2                                                  
NAXIS1  =                  113                                                  
NAXIS2  =                   76                                                  
EXTEND  =                    T                                                  
BSCALE  =   1.000000000000E+00 /PHYSICAL = PIXEL*BSCALE + BZERO                 
BZERO   =   0.000000000000E+00                                                  
BMAJ    =   1.486111200000E-02                                                  
BMIN    =   9.499999700000E-03                                                  
BPA     =   6.000000000000E+00                                                  
BTYPE   = 'Intensity'                                                           
OBJECT  = '        '                                                            
BUNIT   = 'Jy/beam '           /Brightness (pixel) unit                         
EQUINOX =   2.000000000000E+03                                                  
RADESYS = 'FK5     '                                                            
LONPOLE =   1.800000000000E+02                                                  
LATPOLE =   0.000000000000E+00                                                  
PC1_1   =   7.071067811865E-01                                                  
PC2_1   =  -7.071067811865E-01                                                  
PC1_2   =   7.071067811865E-01                                                  
PC2_2   =   7.071067811865E-01                                                  
CTYPE1  = 'RA---SIN'                                                            
CRVAL1  =   0.000000000000E+00                                                  
CDELT1  =  -2.222222308810E-03                                                  
CRPIX1  =   5.600000000000E+01                                                  
CUNIT1  = 'deg     '                                                            
PV2_1   =   0.000000000000E+00                                                  
CTYPE2  = 'DEC--SIN'                                                            
CRVAL2  =   0.000000000000E+00                                                  
CDELT2  =   3.333333234031E-03                                                  
CRPIX2  =   3.800000000000E+01                                                  
CUNIT2  = 'deg     '                                                            
PV2_2   =   0.000000000000E+00                                                  
DATE    = '2019-11-04T06:30:53.875825' /Date FITS file was written              
TIMESYS = 'UTC     '           /Time system for HDU                             
ORIGIN  = 'CASA 5.5.0-149'                           
With ds9, it looks like
[...]

PR #517, opened 2020-04-29
Refactor region handling for WCS groups

### LINKED ISSUE #445: WCS regions: remove Region code from Frame
Currently, a Region belongs to a Frame so all region-handling code is in the Frame.  With WCS ROI, a region may be applied to any Frame in a WCS group, so it needs to be independent of a particular Frame until a data stream is needed. This means the file loaders and statistics calculators must be available to both.

The plan (in progress and subject to change) is documented [here](https://docs.google.com/document/d/1aN2ex1OHZ7_GQNVbagJpDNXnjQn8jIFIyiuIfQ8kq1Y/edit#).

The new architecture will make merges difficult if not impossible.  I plan to wait until major current development items from others are merged into dev, then manually add the changes into the new organization.

pford:
Closes #445 .

Very large architecture change, as indicated this cannot be automatically merged.  Once approved, I will do the merge in my sandbox and then push to dev.

The new architecture and its motivation is described in this doc in the shared drive backend development folder:
https://docs.google.com/document/d/1aN2ex1OHZ7_GQNVbagJpDNXnjQn8jIFIyiuIfQ8kq1Y/edit

Notes:
- I have manually added changes from all PRs merged since this branch was created.  Because of this, I would like to merge this in as soon as possible after basic testing (@kswang1029 ) rather than waiting for exhaustive testing, if the basic architecture change meets approval.
- Most changes are in Session, Frame, and the Region classes.  The HDF5Loader was changed to calculate a section of the spectral profile with the swizzled data, moving the outer loop (and cancellation checks) to the RegionHandler.  Message handling, file browsing, catalog overlays, etc., should not be changed.  Some code was rearranged into (import/export) or out of (statistics/histograms) the Region directory and out of the top directory to new DataStream and ImageStats directories.
- I have tested each feature with the region's reference image (the image it was created in) and will continue to test against the ICD tests to verify the refactor did not break current functionality while this PR is under review.
- Once merged, we can start testing applying regions to other images and note problems in the open related issues #146 and #435 or create new issues as needed.


kswang1029:
will do some testing

kswang1029:
@pford It appears to me that if we have one spectral profiler widget with region = "Active" and if we have two regions, the backend does not terminate calculations and start a new one if we switch from one region to another (need a larger (FITS) cube to see this easily). What I see is it queues up. The backend keeps delivering profiles of the 1st region and only starts to deliver profiles of the 2nd region when the 1st is done (progress=100%). This is tested with TBB=4, OMP=4 with the help from @ajm-asiaa 

In addition, it happened to me once that the backend crashed when I was switching regions to see their profiles.

> Num sessions: 1
*** Error in `./carta_backend': double free or corruption (!prev): 0x0000000000efa130 ***
Segmentation fault (core dumped)

kswang1029:
Somehow if we have two spectral profilers and a set of regions and we check profiles of different regions in different widgets and modify regions or switch regions while profiles are delivering, the backend will crash.

Backtrace (thanks to @ajm-asiaa)

> Num sessions: 1
*** Error in `./carta_backend': double free or corruption (!prev): 0x00000000012f5ef0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x81489)[0x7f99ae989489]
/usr/local/lib/libcasa_measures.so.3(_ZN8casacore9MeasFrameD2Ev+0x95)[0x7f99b18b62a5]
/usr/local/lib/libcasa_measures.so.3(_ZN8casacore11MeasConvertINS_8MDopplerEE6createEv+0x3ef)[0x7f99b18f6f9f]
/usr/local/lib/libcasa_measures.so.3(_ZN8casacore15VelocityMachine4initEv+0xc53)[0x7f99b1916743]
/usr/local/lib/libcasa_measures.so.3(_ZN8casacore15VelocityMachineC1ERKS0_+0x129)[0x7f99b19178a9]
/usr/local/lib/libcasa_coordinates.so.3(_ZN8casacore18SpectralCoordinate4copyERKS0_+0x1e3)[0x7f99b34bcf13]
/usr/local/lib/libcasa_coordinates.so.3(_ZN8casacore18SpectralCoordinateC1ERKS0_+0x15c)[0x7f99b34bd14c]
/usr/local/lib/libcasa_coordinates.so.3(_ZNK8casacore18SpectralCoordinate5cloneEv+0x21)[0x7f99b34bd1d1]
/usr/local/lib/libcasa_coordinates.so.3(_ZN8casacore16CoordinateSystem4copyERKS0_+0x148)[0x7f99b3461998]
/usr/local/lib/libcasa_coordinates.so.3(_ZN8casacore16CoordinateSystemC1ERKS0_+0x59a)[0x7f99b346265a]
/usr/local/lib/libcasa_images.so.3(_ZN8casacore14ImageInterfaceIfEC2ERKS1_+0x71)[0x7f99b271b2d1]
/usr/local/lib/libcasa_images.so.3(_ZN8casacore10PagedImageIfEC2ERKS1_+0x1a)[0x7f99b271b6ca]
/usr/local/lib/libcasa_images.so.3(_ZNK8casacore10PagedImageIfE7cloneIIEv+0x21)[0x7f99b271b7a1]
/usr/local/lib/libcasa_lattices.so.3(_ZN8casacore10SubLatticeIfEC1ERKNS_13MaskedLatticeIfEERKNS_13LatticeRegionENS_13AxesSpecifierE+0x69)[0x7f99b21f9dc9]
/usr/local/lib/libcasa_images.so.3(_ZN8casacore8SubImageIfEC2ERKNS_14ImageInterfaceIfEERKNS_16LattRegionHolderENS_13AxesSpecifierEb+0xbc)[0x7f99b272df2c]
./carta_backend[0x4a2deb]
./carta_backend[0x487420]
./carta_backend[0x4da4b4]
./carta_backend[0x4d8d83]
./carta_backend[0x46664c]
./carta_backend[0x467887]
./carta_backend[0x4678eb]
./carta_backend[0x55a74e]
/lib64/libtbb.so.2(+0x286c9)[0x7f99b3d076c9]
/lib64/libtbb.so.2(+0x2306e)[0x7f99b3d0206e]
/lib64/libtbb.so.2(+0x20f8f)[0x7f99b3cfff8f]
/lib64/libtbb.so.2(+0x1c616)[0x7f99b3cfb616]
/lib64/libtbb.so.2(+0x1c5a6)[0x7f99b3cfb5a6]
/lib64/libpthread.so.0(+0x7dd5)[0x7f99b0610dd5]
/lib64/libc.so.6(clone+0x6d)[0x7f99aea05ead]
======= Memory map: ========
00400000-0071d000 r-xp 00000000 fd:00 644814568                          /var/www/backend/c554d5a/carta_backend
0091c000-0091d000 r--p 0031c000 fd:00 644814568                          /var/www/backend/c554d5a/carta_backend
0091d000-00922000 rw-p 0031d000 fd:00 644814568                          /var/www/backend/c554d5a/carta_backend
00922000-00927000 rw-p 00000000 00:00 0 
01165000-039a9000 rw-p 00000000 00:00 0                                  [heap]
7f97c4000000-7f97c4021000 rw-p 00000000 00:00 0 
7f97c4021000-7f97c8000000 ---p 00000000 00:00 0 
7f97c8000000-7f97c8021000 rw-p 00000000 00:00 0 
7f97c8021000-7f97cc000000 ---p 00000000 00:00 0 
7f97cc000000-7f97cc021000 rw-p 00000000 00:00 0 
7f97cc021000-7f97d0000000 ---p 00000000 00:00 0 
7f97d4000000-7f97d4021000 rw-p 00000000 00:00 0 
7f97d4021000-7f97d8000000 ---p 00000000 00:00 0 
7f97e0000000-7f97e0021000 rw-p 00000000 00:00 0 
7f97e0021000-7f97e4000000 ---p 00000000 00:00 0 
7f97e4000000-7f97e7feb000 rw-p 00000000 00:00 0 
7f97e7feb000-7f97e8000000 ---p 00000000 00:00 0 
7f97e8000000-7f97eac35000 rw-p 00000000 00:00 0 
7f97eac35000-7f97ec000000 ---p 00000000 00:00 0 
7f97ec000000-7f97ec021000 rw-p 00000000 00:00 0 
7f97ec021000-7f97f0000000 ---p 00000000 00:00 0 
7f97f0000000-7f97f144b000 rw-p 00000000 00:00 0 
7f97f144b000-7f97f4000000 ---p 00000000 00:00 0 
7f97f8000000-7f97fbff2000 rw-p 00000000 00:00 0 
7f97fbff2000-7f97fc000000 ---p 00000000 00:00 0 
7f97fc000000-7f97fffd3000 rw-p 00000000 00:00 0 
7f97fffd3000-7f9800000000 ---p 00000000 00:00 0 
7f9800000000-7f9803fff000 rw-p 00000000 00:00 0 
7f9803fff000-7f9804000000 ---p 00000000 00:00 0 
7f9804000000-7f9807f94000 rw-p 00000000 00:00 0 
7f9807f94000-7f9808000000 ---p 00000000 00:00 0 
7f9808000000-7f9808021000 rw-p 00000000 00:00 0 
7f9808021000-7f980c000000 ---p 00000000 00:00 0 
7f980c000000-7f980c021000 rw-p 00000000 00:00 0 
7f980c021000-7f9810000000 ---p 00000000 00:00 0 
7f9814000000-7f9814021000 rw-p 00000000 00:00 0 
7f9814021000-7f9818000000 ---p 00000000 00:00 0 
7f9818000000-7f981bfeb000 rw-p 00000000 00:00 0 
7f981bfeb000-7f981c000000 ---p 00000000 00:00 0 
7f981c000000-7f981ffe4000 rw-p 00000000 00:00 0 
7f981ffe4000-7f9820000000 ---p 00000000 00:00 0 
7f9820000000-7f9820021000 rw-p 00000000 00:00 0 
7f9820021000-7f9824000000 ---p 00000000 00:00 0 
7f9828000000-7f9828021000 rw-p 00000000 00:00 0 
7f9828021000-7f982c000000 ---p 00000000 00:00 0 
7f982c000000-7f982c021000 rw-p 00000000 00:00 0 
7f982c021000-7f9830000000 ---p 00000000 00:00 0 
7f9830000000-7f9830021000 rw-p 00000000 00:00 0 
7f9830021000-7f9834000000 ---p 00000000 00:00 0 
7f9834000000-7f9834021000 rw-p 00000000 00:00 0 
7f9834021000-7f9838000000 ---p 00000000 00:00 0 
7f9838000000-7f9838021000 rw-p 00000000 00:00 0 
7f9838021000-7f983c000000 ---p 00000000 00:00 0 
7f9840000000-7f9840021000 rw-p 00000000 00:00 0 
7f9840021000-7f9844000000 ---p 00000000 00:00 0 
7f9844000000-7f9844021000 rw-p 00000000 00:00 0 
7f9844021000-7f9848000000 ---p 00000000 00:00 0 
7f9848000000-7f984bfda000 rw-p 00000000 00:00 0 
7f984bfda000-7f984c000000 ---p 00000000 00:00 0 
7f9850000000-7f9850021000 rw-p 00000000 00:00 0 
7f9850021000-7f9854000000 ---p 00000000 00:00 0 
7f9858000000-7f985bf88000 rw-p 00000000 00:00 0 
7f985bf88000-7f985c000000 ---p 00000000 00:00 0 
7f985c000000-7f985c021000 rw-p 00000000 00:00 0 
7f985c021000-7f9860000000 ---p 00000000 00:00 0 
7f9860000000-7f9863ffc000 rw-p 00000000 00:00 0 
7f9863ffc000-7f9864000000 ---p 00000000 00:00 0 
7f9868000000-7f9868021000 rw-p 00000000 00:00 0 
7f9868021000-7f986c000000 ---p 00000000 00:00 0 
7f986c000000-7f986ffea000 rw-p 00000000 00:00 0 
7f986ffea000-7f9870000000 ---p 00000000 00:00 0 
7f9870000000-7f9870021000 rw-p 00000000 00:00 0 
7f9870021000-7f9874000000 ---p 00000000 00:00 0 
7f9878000000-7f9878021000 rw-p 00000000 00:00 0 
7f9878021000-7f987c000000 ---p 00000000 00:00 0 
7f9880000000-7f9880021000 rw-p 00000000 00:00 0 
7f9880021000-7f9884000000 ---p 00000000 00:00 0 
7f9888000000-7f9888021000 rw-p 00000000 00:00 0 
7f9888021000-7f988c000000 ---p 00000000 00:00 0 
7f9890000000-7f9890021000 rw-p 00000000 00:00 0 
7f9890021000-7f9894000000 ---p 00000000 00:00 0 
7f9898000000-7f9898021000 rw-p 00000000 00:00 0 
7f9898021000-7f989c000000 ---p 00000000 00:00 0 
7f98a0000000-7f98a0021000 rw-p 00000000 00:00 0 
7f98a0021000-7f98a4000000 ---p 00000000 00:00 0 
7f98a8000000-7f98a8021000 rw-p 00000000 00:00 0 
7f98a8021000-7f98ac000000 ---p 00000000 00:00 0 
7f98b0000000-7f98b0021000 rw-p 00000000 00:00 0 
7f98b0021000-7f98b4000000 ---p 00000000 00:00 0 
7f98b8000000-7f98b8021000 rw-p 00000000 00:00 0 
7f98b8021000-7f98bc000000 ---p 00000000 00:00 0 
7f98bc000000-7f98bc021000 rw-p 00000000 00:00 0 
7f98bc021000-7f98c0000000 ---p 00000000 00:00 0 
7f98c0000000-7f98c0021000 rw-p 00000000 00:00 0 
7f98c0021000-7f98c4000000 ---p 00000000 00:00 0 
7f98c5ffc000-7f98c5ffd000 ---p 00000000 00:00 0 
7f98c5ffd000-7f98c67fd000 rw-p 00000000 00:00 0 
7f98c6ffe000-7f98c6fff000 ---p 00000000 00:00 0 
7f98c6fff000-7f98c77ff000 rw-p 00000000 00:00 0 
7f98c77ff000-7f98c7800000 ---p 00000000 00:00 0 
7f98c7800000-7f98c8000000 rw-p 00000000 00:00 0 
7f98c8000000-7f98c8021000 rw-p 00000000 00:00 0 
7f98c8021000-7f98cc000000 ---p 00000000 00:00 0 
7f98cc000000-7f98cc021000 rw-p 00000000 00:00 0 
7f98cc021000-7f98d0000000 ---p 00000000 00:00 0 
7f98d0000000-7f98d0021000 rw-p 00000000 00:00 0 Aborted (core dumped)

pford:
@kswang1029 I pushed a fix to cancel profiles but could not reproduce the crash (with the fix).

kswang1029:
Just did a quick test of e34d485, cannot reproduce the crash. I will do more tests tomorrow. : )

kswang1029:
it appears that cursor profile request does not have progressive update(?). The widget shows an old one and a new full profile just shows up.

pford:
I fixed the cursor spectral profile for partial updates.

ajm-ska:
@pford This branch seems to have trouble building on MacOS (It is fine on Linux). It seems to be a clang vs gcc problem (The MacOS system is using `Apple clang version 11.0.0`). I see similar similar issues when I search on Google, but I can't figure it out. Would you have a better understanding of how to fix it?

```

[ 69%] Building CXX object CMakeFiles/carta_backend.dir/Region/CrtfImportExport.cc.o

/Users/acdc/jenkins-ci-mac/workspace/_backend_pam_refactor_wcs_region/Region/CrtfImportExport.cc:38:30: error: no viable conversion from 'casacore::SubString' to 'std::__1::basic_string<char>'

            _import_errors = err.getMesg().before("at File");

                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/string:800:5: note: candidate constructor not viable: no known conversion from 'casacore::SubString' to 'const std::__1::basic_string<char> &' for 1st argument

    basic_string(const basic_string& __str);

    ^

/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/string:805:5: note: candidate constructor not viable: no known conversion from 'casacore::SubString' to 'std::__1::basic_string<char> &&' for 1st argument

    basic_string(basic_string&& __str)

    ^

/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/string:818:5: note: candidate constructor template not viable: no known conversion from 'casacore::SubString' to 'const char *' for 1st argument

    basic_string(const _CharT* __s) {

    ^

/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/string:868:5: note: candidate constructor not viable: no known conversion from 'casacore::SubString' to 'initializer_list<char>' for 1st argument

    basic_string(initializer_list<_CharT> __il);

    ^

/usr/local/include/casacore/casa/BasicSL/String.h:66:3: note: candidate function

  operator const string() const { return string(ref_p, pos_p, len_p); }

  ^

/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/string:886:44: note: passing argument to parameter '__str' here

    basic_string& operator=(basic_string&& __str)

                                           ^

1 error generated.

make[2]: *** [CMakeFiles/carta_backend.dir/Region/CrtfImportExport.cc.o] Error 1

make[1]: *** [CMakeFiles/carta_backend.dir/all] Error 2

make: *** [all] Error 2

script returned exit code 2
```

kswang1029:
@pford yes tested cursor profile progressive update 👍 

Just wondering, now do we reuse a region id if there is no ambiguity? For example, I import a region file with 10 regions so region ids are 1...10. If I remove region 5 and I manually create one new region then its id is 11. Now if I remove all 10 regions, and import another region file with 5 regions, the ids are supposed to be 1~5 again?

kswang1029:
@pford I also tested regions with multiple images, including manually creating/removing region and region import, etc. All appear fine so far. 👍

kswang1029:
@pford with TBB=4 and OMP=4, if there are two widgets updating profiles at the same time (one for region 1 and another for region 2) and if I change to flux_density (was the default "mean") in one widget, the backend crashed.

kswang1029:
@pford Progressive update and cancellation of Stokes profiles (e.g. U) seems not working.

pford:
@ajm-asiaa I fixed the MacOS build error but all of the ICD test stages fail; is that expected?
@kswang1029 I improved cancelling the spectral profiles.  Thanks for your testing efforts!

ajm-ska:
@pford Thanks for fixing the MacOS build issue! The ICD test problem was at my end. It looks like backend on the CentOS7 system didn't start up because the port was already occupied for some reason. I just made Jenkins rerun, and this time it passed all the tests.

kswang1029:
@pford I am still able to make the backend crash with ed6aa5f. Here is what I did (TBB=4, OMP=4):
1) load a large cube
2) freeze cursor
3) make two spectral line widgets
4) make two moderate-size regions so that we can "see" progressive update for at least few seconds
5) set widget 1 to region 1 and set region 2 for region 2
6) move the two regions to a new place so that we can see two profiles are updating
7) now change the stats in each widget several times (~2-3 sec in between). e.g. widget1-> fluxdensity, widget2->min, widget1->max, widget2->stddev, etc.
8) by doing so I can see the backend crashes somehow. 


update: sometimes at step5 the backend will crash, like the attached video
[Screen Recording 2020-05-02 at 5.06.49 PM.mov.zip](https://github.com/CARTAvis/carta-backend/files/4567369/Screen.Recording.2020-05-02.at.5.06.49.PM.mov.zip)

kswang1029:
@pford Profile cancellation involving Stokes with ed6aa5f seems to still have issues. It appears to me that new request is not cancelled and further requests are queuing up. 
Screen Shot 2020-05-02 at 4 50 05 PM
This screenshot shows two empty profiles but in the js console we can see two (old request) sets of profiles are streaming.

pford:
@kswang1029 please re-test the spectral profiles with 33309a0, changing stats type, stokes, and region.  The profiles should restart.

kswang1029:
with  33309a0, cursor profile seems broken when there is no other region created (region=Active Stokes=Current). If I select Stokes explicitly (e.g. Q) I can see cursor profile.

with ac673df (the latest commit) I can still make the backend crash by changing region and stats quickly from three widgets.

pford:
@kswang1029 figured out the crash, working on a fix.

ImageData/Hdf5Loader.cc
```diff
@@ -200,206 +200,172 @@ bool Hdf5Loader::UseRegionSpectralData(const std::shared_ptr<casacore::ArrayLatt
     return true;
 }
 
-bool Hdf5Loader::GetRegionSpectralData(int region_id, int config_stokes, int profile_stokes,
-    const std::shared_ptr<casacore::ArrayLattice<casacore::Bool>> mask, IPos origin, std::mutex& image_mutex,
-    const std::function<void(std::map<CARTA::StatsType, std::vector<double>>*, float)>& partial_results_callback) {
-    // config_stokes is the stokes value in the spectral config, could be -1 for CURRENT_STOKES; used to retrieve config from region
-    // profile_stokes is the stokes index to use for slicing image data
+bool Hdf5Loader::GetRegionSpectralData(int region_id, int stokes, const casacore::Array<casacore::Bool>& mask,
+    const casacore::IPosition& origin, std::mutex& image_mutex, std::map<CARTA::StatsType, std::vector<double>>& results, float& progress) {
```
confluence: This is minor, but because `casacore::IPosition` is used so often in this class there is an alias for it (`IPos`). This is already not used consistently; I wonder if it would make sense to use it only in places where the full name is too verbose.

ImageData/Hdf5Loader.cc
```diff
@@ -200,206 +200,172 @@ bool Hdf5Loader::UseRegionSpectralData(const std::shared_ptr<casacore::ArrayLatt
...
-                                kv.second[z] = NAN;
-                                break;
-                        }
-                    }
+                casacore::IPosition mask_pos(mask_shape.size());
+                mask_pos(0) = x;
+                mask_pos(1) = y;
+                mask_pos(_spectral_axis) = z;
+                if (_stokes_axis >= 0) {
+                    mask_pos(_stokes_axis) = 0;
```
confluence: Could you please clarify this change? My understanding is that a region is still a 2D shape in the XY plane projected in the Z direction. So the mask should be 2D, and it should be possible to exclude all Z values for a particular XY point with a check in the middle loop, without the need to iterate over Z.

Region/RegionHandler.cc
```diff
@@ -0,0 +1,1339 @@
...
+        return false;
+    }
+
+    std::unique_lock<std::mutex> ulock(_spectral_mutex);
+    std::unordered_map<ConfigId, RegionSpectralConfig, ConfigIdHash> region_configs;
+    region_configs.insert(_spectral_req.begin(), _spectral_req.end());
+    ulock.unlock();
+
+    bool profile_ok(false);
+    if (region_id > 0) {
```
confluence: The code in these blocks is identical except for the region and file IDs and the check used to filter region configs. Would it be possible to remove the duplication by setting these properties in a conditional (using a lambda for the check)?

confluence (changes requested):
I have a question about the shape of the new region mask. The other comments are minor. In general I'm happy with the architecture changes.

ImageData/Hdf5Loader.cc
pford: Since I am familiar with casacore, I think "casacore::IPosition" when I develop code.  This is defined in FileLoader.h and only used in Hdf5Loader, so you have to search the code to see what this thing is.  So my personal preference is not to use aliases, but I will change HdfLoader to be consistent.

ImageData/Hdf5Loader.cc
pford: In the refactor, the mask is a 3D/4D boolean array which matches the region shape.

In CASA images, pixel values are not changed for "bad" values, a boolean mask is created to indicate which pixels should be included in a calculation.  So there are two masks for a casacore region or SubImage, (1) a mask to indicate whether the pixel is included/excluded in the region, and (2) a mask to indicate whether the pixel is good/bad.

In dev, since the Region's mask is 2D, it ignores the image mask and only indicates whether the pixel is included/excluded (type 1).  In the refactor, the mask is a 3D/4D mask which applies the region mask (type 1) to the image mask (type 2) to get a 3D/4D region mask (type 2).  So the mask IPos must include the z index and stokes index.

The CartaHDF5Image mask (type 2) is set depending on the NaN pixel values.  So the dev/refactor code is equivalent in that NaN values are ignored using "if (std::isfinite(v))" in dev (which I now see should be removed in the refactor code) or the mask in the refactor.  Since the Hdf5Loader is the only user of the mask, I could revise this to use the 2D mask again and check the type 1 mask and pixel values separately as before.
 
If in the future there will be a way to create an image mask for an HDF5 image that does not change the pixel values (interactive clean?), the 2D mask will no longer be sufficient. @kswang1029 what do you think?

Region/RegionHandler.cc
pford: I will improve this.

ImageData/Hdf5Loader.cc
kswang1029: @pford I think for interactive clean with CASA, we will just need to focus on CASA image only as that's the native image format of CASA. 

I think it is better to keep the flexibility of having masks in 3D/4D for HDF5 images. One use case I can think of is moment generator. Usually users can define:
1) a box in 2D so that it defines the spatial size of the output moments.
2) a range of channels to be used for moment calculations
3) a range of pixel values to be included/excluded for the moment calculations

So given the above three items, we will need a 3D/4D mask I suppose?

I remembered that @low-sky mentioned in the moment generator doc that it would be good to have a tool to define masks to be used elsewhere. In this case, we will need to support 3D/4D masks for all image formats (casa, fits, hdf5, miriad)

confluence: Thanks; that makes sense. I think that if it's possible to do a single check using a 3D/4D mask, that's probably better than doing two checks (first 2D mask and later `isfinite`). So I'm happy to leave it like this and remove the second check.

confluence (approved):
Mask query resolved.

ImageData/Hdf5Loader.cc
confluence: Update: Angus pointed out in our local meeting that the 3D/4D mask may be a problem if the image is very large. This may be a good reason to go back to the old behaviour here. Can we check if the entire mask is created in memory at once?

pford:
@kswang1029 ready to re-test. @confluence I reverted the Hdf5Loader spectral profile to its previous behavior (using 2D mask and origin) and verified an identical region has the same spectral profile in this branch and dev.

kswang1029:
@pford Overall a90980d seems fine. Just one time I saw profile stopped updating but I cannot reproduce. I will keep an eye on it.

The cadence of HDF5 region profile progressive update seems much higher than before? Shall we just update every 1s? Others for hdf5 appear fine.

Stokes cubes appear fine too.

kswang1029:
@pford HDF5 profile now updates much less frequently. 👍

veggiesaurus:
@pford when exporting regions in a CRTF region, I'm getting a hard crash, because `_coord_sys->polarizationCoordinateNumber()` returns -1.

kswang1029:
@pford regarding export a region then import it back, it appears that the size of a region (rectangle/ellipse) varies slightly with different formats (world/pixel, crtf/reg). It appears to me that the export region files have a bit different precision in coordinates/sizes/angle. Could you check and make them consistent and "precise" (sub-milli-arcsecond level)?

pford:
I have done more manual import/export testing with all the datasets I have downloaded.  I think it works as well as the dev branch, and has the same failures for rectangle regions.  The casaviewer does not support exporting CRTF centerbox or rotbox, likely because of the issues with certain images; it only exports a box with blc, trc coordinates.

- 18h03m_tabular_axis.image: when a rectangle's corners are converted to world coordinates (and back to pixel), it is no longer a rectangle due to the width changing by cos(declination).  Using the corners of the polygon (normally blc=corner 0 and trc=corner 2) for center, width, and height does not return the original rectangle.
- rotated.fits (attached to #372): imageanalysis and SAOImage are smart enough to know the image is rotated, so the imported box is rotated to the world axes instead of pixel axes.  As in previous, calculating rect control points from polygon corners is incorrect.
- When the image has a casacore::LinearCoordinate instead of a casacore::DirectionCoordinate (e.g. LMCq.fits), odd values are returned when converting pixel to world coordinates; will investigate.  For now, as in dev, CRTF import/export and DS9 world export fails for images with no direction.  Can import/export DS9 in pixel coordinates.

pford:
Closing this PR for more work on the rectangle import/export based on the CARTA_region_wcs_projections doc.

PR #521, opened 2020-05-07
Remove cached file info from the session object

confluence:
The file info which is cached on the session object is not invalidated correctly if the file browser is bypassed (which happens when a scripting command is used to open or append a file) and the file name hasn't changed. This leads to unexpected scripting behaviour when an attempt is made to load the same file multiple times, even if it was closed in the interim, or even to load a file with a different path but the same filename, including non-existent files.

The latter issue could be solved if we also cached and checked the directory, but this would not solve the underlying issue with loading the exact same file again. We could possibly invalidate the cached data after loading each file, but the simplest solution is to remove the caching, which doesn't save a lot of bandwidth. Apart from the scripting issue, caching could cause a problem if the user overwrites a file with a different file while the file info is open in the browser. Please comment if you think that it's worth keeping the caching and fixing it.

This PR removes the caching and (for consistency) converts some pointers to file info message objects passed by value into the underlying objects passed by reference.

### DIFF
--- a/FileList/FileExtInfoLoader.cc
+++ b/FileList/FileExtInfoLoader.cc
@@ -22,9 +22,9 @@ using namespace carta;
 FileExtInfoLoader::FileExtInfoLoader(carta::FileLoader* loader) : _loader(loader) {}
 
 bool FileExtInfoLoader::FillFileExtInfo(
-    CARTA::FileInfoExtended* extended_info, const std::string& filename, const std::string& hdu, std::string& message) {
+    CARTA::FileInfoExtended& extended_info, const std::string& filename, const std::string& hdu, std::string& message) {
     // set name from filename
-    auto entry = extended_info->add_computed_entries();
+    auto entry = extended_info.add_computed_entries();
     entry->set_name("Name");
     entry->set_value(filename);
     entry->set_entry_type(CARTA::EntryType::STRING);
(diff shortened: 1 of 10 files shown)


veggiesaurus (approved):
this looks fine to me. We seem to use pointers and references somewhat inconsistently for `Fill ... ` functions. I don't have a particularly strong opinion on which is more sensible, but a quick search for `::Fill` suggests that we're now _mostly_ using references

pford:
Usually the reference was for a response message and the pointer was for a response submessage, since that is what protobuf returns when you add the submessage to the message.

FileInfo and FileInfoExtended were created for the file browser and reused for OpenFileAck, since OpenFile always followed FileInfoRequest for the same image.  Even a double-click on the file to open it sends FileInfoRequest first.  At one point FileInfoExtended was also passed through the Frame to the Hdf5Loader for the CartaHdf5Image constructor to use the headers for CoordinateSystem and ImageInfo creation.

I have no problem with these changes.

pford (approved):

pford:
The build/test failures should be reviewed since they passed on my refactor_wcs_region branch.

confluence:
@ajm-asiaa says the tests are still broken at the moment.

ajm-ska:
@pford I'm sorry about the trouble here. I have put the ICD tests back to their previous state and have rerun them on the 'dev' branch, this 'confluence/remove_file_info_caching branch', and your 'refactor_wcs_region branch'. It is possible for them all to pass.

Issue #489, opened 2020-04-08
Implement basic end-to-end generic scripting functionality (backend component)


confluence:
Add support for using the backend as a proxy for a scripting wrapper to call actions on the frontend generically and receive data and other responses.

A basic test case: open an image and receive and save image data.

This issue covers these components:

- [ ] a generic protocol buffer interface between wrapper and backend
- [ ] a generic gRPC service in the backend
- [ ] a generic protocol buffer interface between backend and frontend

The wrapper component is described separately in an issue in the prototype wrapper repository.

PR #523, opened 2020-05-08
Confluence/generic scripting

confluence:
This is an implementation of generic front-end scripting in the backend (#489). It uses @veggiesaurus 's frontend branch [here](https://github.com/CARTAvis/carta-frontend/tree/angus/test_debug_execution) and the corresponding protobuf branch [here](https://github.com/CARTAvis/carta-protobuf/tree/angus/generic_scripting). A prototype Python wrapper which uses generic scripting requests to control the frontend is being developed [here](https://github.com/idia-astro/carta-python).

I have kept this as compatible as possible with @pford's [branch for the gRPC service for the CASA wrapper](https://github.com/CARTAvis/carta-backend/tree/pam/415_grpc_server) so that the two approaches can be merged in the same service. The major difference is the way that I build the gRPC library (from the gRPC repository, not in CASA). I have included build commands in the dockerfile, but these have to be tested.

At the moment the requests are blocking in the backend, and poll the local state for responses until they either find them or time out. We could probably reimplement this in a more elegant, non-blocking way. Should we do that before merging this, or is the current method good enough for now?

(Most of the specific behaviour of the prototype wrapper is outside the scope of this PR because it can be changed independently of this generic backend functionality -- but if you have comments or ideas about that, please file issues in the wrapper repo! The wrapper is very experimental and rapidly changing; I will try only to commit working snapshots.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,7 @@ INCLUDE_DIRECTORIES(${HDF5_INCLUDE_DIR})
 INCLUDE_DIRECTORIES(${LIBXML2_INCLUDE_DIR})
 
 ADD_SUBDIRECTORY(carta-protobuf)
+ADD_SUBDIRECTORY(cartavis)
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/ImageData)
 INCLUDE_DIRECTORIES(/usr/include/casacore)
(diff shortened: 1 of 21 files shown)


ajm-ska (approved):
I was able to run it natively on MacOS as well as with the carta-backend and carta-python running in the Ubuntu Docker container. Thanks for taking my comments on Slack into account.

(The CI tests are currently failing because I messed the MacOS CI system up installing gRPC. It seems to have caused conflicts with SSL versions. I'm currently trying to fix it. But it was working before. So hopefully it will still allow it to be merged.)

Update: MacOS CI system fixed. Now passing the tests.

pford:
Your branch also shows all of the unrelated file info changes as new; merge dev?

pford (approved):

veggiesaurus (approved):

PR #526, opened 2020-05-15
ajm/cmake improvement

### LINKED ISSUE #439: Dynamically detect location of casacore / casacode in cmake
At the moment the `casacode` include location in  `CMakeLists.txt` is incompatible with the install locations of the Ubuntu package, which should now be the preferred way of installing the casa dependencies in Ubuntu. So to install the backend while using the package, we have to either pass an additional flag to cmake or edit the file. This should not be necessary. We haven't edited the file permanently because (to the best of my knowledge) the MacOS and CentOS installs still use casacode installed in `/usr/local`.

It should be possible to modify the file to cover both cases. The correct way to do this is probably to create a cmake extension to locate casa; we may be able to repurpose something like [this](https://github.com/casacore/casarest/blob/master/cmake/FindCasaCore.cmake). But we may be able to do something much simpler, like add both directories, but this requires some investigation (if they both exist, which one will be prioritised? Etc.).

ajm-ska:
These additions to CMakeLists.txt allow us to simply use `cmake ..` to build the carta-backend on RHEL, Ubuntu, and MacOS.

When we build carta-casacore from source, the default install location for 'casacore' and 'casacode' is `/usr/local/include`. That is covered by:

> INCLUDE_DIRECTORIES(/usr/local/include/casacore /usr/local/include/casacode)

When we use the carta-casacore PPA on ubuntu, the default install location is `/usr/include`. That is covered by:
> INCLUDE_DIRECTORIES(/usr/include/casacore /usr/include/casacode)

On MacOS, I believe the expected method to install most 3rd party packages is Homebrew. OpenSSL and libomp are two such packages needed by carta-backend. We can use the following lines to find them and do the linking correctly:
> if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
> INCLUDE_DIRECTORIES(/usr/local/include)
> INCLUDE_DIRECTORIES(/usr/local/opt/openssl/include)
> INCLUDE_DIRECTORIES(/usr/local/opt/libomp/include)
> LINK_DIRECTORIES(/usr/local/lib /usr/local/opt)
> endif (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")

The if-loop around this section is not necessary, but doesn't do any harm, and might help users understand that these lines are MacOS related.

Advanced users who decide to install libraries in non-default locations are still free to use the additional -DCMAKE_CXX_FLAGS=" " and -DCMAKE_CXX_STANDARD_LIBRARIES=" " flags in their cmake command if they so wish.


confluence:
If this has been tested and it works, I'm happy with it -- if we discover a more elegant solution (like a custom config file which we can use with `FIND_PACKAGE`), we can change this later.

Would this PR be a good place to add a directory with all the dockerfiles that we use?

confluence (approved):

ajm-ska:
Here are Dockerfiles for ubuntu 16.04, ubuntu 20.04, CentOS6, CentOS7, and CentOS8. 
They all include gRPC installation for future the scripting-interface. I tested each one and they seem to work fine.

Maybe when the casacore change is officially made for Mark's moment-generator branch, I'll try to redo the PPAs as a single PPA and then update the primary Dockerfile.

ajm-asiaa:
The CentOS7 CI build is currently failing because I tried to install Mark's modified carta-casacore so that the CI tests would pass for his branch. But it seems I have messed up somehow.... I'll try to fix it.

pford (approved):

PR #527, opened 2020-05-19
Angus/build test

veggiesaurus:
improves build process compatibility with gRPC built from source and gRPC installed via `apt`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,21 +9,8 @@ RUN \
     libcfitsio-dev libfmt-dev libgtest-dev libhdf5-dev liblapack-dev libncurses-dev \
     libreadline-dev libssl-dev libstarlink-ast-dev libtbb-dev libtool libxml2-dev \
     libzstd-dev libgsl-dev man pkg-config python-pip python3-pip \
-    software-properties-common unzip vim wcslib-dev wget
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -49,21 +36,18 @@ RUN \
...
+  apt-get update && \
+  apt-get install -y libprotobuf-dev protobuf-compiler libgrpc++-dev libgrpc-dev protobuf-compiler-grpc googletest
 
 # Build carta-backend (currently checkouts the confluence/generic-scripting branch)
 RUN \
   apt-get -y install libxml2-dev && \
   git clone https://github.com/CARTAvis/carta-backend.git && \
   cd carta-backend && \
-  git checkout confluence/generic-scripting && \
+  git checkout angus/build_test && \
```
confluence, line 50: How does this work when we merge? Shouldn't we remove it when we eventually merge this into `dev`?

confluence (approved):
I'm generally happy, but I have a query about the hardcoded branch name in the Dockerfile.

Dockerfile
veggiesaurus: I guess the sustainable way is to rather use the host's checkout of carta and map it to a `src` directory or something.

PR #530, opened 2020-05-24
Angus/catalog alt implementation

### LINKED ISSUE #500: support sorting of catalog table
When users click the column head in the catalog table, sorting should be applied to the catalog.
If no filter is applied, sorting should be applied to _all_ data and backend delivers partial data of the sorted ones.
If there is filter applied, sorting should be applied to the _filtered_ data and backed delivers partial data of the filtered and sorted ones.

related to frontend issue https://github.com/CARTAvis/carta-frontend/issues/878

### LINKED ISSUE #509: performance optimization of loading catalog file
With the current dev backend, loading a 100k source catalog (~240MB) took ~38 seconds from pressing the load button to seeing the catalog widget showing up. When loading a 20k source catalog (~48MB), it took ~8 seconds. 

Comparing to the desktop application TOPCAT, it appears there is room for performance optimization of catalog loading. "psrecord" charts are attached when loading 20k-source and 100k-source catalog. 

20k (~1.5s)
topcat_20k_psrecord

100k (~5s)
topcat_100k_psrecord

### LINKED ISSUE #511: support catalog files in FITS format
The backend needs to support catalogue files in FITS format. Example FITS table is accessible [here](https://drive.google.com/open?id=1v4Hv5Ve7HsW3YtNRNmyb8CxEdXhu3UkT).

### LINKED ISSUE #519: Explore alternative XML libraries to improve catalog performance
The current catalog loading implementation is significantly slower than TopCAT. Alternative approaches to the libXML-based approach should be explored in order to speed up catalog loading speeds.

veggiesaurus:
- Optimisation of catalog loading by using an alternative implementation for XML-based files (Closes #509) by using pugiXML (Closes #519 )
- Adds support for FITS tables (Closes #511)
- Reworks file listing using `std::filesystem`
- Contains an ICD change to simplify handling of column types
- Expands support for more column types, and sends 8- and 16-bit columns as the approriate data type, rather than as 32-bit ints
- Sorts filtered data by `sort_column` using `std::sort` (flags added to allow for parallel sorting) (Closes #500)
- (bonus) Adds a `Timer` class for benchmarking purposes. See included unit tests for example usage.

As there is an ICD change, this will require a companion PR in the frontend.
As this is a large re-write of the catalog work, @kswang1029 please re-test for correctness when it comes to filtering.


CMakeLists.txt
```diff
@@ -108,6 +111,9 @@ set(LINK_LIBS
         zfp
         zstd
         tbb
+        stdc++fs
```
markccchiang: For Mac, the `filesystem` is already available. This line will cause a building error. Maybe can move it to the lin 170 as 
```
if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
    target_link_libraries(carta_backend stdc++fs ${LINK_LIBS})
endif (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
```

veggiesaurus: > For Mac, the `filesystem` is already available. This line will cause a building error. Maybe can move it to the lin 170 as 
> ```
> if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
>     target_link_libraries(carta_backend stdc++fs ${LINK_LIBS})
> endif (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
> ```

Good point. I'll push a fix along those lines. It might be more related to clang vs GCC, rather than linux versus Darwin

veggiesaurus: I've added a `if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")` switch to `CMakeLists.txt` to only include `stdc++fs` when using `GCC`

test/Makefile
```diff
@@ -1,3 +0,0 @@
-
-test_message_format: 
-	g++ -I .. test_message_format.cc ../build/carta-protobuf/libcarta-protobuf.a -l protobuf
```
confluence, line 3: Did this and the message format test get deleted by mistake? We should probably convert them to the same kind of test for consistency. Did we ever decide on a clean way to build tests with dependencies on the rest of the CARTA code without rebuilding it in all the individual subdirectories? Possibly a single test build with some kind of top-level suite that can run any subset of the tests?

confluence (commented):
I have a query about some deleted test files (and the structure of our tests in general) -- apart from that this looks fine.

test/Makefile
veggiesaurus: it wasn't entirely by accident, but I removed it because it seemed like an inconsistent test that should probably be updated to a proper unit test if its still required.

veggiesaurus: perhaps I should leave that cleanup for another PR :smile:

confluence (approved):

markccchiang (approved):

Issue #528, opened 2020-05-20
Cannot import ICRS regions to FK5 image


veggiesaurus:
Importing ICRS regions into an FK5 image results in a backend error message:  `The error message when using CRTF or DS9 region files is Unsupported coordinate frame for regions ICRS`

Steps to reproduce:
1. Open an ICRS image
2. Create some regions
3. Export regions in either CRTF or DS9 format, with world coordinates. These will get exported in ICRS
4. Open an FK5 image
5. Import created regions

Will leave it up to @kswang1029 to decide whether this counts as critical or not. It doesn't result in a hard crash, but as far as I can see, there's no workaround for getting the regions into an FK5 image, short of manually converting them to FK5 elsewhere.

kswang1029:
Hmm this indeed causes bad UX. The only possible solution is to manually modify the region text file with correct world coordinates at a specific reference system. Not very critical though. I suppose when wcs region is in place, this should be solved. For example, a region file in galactic coordinates could be applied to few matched images with different coordinate systems such as FK5, Ecliptic, etc).

veggiesaurus:
> I suppose when wcs region is in place, this should be solved. 

I'm not sure that will fix the issue. This occurs when loading a single set of regions onto a single open image. 

Am I misunderstanding the issue? It seems like quite a fundamental limitation (not being able to load ICRS regions onto anything other than ICRS images)

pford:
This is thrown by imageanalysis code (Annotations/AnnotationBase.cc).  When the region coordinates must be converted because the file frame (_directionRefFrame) does not match the image coordinate system (_csys) frame, only the following types are supported:

	if (_directionRefFrame != _csys.directionCoordinate().directionType(false)
		&& _directionRefFrame != MDirection::B1950
		&& _directionRefFrame != MDirection::B1950_VLA
		&& _directionRefFrame != MDirection::BMEAN
		&& _directionRefFrame != MDirection::DEFAULT
		&& _directionRefFrame != MDirection::ECLIPTIC
		&& _directionRefFrame != MDirection::GALACTIC
		&& _directionRefFrame != MDirection::J2000
		&& _directionRefFrame != MDirection::SUPERGAL) {
		throw AipsError(preamble + "Unsupported coordinate frame for regions "
			+ MDirection::showType(_directionRefFrame));
	}

veggiesaurus:
[...]

Issue #505, opened 2020-04-17
very high memory usage and poor performance when loading DS9 regions


veggiesaurus:
Note: there will be a (far more critical) frontend issue created related to this, but I thought I'd separate the issues.

When loading a 2.3 MB DS9 region file with 15 000 regions,  the backend requires upwards of 2.5 GB to load the region file. The loading process takes about 30 seconds on my machine, and after loading, about 1 GB of memory is required. 

I ran some profiling with `perf` and it seems a lot of the CPU time is spent re-allocating memory in casacore. I wonder if pre-allocating the correct amount of memory would help the situation. 

![image](https://user-images.githubusercontent.com/592504/79595059-7f983d00-80de-11ea-8eb7-a2d0fbf25867.png)

15 000 regions does seem like a worst case scenario, but I do think we need to improve this quite significantly. DS9 manages to load all the regions in about 2 seconds, with 60 MB memory usage.

veggiesaurus:
@veggiesaurus to re-test with same file

veggiesaurus:
this is now fixed in the backend, and now requires the frontend to be sorted out (frontend doesn't seem to handle large region files nicely at all!)

PR #539, opened 2020-06-09
Refactor backend for wcs regions

### LINKED ISSUE #334: Cannot import/export regions when image has invalid or non-existent WCS headers

### LINKED ISSUE #445: WCS regions: remove Region code from Frame

pford:
Closes #445.  Builds on previous (closed without merge) PR #517 with additional workarounds for region import/export.

Should also:
- fix #334 (region import/export for image with no direction coordinate)
- improve #505 (performance loading DS9 regions)

Does not fix:
- #528 (import ICRS region to FK5 ), pending casa6 PR then carta-casacore submodule update
- regions for images with rotation/non-square pixels (will create new issue)


Frame.cc
```diff
@@ -144,543 +135,94 @@ int Frame::CurrentStokes() {
...
-        } else {
-            auto region = std::unique_ptr<carta::Region>(
-                new carta::Region(name, type, points, rotation, _image_shape, _spectral_axis, _stokes_axis, coord_sys));
-            if (region->IsValid()) {
-                _regions[region_id] = move(region);
-                region_set = true;
-            }
-        }
+    // Normalize any constants
+    ChannelRange range(chan_range.from, chan_range.to);
```
confluence: Very minor nitpick: if we're going to copy `ChannelRange`s like this, we should add a constructor with a single `ChannelRange` parameter.

confluence (approved):
It looks like there aren't any additional changes to the HDF5 code (beyond what was discussed in the previous PR) -- I don't have any other specific comments.

Frame.cc
pford: Changed to use int variables instead for image slicer.

kswang1029 (approved):
tested region import/export. 👍

veggiesaurus (approved):

PR #546, opened 2020-06-24
Angus/boost fs test

veggiesaurus:
Some compilers support C++17, but don't fully support the C++ std libs. In particular, `std::filesystem` support is problematic. This PR allows you to switch to `boost::filesystem` (`libboost-filesystem-dev` on ubuntu, can also be installed via `homebrew` on mac) by using the cmake flag `UseBoostFilesystem`. A couple of changes were needed to make the code compatible between the two libraries (e.g. explicitly casting `path` objects to `string`)


confluence (approved):

kswang1029:
tested the Mojave package and as far as I can see it works fine.

Issue #531, opened 2020-05-29
Remove deprecated TBB features


pford:
https://www.threadingbuildingblocks.org/docs/help/index.htm#reference/appendices/deprecated_features/cpp_obsolete/atomic_cls.html

Warnings when building backend for:
- tbb/atomic.h
- tbb/mutex.h
- tbb/task_scheduler_init.h
- tbb/tbb.h (contains deprecated functionality)

For example:
/usr/include/tbb/mutex.h:21:140: note: #pragma message: TBB Warning: tbb/mutex.h is deprecated. For details, please see Deprecated Features appendix in the TBB reference manual.

veggiesaurus:
`task_scheduler_init`  is the tricky one to replace, I think :thinking:  The other ones can be replaced trivially by `std::atomic` and `std::mutex`

veggiesaurus:
Assigning this to @robsimmonds because his `tbb-be-gone` branch should solve this.

PR #548, opened 2020-06-26
switched to std::mutex and std::atomic, fixed tests

veggiesaurus:
Partial fix for #531. Still need to work out how to remove the task scheduler bit.

### DIFF
--- a/test/TestFITS.cc
+++ b/test/TestFITS.cc
@@ -125,7 +125,7 @@ TEST(Filtering, FailOnWrongFilterType) {
     EXPECT_FALSE(table.View().StringFilter(table["dummy"], "N 224"));
     EXPECT_FALSE(table.View().StringFilter(table["RA"], "N 224"));
 
-    EXPECT_FALSE(table.View().NumericFilter(table["dummy"], CARTA::RangeOpen, 0, 100));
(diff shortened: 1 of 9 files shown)


test/TestFITS.cc
```diff
@@ -125,14 +125,14 @@ TEST(Filtering, FailOnWrongFilterType) {
     EXPECT_FALSE(table.View().StringFilter(table["dummy"], "N 224"));
     EXPECT_FALSE(table.View().StringFilter(table["RA"], "N 224"));
 
-    EXPECT_FALSE(table.View().NumericFilter(table["dummy"], RANGE_INCLUSIVE, 0, 100));
-    EXPECT_FALSE(table.View().NumericFilter(table["Name"], RANGE_INCLUSIVE, 0, 100));
+    EXPECT_FALSE(table.View().NumericFilter(table["dummy"], CARTA::RangeOpen, 0, 100));
```
confluence: Has this deliberately been changed from closed to open?

confluence (commented):
One of the ranges seems to have changed.

test/TestFITS.cc
veggiesaurus: good catch, it should also be `RangeClosed`. Fixed

confluence (approved):

PR #554, opened 2020-07-08
Send cursor spatial message to update value

### LINKED ISSUE #553: pixel value does not update if the initial layout does not contain spatial profile widgets
to reproduce the issue:
1) create a layout without any spatial profile widgets, save and set as default layout
2) reload CARTA and load an image
3) move cursor around and observe the pixel value in the cursor info bar, which does not update while cursor is moving.

pford:
Closes #553. 

After set cursor, backend will send spatial profile message regardless of requirements.  Tested with mix n match.


veggiesaurus:
confirmed to work when moving cursor _and_ when changing channels :+1:

veggiesaurus (approved):

PR #555, opened 2020-07-14
Jason/line id overlay

duidae:
backend part for line id overlay

CI failure is due to incorrect protobuf branch, should be gone when line id overlay protobuf merged.


SpectralLineRequest.cc
```diff
@@ -0,0 +1,163 @@
...
+    "displaySLAIM=displaySLAIM&displayToyaMA=displayToyaMA&displayOSU=displayOSU&displayRecomb=displayRecomb&displayLisa=displayLisa&"
+    "displayRFI=displayRFI&ls1=ls1&ls2=ls2&ls3=ls3&ls4=ls4&ls5=ls5&el1=el1&el2=el2&el3=el3&el4=el4&show_unres_qn=show_unres_qn&"
+    "submit=Export&export_type=current&export_delimiter=tab&offset=0&limit=100000&range=on";
+
+SpectralLineRequest::SpectralLineRequest() {}
+
+SpectralLineRequest::~SpectralLineRequest() {}
+
+/* Referenced from curl example https://curl.haxx.se/libcurl/c/getinmemory.html */
+void SpectralLineRequest::SendRequest(const CARTA::DoubleBounds& frequencyRange, CARTA::SpectralLineResponse& spectral_line_response) {
```
veggiesaurus: I think [this approach](https://gist.github.com/alghanmi/c5d7b761b2c9ab199157#file-curl_example-cpp) looks a little more C++-friendly. I'd like to avoid `malloc`, `realloc` and `free` if possible

SpectralLineRequest.cc
```diff
@@ -0,0 +1,163 @@
...
+    chunk.memory = (char*)malloc(1);
+    chunk.size = 0;
+
+    curl_global_init(CURL_GLOBAL_ALL);
+
+    /* init the curl session */
+    curl_handle = curl_easy_init();
+
+    /* specify URL to get */
+    std::string frequencyRangeStr =
```
veggiesaurus: the rest of the backend uses `fmt::format` for templated strings like this

SpectralLineRequest.cc
```diff
@@ -0,0 +1,163 @@
...
+    CURL* curl_handle;
+    CURLcode res;
+
+    struct MemoryStruct chunk;
+
+    /* will be grown as needed by the realloc above */
+    chunk.memory = (char*)malloc(1);
+    chunk.size = 0;
+
+    curl_global_init(CURL_GLOBAL_ALL);
```
veggiesaurus: can't we just init and cleanup curl on start and end, rather than for every request?

Table/Columns.cc
```diff
@@ -52,6 +52,22 @@ std::unique_ptr<Column> Column::FromField(const pugi::xml_node& field) {
     return column;
 }
 
+std::unique_ptr<Column> Column::FromValues(const std::vector<string>& values, string name) {
+    unique_ptr<Column> column;
+
+    column = make_unique<DataColumn<string>>(name);
+    column->Resize(values.size());
+    for (auto row_index = 0; row_index < values.size(); row_index++) {
+        (dynamic_cast<DataColumn<string>*>(column.get()))->SetFromValue(values[row_index], row_index);
```
veggiesaurus: should just do the dynamic cast one, before the loop, and should also check that it is valid after casting

SpectralLineRequest.cc
```diff
@@ -0,0 +1,163 @@
...
+
+    mem->memory = ptr;
+    memcpy(&(mem->memory[mem->size]), contents, realsize);
+    mem->size += realsize;
+    mem->memory[mem->size] = 0;
+
+    return realsize;
+}
+
+void SpectralLineRequest::ParsingQueryResult(const MemoryStruct& results, CARTA::SpectralLineResponse& spectral_line_response) {
```
veggiesaurus: Would `ParseQueryResult` be better?

SpectralLineRequest.cc
```diff
@@ -0,0 +1,163 @@
...
+#include <string>
+
+#include "Table/Columns.h"
+
+using namespace carta;
+
+#define REST_FREQUENCY_COLUMN_INDEX 2
+
+const std::string SpectralLineRequest::SplatalogueURL =
+    "https://www.cv.nrao.edu/php/splat/"
```
veggiesaurus: I wonder if we should build the query string from a base URL and a number of required columns. Might be much easier to understand what's going on than a long URL. In future we may want to be able to customise the base URL for the server build, if it is hosted locally

veggiesaurus (changes requested):
few minor comments

duidae:
fixed, please have a look, thanks.

CMakeLists.txt
```diff
@@ -120,6 +120,7 @@ set(LINK_LIBS
         tbb
         cfitsio
         pugixml
+        curl
```
veggiesaurus, line 123: please can you add curl to the list of dependencies in the readme

Session.cc
```diff
@@ -961,6 +962,13 @@ void Session::OnCatalogFilter(CARTA::CatalogFilterRequest filter_request, uint32
     });
 }
 
+void Session::OnSpectralLineRequest(CARTA::SpectralLineRequest spectral_line_request, uint32_t request_id) {
+    CARTA::SpectralLineResponse spectral_line_response;
+    carta::SpectralLineRequest requester;
```
veggiesaurus: I think this is very confusing. `CARTA::SpectralLineRequest` and `carta::SpectralLineRequest` are very different objects. I'm not even sure why you need to create an instance of this, rather than just using the static functions. The `carta::SpectralLineRequest` class doesn't have any instance variables. Maybe `carta::SpectralDbQuery` or something like that would make more sense .

Table/Columns.cc
```diff
@@ -52,6 +52,25 @@ std::unique_ptr<Column> Column::FromField(const pugi::xml_node& field) {
     return column;
 }
 
+std::unique_ptr<Column> Column::FromValues(const std::vector<string>& values, string name) {
+    unique_ptr<Column> column;
```
veggiesaurus: this might be simpler:

```
std::unique_ptr<Column> Column::FromValues(const std::vector<string>& values, string name) {
    auto data_column_ptr = make_unique<DataColumn<string>>(name);
    data_column_ptr->Resize(values.size());
    if (data_column_ptr != nullptr) {
        for (auto row_index = 0; row_index < values.size(); row_index++) {
            data_column_ptr->SetFromValue(values[row_index], row_index);
        }
    }

    data_column_ptr->id = "";
    data_column_ptr->description = "";
    data_column_ptr->unit = "";
    data_column_ptr->ucd = "";
    return data_column_ptr;
}
```
(just double-check it produces the same results)

duidae:
Change class name from SpectralLineRequest to SpectralLineCrawler & change SendRequest() to static.

SpectralLineCrawler.cc
```diff
@@ -0,0 +1,142 @@
+#include "SpectralLineCrawler.h"
```
confluence: Could this class not go in the Table subdirectory? We should avoid putting anything in the top-level directory unless it's very general, and this looks like it's specific to catalogues.

confluence (changes requested):
See the comment re: the location of the new files.

duidae:
moved files to SpectralLine/

confluence (approved):
Looks fine to me.

veggiesaurus (approved):

PR #556, opened 2020-07-14
Region wcs matching with different coordinate frame

### LINKED ISSUE #552: Problem getting statistics for region when image files have differing coordinate systems
PR #539 made it possible to request analysis for a region that is defined in a different image, which is exactly what we need for the WCS region feature. However, it seems to struggle when the reference image's coordinate system doesn't match the secondary image's coordinate system. 

To reproduce in my test branch `angus/shared_regions` (this branch supports _displaying_ of point and polygon regions in spatially matched images, but not editing them):
1. Open a file (e.g. C.fits)
2. Create a polygonal region
3. Open a second file (e.g. C_GAL.fits)
4. Enable spatial matching. The polygonal region should now be visible
5. Switch to the first (reference file) and select the polygonal region
6. Open the stats widget. The stats for that region should be displayed correctly
7. Switch to the second file. The stats come up as empty, and the backend displays `Error applying region to file 1 (toLCRegion): WCRegion::toLCRegion - a region axis is unknown or inconsistent in target coordinate system`

With the debug execution stuff (Help -> Debug execution), you can just run this script to get things set up before opening the stats widget:

```
openFile("C.fits");
activeFrame.regionSet.addPolygonalRegion([{"x": 200, "y": 200}, {"x": 300, "y": 200}, {"x": 300, "y": 300}, {"x": 200, "y": 300}]);
appendFile("C_GAL.fits");
setSpatialMatchingEnabled($activeFrame, true);
setActiveFrameByIndex(0);
activeFrame.regionSet.selectRegionByIndex(1);
```
(Files included below)

This seems to work ok for images that have scaling and/or offsets, but the same coordinate system (e.g. `C.fits` and `C_pxScaled_0.05.fits`)

[test_images_wcs.zip](https://github.com/CARTAvis/carta-backend/files/4889642/test_images_wcs.zip)

pford:
Closes #552.

When the reference image wcs region cannot be converted to a lattice region in the matched image because of a different direction frame, the control points are converted to the new frame, put into a record, and the lattice region is created from the record.

### DIFF
--- a/Region/Region.cc
+++ b/Region/Region.cc
@@ -7,6 +7,7 @@
 
 #include <casacore/casa/Quanta/Quantum.h>
 #include <casacore/coordinates/Coordinates/DirectionCoordinate.h>
+#include <casacore/images/Regions/ImageRegion.h>
 #include <casacore/images/Regions/WCBox.h>
 #include <casacore/images/Regions/WCEllipsoid.h>
 #include <casacore/images/Regions/WCPolygon.h>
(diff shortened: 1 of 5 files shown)


veggiesaurus:
The example I used to find #552 is now working nicely. @kswang1029 can you give this a few tries using the `angus/shared_regions` branch? (just be sure to only create and edit the region in the _reference_ image for now, otherwise really weird things happen.

Issue #485, opened 2020-04-01
Export generated moment images in FITS image format


pford:
imageanalysis supports exporting moment images in CASA PagedImage format.  For 1.4, this should also support FITS using the casacore ImageFITSConverter.

Requirements doc: https://docs.google.com/document/d/1RciBmqsa_bZw1vgC4J0-YSOxnlIMDQyP-MINL81MYOs/edit#heading=h.pd8c2kw8bdfy

pford:
Reassigning to Mark.  Or you can close this issue and include it in #154.

Issue #154, opened 2019-04-04
Moments generator


Kechil:
To define a rectangle region, a range of channels, and a clip level to create moment images.
moments include (c.f. casa "immoments"):
moments=-1  - mean value of the spectrum  
moments=0   - integrated value of the spectrum  
moments=1   - intensity weighted coordinate;traditionally used to get ’velocity fields’  
moments=2   - intensity weighted dispersion of the coordinate; traditionally used to get "velocity dispersion"  
moments=3   - median of I  
moments=4   - median coordinate 
moments=5   - standard deviation about the mean of the spectrum  
moments=6   - root mean square of the spectrum  
moments=7   - absolute mean deviation of the spectrum  
moments=8   - maximum value of the spectrum  
moments=9   - coordinate of the maximum value of the spectrum 
moments=10  - minimum value of the spectrum  
moments=11  - coordinate of the minimum value of the spectrum

pford:
Requirements doc: https://docs.google.com/document/d/1RciBmqsa_bZw1vgC4J0-YSOxnlIMDQyP-MINL81MYOs/edit#heading=h.pd8c2kw8bdfy

PR #557, opened 2020-07-21
Support image moments (#154) and CASA/FITS file conversion (#485)

markccchiang:
The sequence diagram for the image moments process is shown below:

set_image_moments

It needs to build with a modified version of the CASA code. One can use the following commands to download and build with it:

```
git clone https://github.com/CARTAvis/carta-casacore.git --branch mark/for_moment_generator

cd carta-casacore
git submodule init && git submodule update

cd casa6
git submodule init && git submodule update
```

The class `carta::ImageMoments` inherits from the `casa::MomentsBase`. It is a rewrite version of `casa::ImageMoments` and a cancellation mechanism is added under it. A new class `ImagePtrLoader` which inherits from the `FileLoader` is used to load a moment image from the memory. The `FileConverter` class is used to convert the image format between CASA/FITS, and also used to save the moment images to the disk.


ImageData/CasaLoader.h
```diff
@@ -17,11 +17,12 @@ class CasaLoader : public FileLoader {
     ImageRef GetImage() override;
 
 private:
-    std::string _filename;
     std::unique_ptr<casacore::PagedImage<float>> _image;
 };
 
-CasaLoader::CasaLoader(const std::string& filename) : _filename(filename) {}
+CasaLoader::CasaLoader(const std::string& filename) {
+    _filename = filename;
```
confluence: What is the purpose of moving this assignment out of the initializer list and into the body of the constructor in all these subclasses?

Since the filename is now a member of the parent class, it may also be a good idea to initialize it in the parent constructor and delegate to that constructor from the other constructors.

ImageData/ImagePtrLoader.h
```diff
@@ -0,0 +1,56 @@
...
+    ImageRef GetImage() override;
+
+private:
+    std::shared_ptr<casacore::ImageInterface<float>> _image;
+};
+
+ImagePtrLoader::ImagePtrLoader(std::shared_ptr<casacore::ImageInterface<float>> image) {
+    _image = image;
+    if (_image) {
+        _num_dims = _image->shape().size();
```
confluence: I would assign these both in an initializer list, and move the check for the null pointer into the calling function (`FileLoader::GetLoader`).

Moment/ImageMoments.tcc
```diff
@@ -0,0 +1,695 @@
+//
+// Re-write from the file: "casa/code/imageanalysis/ImageAnalysis/ImageMoments.tcc"
```
confluence, line 2: What is the eventual plan for this code? Are we attempting to have these changes merged back into casa and casacore?

Moment/ImageMoments.tcc
```diff
@@ -0,0 +1,695 @@
...
+                if (mask_out.isWritable()) {
+                    if (!keep_axis) {
+                        result_array_masks[k].removeDegenerate(display_axes);
+                    }
+                    mask_out.putSlice(result_array_masks[k], result_pos);
+                }
+            }
+        }
+
+        if (tell_progress != 0) {
```
veggiesaurus, line 648: why don't we move the `if(_stop)` check to here? So that it's consistent with the `tell_progress` callback

Moment/MomentGenerator.cc
```diff
@@ -0,0 +1,355 @@
...
+
+    // Set moment calculation progress monitor
+    _image_moments->SetProgressMonitor(this);
+}
+
+int MomentGenerator::GetMomentMode(CARTA::Moment moment) {
+    int mode(-1);
+    switch (moment) {
+        case CARTA::Moment::MEAN_OF_THE_SPECTRUM: {
+            mode = ImageMoments<casacore::Float>::AVERAGE;
```
veggiesaurus: don't need `mode`, just `return` the value directly. 

Also, since this is just mapping CARTA moment types to casa types, why don't we just use a map?

Moment/MomentGenerator.cc
```diff
@@ -0,0 +1,355 @@
...
+    // Make an ImageMoments object (and overwrite the output file if it already exists)
+    _image_moments.reset(new ImageMoments<casacore::Float>(casacore::SubImage<casacore::Float>(*_sub_image), os, true));
+
+    // Set moment calculation progress monitor
+    _image_moments->SetProgressMonitor(this);
+}
+
+int MomentGenerator::GetMomentMode(CARTA::Moment moment) {
+    int mode(-1);
+    switch (moment) {
```
confluence: Could we replace this by a lookup in an unordered map?

We could also cast from the casa enum to ours if we number ours in the same way. But a map would be more explicit and less magical.

Moment/MomentGenerator.cc
```diff
@@ -0,0 +1,355 @@
...
+        case ImageMoments<casacore::Float>::MAXIMUM_COORDINATE: {
+            suffix = "maximum_coord";
+            break;
+        }
+        case ImageMoments<casacore::Float>::MINIMUM: {
+            suffix = "minimum";
+            break;
+        }
+        case ImageMoments<casacore::Float>::MINIMUM_COORDINATE: {
+            suffix = "minimum_coord";
```
veggiesaurus: same as above

Moment/MomentGenerator.cc
```diff
@@ -0,0 +1,355 @@
...
+            std::cerr << "Unknown moment mode: " << moment << std::endl;
+            break;
+        }
+    }
+    return mode;
+}
+
+casacore::String MomentGenerator::GetMomentSuffix(casacore::Int moment) {
+    casacore::String suffix;
+    switch (moment) {
```
confluence: This could also be a map lookup.

confluence (changes requested):

FileConverter.cc
```diff
@@ -0,0 +1,94 @@
...
+    output_filename = abs_path + "/" + output_filename;
+
+    if (output_filename == in_file) {
+        message = "The source file can not be overwritten!";
+        save_file_ack.set_success(success);
+        save_file_ack.set_message(message);
+        return;
+    }
+
+    // Check the writing permission
```
pford, line 39: For consistency can we use std::filesystem for this check?

Frame.cc
```diff
@@ -1227,6 +1228,96 @@ casacore::LCRegion* Frame::GetImageRegion(int file_id, std::shared_ptr<carta::Re
     return image_region;
 }
 
+bool Frame::GetImageRegion(const ChannelRange& chan_range, int stokes, casacore::ImageRegion& image_region) {
```
pford, line 1231: The ImageRegion is passed from the MomentController to the MomentGenerator to create a SubImage.  Couldn't this be done more simply in the Frame with a Slicer as in Frame::GetSlicerStats?  I think this needs the protection of the _image_mutex as well (to avoid casacore crashes with multithreading), which is why the RegionHandler routes calls through the Frame to get region data and stats.

Frame.h
```diff
@@ -156,6 +157,23 @@ class Frame {
     bool GetLoaderSpectralData(int region_id, int stokes, const casacore::ArrayLattice<casacore::Bool>& mask,
         const casacore::IPosition& origin, std::map<CARTA::StatsType, std::vector<double>>& results, float& progress);
 
+    // Get the full name of image file
```
pford, line 159: This is exposing the internals of the Frame and FileLoaders. I would rather see the Frame managing the moment generation for the image plane (with callbacks to Session for progress updates) rather than handing a Frame pointer and ImageRegion to the MomentController.  The Frame can also manage its processes if it needs to close.

ImageData/ImagePtrLoader.h
```diff
@@ -0,0 +1,56 @@
+#ifndef CARTA_BACKEND_IMAGEDATA_IMAGEPTRLOADER_H_
+#define CARTA_BACKEND_IMAGEDATA_IMAGEPTRLOADER_H_
+
+#include "FileLoader.h"
+
+namespace carta {
+
+class ImagePtrLoader : public FileLoader {
+public:
+    ImagePtrLoader(std::shared_ptr<casacore::ImageInterface<float>> image);
```
pford, line 10: Is this image type a PagedImage like a CASA image?  Wondering if we can just add another constructor to CasaLoader.

Region/RegionHandler.h
```diff
@@ -65,6 +65,9 @@ class RegionHandler {
         std::function<void(CARTA::SpectralProfileData profile_data)> cb, int region_id, int file_id, bool stokes_changed);
     bool FillRegionStatsData(std::function<void(CARTA::RegionStatsData stats_data)> cb, int region_id, int file_id);
 
+    // Get an image region with respect to the region id, file id, channels range and a stoke type
```
pford, line 68: As with the Frame, I would prefer to keep this private and have the RegionHandler create the ImageRegion then call a Frame method to create the SubImage (with a mutex) and call the moments calculator generator.

pford (commented):
I have concerns about the architecture, making things public so they can be accessed by the Session and handed to the moments calculator.  But it may be too close to code freeze to make any large changes.

FileConverter.cc
veggiesaurus: specifically, we should use `fs`:

```
#ifdef _BOOST_FILESYSTEM_
#include <boost/filesystem.hpp>
namespace fs = boost::filesystem;
#else
#include <filesystem>
namespace fs = std::filesystem;
#endif
```
since the MacOS 10.14 C++17 implementation is not feature-complete.

Moment/ImageMoments.tcc
markccchiang: Yes. To avoid the repetition of codes. It is better to change them in the casa and casacore.

ImageData/ImagePtrLoader.h
markccchiang: Using `doTemp` option the moment generator creates an image type`TempImage`, which inherits from the `ImageInterface`. So I think it is not the `PagedImage` and the `TempImage` can not convert to this type.

markccchiang:
As for @pford's comments about the architecture. Allow me to spend one more day and see if I can modify the code properly.

pford:
@markccchiang use your implementation for 1.4, and we will discuss if refactoring is needed for 1.5.

veggiesaurus:
> @markccchiang use your implementation for 1.4, and we will discuss if refactoring is needed for 1.5.

we can try to combine this with PRs to imageanalysis to remove code repetition as well :+1:

confluence (approved):
I'm also happy to use this as a stopgap for 1.4, provided that we revisit this to discuss moving the casa code back into casa and other refactoring issues before 1.5.

pford (approved):

PR #561, opened 2020-07-22
Angus/issue 463

### LINKED ISSUE #463: Handling of NaN values during Gaussian smoothing
When smoothing, there are a few options for how to handle `NaN` values:
1. Propagate them, so that every pixel that has a `NaN` within its smoothing kernel becomes a `NaN`. This is the route DS9 takes, and it's not a good idea, because you lose a lot of real data around NaNs
2. Suppress them, so that `NaN` pixels are skipped during the smoothing. Only pixels that have _no_ non-`NaN` pixels within its smoothing kernel persist as `NaN`. This is the default behaviour of astropy (see [here](http://docs.astropy.org/en/stable/convolution/) for a detailed discussion on this issue). This is the current approach used in CARTA. 
3. Suppress them, but persist `NaN` values where the original data is `NaN`.

Option 2 is quite nice for _displaying_ an image, because it "recovers" data that is missing due to `NaN`s, and doesn't generate a whole bunch of contours surrounding `NaN` values, as shown in the top image in each set below. However, there's an issue with this approach: it can result in contours that pass through what is originally `NaN` data, which is misleading, as shown in the bottom image in each set below. For block smoothing, since we are effectively reducing the image size, there's no way to implement Option 3, but for Gaussian smoothing, we can put `NaN` values in their original location after smoothing, ensuring that contours never cross a `NaN` value.

@kswang1029 and I have discussed this and settled on changing it to Option 3, but I never got round to it (until today), and thought we should check with @arusstaylor @low-sky and @jott3077 before making this change (does CF not have GitHub?)

### No smoothing (`dev`):
image
image

### 4x Block smoothing (`dev`, Option 2):
image
image

### 4x Gaussian smoothing (`dev`, Option 2):
image
image

### 4x Gaussian smoothing (`angus/issue_463`, Option 3):
image
image

veggiesaurus:
Closes #463 

(I totally forgot that this hadn't been merged already!)


pford (approved):

kswang1029:
ICD tests for contours need to be re-designed to support this.

PR #562, opened 2020-07-26
attempt to get remote address from x-forwarded-for header

veggiesaurus:
very minor update to check the `x-forwarded-for` header for the remote address first, before assuming it is the same as the WebSocket connection address. This prevents us from always logging `127.0.0.1` as the remote address.


confluence (approved):

veggiesaurus:
@confluence I made some minor changes, please re-check

confluence (approved):

PR #565, opened 2020-07-28
Check file is readable before exists for file list

### LINKED ISSUE #564: Check folder read permissions before checking whether folder is a CASA image
Currently, our CARTA backend logs are filled with errors such as

`WARN	File::exists	lstat failed for /users/angus/table.dat: errno=13'Permission denied'`

These crop up when a user navigates to the `/users` folder, and then only has access to one subfolder. However, we're attempting to check _every_ subfolder to see if it's a CASA image folder. We should first check to see if that folder is accessible at all before attempting to check for `table.dat` inside it.

pford:
Closes #564


pford:
@veggiesaurus please test, could not reproduce issue running as root in Docker.

veggiesaurus (approved):

Issue #475, opened 2020-03-23
[question] logging ICD message history in verbose mode


kswang1029:
Currently when verbose mode is enabled, the log seems not providing sufficient information when the backend crashed (well, it is seg fault but it is not trivial to see which step caused the crash). I am wondering if it is possible to log all ICD messages in the log as well? @veggiesaurus @pford @ajm-asiaa

Kechil:
Multiple verbosity levels?

kswang1029:
possibly one level for logging messages and another level for performance timers.

kswang1029:
@markccchiang could you help on getting this implemented for v1.4?

kswang1029:
@veggiesaurus other requirements from server point of view are needed. Please help to address those.

veggiesaurus:
I think the main thing we need from logging is a number of verbosity levels and support for tags. [Boost.log](https://www.boost.org/doc/libs/1_73_0/libs/log/doc/html/log/tutorial/formatters.html) provides that, as well as a fair amount of other nice things (automatically rotating log files, etc). We could also use it to eventually have optional integration with something like syslog.

I’m not sure that just logging ICD messages is helpful other than in the most verbose log level

markccchiang:
So we have a log file with a fixed maximum size. In a log file, we have a tag for ICD messages, a tag for performance (like elapsed time for some expensive calculations), and the other tags if we are interested. Is that the way?

veggiesaurus:
> So we have a log file with a fixed maximum size. In a log file, we have a tag for ICD messages, a tag for performance (like elapsed time for some expensive calculations), and the other tags if we are interested. Is that the way?

Yes. Messages can have multiple tags, so they can be ICD, performance, etc and also tagged with the class they're emitted from

markccchiang:
Besides Boost.log, another popular tool is [spdlog](https://github.com/gabime/spdlog). It seems this library also fits our requirements. Will choose one of them for implementation.

veggiesaurus:
Spdlog looks good. Does it support multiple tags?

markccchiang:
> Spdlog looks good. Does it support multiple tags?

[...]

PR #570, opened 2020-07-31
Log ICD messages in a file while the verbose mode is on

markccchiang:
For issue #475, use the logging package [Spdlog](https://github.com/gabime/spdlog) to save ICD messages in a log file. The default log file path name is `$HOME/CARTA/log/carta.log`. Users can also manually set the file path name by the command line: `./carta_backend verbose=true log_dir=/path/to/the/log`. An error will report if the log file path name is not valid or users have no write permission. The maximum size of the rotated log file is 5 Mb. The pattern of the log file for ICD messages is like following:

```
[2020-07-31 14:25:54.240] [<<==] ">>>>>>>>> Start the RECEIVE logger <<<<<<<<<"
[2020-07-31 14:25:54.240] [==>>] ">>>>>>>>> Start the SEND logger <<<<<<<<<"
[2020-07-31 14:26:16.656] [<<==] REGISTER_VIEWER
[2020-07-31 14:26:16.656] [==>>] REGISTER_VIEWER_ACK
[2020-07-31 14:26:17.626] [<<==] FILE_LIST_REQUEST
[2020-07-31 14:26:17.631] [==>>] FILE_LIST_RESPONSE
[2020-07-31 14:26:21.862] [<<==] FILE_LIST_REQUEST
[2020-07-31 14:26:21.872] [==>>] FILE_LIST_RESPONSE
[2020-07-31 14:26:29.526] [<<==] FILE_LIST_REQUEST
[2020-07-31 14:26:29.581] [==>>] FILE_LIST_RESPONSE
[2020-07-31 14:26:30.502] [<<==] FILE_INFO_REQUEST
[2020-07-31 14:26:30.522] [==>>] FILE_INFO_RESPONSE
[2020-07-31 14:26:31.528] [<<==] CLOSE_FILE
[2020-07-31 14:26:31.530] [<<==] OPEN_FILE
[2020-07-31 14:26:31.620] [==>>] OPEN_FILE_ACK
[2020-07-31 14:26:31.698] [==>>] REGION_HISTOGRAM_DATA
[2020-07-31 14:26:32.002] [<<==] ADD_REQUIRED_TILES
[2020-07-31 14:26:32.002] [==>>] RASTER_TILE_SYNC
[2020-07-31 14:26:32.003] [<<==] SET_CURSOR
[2020-07-31 14:26:32.023] [==>>] RASTER_TILE_DATA
[2020-07-31 14:26:32.024] [==>>] RASTER_TILE_DATA
[2020-07-31 14:26:32.031] [==>>] RASTER_TILE_DATA
```
Where the timestamp precision is up to milliseconds.


Logger/Logger.cc
```diff
@@ -0,0 +1,330 @@
...
+void LogReceiveEventType(uint16_t event_type) {
+    std::string logger_name = "<<==";
+    std::shared_ptr<spdlog::logger> receive_logger = spdlog::get(logger_name);
+    if (!receive_logger) {
+        spdlog::error("Fail to get the logger {0}", logger_name);
+        return;
+    }
+
+    switch (event_type) {
+        case CARTA::EventType::REGISTER_VIEWER: {
```
veggiesaurus: please use an `unordered_map<CARTA::EventType, string>` instead of a very long switch

Logger/Logger.h
```diff
@@ -0,0 +1,19 @@
...
+#include <fmt/format.h> // This header must be before spdlog headers
+#include <spdlog/sinks/rotating_file_sink.h>
+#include <spdlog/sinks/stdout_color_sinks.h>
+#include <spdlog/spdlog.h>
+
+// Create loggers
+void CreateLoggers(std::string log_filename = "");
+
+// Log event types for receiving from the frontend
+void LogReceiveEventType(uint16_t event_type);
```
veggiesaurus: Rename to `LogReceivedEventType` and `LogSentEventType`

veggiesaurus (changes requested):
Please also adjust the readme to included spdlog as one of the dependencies

kswang1029:
@markccchiang shall we swap "==>" and "<==" to be consistent to what we see in the js console? It would be more user friendly this way. [please ignore this part]

 is carta.log dedicated to collect ICD messages only? If so, shall we name it specifically? (eg icd_msg.log)

where would those performance-related log go?

veggiesaurus:
> @markccchiang shall we swap "==>" and "<==" to be consistent to what we see in the js console? It would be more user friendly this way.
> 
> is carta.log dedicated to collect ICD messages only? If so, shall we name it specifically? (eg icd_msg.log)
> 
> where would those performance-related log go?

shouldn't they be the opposite, since the backend is sending and the frontend is receiving? or am I misunderstanding you?

kswang1029:
uh.. I was referencing js console, such as
Screen Shot 2020-07-31 at 5 18 46 PM
Frontend is on the left hand side so outgoing is ==> and incoming is <==.

If we put a hypothetical backend on the right hand side, the same notations should work consistently. 

Screen Shot 2020-07-31 at 5 22 56 PM

veggiesaurus:
> uh.. I was referencing js console, such as
> Screen Shot 2020-07-31 at 5 18 46 PM
> Frontend is on the left hand side so outgoing is ==> and incoming is <==.
> 
> If we put a hypothetical backend on the right hand side, the same notations should work consistently.
> 
> Screen Shot 2020-07-31 at 5 22 56 PM

oh I see what you mean. So changing from "==>>"  to "==>". I thought you meant switching the directions around

kswang1029:
I do mean switching the directions. In the backend log

> [2020-07-31 14:26:16.656] [==>>] REGISTER_VIEWER_ACK

ACK points to the right but in js console it points to the left.

Or, the meaning of "==>" and "<==" is a kind of standard notation? If so, please forget about what I was suggesting :)

veggiesaurus:
> I do mean switching the directions. In the backend log
> 
> > [2020-07-31 14:26:16.656] [==>>] REGISTER_VIEWER_ACK
> 
> ACK points to the right but in js console it points to the left.
> 
> Or, the meaning of "==>" and "<==" is a kind of standard notation? If so, please forget about what I was suggesting :)

ah, I see what you mean now. I think it makes sense for them to be pointing in opposite directions, since an incoming message in the one is an outgoing message in the other

kswang1029:
ok. so "==>" always means outgoing and "<==" means incoming. Now I see they do have meanings. 👍

veggiesaurus:
I agree that you should change the file name to something indicating it's an ICD log, at least until _everything_ is logged using the same log library (that's the end goal, hence the need for tags and multiple verbosity levels). As it stands now, writing to `~/CARTA/log/carta.log` will overwrite the log file that @ajm-asiaa creates to log `stdout` from `carta_backend`.

ajm-ska:
@veggiesaurus It will be OK. The run.sh script that I use to start the carta_backend redirects the `stdout` output to a file with the date and time information. i.e.

> logfilename=$HOME/CARTA/log/$(date +"%Y%m%d_%H%M%S_%Z").log
> 
> $dirname/carta_backend verbose=true base=$1 root=$2 port=$3 threads=$4 omp_threads=$5 >> $logfilename 2>&1

veggiesaurus:
> @veggiesaurus It will be OK. The run.sh script that I use to start the carta_backend redirects the `stdout` output to a file with the data and time information. i.e.
> 
> > logfilename=$HOME/CARTA/log/$(date +"%Y%m%d_%H%M%S_%Z").log
> > $dirname/carta_backend verbose=true base=$1 root=$2 port=$3 threads=$4 omp_threads=$5 >> $logfilename 2>&1

ok, good. @markccchiang I do think we should include the prefix `icd_msg` or something like that until we've properly unified our logging

markccchiang:
Ok. I will change the log file name as `icd_msg.log`. And users can specify the log file directory by setting the variable `log_dir`. But they can not specify the log file name. I think in the future we can save another tag messages in a different log file name if we are interested in. Or just unify them in a single log file, it depends.

markccchiang:
Have pushed new codes. Now the ICD log file name is fixed as `icd_msg.log`. The log pattern is like the following:
```
[2020-08-01 01:03:37.495] [<==] ">>>>>>>>> Start the incoming logger <<<<<<<<<"
[2020-08-01 01:03:37.495] [==>] ">>>>>>>>> Start the outgoing logger <<<<<<<<<"
[2020-08-01 01:03:46.366] [<==] REGISTER_VIEWER
[2020-08-01 01:03:46.367] [==>] REGISTER_VIEWER_ACK
[2020-08-01 01:03:47.141] [<==] FILE_LIST_REQUEST
[2020-08-01 01:03:47.147] [==>] FILE_LIST_RESPONSE
[2020-08-01 01:03:50.335] [<==] FILE_LIST_REQUEST
[2020-08-01 01:03:50.341] [==>] FILE_LIST_RESPONSE
[2020-08-01 01:03:52.480] [<==] FILE_LIST_REQUEST
[2020-08-01 01:03:52.532] [==>] FILE_LIST_RESPONSE
[2020-08-01 01:03:54.998] [<==] FILE_INFO_REQUEST
[2020-08-01 01:03:55.011] [==>] FILE_INFO_RESPONSE
[2020-08-01 01:03:58.032] [<==] CLOSE_FILE
[2020-08-01 01:03:58.034] [<==] OPEN_FILE
[2020-08-01 01:03:58.064] [==>] OPEN_FILE_ACK
[2020-08-01 01:03:58.069] [==>] REGION_HISTOGRAM_DATA
[2020-08-01 01:03:58.207] [<==] ADD_REQUIRED_TILES
[2020-08-01 01:03:58.207] [==>] RASTER_TILE_SYNC
[2020-08-01 01:03:58.213] [==>] RASTER_TILE_DATA
[2020-08-01 01:03:58.216] [==>] RASTER_TILE_DATA
[2020-08-01 01:03:58.218] [<==] SET_CURSOR
[2020-08-01 01:03:58.221] [==>] RASTER_TILE_DATA
[2020-08-01 01:03:58.224] [==>] RASTER_TILE_DATA
[2020-08-01 01:03:58.225] [==>] RASTER_TILE_SYNC
[2020-08-01 01:03:58.225] [==>] SPATIAL_PROFILE_DATA
[2020-08-01 01:03:58.627] [<==] SET_SPATIAL_REQUIREMENTS
[2020-08-01 01:03:58.627] [==>] SPATIAL_PROFILE_DATA
[2020-08-01 01:03:59.238] [<==] SET_CURSOR
[2020-08-01 01:03:59.239] [==>] SPATIAL_PROFILE_DATA
```

veggiesaurus:
@ajm-asiaa after building spdlog from source, don't we still need to use `FIND_PACKAGE`? I don't think we should just get rid of it, since for 20.04 we can use the APT version without any major issue

ajm-ska:
@veggiesaurus I'm a bit behind on testing. It seems to work on 20.04 without FIND_PACKAGE. If `libspdlog-dev` is not installed, it is noticed in the make procedure. If `libspdlog-dev` is installed, it builds successfully. But if using FIND_PACKAGE in CMakelists.txt is the more proper way of doing things, please add it back @markccchiang 

It looks like ubuntu 20.04 is the only OS version that we can install spdlog using the package manager. Surprisingly it seems we can't use it in CentOS8. CentOS8 installs `spdlog-devel-1.5.0-2.el8.x86_64`(seemingly the same version as ubuntu 20.04), but the build fails with:

> /usr/include/spdlog/fmt/fmt.h:22:10: fatal error: spdlog/fmt/bundled/core.h: No such file or directory
>  #include <spdlog/fmt/bundled/core.h>
>           ^~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.

That 'bundled' folder doesn't exist in the CentOS8 version, but does on ubuntu 20.04.

veggiesaurus:
> @veggiesaurus I'm a bit behind on testing. It seems to work on 20.04 without FIND_PACKAGE. If `libspdlog-dev` is not installed, it is noticed in the make procedure. If `libspdlog-dev` is installed, it builds successfully. But if using FIND_PACKAGE in CMakelists.txt is the more proper way of doing things, please add it back @markccchiang
> 
> It looks like ubuntu 20.04 is the only OS version that we can install spdlog using the package manager. Surprisingly it seems we can't use it in CentOS8. CentOS8 installs `spdlog-devel-1.5.0-2.el8.x86_64`(seemingly the same version as ubuntu 20.04), but the build fails with:
> 
> > /usr/include/spdlog/fmt/fmt.h:22:10: fatal error: spdlog/fmt/bundled/core.h: No such file or directory
> > #include <spdlog/fmt/bundled/core.h>
> > ^~~~~~~~~~~~~~~~~~~~~~~~~~~
> > compilation terminated.
> 
> That 'bundled' folder doesn't exist in the CentOS8 version, but does on ubuntu 20.04.

That's odd. I'm wondering why it doesn't try to use the `fmt` includes from elsewhere :thinking:

markccchiang:
Spdlog has its own fmt bundled inside. Looking at its [CMakeLists.txt](https://github.com/gabime/spdlog/blob/v1.x/CMakeLists.txt), you can choose to turn ON the variable `SPDLOG_FMT_EXTERNAL` or `SPDLOG_FMT_EXTERNAL_HO` while building. Then it will build with the fmt package elsewhere. By default, these two variables are turned OFF. My laptop is Mac OS and I did not try to build it on the Linux. Sorry about that!

ajm-ska:
That's OK. It suggests the CentOS8 package manger version of spdlog was built without the bundled `fmt`, but the Ubuntu 20.04 version was.

Anyway, there seems to be a different issue with CentOS7 now. I install spdlog normally from Github (using our Dockerfile):

> RUN \
>   cd /root && \
>   git clone https://github.com/gabime/spdlog.git && \
>   mkdir -p spdlog/build && cd spdlog/build && \
>   cmake .. && make && make install && \
>   cd /root && rm -rf spdlog

But then the carta-backend build fails with many errors starting with:

> 
> /usr/local/include/spdlog/common.h:103:27: error: 'basic_memory_buffer' in namespace 'fmt' does not name a template type
>  using memory_buf_t = fmt::basic_memory_buffer<char, 250>;
>                            ^~~~~~~~~~~~~~~~~~~
> /usr/local/include/spdlog/common.h:103:22: note: suggested alternative: 'BasicMemoryWriter'
>  using memory_buf_t = fmt::basic_memory_buffer<char, 250>;
>                       ^~~
>                       BasicMemoryWriter


Then I try to build spdlog with the external fmt option, `cmake .. -DSPDLOG_FMT_EXTERNAL=ON` it says 

> CMake Error at CMakeLists.txt:174 (find_package):
>   Could not find a configuration file for package "fmt" that is compatible
>   with requested version "5.3.0".
> 
>   The following configuration files were considered but not accepted:
> 
>     /usr/share/cmake/fmt/fmt-config.cmake, version: 3.0.2
> 
> 
> 
> -- Configuring incomplete, errors occurred!

CentOS7 uses `fmt-3.0.2-1.el7.x86_64`

So I install the latest`fmt` from Github, and reinstall `spdlog` with `cmake .. -DSPDLOG_FMT_EXTERNAL=ON`. Then the carta_backend build gets further, fails with many linking errors. So I put in the steps to build `fmt` and `spdlog` from source into the Dockerfile, and now the carta_backend builds and runs successfully.

veggiesaurus:
These are troubling complications. I wonder if we could try avoid these issues somehow. We should at least be able to use built in CentOS 8 packages without resorting to building from source

ajm-ska:
I found an alternative solution for CentOS8 [here](https://github.com/gabime/spdlog/issues/658)
We can modify `Logger/Logger.h` by adding the following two lines before the spdlog includes:

> **#define SPDLOG_FMT_EXTERNAL**
> **#define FMT_HEADER_ONLY**

> #include <spdlog/sinks/rotating_file_sink.h>
> #include <spdlog/sinks/stdout_color_sinks.h>
> #include <spdlog/spdlog.h>

Then the carta-backend builds and runs with package manager installed `spdlog-devel` and `fmt-devel`!

veggiesaurus:
> I found an alternative solution for CentOS8 [here](https://github.com/gabime/spdlog/issues/658)
> We can modify `Logger/Logger.h` by adding the following two lines before the spdlog includes:
> 
> > **#define SPDLOG_FMT_EXTERNAL**
> > **#define FMT_HEADER_ONLY**
> 
> > #include <spdlog/sinks/rotating_file_sink.h>
> > #include <spdlog/sinks/stdout_color_sinks.h>
> > #include <spdlog/spdlog.h>
> 
> Then the carta-backend builds and runs with package manager installed `spdlog-devel` and `fmt-devel`!

Does this still work with Ubuntu 20.04?

ajm-ska:
> Does this still work with Ubuntu 20.04?

Confirmed. With the two new lines added to Logger.h, it builds and appears to run successfully on Ubuntu 20.04.

Next I'll try ubuntu 18.04.

ajm-ska:
To work on ubuntu 18.04, we need to install the newer `fmt` plus `spdlog` from source. It can be simply added to the Dockerfile as follows:

```
  RUN \
  cd /root && \
  git clone https://github.com/fmtlib/fmt && \
  mkdir -p fmt/build && cd fmt/build && \
  cmake .. && make && make install && \
  cd /root && rm -rf fmt

  RUN \
  cd /root && \
  git clone https://github.com/gabime/spdlog && \
  mkdir -p spdlog/build && cd spdlog/build && \
  cmake .. -DSPDLOG_FMT_EXTERNAL=ON && \
  make && make install && \
  cd /root && rm -rf spdlog
```
I can't find any PPA, unless we consider making one ourselves.

Also the two extra lines need to be added to `Logger/Logger.h`:
```
    #define SPDLOG_FMT_EXTERNAL
    #define FMT_HEADER_ONLY
```

ajm-ska:
I also checked adding:
```
#define SPDLOG_FMT_EXTERNAL
#define FMT_HEADER_ONLY
```
to `Logger/Logger.h` on MacOS. It still appears to build and run fine.

markccchiang:
@ajm-asiaa Thank you for the nice work! It is better that you push the new change of these codes rather than by me.

ajm-ska:
@veggiesaurus Don't merge just yet. I want to double check all the Dockerfiles from scratch first, plus maybe add back the the `FIND_PACKAGE` to CMakelists.txt.

This is also a general Dockerfile update as `pugixml` had been missing. For the main Dockerfile, I added update-alternatives to set gcc-8 as the default compiler so that you don't need to do the `CXX=g++-8 CC=gcc-8 cmake ..` command. I also removed the `git checkout angus/dev` line. I hope that doesn't mess anything up for you? I made `Dockerfiles/Dockerfile-ubuntu-18.04` a symlink back to the main Dockerfile so that people know it is for 18.04.

markccchiang:
@ajm-asiaa There are some conflicts of Dockerfiles with the dev branch. Not sure are they fixed now? I will fix the other conflicts in the "Main.cc".

ajm-ska:
@markccchiang It looks like I can't just resolve the conflict in the two Dockerfiles and leave Main.cc (using the web interface). Otherwise it doesn't seem to save. The problem with the Dockerfiles is that @veggiesaurus has since removed `libjsoncpp-dev`. So `libjsoncpp-dev` just needs to be removed from both files.

markccchiang:
@ajm-asiaa Ok, then I will try to merge them.

ajm-ska:
@markccchiang I changed the Dockerfiles back to effectively undo your `d758aad` commit. Simply changing the package installation section of all the Dockerfiles to make them consistent with the current `dev` branch would not work because I had purposely modified them to work with the new `spdlog` package plus some other essential updates (e.g. installing pugixml). I also added back the handy `Dockerfiles/Dockerfile-ubuntu-18.04` symlink to the main Dockerfile.

PR #573, opened 2020-08-03
removes deprecated auth support

veggiesaurus:
this PR removes all deprecated auth support in preparation for the carta-node-server implementation rolling out with 1.4. The following has been removed:

- API key and permissions checking
- Auth token and JSON file reading
- mongo support, as well as user/layout getters and setters

This should make building the backend simpler, as mongo and json-cpp are no longer required


confluence (changes requested):
It looks like there's some weird whitespace in here, and the format check is failing.

confluence:
> It looks like there's some weird whitespace in here, and the format check is failing.

I see you fixed that while I was typing. :)

confluence (approved):
Looks OK.

veggiesaurus:
@confluence some additional changes (the auth feature we discussed in today's meeting). Please re-review

confluence (approved):
Looks OK.

PR #584, opened 2020-08-12
Angus/catalog info strings

veggiesaurus:
Adds additional catalog info in the file info description field:
- Column count
- Row count (if it is defined)
- Parameters (from VOTable `PARAM` nodes)
- Coordinate system info (from VOTable `COOSYS` node)

### DIFF
--- a/Table/Table.cc
+++ b/Table/Table.cc
@@ -75,13 +75,14 @@ bool Table::ConstructFromXML(bool header_only) {
             return false;
         }
     }
-
     auto votable = doc.child("VOTABLE");
 
     if (!votable) {
         _parse_error_message = "Missing XML element VOTABLE!";
         return false;
     }
+
+    PopulateCoosys(votable);
 
     auto resource = votable.child("RESOURCE");
     if (!resource) {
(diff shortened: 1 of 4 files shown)


markccchiang (approved):

Table/Table.cc
```diff
@@ -216,28 +281,29 @@ bool Table::ConstructFromFITS(bool header_only) {
...
-#pragma omp parallel for default(none) schedule(dynamic) shared(num_cols, buffer, _num_rows, total_width)
+#pragma omp parallel for default(none) schedule(dynamic) shared(num_cols, buffer, allocated_rows, total_width)
         for (auto i = 0; i < num_cols; i++) {
-            _columns[i]->FillFromBuffer(buffer.get(), _num_rows, total_width);
+            _columns[i]->FillFromBuffer(buffer.get(), allocated_rows, total_width);
         }
     } else {
         fits_close_file(file_ptr, &status);
     }
+    _num_rows = rows;
```
confluence: Is this intentionally set to `rows` and not `allocated_rows`?

veggiesaurus: yes. `allocated_rows` is the number of rows to allocate for reading. When reading the header, this is zero. I'll add a comment

veggiesaurus: actually, this is misleading. We need a new field to indicate the total rows available, not the total rows read into the table

veggiesaurus:
@confluence please have another look, I've made some adjustments

confluence (approved):

PR #585, opened 2020-08-13
Check file id before proceeding to send tiles

### LINKED ISSUE #581: hanging tile stream
to reproduce the issue with the dev backend + dev frontend
1) load set_HD163296/HD163296_CO_2_1.fits
2) load set_HD163296/HD163296_13CO_2_1.fits
3) load set_HD163296/HD163296_C18O_2_1.fits
4) match them all in xy and z
5) switch to HD163296_CO_2_1.fits (image 0)
6) close it so all matched image becomes unmatched
7) match all image again in xy and z
8) use animator to switch to other channels. The first trial 'may' succeed but the second one will result in hanging tile stream

pford:
Closes #581 

Message handling for AddRequiredTiles hangs if image has been closed.  Returns if file id is not in the frames map.

### DIFF
--- a/Session.cc
+++ b/Session.cc
@@ -366,6 +366,10 @@ void Session::DeleteFrame(int file_id) {
 
 void Session::OnAddRequiredTiles(const CARTA::AddRequiredTiles& message, bool skip_data) {
     auto file_id = message.file_id();
+    if (!_frames.count(file_id)) {
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


kswang1029:
Confirmed. The issue is gone with this fix! 👍

veggiesaurus (approved):

PR #589, opened 2020-08-13
Approximate shared region as polygon in non-reference image

pford:
Creates polygon region with 1000 points to convert to another image's coordinate system, to avoid projection issues.  Later, the number of points will be configurable.


pford:
@kswang1029, @veggiesaurus added a mutex to solve the weird/non-existent spectral profile issue.

kswang1029 (approved):

PR #590, opened 2020-08-14
fixes compile issues when using boost::filesystem

veggiesaurus:
will do a backport if this works out ok


markccchiang (approved):

ajm-ska (approved):
It seems OK. It appears to build and run successfully on MacOS 10.14 with the `-DUseBoostFilesystem=True` flag, plus builds and runs successfully on MacOS 10.15 without that flag.

PR #597, opened 2020-08-14
attempt to fix spectral profile crash with HDF5 files

veggiesaurus:
@confluence suggested this fix to prevent a hard crash when requesting a spectral profile from a region of an HDF5 file


pford (approved):

pford:
This should handle both cases (native region or shared region).  I will review Region after the release since things did not work the way I thought they would and could use some refactoring.

confluence (approved):

PR #600, opened 2020-08-18
Added gRPC and pugixml dependencies to instructions

confluence:
Is this detailed enough?


ajm-ska (approved):
Looks good to me. Short and concise as it should be in that section.

I think I see an error in the `fmt` line. I just tried but `libfmt3-dev` doesn't seem to exist on 18.04 or 20.04. It should be changed to `libfmt-dev`.
If you decide to change it, and as you are mentioning 16.04 for the other packages, maybe you could also add that libfmt-dev does not exist on Ubuntu versions 16.04 and needs to be installed from source (https://github.com/fmtlib/fmt.git)?

PR #601, opened 2020-08-19
Remove redundancy codes and minor correction

zarda:
Minor correction of the codes for measuring performance

### DIFF
--- a/Region/RegionHandler.cc
+++ b/Region/RegionHandler.cc
@@ -1123,8 +1123,8 @@ bool RegionHandler::GetRegionSpectralData(int region_id, int file_id, std::strin
             if (_verbose) {
                 auto t_end_spectral_profile = std::chrono::high_resolution_clock::now();
                 auto dt_spectral_profile =
-                    std::chrono::duration_cast<std::chrono::milliseconds>(t_end_spectral_profile - t_start_spectral_profile).count();
-                fmt::print("Fill spectral profile in {} ms\n", dt_spectral_profile);
+                    std::chrono::duration_cast<std::chrono::microseconds>(t_end_spectral_profile - t_start_spectral_profile).count();
+                fmt::print("Fill spectral profile in {} ms\n", dt_spectral_profile * 1e-3);
             }
 
             _frames.at(file_id)->DecreaseZProfileCount();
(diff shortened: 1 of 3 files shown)


Region/RegionHandler.cc
```diff
@@ -881,7 +881,8 @@ bool RegionHandler::GetRegionHistogramData(
         auto t_end_region_histogram = std::chrono::high_resolution_clock::now();
         auto dt_region_histogram =
             std::chrono::duration_cast<std::chrono::microseconds>(t_end_region_histogram - t_start_region_histogram).count();
```
pford, line 883: Out of curiosity: why do you get the count in microseconds then convert ( * 1e-3) to milliseconds?  You can get the count in milliseconds: `std::chrono::duration_cast<std::chrono::milliseconds>(t_end - t_start).count();`

pford (approved):
Thank you for fixing this mess.  I hope it was a merge issue and I did not repeat code this badly!

Region/RegionHandler.cc
veggiesaurus: > Out of curiosity: why do you get the count in microseconds then convert ( * 1e-3) to milliseconds? You can get the count in milliseconds: `std::chrono::duration_cast<std::chrono::milliseconds>(t_end - t_start).count();`

if you get it in milliseconds, doesn't it round it to the nearest millisecond?

pford: Yes, you are right.

Session.cc
```diff
@@ -947,12 +947,6 @@ void Session::OnResumeSession(const CARTA::ResumeSession& message, uint32_t requ
...
     // Measure duration for resume
-    if (_verbose_logging) {
-        auto t_end_resume = std::chrono::high_resolution_clock::now();
-        auto dt_resume = std::chrono::duration_cast<std::chrono::microseconds>(t_end_resume - t_start_resume).count();
-        fmt::print("Resume in {} ms\n", dt_resume * 1e-3);
-    }
-
     if (_verbose_logging) {
         auto t_end_resume = std::chrono::high_resolution_clock::now();
         auto dt_resume = std::chrono::duration_cast<std::chrono::milliseconds>(t_end_resume - t_start_resume).count();
```
pford: Change to microseconds and adjust in fmt::print

pford (changes requested):

Session.cc
pford: There were quite a few of these so I fixed them myself.

pford (approved):

veggiesaurus:
closing as this PR's changes were included in #609

Issue #593, opened 2020-08-14
moment generator: no progress update or crash


kswang1029:
in some cases of long calculations, we don't have progress report after the first 5s. And if we cancel the request and request a new one, the backend may crash. 

Backport for v1.4 final

markccchiang:
@kswang1029 Is there an example case that can be reproduced? I am trying to fix it.

kswang1029:
try set_S255/S255_IR_sci.spw29.cube.I.pbcor.fits
1. enable moment calculations by pressing the generate button directly (using entire cube)
2. when you see the first progress update, press cancel
3. enable a new moment calculations by pressing the generate button directly
4. backend will crash

kswang1029:
about the issue of no progress update
try set_HDF5/S255_IR_sci.spw29.cube.I.pbcor.hdf5
1. enable moment calculations by pressing the generate button directly (using entire cube)
2. there is no progress update
3. cancel the calculation and make a new one
4. backend may crash (the root cause might be the same as the one above) threading?

kswang1029:
Found a more robust way to reproduce the crash event. It appears that the backend accidentally was working on streaming spectral profile and moment generation at the same time.
1. load a large cube with lots of channels (need profile streaming)
2. lock cursor
3. enable moment calculations and cancel
4. unlock cursor and click the generate button quickly before new cursor profile is streaming.
5. the backend should crash

duidae:
about the progress, can turn on the progress message:

![image](https://user-images.githubusercontent.com/25971696/90731443-4aa22e00-e2fc-11ea-84f3-295cf78eb7b7.png)
![image](https://user-images.githubusercontent.com/25971696/90732197-53473400-e2fd-11ea-9574-1a1400e62706.png)


tried set_HDF5/S255_IR_sci.spw29.cube.I.pbcor.hdf5 with my desktop(ubuntu) yesterday, seems no message was sent yet & results in empty progress, and Mark tried with server, there will be progress response, so probably missing progress is due to pending message.

markccchiang:
The progress bar for `set_HDF5/S255_IR_sci.spw29.cube.I.pbcor.hdf5` is OK if you test it on the server. However, @duidae tested it on the desktop and not work. It is possible that if the calculation time for a chunk is more than 5 secs, then you will not see the progress report in the first 5 secs.

veggiesaurus:
@kswang1029 please confirm this is fixed in the latest `release/1.4` branch

[...]

PR #606, opened 2020-08-21
Fix moment generation problems

markccchiang:
Fix the issue #593, code changes are:
- Add a lock for moment and spectral profile calculations.
- Make sure to stop the moment calculation while the spectral profile begins
- Make sure to stop the old moment calculation while the new one begins
- Add a bool variable `cancel` in the MomentResponse message to explicitly inform the frontend that the moment cancellation is done
- Check the progress every slice instead of every chunk, and the other minor changes


Session.cc
```diff
@@ -1279,6 +1284,8 @@ bool Session::SendSpectralProfileData(int file_id, int region_id, bool stokes_ch
     if ((region_id > CURSOR_REGION_ID) || (region_id == ALL_REGIONS) || (file_id == ALL_FILES)) {
         // Region spectral profile
         CARTA::SpectralProfileData profile_data;
+        _moment_controller->StopCalculation(file_id); // Stop the moment calculations while spectral profile begins
```
pford: As shown in the if statement, file_id could be ALL_FILES (-1) which would not be in the moment controller's map.  It is ALL_FILES when the region changes so that the spectral profiles for all images with that region are recalculated.  In RegionHandler::FillSpectralProfileData, it iterates through the spectral configs (_spectral_req) to start a new profile when the requirement contains the region_id.  This is why I had recommended that the RegionHandler handle the moments for regions, as it could call StopCalculation for each actual file_id in this loop before doing the spectral profile.  The simplest thing to do for ALL_FILES at this point (before the refactor) is to add a RegionHandler function to return the file ids for a given region id by checking the _spectral_req ConfigId keys.  ConfigId is a struct containing file_id and region_id.  Also note that the key may exist with no SpectralConfigs assigned; the key is removed when the region or frame is closed.

Session.cc
```diff
@@ -1279,6 +1284,8 @@ bool Session::SendSpectralProfileData(int file_id, int region_id, bool stokes_ch
...
+        std::unique_lock<std::mutex> ulock(_image_mutexes[file_id]);
```
pford: Again, file_id could be ALL_FILES (for that region) which would not be in the image mutex map.

pford (changes requested):

Session.cc
markccchiang: Thank you for the suggestion! Have modified the code accordingly. Please check is that ok.

confluence (approved):

pford (approved):
Approving this for now, since refactor is planned.  RegionHandler::GetFileIds is vague, as it applies only to the spectral requirements.  Moving the moment generator handling code from Session to Frame will enable stopping moment generation and locking the image as each spectral profile is calculated, instead of a bulk stop/lock/unlock for all images around the calculations.

PR #626, opened 2020-09-04
Fix floating paramter issue

duidae:
query range 348700 ~ 349500 MHz, intensity = -5, should return 2665 lines


veggiesaurus (approved):

kswang1029 (approved):
tested 👍

Issue #586, opened 2020-08-13
Change the architecture of moments generation code


markccchiang:
Things to do:

- [x] Remove the MomentController and move the MomentGenerator under the Frame
- [x] Move the FileConverter under the Frame
- [x] Keep the file name as a private parameter in the Frame and FileLoader
- [x] Keep the function `ApplyRegionToFile()` as a private function in the RegionHandler
- [x] Apply an `_image_mutex` in the Frame while calculating image moments

Also, the `ImageMoments` should be removed after changes in casa and casacore are done.

veggiesaurus:
@markccchiang since the PR has been approved, I assume this will be a separate PR, not necessarily for the code freeze?

markccchiang:
Yes. I will do it in another PR.

PR #634, opened 2020-09-15
Refactor the moment generator

markccchiang:
This follows the refactoring plan in #586 that removes the MomentController and moves the MomentGenerator to the Frame. FileConverter is also removed and the saving/converting file function is moved to the Frame. These functions are protected by the `_image_mutex` from the Frame. The execution of moments calculation is through objects Session -> RegionHandler -> Frame -> MomentGenerator. The image loader (FileLoader) can only be accessed by the Frame, and the function `ApplyRegionToFile()` is restored as a private function for the RegionHandler. The function to get a full 2D image region `Frame::GetImageRegion()` is simplified.


Frame.cc
```diff
@@ -31,11 +39,12 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
       _stokes_index(-1),
       _num_channels(1),
       _num_stokes(1),
-      _z_profile_count(0) {
+      _z_profile_count(0),
+      _moment_generator(std::make_unique<MomentGenerator>(GetFileName(), GetImage())) {
```
pford: Why not set to nullptr and construct MomentGenerator when moments are requested?

Frame.cc
```diff
@@ -1225,39 +1237,15 @@ bool Frame::GetImageRegion(int file_id, const ChannelRange& chan_range, int stok
     if (!CheckChannel(chan_range.from) || !CheckChannel(chan_range.to) || !CheckStokes(stokes)) {
         return false;
     }
-
     // Get the image shape and coordinate system
```
pford: This could be simplified further using a Slicer:
`casacore::Slicer slicer = GetImageSlicer(chan_range, stokes);
casacore::LCSlicer* lcslicer = new LCSlicer(slicer);
casacore::ImageRegion this_region(lcslicer);`

Region/RegionHandler.cc
```diff
@@ -713,6 +713,36 @@ bool RegionHandler::ApplyRegionToFile(
...
+    // Get an image region
+    casacore::ImageRegion image_region;
+    int region_id(moment_request.region_id());
+    int chan_min(moment_request.spectral_range().min());
+    int chan_max(moment_request.spectral_range().max());
+    bool image_region_ok(false);
+
+    frame->IncreaseZProfileCount(); // Use z profile count to control the destruction for the Frame
+
+    if (region_id > 0) {
```
pford: I would prefer to move this logic back to the Session; if region_id > 0 then go through the RegionHandler, else call a Frame method directly to GetImageRegion then CalculateMoments.

Region/RegionHandler.cc
```diff
@@ -713,6 +713,36 @@ bool RegionHandler::ApplyRegionToFile(
...
+        _regions.at(region_id)->IncreaseZProfileCount(); // Use z profile count to control the destruction for the Region
```
pford: The zprofile count should be renamed in Frame and Region if it is applied to other calculations

Session.cc
```diff
@@ -1011,38 +1004,16 @@ void Session::OnMomentRequest(const CARTA::MomentRequest& moment_request, uint32
             SendEvent(CARTA::EventType::MOMENT_PROGRESS, request_id, moment_progress);
         };
 
-        casacore::ImageRegion image_region;
+        // Do calculations
```
pford, line 1007: region_id (above) is not used but should be, to determine whether to go through RegionHandler or directly to Frame

Session.cc
```diff
@@ -1279,11 +1249,6 @@ bool Session::SendSpectralProfileData(int file_id, int region_id, bool stokes_ch
     if ((region_id > CURSOR_REGION_ID) || (region_id == ALL_REGIONS) || (file_id == ALL_FILES)) {
         // Region spectral profile
         CARTA::SpectralProfileData profile_data;
```
pford, line 1264: StopCalculation has been removed; is this now controlled by the frontend (StopMomentCalc message)?

pford (changes requested):

Session.cc
markccchiang: Yes. The spectral profile calculation will not be triggered until the moment calculation is done or canceled by the frontend.

pford (approved):

Frame.cc
```diff
@@ -1417,3 +1359,116 @@ bool Frame::GetLoaderSpectralData(int region_id, int stokes, const casacore::Arr
...
+    // Set response message
+    int file_id(save_file_msg.file_id());
+    save_file_ack.set_file_id(file_id);
+    bool success(false);
+    casacore::String message;
+
+    // Get the full resolved name of the output image
+    std::string temp_path = root_folder + "/" + directory;
+    std::string abs_path = fs::absolute(fs::path(temp_path)).string();
+    output_filename = abs_path + "/" + output_filename;
```
confluence: This is a minor issue, but it would be cleaner to [use the `/` operator on path objects](https://en.cppreference.com/w/cpp/filesystem/path/operator_slash) here instead of concatenating raw strings with a hardcoded separator.

confluence (approved):

veggiesaurus (approved):

veggiesaurus:
@markccchiang can this be safely merged into `dev`?

markccchiang:
Yes, it can be merged safely . Thank you!

Angus Comrie <notifications@github.com>於 2020年11月2日 週一，下午7:54寫道：

Issue #640, opened 2020-10-05
reduce the progress update cadence of cube histogram calculations


kswang1029:
Probably due to the previous refactoring of parallelization from TBB to OMP, the progress update cadence of cube histogram calculations now is very high. Updating every 1s should be sufficient.
![Oct-05-2020 17-33-22](https://user-images.githubusercontent.com/20819712/95063655-3346c500-0731-11eb-997a-0d84a124ee08.gif)

kswang1029:
@markccchiang could you help on this issue?

markccchiang:
OK.

PR #641, opened 2020-10-06
Reduce the progress update cadence of cube histogram calculations

markccchiang:
For issue #640, we now change the update of cube histogram progress by every 1 second.


kswang1029 (approved):
tested with some fits and hdf5 images. It works well.

pford (approved):

PR #645, opened 2020-10-19
enforce 16-byte array alignment in order to utilise SIMD correctly

veggiesaurus:
this seems to fix the issue with contour compression on some compilers


ajm-ska (approved):
It works. Contours now render with compression without crashing on carta_backends compiled on MacOS with Clang.

I tried four combinations on MacOS using the Clang compiler:
-DEnableAvx=OFF -DCMAKE_BUILD_TYPE=Debug
-DEnableAvx=ON -DCMAKE_BUILD_TYPE=Debug
-DEnableAvx=OFF -DCMAKE_BUILD_TYPE=Release
-DEnableAvx=ON -DCMAKE_BUILD_TYPE=Release
All combinations now work whereas previously the first two failed. 

It appears to it continue to work on Linux too without problems.

PR #653, opened 2020-11-03
Set fileInfo response name to symlink not resolved name

### LINKED ISSUE #358: Inconsistent handling of symlinks
File info hangs on file symlink outside of root filesystem, but is listed and opens on double-click.  Make this consistent by allowing file info.

pford:
Closes #358. Must match fileInfoRequest name.


markccchiang (approved):

kswang1029 (approved):
Tested with symlinked images both inside root and outside root. The symlinked images are displayed in the file list and their file info can be generated and they can be loaded. 👍 Comparing to the dev branch, symlinked images (either inside root or outside root) result in hanging file info.

Issue #637, opened 2020-09-23
Improperly reading ds9 region files


dentalfloss1:
Attached is a ds9 region file that I am trying to load in. Before the 1.4 release, it worked fine. After the 1.4 release it loads in with the wrong coordinates. See attached region file and screenshot. 
![image](https://user-images.githubusercontent.com/24629883/94028438-dee33300-fd89-11ea-860e-afffa0c9b29c.png)

[GRB200219A_reg.txt](https://github.com/CARTAvis/carta/files/5268910/GRB200219A_reg.txt)

kswang1029:
To summarize the issue, we need a fix for:
- [ ] region name parsing
- [ ] support region file in dos format

dentalfloss1:
The above image does indeed look correct

veggiesaurus:
@dentalfloss1 we'll fix this and deploy the fix to carta-testing.idia.ac.za once ready. In the meantime, please use the workaround provided

zarda:
I got a similar bug for ds9 region importing. Please try it by import the region on demo server: /set_QA/set_QA_regionTest/set_QA/M17_SWex_testRegions_pix.reg
We will got a wrong PA for the last ellipse region: it should be 45 but it returns 315.

PR #658, opened 2020-11-14
Fix DS9 region import

pford:
Fix issues in #637: region name parsing and support region file in dos format.  The region parameters parser is used by both DS9 and CRTF importers, when the imageanalysis casa::RegionTextList cannot be used for CRTF.


confluence (approved):
Looks OK to me.

markccchiang (approved):

kswang1029 (approved):
This works nicely! 👍

PR #662, opened 2020-11-19
438 extrema zprofile

### LINKED ISSUE #438: add "extrema" to stats drop down options
In the case that I am looking at  z-profile in a Stokes cube values of Q, U and V can be negative.  So if I want to look at a profile of maximum Q, U or V polarised intensity in a region I cannot do that by selecting "max".  We need another option "extrema" which will plot the value of the extrema in a region whether it is positive or negative.

pford:
Closes #438 .  Add support for extrema stats in spectral profile.

Merge with frontend PR #1255 and protobuf pam/extrema_stats

### DIFF
--- a/EventHeader.h
+++ b/EventHeader.h
@@ -8,7 +8,7 @@
 #define CARTA_BACKEND__EVENTHEADER_H_
 
 namespace carta {
-const uint16_t ICD_VERSION = 17;
+const uint16_t ICD_VERSION = 18;
 struct EventHeader {
     uint16_t type;
     uint16_t icd_version;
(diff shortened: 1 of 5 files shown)


veggiesaurus:
@confluence can you look at this PR? There's something weird going on with our access control, so I can't currently assign you as a reviewer :thinking:

confluence (changes requested):
The code for adding this stat to the HDF5 Z profile looks correct. However, for completeness I think that we should also add the extrema stat to the HDF5 full image statistics (which are loaded in `FileLoader::LoadImageStats`).

pford:
> for completeness I think that we should also add the extrema stat to the HDF5 full image statistics (which are loaded in `FileLoader::LoadImageStats`).

If this stats type is not going to be added to the RegionStatsData, I think there is no need to calculate it in the FileLoader.  If it is added, that requires additional development for the backend (to calculate it, HDF5 and non-HDF5, and add to RegionStatsData) and the frontend (to display it in the Stats widget).  But the user can easily determine this from the existing values for Min and Max since they are single values adjacent in the table.

confluence:
Fair enough. In that case I think this is fine -- I'll approve the PR.

confluence (approved):

PR #664, opened 2020-11-23
402 Add beam table to OpenFileAck

### LINKED ISSUE #402: Multi-beam image: file info and beam table
Followup work for #46 per-plane-beam images:

File info should show median beam instead of restoring beam.  When the user loads the image, the full beam table should be sent so that the beam for each image plane may be shown accurately.  I have proposed adding this to the OpenFileAck message in this [doc](https://docs.google.com/document/d/1nFa1OoMKz4a0i-VId6_1qXjkzxq1U7W6oPDen4JxjUk/edit).

pford:
Closes #402.  Adds optional beam information to OpenFileAck using a new repeated Beam proto definition (channel, stokes, major_axis, minor_axis, pa).

Merge with frontend PR #1265 and protobuf pam/beam_table


ImageData/FileLoader.cc
```diff
@@ -234,6 +234,47 @@ bool FileLoader::GetSubImage(const casacore::LattRegionHolder& region, casacore:
...
+                    carta_beam.set_major_axis(gaussian_beam.getMajor("arcsec"));
+                    carta_beam.set_minor_axis(gaussian_beam.getMinor("arcsec"));
+                    carta_beam.set_pa(gaussian_beam.getPA("deg").getValue());
+                    beams.push_back(carta_beam);
+                }
+            }
+        }
+        success = true;
+    } catch (casacore::AipsError& err) {
+        // beams failed somehow
```
confluence: Should we log a warning here?

confluence (approved):

ImageData/FileLoader.cc
pford: The FileLoader returns the error (if any) which is logged by the Frame. since it knows the session id.

markccchiang (approved):

Issue #224, opened 2019-05-22
Migrate to latest stable release of uWebSockets


veggiesaurus:
The latest version of uWebSockets has some performance improvements and removes the use of deprecated OpenSSL functions. We should migrate to this version (0.15.6 as of time of writing)

The API has changed quite a bit, but our use of the API itself is quite limited, so this shouldn't be too much of an issue.

zarda:
The last master branch 0.15.6 is not lib.so any more. The branch 0.15.6 is header only api which rely on uSocket. This means we need lot of modification about building. At least the dependency of libuv or epoll need to be addressed.

veggiesaurus:
this shouldn't be too hard to change. uSocket is a pretty minor dependency

alexhultman:
Hello,

New code is header only (just include it) and cross-platform C (without any dependencies) with optionally OpenSSL and zlib as dependencies. Windows is the only platform which need libuv

veggiesaurus:
> Hello,
> 
> New code is header only (just include it) and cross-platform C (without any dependencies) with optionally OpenSSL and zlib as dependencies. Windows is the only platform which need libuv

@alexhultman thanks for the info!

zarda:
Do we need to include `uWebSockets` in submodules?

veggiesaurus:
> Do we need to include `uWebSockets` in submodules?

that's probably a good idea. However, we first need to check if the backend can work nicely with C++17 compilers. I'm a bit concerned about RedHat support in that regard.

markccchiang:
I have created a test branch [mark/upgrade_uWebSockets](https://github.com/CARTAvis/carta-backend/tree/mark/upgrade_uWebSockets), which applies the latest version of uWebSockets. In the new uWebSockets, the `libuv` is wrapped in the submodule `uSockets`. Since it lacks detailed instruction on how to apply the async mechanism for our needs. The current solution is to slightly modify the `uSockets` (as well as `uWebSockets`) in order to access the `uv loop` directly. The installation script for the modified version of `uWebSockets` is [here](https://github.com/CARTAvis/carta-backend/blob/mark/upgrade_uWebSockets/scripts/install_uWebsockets.sh). Modifying the uWebSockets maybe not a good way, unless there is a significant improvement of the performance (it needs tests). I will keep understanding the new websockets code and try to avoid changing it.

veggiesaurus:
[...]

PR #667, opened 2020-12-03
Migrate to the new version of uWebSockets (#224)

markccchiang:
The installation of the new uWebSockets can refer to the script `scripts/install_uWebSockets.sh`. The flag `-flto` is removed in the `uSockets` building because on `RedHat7` it will cause the linking problem in the carta backend building.


Main.cc
```diff
@@ -46,60 +45,81 @@ static bool verbose;
...
     if (!auth_token.empty()) {
-        auto token_header_entry = http_request.getHeader("carta-auth-token");
-        if (token_header_entry) {
-            string token_header_value = token_header_entry.toString();
+        auto token_header_entry = http_request->getHeader("carta-auth-token");
+        if (!token_header_entry.empty()) {
+            string token_header_value(token_header_entry);
             if (token_header_value != auth_token) {
-                ws->close(1000, "Header auth failed");
+                std::cerr << "Header auth failed!\n";
```
confluence: I think we've standardised on `fmt::print` for these messages.

scripts/install_uWebSockets.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
confluence: What is the purpose of this script? Why not insert this code directly into the Dockerfiles?

scripts/install_uWebSockets.sh
```diff
@@ -0,0 +1,27 @@
...
+
+git clone https://github.com/uNetworking/uWebSockets.git
+
+cd $DIR
+git checkout 72e9951
+git submodule init && git submodule update
+
+## remove the -flto flag while compiling the uSockets
+cd uSockets
+sed 's/-flto//g' Makefile > Makefile.new && mv Makefile.new Makefile
```
confluence: You can use `sed -i` here instead.

scripts/install_uWebSockets.sh
```diff
@@ -0,0 +1,27 @@
...
+cd ..
+
+make
+
+## install the uWebSockets
+sudo make install
+
+## install the uSockets
```
confluence: This should not be necessary. Can we not use the makefile inside the submodule to install these?

confluence (changes requested):

Main.cc
```diff
@@ -440,7 +452,7 @@ void OnMessage(uWS::WebSocket<uWS::SERVER>* ws, char* raw_message, size_t length
             }
         }
     } else if (op_code == uWS::OpCode::TEXT) {
-        if (strncmp(raw_message, "PING", 4) == 0) {
+        if (strncmp(message.data(), "PING", 4) == 0) {
```
veggiesaurus: can't we just use `message == "PING"`?

Main.cc
```diff
@@ -517,6 +532,9 @@ int main(int argc, const char* argv[]) {
             inp.create("root", root_folder, "set top-level folder for data files", "String");
             inp.create("exit_after", "", "number of seconds to stay alive after last sessions exists", "Int");
             inp.create("init_exit_after", "", "number of seconds to stay alive at start if no clents connect", "Int");
+            inp.create("max_payload_length", to_string(max_payload_length), "set uWebSockets max payload length", "Int");
```
veggiesaurus: do these need to be command line arguments? Can't we just use the default ones? Weren't we doing that before?

Main.cc
```diff
@@ -567,19 +588,36 @@ int main(int argc, const char* argv[]) {
...
-
-        if (websocket_hub.listen(port)) {
-            fmt::print("Listening on port {} with root folder {}, base folder {}, {} threads in worker thread pool and {} OMP threads\n",
-                port, root_folder, base_folder, thread_count, omp_thread_count);
-            websocket_hub.run();
-        } else {
-            fmt::print("Error listening on port {}\n", port);
-            return 1;
-        }
+        uWS::SSLApp({/* Set certificates in uWebSockets.js repo */
```
veggiesaurus: we don't need to use `SSLApp`, use `uWS::App` instead

Main.cc
```diff
@@ -567,19 +588,36 @@ int main(int argc, const char* argv[]) {
...
+                        .key_file_name = "",
+                        .cert_file_name = "",
+                        .passphrase = ""})
+            .ws<PerSocketData>("/*",
+                {/* Settings */
+                    .compression = uWS::SHARED_COMPRESSOR,
+                    .maxPayloadLength = max_payload_length,
+                    .idleTimeout = idle_timeout,
```
veggiesaurus: these can probably be left to default values

Main.cc
```diff
@@ -567,19 +588,36 @@ int main(int argc, const char* argv[]) {
...
+                    .maxBackpressure = max_backpressure,
+                    /* Handlers */
+                    .upgrade = OnUpgrade,
+                    .open = OnConnect,
+                    .message = OnMessage,
+                    .drain = [](auto* ws) { fmt::print("uWebSockets on drain: buffer amount {}\n", ws->getBufferedAmount()); },
+                    .ping = [](auto* ws) { fmt::print("uWebSockets on ping\n"); },
```
veggiesaurus: we don't need these ping/pong handlers

Main.cc
```diff
@@ -567,19 +588,36 @@ int main(int argc, const char* argv[]) {
...
+                {/* Settings */
+                    .compression = uWS::SHARED_COMPRESSOR,
+                    .maxPayloadLength = max_payload_length,
+                    .idleTimeout = idle_timeout,
+                    .maxBackpressure = max_backpressure,
+                    /* Handlers */
+                    .upgrade = OnUpgrade,
+                    .open = OnConnect,
+                    .message = OnMessage,
+                    .drain = [](auto* ws) { fmt::print("uWebSockets on drain: buffer amount {}\n", ws->getBufferedAmount()); },
```
veggiesaurus: not sure why we need `drain` handler

veggiesaurus (changes requested):

Main.cc
pford: There are many std::cerr in the code, and many fmt::print.  I was not aware of this standard.

confluence: Hmm. I could have sworn that we made an attempt to switch over at some point in the past -- it's why we added `fmt` as a dependency. It does look very inconsistent right now. I also noticed that we're using `fmt::print` for error messages without specifying `stderr` as the stream, which prints them to `stdout`.

I think we should make a separate issue for this, and maybe discuss it in the next Slack meeting.

pford: I agree, including any other standardised code usage (?).  We should create a document in the backend architecture folder in our team drive.

Main.cc
```diff
@@ -567,19 +579,30 @@ int main(int argc, const char* argv[]) {
...
-            fmt::print("Listening on port {} with root folder {}, base folder {}, {} threads in worker thread pool and {} OMP threads\n",
-                port, root_folder, base_folder, thread_count, omp_thread_count);
-            websocket_hub.run();
-        } else {
-            fmt::print("Error listening on port {}\n", port);
-            return 1;
-        }
+        uWS::App()
+            .ws<PerSocketData>("/*", {.compression = uWS::SHARED_COMPRESSOR,
+                                         .maxPayloadLength = 16 * 1024,      // use default setting
```
veggiesaurus: if we use designated initialisers, then we can leave out those properties that are initialised to default, can't we?

note that designated initialisers are part of the c++20 spec, but are supported on all our target compilers, as far as I can see

veggiesaurus:
@markccchiang which specific release of uWebSockets is this using?

Can you update the readme as well?

Main.cc
markccchiang: It seems to me still got error messages on CentOS (using gcc8.4) if I remove `.maxPayloadLength`, `.idleTimeout` and `.maxBackpressure` settings.

veggiesaurus: does it work on clang? @ajm-asiaa what version of GCC do we build with?

markccchiang: Yes, it works on clang.

ajm-ska: We usually use GCC 8.4.0

Main.cc
```diff
@@ -440,7 +452,7 @@ void OnMessage(uWS::WebSocket<uWS::SERVER>* ws, char* raw_message, size_t length
             }
         }
     } else if (op_code == uWS::OpCode::TEXT) {
-        if (strncmp(raw_message, "PING", 4) == 0) {
+        if (sv_message == "PING") {
             ws->send("PONG");
```
veggiesaurus: with the new version, it seems we need to explicitly specify a text output as well:

`ws->send("PONG", uWS::OpCode::TEXT);`

veggiesaurus (changes requested):

ajm-ska:
@markccchiang I notice that the IP address information is missing from the backend's session connection information:

For example, when running ICD tests it shows:

> Session 1 (Fri Dec 11 10:53:51 2020): Client 1 [???m?] Connected. Num sessions: 1
> Session 1 (Fri Dec 11 10:53:52 2020): Client 1 [???m?] Disconnected. Remaining sessions: 1


In comparison, when running ICD tests on the `dev` branch it shows the IP address:

> Session 1 (Fri Dec 11 10:58:57 2020): Client 1 [::ffff:140.109.176.28] Connected. Num sessions: 1

Would it be easy to fix?

ajm-ska:
I know this branch has been consistently failing in the Animator ICD tests. It is strange because other branches can pass those tests without problems. 
It was apparently failing on the `ANIMATOR_CONTOUR_MATCH` test, but after investigating further, it was actually failing at the end of the test before it which is the `ANIMATOR_CONTOUR` test. Somehow once the test finishes, it causes a segmentation fault in the backend.
For now I have switched the order of the tests around so that the problematic `ANIMATOR_CONTOUR` test runs last. So now the set of Animator ICD tests apparently pass.
Meanwhile we will investigate the cause of the `ANIMATOR_CONTOUR` test crash.

markccchiang:
> @markccchiang I notice that the IP address information is missing from the backend's session connection information:
> 
> For example, when running ICD tests it shows:
> 
> > Session 1 (Fri Dec 11 10:53:51 2020): Client 1 [???m?] Connected. Num sessions: 1
> > Session 1 (Fri Dec 11 10:53:52 2020): Client 1 [???m?] Disconnected. Remaining sessions: 1
> 
> In comparison, when running ICD tests on the `dev` branch it shows the IP address:
> 
> > Session 1 (Fri Dec 11 10:58:57 2020): Client 1 [::ffff:140.109.176.28] Connected. Num sessions: 1
> 
> Would it be easy to fix?

@ajm-asiaa I have pushed a new commit. Does it solve the problem in your test? Also, I will check with the contour animation.

ajm-ska:
> > @markccchiang I notice that the IP address information is missing from the backend's session connection information:
> > For example, when running ICD tests it shows:
> > > Session 1 (Fri Dec 11 10:53:51 2020): Client 1 [???m?] Connected. Num sessions: 1
> > > Session 1 (Fri Dec 11 10:53:52 2020): Client 1 [???m?] Disconnected. Remaining sessions: 1
> > 
> > 
> > In comparison, when running ICD tests on the `dev` branch it shows the IP address:
> > > Session 1 (Fri Dec 11 10:58:57 2020): Client 1 [::ffff:140.109.176.28] Connected. Num sessions: 1
> > 
> > 
> > Would it be easy to fix?
> 
> @ajm-asiaa I have pushed a new commit. Does it solve the problem in your test? Also, I will check with the contour animation.

Thanks @markccchiang. The IP address appears now, although it is in IPv6 format. Would it be easy for it to be in IPv4 format?

`Session 1 (Sat Dec 12 16:28:07 2020): Client 1 [0000:0000:0000:0000:0000:ffff:8c6d:b01c] Connected. Num sessions: 1`

markccchiang:
> Thanks @markccchiang. The IP address appears now, although it is in IPv6 format. Would it be easy for it to be in IPv4 format?
> 
> `Session 1 (Sat Dec 12 16:28:07 2020): Client 1 [0000:0000:0000:0000:0000:ffff:8c6d:b01c] Connected. Num sessions: 1`

I am not sure is it suitable to convert IPv6 format to IPv4. If I shorten the IP address, say convert `0000:0000:0000:0000:0000:ffff:8c6d:b01c` to `:::::ffff:8c6d:b01c`. Is that ok?

ajm-ska:
> > Thanks @markccchiang. The IP address appears now, although it is in IPv6 format. Would it be easy for it to be in IPv4 format?
> > `Session 1 (Sat Dec 12 16:28:07 2020): Client 1 [0000:0000:0000:0000:0000:ffff:8c6d:b01c] Connected. Num sessions: 1`
> 
> I am not sure is it suitable to convert IPv6 format to IPv4. If I shorten the IP address, say convert `0000:0000:0000:0000:0000:ffff:8c6d:b01c` to `:::::ffff:8c6d:b01c`. Is that ok?

OK. Thanks. Too bad IPv4 is not possible, but at least the shortened IPv6 looks a bit neater.

ajm-ska:
@markccchiang Here is a very minor thing. During the build process on MacOS, a bunch of warnings are issued:

> [ 12%] Linking C static library libuSockets.a
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(openssl.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(wolfssl.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(gcd.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(libuv.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(openssl.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(wolfssl.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(gcd.c.o) has no symbols
> /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libuSockets.a(libuv.c.o) has no symbols
> [ 12%] Built target uSockets

They don't appear to occur on Linux. Maybe we should use this opportunity to eliminate them?

I found one solution [here](https://stackoverflow.com/questions/4929255/building-static-libraries-on-mac-using-cmake-and-gcc#answer-4953904). So I propose we could add the following lines between line 51-52 of our [CMakeLists.txt](https://github.com/CARTAvis/carta-backend/blob/dev/CMakeLists.txt)

```
SET(CMAKE_C_ARCHIVE_CREATE   "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
SET(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
SET(CMAKE_C_ARCHIVE_FINISH   "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
SET(CMAKE_CXX_ARCHIVE_FINISH "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
```

When added, the warnings do not appear.

markccchiang:
> I found one solution [here](https://stackoverflow.com/questions/4929255/building-static-libraries-on-mac-using-cmake-and-gcc#answer-4953904). So I propose we could add the following lines between line 51-52 of our [CMakeLists.txt](https://github.com/CARTAvis/carta-backend/blob/dev/CMakeLists.txt)
> 
> ```
> SET(CMAKE_C_ARCHIVE_CREATE   "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
> SET(CMAKE_CXX_ARCHIVE_CREATE "<CMAKE_AR> Scr <TARGET> <LINK_FLAGS> <OBJECTS>")
> SET(CMAKE_C_ARCHIVE_FINISH   "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
> SET(CMAKE_CXX_ARCHIVE_FINISH "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
> ```
> 
> When added, the warnings do not appear.

@ajm-asiaa Thank you! I have added these lines in the CMakeLists.txt. As for the IP address, I can manually remove the first `::::` if the first four octets are empty. So it will represent in IPv4 format.

veggiesaurus:
> I can manually remove the first `::::` if the first four octets are empty. So it will represent in IPv4 format.

That won't result in the correct IPv4 address. You need to convert the remaining 16-bit numbers to groups of 8-bit numbers and print them as decimals

veggiesaurus (commented):
@markccchiang why not just use the IP address in data array format instead of converting from text?

Also, you can use `fmt::print` to print hexadecimal directly, no need for sscanf and sprintf: https://fmt.dev/latest/syntax.html

ajm-ska:
Thanks @markccchiang! It now works exactly as it did with the old uWebSockets on the `dev` branch etc.

> Session 1 (Tue Dec 15 14:40:53 2020): Client 1 [::ffff:140.109.176.28] Connected. Num sessions: 1
> Session 1 (Tue Dec 15 14:40:53 2020): Client 1 [::ffff:140.109.176.28] Disconnected. Remaining sessions: 1

veggiesaurus:
> @markccchiang why not just use the IP address in data array format instead of converting from text?



> Thanks @markccchiang! It now works exactly as it did with the old uWebSockets on the `dev` branch etc.
> 
> > Session 1 (Tue Dec 15 14:40:53 2020): Client 1 [::ffff:140.109.176.28] Connected. Num sessions: 1
> > Session 1 (Tue Dec 15 14:40:53 2020): Client 1 [::ffff:140.109.176.28] Disconnected. Remaining sessions: 1

wait, did it previously also show the `::ffff:` prefix? if it's an IPv4 address, then we can just leave that out

ajm-ska:
> > @markccchiang why not just use the IP address in data array format instead of converting from text?
> 
> > Thanks @markccchiang! It now works exactly as it did with the old uWebSockets on the `dev` branch etc.
> > > Session 1 (Tue Dec 15 14:40:53 2020): Client 1 [::ffff:140.109.176.28] Connected. Num sessions: 1
> > > Session 1 (Tue Dec 15 14:40:53 2020): Client 1 [::ffff:140.109.176.28] Disconnected. Remaining sessions: 1
> 
> wait, did it previously also show the `::ffff:` prefix? if it's an IPv4 address, then we can just leave that out

Yes. The previous version always showed the `::ffff:` prefix for some reason!

veggiesaurus:
Weird!

Util.cc
```diff
@@ -229,3 +229,22 @@ void ConvertCoordinateToAxes(const std::string& coordinate, int& axis_index, int
         stokes_index = -1;
     }
 }
+
+std::string IPAsText(std::string_view binary) {
+    static thread_local char buf[64];
```
veggiesaurus: what is this for?

veggiesaurus (changes requested):
@markccchiang one or two minor changes still required.

Also, can you update the readme to indicate that you'll need to do a recursive clone in order to checkout uWebSockets and uSockets?

veggiesaurus (approved):

PR #668, opened 2020-12-03
Script for checking and adding copyright and licence headers, and first pass to add them

### LINKED ISSUE #40: Add copyright heading to source code
Once the text is agreed upon

confluence:
This fixes #40. The script can check for files with missing or incorrect headers (`check` command), and can replace or add them (`fix` command). The idea would be to add this to the same build test as the format check. When we want to update the header (e.g. to add a year) we should be able to change the correct header in the script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
       - run:
           name: "checking the copyright and licence headers in the carta_backend code"
           command: |
-            python3 scripts/headers.py check
+            ./scripts/headers.py check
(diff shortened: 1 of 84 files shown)


ajm-ska (approved):
It's a good idea and seems to work. 
We just need to inform backend developers to keep an eye on the Circle CI format check and run `./scripts/headers.py fix` if there are any problems.

pford (approved):

veggiesaurus (approved):

PR #672, opened 2020-12-09
357 Move FITS hdus to file info response

### LINKED ISSUE #357: Remove FITS hdus from file list and add to file info
The file list takes too long with many FITS files because they are opened to find the image HDUs by name.  Instead, the file list should return file type (FITS) and the file size only.  The file info for each hdu should be sent in the file info message as a repeated field.

pford:
Closes #357 .

Removes FITS HDU list from file list and moves them into the FileInfoResponse map<hdu, FileInfoExtended>.  This will speed up the initial file list.

Merge with frontend PR #1274 and protobuf pam/fits_hdu_info.


confluence (approved):

confluence (changes requested):
I think that you need to update the protobuf directory.

pford:
Merged dev into protobuf and backend branches.

confluence (approved):

markccchiang (approved):

veggiesaurus (changes requested):
@pford I was about to merge this in, but then discovered a critical issue:

If the user double-clicks on a file in the file list, the _default_ HDU is opened, which sends an `OpenFile` message with the HDU value left as default (i.e. an empty string). In this case, the returned `OpenFileAck` message has an empty list of HDUs and empty values for the `computedEntries` and `headerEntries`

veggiesaurus:
@markccchiang @confluence I've made some changes to fix the issues when receiving an `OpenFile` message with no HDU. Please re-review. @kswang1029 can you confirm this fixes the double-click and multi-load issue in https://github.com/CARTAvis/carta-frontend/pull/1286?

kswang1029:
@veggiesaurus the missing header issue is gone! Now multi-load with some invalid images works properly. 👍

markccchiang (approved):

PR #673, opened 2020-12-11
Fix extrema stat for whole-image regions of HDF5 files

confluence:
The stats widget uses the whole-image stats for HDF5 files.

### DIFF
--- a/ImageData/FileLoader.cc
+++ b/ImageData/FileLoader.cc
@@ -505,7 +505,7 @@ void FileLoader::LoadImageStats(bool load_percentiles) {
     // Remove this check when we drop support for the old schema.
     // We assume that checking for only one of these datasets is sufficient.
     bool full(HasData(FileInfo::Data::STATS_2D_SUM));
-    double sum, sum_sq;
+    double sum, sum_sq, min, max;
     uint64_t num_pixels;
 
     if (HasData(FileInfo::Data::STATS)) {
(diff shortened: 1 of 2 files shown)


kswang1029:
@confluence @veggiesaurus I tested with the test hdf5 images on the mix server and it works. Sorry not spotting this in the previous PR🙈

confluence:
This should now also fix #674.

kswang1029 (approved):
tested with sample HDF5 images on mix n match and it works. 

Just curious, for a single-pixel region, we see an empty plot with stat=stdDev with an HDF5 cube. With a fits cube, we see a "line" at y=0. Does this imply in the HDF5 case the backend does not supply an array with all elements 0? Note this difference does not have usability impact I think.

confluence:
The formula we use for standard deviation involves division by `num_pixels - 1`, so if there's only one pixel we get NAN. I'm not sure if the difference in behaviour between FITS and HDF5 is a difference in the formula that casacore uses (I thought that we had matched these exactly) or an artifact of zero being used as a default value somewhere instead of NAN. I see that in the stats widget both files show the value as zero, so I suspect the latter.

confluence:
After a discussion with @veggiesaurus I have removed the entire lazy min/max optimisation, and simplified the code.

veggiesaurus (approved):
@kswang1029 please re-test after the latest commits

kswang1029:
@veggiesaurus ready for merge. 👍

PR #679, opened 2020-12-22
678 Retain spectral convention when saving a FITS image

### LINKED ISSUE #678: retain spectral convention when saving as a FITS image
Thanks @zarda to spot this issue.

To reproduce:
1) load the FITS image M17_SWex.fits
2) use the "save image" menu to save the image as another FITS image
3) load the image and check the header. The spectral axis should be in FREQ but the saved FITS is in VRAD.  

Saving as the CASA format does not have this issue.

markccchiang:
Closes #678. Retain spectral convention when saving as a FITS image.


zarda (approved):

kswang1029 (approved):
Tested with images with the spectral convention as FREQ, VRAD, and WAVE. All seem to be exported properly with correct spectral convention. 👍

Frame.cc
```diff
@@ -1457,7 +1457,7 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
         success = true;
     } else if (output_file_type == CARTA::FileType::FITS) {
         // Remove the old image file if it has a same file name
-        casacore::Bool ok = casacore::ImageFITSConverter::ImageToFITS(message, *image, output_filename.string(), 64, casacore::True,
+        casacore::Bool ok = casacore::ImageFITSConverter::ImageToFITS(message, *image, output_filename.string(), 64, casacore::False,
             casacore::True, -32, 1.0, -1.0, casacore::True, casacore::False, casacore::True, casacore::False, casacore::False,
```
pford: This now sets preferVelocity=False, opticalVelocity=True.  The comment for this parameter says "If writing a velocity, use the optical definition (otherwise use radio)".  To get the correct FITS headers in FileExtInfoLoader.cc, I checked this with:
    casacore::MDoppler::Types vel_doppler(coord_sys.spectralCoordinate().velocityDoppler());
    if ((vel_doppler == casacore::MDoppler::Z) || (vel_doppler == casacore::MDoppler::OPTICAL)) {
        optical_velocity = true;
    }
But @kswang1029 you said the VRAD headers were correct?

kswang1029: uh... I will need to double check again the case of optical velocity definition!

pford: Tracked the use of the opticalVelocity parameter down to casacore::FITSSpectralUtil::toFITSHeader:
```
	    // OPTICAL
	    refVelocity = -C::c * (1.0 - restfreq / refFrequency);
	    velocityIncrement = -C::c * (1.0 - restfreq / (refFrequency + freqIncrement)) - refVelocity;

	    // RADIO
	    refVelocity = -C::c * (refFrequency/restfreq - 1.0);
	    velocityIncrement = -C::c * ((refFrequency + freqIncrement)/restfreq - 1.0) - refVelocity;
```

kswang1029: The header parser in CARTA is doing something unexpectedly when loading an image with VOPT axis (test image at carta_save_image/M17_SWex_VOPT.fits which is generated with CASA-exportfits from the image with FREQ). In CARTA we see the axis is somehow converted to VRAD.... Below are screenshots from ds9 and CARTA with the same image "M17_SWex_VOPT.fits"
Screen Shot 2020-12-24 at 10 01 08 AM
.
Screen Shot 2020-12-24 at 10 01 37 AM

We will need to fix this first to see the "correct" header first.

From the implementation point of view, if the input cube is VOPT, the subimage export needs to be in VOPT too. In short, keep the spectral convention as it is. We will need to check the input header first and apply proper parameters when exporting.

pford (changes requested):
This should check SpectralCoordinate as in FileExtInfoLoader for setting the "prefer" parameters.  This common code should be moved to Util.cc.

markccchiang:
@pford Is the newly modified code proper in your opinion?

pford (approved):

pford:
@kswang1029  would you like to test this PR before it is merged?  It is changing the same code as PR #739 , so it would be good to get it merged into dev.

pford:
@ajm-asiaa  build/test errors were due to "no space left on device"

ajm-ska:
> @ajm-asiaa build/test errors were due to "no space left on device"

I'm sorry about it. It is an embarrassing ongoing problem.

The issue is that every build directory ends up quite large, particularly now with -Dtest=on. The total size is ~1.6GB.
The biggest files/folders are:
161M	carta_backend
302M	carta-backend-ICD-test
394M	CMakeFiles
567M	test
In `test`, the `CMakeFiles` folder is 410M. In `carta-backend-ICD-test` the `node_modules` folder is 294M.

The local /home partition on that server is only 10GB in size.

I thought I improved it by putting:

> rm -rf CMakeFiles
> rm -rf test/CMakeFiles

at the end of the build procedure (It would remove ~700MB), plus reduce the number of builds kept by Jenkins, but I must have made a mistake somewhere.

Anyway, now I have just tried a different approach. I used the command `usermod -m -d /newhome/username username` to permanently move the Jenkins user home directory into the root partition of that server. Now it has 130+GB free, so any potential space issues should now finally be eliminated.

veggiesaurus:
@kswang1029 please retest

kswang1029 (approved):
re-tested and it works properly. 👍 Ready for merging.

PR #686, opened 2021-01-11
Fix clang-format settings and update directories to exclude in all scripts

### LINKED ISSUE #677: Fix configuration of newer clang-format
The default clang-format version in Ubuntu 20.04 introduces default settings which break our expected formatting. We should try to fix this by adding more specific configuration (instead of using an older clang-format).

confluence:
This fixes #677.

There were three separate issues: since `clang-format` 9, included headers were being reordered by default. I found @pford 's Slack comment about adding `IncludeBlocks: Preserve` to the config -- this fixes the reordering and is backwards-compatible, so there's no reason not to commit this change.

The other formatting issue was the handling of a particular set of `default` switch statement cases in one file. I believe that `clang-format` was handling this incorrectly before version 8 -- we have `AllowShortCaseLabelsOnASingleLine: false` in the config, but these particular labels were being contracted to a single line. This behaviour was triggered by the braces around the `default` blocks somehow. These braces are unnecessary, so I have removed them -- but the issue could recur in the future as long as anyone uses pre-8 `clang-format`, so I would suggest that we switch to 8 as the minimum version. It can be installed in Ubuntu 18.04.

A last, unrelated issue: because all our code is in the root directory of the repo, we have to explicitly exclude every submodule as well as the build directory from every script that we want to apply to our code files. We did not do this after the recent addition of the uWebSockets submodule, so files in this subdirectory were being affected. I have fixed these paths in all the scripts, but in the long term I think we should really consider moving our code to a subdirectory. I will create a separate issue for this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -21,3 +21,4 @@ AllowShortCaseLabelsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
 AlignAfterOpenBracket: DontAlign
 AccessModifierOffset: -4
+IncludeBlocks: Preserve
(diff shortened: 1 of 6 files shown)


pford (approved):
Thank you for fixing this.  Excellent idea to move source code.

PR #688, opened 2021-01-12
Move source files to src, update scripts, reformat and add copyright statements to forgotten tcc files

### LINKED ISSUE #687: Move all our source files into a src subdirectory
Keeping the source files in the root directory of the repository causes multiple ongoing problems. All of our style and formatting scripts have to exclude the build directory and every submodule separately -- every time we add a new submodule containing `.cc` or `.h` files we have to update all the scripts to exclude it (see explanation in #686). This layout also makes it difficult to grep through our codebase and use other commandline tools to interact with it. A lot of this would be simplified if we had a single unambiguous subdirectory where all the source files were located.

confluence:
Fixes #687. Also applies formatting and header scripts to `.tcc` files, which we previously forgot about.

I didn't do anything fancy to find the source files; I just prepended the subdirectory to all entries in the list. Using a glob to find these is not recommended, because if the file doesn't change when a new source file is added, cmake doesn't know that a rebuild is required.


ajm-ska (approved):
Nice work. This helps tidy things up. It still seems to build fine everywhere.

If you want to do extra tidying up, feel free to delete the hidden `.travis.yml` file because we are not currently using Travis CI. 

What about the `test` directory? Is it still being used for something?

Maybe I'm going too far, but do you think the `uWebSockets`, `carta-scripting-grpc`, and `carta-protobuf` directories could be removed too? Those directories are empty at checkout and if not present, they are created by the `git submodule update --init --recursive` command anyway. Or maybe it is proper submodule programming practice to show those directories in the repository?

confluence:
The `test` directory contains some tests. We don't make a lot of use of them, but in future we should increase coverage rather than remove them.

The submodules aren't empty directories; they're links to submodules. You can't track an empty directory in git.

I'll remove the Travis file -- if it's not been used, it's unmaintained, and if we ever use Travis we can get the old version from the history.

veggiesaurus (approved):

PR #691, opened 2021-01-14
Angus/UUID token

### LINKED ISSUE #670: allow backend to host static frontend files on same port
We need to utilise the new HTTP server features of uWebSockets to serve static frontend files (if found) on the same port. This will allow users to _just_ run `carta-backend` and get a simple single-port solution for remote visualisation. This should remove the need for the "remote" version or the remote flag in the AppImage.

We should display the frontend URL (e.g. `http://localhost:3002`) once the server has been initialised. This will result in a user experience similar to running jupyter-notebook from a terminal, which astronomers are quite comfortable with. For reference:
```
angus@angus-ryzen:~$ jupyter-notebook 
[I 07:58:48.605 NotebookApp] Serving notebooks from local directory: /home/angus
[I 07:58:48.605 NotebookApp] The Jupyter Notebook is running at:
[I 07:58:48.605 NotebookApp] http://localhost:8888/?token=733d04f23175f9b2f3c7bf014bcc583d29e1d02b37f38c06
[I 07:58:48.605 NotebookApp]  or http://127.0.0.1:8888/?token=733d04f23175f9b2f3c7bf014bcc583d29e1d02b37f38c06
[I 07:58:48.605 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[C 07:58:48.627 NotebookApp] 
    
    To access the notebook, open this file in a browser:
        file:///home/angus/.local/share/jupyter/runtime/nbserver-16411-open.html
    Or copy and paste one of these URLs:
        http://localhost:8888/?token=733d04f23175f9b2f3c7bf014bcc583d29e1d02b37f38c06
     or http://127.0.0.1:8888/?token=733d04f23175f9b2f3c7bf014bcc583d29e1d02b37f38c06
```

My guess is that we want to check two paths for a `carta-frontend` folder to serve by default: `/usr/local/share/carta` (if the backend is built and installed from source) and `/usr/share/carta` (if the backend is installed via `apt`)

I think we should consider running the HTTP server by _default_ and introduce a `standalone` flag in order to run it without the HTTP server. This should allow us to move closer to the desired approach of being able to simply `apt get install carta` and then run CARTA from the terminal with minimal issues

### LINKED ISSUE #676: Generate UUID token on the backend to prevent unauthorized access to manually started backend processes
This issue deals with an important change required in order to make the execution of CARTA (outside of the carta-node-server case) more secure and more familiar to astronomers. The approach is based on that of [jupyter-notebook](https://jupyter-notebook.readthedocs.io/en/stable/security.html).

The backend currently accepts a pre-generated token via the `CARTA_AUTH_TOKEN` environmental variable. If this is missing, the backend will accept any and all incoming websocket connections. This presents a security risk when running in what is currently known as "remote mode", where a backend is started on a server and accessed by the user in a browser.

Instead we should generate a UUID token (random-based, rather than time or hardware address based) if no token is passed in through the `CARTA_AUTH_TOKEN` environmental variable. 

This feature, along with #670 and #671 will allow us to print out a custom URL for the user to connect to as soon as the backend is started. For reference, this is what jupyter-notebook does:

```
angus@angus-ryzen:~$ jupyter-notebook 
[I 14:32:54.017 NotebookApp] Serving notebooks from local directory: /home/angus
[I 14:32:54.017 NotebookApp] The Jupyter Notebook is running at:
[I 14:32:54.017 NotebookApp] http://localhost:8888/?token=2be16b2a306f1d9288b6c1fd9153357d5a0dc229117bb19d
[I 14:32:54.017 NotebookApp]  or http://127.0.0.1:8888/?token=2be16b2a306f1d9288b6c1fd9153357d5a0dc229117bb19d
[I 14:32:54.017 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
```

This presensts the user with a simple and (more) secure way of running CARTA without relying on "remote mode" tacked on to the electron app.

The backend should also have a `--no-token` command-line argument that can be used to prevent the token from being generated, for debug/development purposes.

veggiesaurus:
closes #670 and closes #676.

Companion PR of https://github.com/CARTAvis/carta-frontend/pull/1294

This PR adds a basic HTTP server (using uWebSockets) on the same port as the websocket connection. This allows us to simplify deployment quite a bit, as we can now serve the frontend through the backend, and only need a single port for both the HTTP. This also means that we don't need a "remote" mode with the bundled app image, but can rather encourage users who need remote access to simply install the carta components. Users who run CARTA locally may also prefer this approach to the bundled version, as it closely mirrors the approach taken by jupyter-notebook.

By default, the backend attempts to host frontend files found in the path
`<location of carta_backend executable>../share/carta/frontend`.

This covers three main use cases:
- CARTA installed via a package manager under `/usr/`
- CARTA manaually installed (or via a script) under `/usr/local`
- CARTA running from a folder extracted from a tarbal (e.g. `~/Downloads/carta2.0/`)

There are a number of new command-line arguments introduced:
- `no_http` for disabling the HTTP server altogether
- `no_browser` for preventing automatically opening the frontend URL in the user's browser
- `debug_no_auth` for disabling any checking of auth tokens. This should only be done for debug purposes, as it will accept all incoming WebSocket connections
- `host` to allow incoming connections on a specific inferface only
- `frontend_folder` for specifying a custom frontend folder location

In addition to the HTTP server, this PR also adds simple UUID token generation for auth of incoming WebSocket connections. When running the backend, a random UUID token will be generated. This token will be passed to the frontend via the `token` URL query parameter. The full URL is displayed in the console, and opened in the user's default browser using `xdg-open` or `open` (for linux and MacOS respectively).

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -533,7 +533,7 @@ int main(int argc, const char* argv[]) {
         int omp_thread_count = OMP_THREAD_COUNT;
         string http_root_folder;
         bool no_http = false;
-        bool no_token = false;
+        bool debug_no_auth = false;
         bool no_browser = false;
 
         { // get values then let Input go out of scope
@@ -542,7 +542,7 @@ int main(int argc, const char* argv[]) {
             inp.create("verbose", "False", "display verbose logging", "Bool");
             inp.create("perflog", "False", "display performance logging", "Bool");
             inp.create("no_http", "False", "disable CARTA frontend HTTP server", "Bool");
-            inp.create("no_token", "False", "disable token validation on connection", "Bool");
+            inp.create("debug_no_auth", "False", "accept all incoming WebSocket connections (insecure, reserved for development and debugging", "Bool");
             inp.create("no_browser", "False", "prevent the frontend from automatically opening in the default browser on startup", "Bool");
             inp.create("port", to_string(port), "set server port", "Int");
             inp.create("grpc_port", to_string(grpc_port), "set grpc server port", "Int");
@@ -559,7 +559,7 @@ int main(int argc, const char* argv[]) {
             verbose = inp.getBool("verbose");
             perflog = inp.getBool("perflog");
             no_http = inp.getBool("no_http");
-            no_token = inp.getBool("no_token");
+            debug_no_auth = inp.getBool("debug_no_auth");
             no_browser = inp.getBool("no_browser");
             port = inp.getInt("port");
             grpc_port = inp.getInt("grpc_port");
@@ -582,7 +582,7 @@ int main(int argc, const char* argv[]) {
             return 1;
         }
 
-        if (!no_token) {
+        if (!debug_no_auth) {
             auto env_entry = getenv("CARTA_AUTH_TOKEN");
             if (env_entry) {
                 auth_token = env_entry;
@@ -644,7 +644,7 @@ int main(int argc, const char* argv[]) {
         }
 
         string ws_pattern;
-        if (no_token) {
+        if (debug_no_auth) {
             ws_pattern = "/*";
         } else {
             ws_pattern = "/token/:token";
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -637,7 +637,17 @@ int main(int argc, const char* argv[]) {
                     }
                 });
 
-                string frontend_url = fmt::format("http://{}:{}", host.empty() ? "localhost" : host, port);
+                string default_host_string = host;
+                if (host.empty()) {
+                    auto server_ip_entry = getenv("SERVER_IP");
+                    if (server_ip_entry) {
+                        default_host_string = server_ip_entry;
+                    } else {
+                        default_host_string = "localhost";
+                    }
+                }
+
+                string frontend_url = fmt::format("http://{}:{}", default_host_string, port);
                 if (!auth_token.empty()) {
                     frontend_url += fmt::format("?token={}", auth_token);
                 }
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -630,7 +630,7 @@ int main(int argc, const char* argv[]) {
                 }
                 if (!no_browser) {
 #if defined(__APPLE__)
-                    string open_command = "open"
+                    string open_command = "open";
 #else
                     string open_command = "xdg-open";
 #endif
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -690,14 +690,14 @@ int main(int argc, const char* argv[]) {
                                          .close = OnDisconnect})
             .listen(port, LIBUS_LISTEN_EXCLUSIVE_PORT,
                 [=](auto* token) {
-                    if (token) {
-                        fmt::print(
-                            "Listening on port {} with root folder {}, base folder {}, {} threads in worker thread pool and {} OMP "
-                            "threads\n",
-                            port, root_folder, base_folder, thread_count, omp_thread_count);
-                    } else {
-                        fmt::print("Error listening on port {}\n", port);
-                    }
+            if (token) {
+                fmt::print(
+                    "Listening on port {} with root folder {}, base folder {}, {} threads in worker thread pool and {} OMP "
+                    "threads\n",
+                    port, root_folder, base_folder, thread_count, omp_thread_count);
+            } else {
+                fmt::print("Error listening on port {}\n", port);
+            }
                 })
             .run();
     } catch (exception& e) {
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -74,14 +74,22 @@ void OnUpgrade(uWS::HttpResponse<false>* http_response, uWS::HttpRequest* http_r
 
     // Check if there's a token to be matched
     if (!auth_token.empty()) {
-        // First try the URL parameter
-        auto req_token = http_request->getParameter(0);
+        string req_token;
+        // First try the cookie auth token
+        auto cookie_header = http_request->getHeader("cookie");
+        if (!cookie_header.empty()) {
+            auto cookie_auth_value = GetAuthTokenFromCookie(string(cookie_header));
+            if (!cookie_auth_value.empty()) {
+                req_token = cookie_auth_value;
+            }
+        }
+        // Then try the auth header
         if (req_token.empty()) {
             req_token = http_request->getHeader("carta-auth-token");
         }
+
         if (!req_token.empty()) {
-            string token_header_value(req_token);
-            if (token_header_value != auth_token) {
+            if (req_token != auth_token) {
                 fmt::print("Incorrect auth token supplied! Closing WebSocket connection\n");
                 http_response->close();
                 return;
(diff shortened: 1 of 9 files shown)


ajm-ska:
@veggiesaurus I see that the `ld: library not found for -luuid` build issue on MacOS can be simply fixed by removing `uuid` from the `set LINKS_LIBS` in the **CMakeLists.txt**. 
But unfortunately we can not do the same for Linux.
However, I found a fairly simple solution. We just need to add the following line between lines 173 and 174 of **CMakeLists.txt**:

> list(REMOVE_ITEM LINK_LIBS uuid)

That is conveniently located inside an existing "Darwin" loop just before LINK_LIBS is used. So this will allow successful builds on both MacOS and Linux.

ajm-ska:
@veggiesaurus I see this will very conveniently replace the existing --remote mode. However, the usual use case of --remote mode is running the carta_backend on a remote server and connecting to it from a different machine on the same network. Therefore, would it be possible to find and display the local IP address or hostname rather than `localhost`?

ajm-ska:
@veggiesaurus For a packaged distributable CARTA, it would be convenient if the default http_root_folder location could be relative to the carta_backend executable. Assuming we usually put the carta_backend executable in a `bin` directory then maybe something similar to `current_path()/../etc/frontend` or `current_path()/../frontend` ?

Maybe the above could be the first place to look and `/usr/local/share/carta/frontend` could be the second? 
Or have a flag at build time for setting `_http_root_folder`?
Or if you don't agree then I suppose whoever is building the package can easily modify the path in `SimpleFrontendServer.cc`.

Although with that said, if we retire the current standalone remote CARTA version, we would be using the --remote flag through the Electron Desktop version, so then we would be going through a wrapper script that would be defining `http_root` anyway.

veggiesaurus:
> @veggiesaurus I see this will very conveniently replace the existing --remote mode. However, the usual use case of --remote mode is running the carta_backend on a remote server and connecting to it from a different machine on the same network. Therefore, would it be possible to find and display the local IP address or hostname rather than `localhost`?

Good point. I think jupyter-notebook does this actually.

veggiesaurus:
> Or have a flag at build time for setting `_http_root_folder`?

I like this approach.

> Although with that said, if we retire the current standalone remote CARTA version, we would be using the --remote flag through the Electron Desktop version, so then we would be going through a wrapper script that would be defining `http_root` anyway.

The aim is to move away from using the electron desktop version for anything other than its primary purpose (a standalone desktop application). Other use cases are better server with the simpler approach taken by jupyter-notebook, and mean we won't need any wrapper scripts

src/Main.cc
```diff
@@ -522,29 +531,46 @@ int main(int argc, const char* argv[]) {
...
+            inp.create("debug_no_auth", "False", "accept all incoming WebSocket connections (insecure, use with caution!)", "Bool");
+            inp.create("no_browser", "False", "prevent the frontend from automatically opening in the default browser on startup", "Bool");
+            inp.create("host", "", "only listen on the specified interface");
             inp.create("port", to_string(port), "set server port", "Int");
             inp.create("grpc_port", to_string(grpc_port), "set grpc server port", "Int");
             inp.create("threads", to_string(thread_count), "set thread pool count", "Int");
             inp.create("omp_threads", to_string(omp_thread_count), "set OMP thread pool count", "Int");
             inp.create("base", base_folder, "set folder for data files", "String");
             inp.create("root", root_folder, "set top-level folder for data files", "String");
+            inp.create("frontend_folder", frontend_folder, "set folder to serve frontend file from", "String");
```
confluence: For consistency, should this not be called `frontend`? But I don't feel that strongly about it.

Typo: file -> files

src/Main.cc
```diff
@@ -584,14 +617,70 @@ int main(int argc, const char* argv[]) {
...
+                    string open_command = "xdg-open";
+#endif
+                    auto open_result = system(fmt::format("{} {}", open_command, frontend_url).c_str());
+                    if (open_result) {
+                        fmt::print("Failed to open the default browser automatically.\n");
+                    }
+                }
+                fmt::print("CARTA is accessible at {}\n", frontend_url);
+            } else {
+                fmt::print("Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument\n");
```
confluence, line 675: This error message should explain more clearly that the frontend files could not be found, and print the location where we tried to look.

src/Main.cc
```diff
@@ -522,29 +531,46 @@ int main(int argc, const char* argv[]) {
...
         { // get values then let Input go out of scope
             casacore::Input inp;
             inp.version(VERSION_ID);
             inp.create("verbose", "False", "display verbose logging", "Bool");
             inp.create("perflog", "False", "display performance logging", "Bool");
+            inp.create("no_http", "False", "disable CARTA frontend HTTP server", "Bool");
+            inp.create("debug_no_auth", "False", "accept all incoming WebSocket connections (insecure, use with caution!)", "Bool");
+            inp.create("no_browser", "False", "prevent the frontend from automatically opening in the default browser on startup", "Bool");
+            inp.create("host", "", "only listen on the specified interface");
             inp.create("port", to_string(port), "set server port", "Int");
```
confluence: Should the description be updated to clarify that the same port will be used by the backend and frontend?

Do we want to make these separately configurable? Is there a reason to do that?

src/Main.cc
```diff
@@ -522,29 +531,46 @@ int main(int argc, const char* argv[]) {
...
+
         { // get values then let Input go out of scope
             casacore::Input inp;
             inp.version(VERSION_ID);
             inp.create("verbose", "False", "display verbose logging", "Bool");
             inp.create("perflog", "False", "display performance logging", "Bool");
+            inp.create("no_http", "False", "disable CARTA frontend HTTP server", "Bool");
+            inp.create("debug_no_auth", "False", "accept all incoming WebSocket connections (insecure, use with caution!)", "Bool");
+            inp.create("no_browser", "False", "prevent the frontend from automatically opening in the default browser on startup", "Bool");
+            inp.create("host", "", "only listen on the specified interface");
```
confluence: If this restriction only applies to serving the frontend, should the name and/or description reflect that?

src/Main.cc
veggiesaurus: I've fixed the typo, but I feel like `base` and `root` somewhat imply a directory, whereas `frontend` doesn't. If someone _does_ feel strongly about it, I'm happy to change :smile:

src/Main.cc
veggiesaurus: `SimpleFrontendServer` Line 127 does exactly that. Example output:

```
Could not find CARTA frontend files in directory /home/angus/repos/carta-backend/cmake-build-relwithdebinfo/../share/carta/frontend.
Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument
```

src/Main.cc
veggiesaurus: WebSocket connections are actually an HTTP connection that is upgraded to WebSocket via the [UPGRADE header](https://en.wikipedia.org/wiki/HTTP/1.1_Upgrade_header), so it's really the same server hosting both the HTTP connections to server the frontend _and_ the WebSocket connections to the backend. I don't think it's really possible to make those ports separately configurable unless we run two separate servers.

veggiesaurus: I've added some additional text for the `port` argument

confluence (approved):

ajm-ska (approved):

PR #693, opened 2021-01-20
Try alternative ports if no port is specified and default port is in use

### LINKED ISSUE #671: try alternative ports if no port is specified and default port is in use
Currently, we default to port 3002. If there's something else (such as another carta-backend process) running on this port, we currently bail out. I propose that after #670 has been implemented, we also automatically try alternative ports, so that users can easily run multiple instances of carta-backend if they absolutely insist on doing so (they really should just be opening multiple tabs in their browser)

Using jupyter-notebook as an example: 

```
angus@angus-ryzen:~$ jupyter-notebook 
[I 08:04:55.573 NotebookApp] The port 8888 is already in use, trying another port.
[I 08:04:55.573 NotebookApp] The port 8889 is already in use, trying another port.
Serving notebooks from local directory: /home/angus
[I 08:04:55.575 NotebookApp] The Jupyter Notebook is running at:
[I 08:04:55.575 NotebookApp] http://localhost:8890/?token=149ba8329501160c9657eccaca73fb943283d0827370a1e3
```

No need for users to manually specify ports or anything.

markccchiang:
Closes #671. Use the `uWS::App::listen` function to get another free port if the default port 3002 is not available.


veggiesaurus:
@markccchiang minor merge conflicts, otherwise looks good to me

src/InterfaceConstants.h
```diff
@@ -73,4 +73,8 @@
 // Shared region polygon approximation
 #define DEFAULT_VERTEX_COUNT 1000
 
+// socket port
+#define DEFAULT_SOCKET_PORT 3002
```
veggiesaurus, line 78: @pford we originally wanted this file for ICD-related constants. Should we move the other `#defines` out of this file to a separate file?

pford: I would rather rename the file and keep them in one place.

pford: It looks like we no longer use the input parameter "threads",  TBB_THREAD_COUNT (its default value), or its variable thread_count in Main.cc.

pford: @markccchiang thanks for cleaning this up!

PR #694, opened 2021-01-20
Angus/simd downsampling

veggiesaurus:
This PR finally merges in my improved downsampling code, which uses SSE2 instructions for smoothing factor of 4 (or multiples of 4 when building without AVX) and AVX instructions for smoothing factors of 8 or higher. 

Speedups of the order of 2-3x when downsampling large images.

Wondering if we can safely make the _default_ build setting `EnableAvx` yet? The only CPUs I know of that don't support this are the Atom and ARM CPUs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,8 +108,7 @@ set(LINK_LIBS
         ${PROTOBUF_LIBRARY}
         fmt
         uSockets
-        ssl
-        crypto
+        uuid
         z
         zfp
         zstd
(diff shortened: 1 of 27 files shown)


veggiesaurus:
I considered writing some unit tests for this, but I got stuck, because the results won't be identical due to floating point errors. But this sort of optimisation PR really should come with some unit tests confirming its accuracy :thinking: 

Marking this as a draft PR until this has been discussed properly

confluence:
You could write unit tests which check for equality within some margin of error.

confluence:
It may also be useful to visualise the error as an image -- that may highlight any weird edge cases at block boundaries, etc..

veggiesaurus:
I've added a bunch of tests (and fixed a couple of path issues with the old tests), now ready for review

confluence (approved):

veggiesaurus:
@kswang1029 could you give this a try and see if you can notice any odd behaviour?

kswang1029:
will do

kswang1029:
I tried some test images (w/ and w/o NaN, images with exceptional values such as 1E-20, 1E30, images with 1 and 0 having regular spacing on the image, etc) and all appear working like before. 👍

https://github.com/CARTAvis/carta-frontend/issues/1223 remains but I think it is a separate issue.

kswang1029:
hold on. I got a backend crash with a 80000x80000 image. Investigating now

kswang1029:
after few more tests, the crash I saw seems to be something else non-trivial to reproduce. This PR is ready to go.

PR #695, opened 2021-01-21
Angus/test histograms

### LINKED ISSUE #642: OMP threads and Cube Histogram
Edit from @veggiesaurus: This seems to occur on all platforms, for channel or cube histograms!

We have discovered an issue with the Cube Histogram calculation on MacOS where it can produce a slightly different result than on Linux (RedHat7 and Ubuntu).

A workaround is to define the environment variable `OMP_NUM_THREADS` in the shell on MacOS before starting the carta_backend. However, it only works properly when it is set in a factor of two sequence; 2, 4, 8, 16, 32, 64 etc. 
If it is set as 6, 12, or 50, for example, it exhibits the same issue.

Setting `OMP_NUM_THREADS` on Linux is apparently not needed.

The carta_backend flag `-omp_threads` has no noticeable effect.

The Cube Histogram calculation should give identical results on MacOS and Linux without needing OMP_NUM_THREADS defined in advance.

Screen Shot 2020-10-14 at 8 38 44 PM

veggiesaurus:
fixes #642 

Replaced histogram calculations with a more conventional OpenMP approach, and added some unit tests (which previously failed) testing its behaviour with various thread counts)

edit: I should have targeted `dev` instead of my other PR! Both are reasonable isolated PRs, so it shouldn't be an issue

### DIFF
--- a/src/ImageStats/Histogram.cc
+++ b/src/ImageStats/Histogram.cc
@@ -10,8 +10,6 @@
 #include <cmath>
 
 #include <omp.h>
-#include <tbb/parallel_for.h>
-#include <tbb/parallel_reduce.h>
 
 using namespace carta;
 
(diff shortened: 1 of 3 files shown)


confluence (approved):

kswang1029:
@acdo2002 is testing with our existing ICD test script to try different omp_threads and different OS. We should expect all results identical and correct.

acdo2002:
perCubeOmp4
perChannel116Omp4

I tested on the CentOS and MacOS server with the backend branch` angus/test-histogram (commit#: ad4ea10)`
Even with the different OS, different OMP (4,5,8), the ICD tests received the sequence of the bins are all the same. 
The attached plots show the issue reported difference disappear (because the identical bins you only can see the latter plotted red colour), issue solved!!!

veggiesaurus:
Closing in favour of #698

PR #699, opened 2021-01-22
Pam/685 region style

### LINKED ISSUE #685: incorrect parsed line width from a ds9 region file
The line width seen in the message from the backend is not correct, resulting the frontend renders an unexpected region.

The line width of the region in the region file is not defined (meaning just default) thus it appears that the backend return a garbage value instead of applying a default value in the message.

https://user-images.githubusercontent.com/20819712/103845623-62295a00-50d7-11eb-84ef-deebb9f3f4cf.mp4

Screen Shot 2021-01-07 at 10 55 28 AM

pford:
Closes #685.

Use default style parameters for region import if not defined in region file: carta defaults for DS9 import, CRTF defaults for CRTF import.

### DIFF
--- a/src/InterfaceConstants.h
+++ b/src/InterfaceConstants.h
@@ -68,9 +68,9 @@
 #define OUTPUT_ID_MULTIPLIER 1000
 
 // CARTA default region style
-#define STYLE_COLOR "#2EE6D6"
-#define STYLE_DASH_LENGTH 2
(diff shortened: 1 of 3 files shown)


confluence (approved):

kswang1029 (approved):
tested and it works well. 👍

markccchiang (approved):

PR #700, opened 2021-01-25
re-introduce threads argument with deprecation text

veggiesaurus:
to prevent errors when running `carta_backend` with the deprecated `threads` arguement


ajm-ska (approved):
It works. I'm glad the carta_backend can now still run even if various wrapper scripts still use the -threads flag.

Although an additional message in the terminal would be nice to inform the user that the -threads flag no longer does anything if called, I guess if pre-existing wrapper scripts are being used, the message may not be seen anyway. So mentioning it in --help and User Manual is probably enough.

PR #703, opened 2021-01-27
Angus/deprecated threads arg

veggiesaurus:
As requested by @robsimmonds, I've re-introduced the old behaviour of the `threads` argument, because setting the scheduler's thread limit might be required in order to queue blocking messages, rather than trying to handle them all at once.

I've also deferred printing of the frontend URL until the port has been determined.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -574,8 +574,8 @@ int main(int argc, const char* argv[]) {
             port = inp.getInt("port");
             host = inp.getString("host");
             grpc_port = inp.getInt("grpc_port");
-            omp_thread_count = inp.getInt("threads");
-            thread_count = inp.getInt("omp_threads");
+            omp_thread_count = inp.getInt("omp_threads");
+            thread_count = inp.getInt("threads");
             base_folder = inp.getString("base");
             root_folder = inp.getString("root");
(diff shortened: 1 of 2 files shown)


src/Main.cc
```diff
@@ -573,7 +574,8 @@ int main(int argc, const char* argv[]) {
             port = inp.getInt("port");
             host = inp.getString("host");
             grpc_port = inp.getInt("grpc_port");
-            omp_thread_count = inp.getInt("omp_threads");
+            omp_thread_count = inp.getInt("threads");
```
pford, line 577: Inputs assigned to wrong variables

pford (changes requested):

src/Main.cc
```diff
@@ -573,7 +574,8 @@ int main(int argc, const char* argv[]) {
...
+            thread_count = inp.getInt("omp_threads");
```
pford: Inputs assigned to wrong variables

pford (changes requested):

src/Main.cc
veggiesaurus: Whoops!

markccchiang (approved):

PR #705, opened 2021-01-28
#689 Fix region histogram

### LINKED ISSUE #689: region histogram widget not updating when updating the angle of an ellipse
user feedback of a bug

It appears to work with a rotated rectangle but not a rotated ellipse. Initially it updates properly but not the second time onward. Tracing down the messages and it seems to be that the backend does receive an update of region and does send out a new region_histogram_data but the histogram data itself is not updated.

pford:
Closes #689 .

Discovered when the region histogram was not updating for a rotated ellipse, but this was an issue for all regions.  The histogram was computed from the data for the entire subimage, not for the masked data indicating the pixels included in the region.  Now the histogram is updated, and the mean and stdDev stats sent in the region histogram message match the values for the region in the Statistics widget.

### DIFF
--- a/src/Frame.cc
+++ b/src/Frame.cc
@@ -1276,7 +1276,19 @@ bool Frame::GetRegionData(const casacore::LattRegionHolder& region, std::vector<
         casacore::Slicer slicer(start, count); // entire subimage
         std::unique_lock<std::mutex> ulock(_image_mutex);
         sub_image.doGetSlice(tmp, slicer);
+
+        // Get mask that defines region in subimage bounding box
+        casacore::Array<bool> tmpmask;
+        sub_image.doGetMaskSlice(tmpmask, slicer);
         ulock.unlock();
+
+        // Apply mask to data
+        std::vector<bool> datamask = tmpmask.tovector();
+        for (size_t i = 0; i < data.size(); ++i) {
+            if (!datamask[i]) {
+                data[i] = NAN;
+            }
+        }
 
         if (_perflog) {
             auto t_end_get_subimage_data = std::chrono::high_resolution_clock::now();


veggiesaurus:
are there any memory implications for this? I guess since you're using a `vector<bool>` this will be 1 bit per pixel, compared to 32 bits for the actual data. 

On that note, is the region subimage extracted as a copy before the region calculations? So if I have a 20 GB 2D image and select half of it, my memory use will be 30 GB?

Edit: it appears to work ok  :+1: 
selecting a large (40k x 40k) region in an 80k x 80k image:
image

src/Frame.cc
```diff
@@ -1276,8 +1276,20 @@ bool Frame::GetRegionData(const casacore::LattRegionHolder& region, std::vector<
...
         std::unique_lock<std::mutex> ulock(_image_mutex);
         sub_image.doGetSlice(tmp, slicer);
+
+        // Get mask that defines region in subimage bounding box
+        casacore::Array<bool> tmpmask;
+        sub_image.doGetMaskSlice(tmpmask, slicer);
         ulock.unlock();
 
+        // Apply mask to data
+        std::vector<bool> datamask = tmpmask.tovector();
```
confluence, line 1286: I assume that this is to get a 1D mask to match the shape of `data`. I *think* from the casacore code that this doesn't unnecessarily copy data if the storage is contiguous. But could we not access `data` here through the 2D `tmp` array instead, matching the shape of the mask?

pford: I would have to use casacore::ArrayIterator, which according to the perflog output adds ~50 ms per region histogram.  But if copying is necessary for contiguous storage it would copy a 'cursor' at a time instead of the entire array.

pford: I pushed branch pam/689_histogram_iter

confluence: If the array iteration is less efficient, the current approach is probably fine -- I was concerned about possible copying of data when the region is very large.

confluence (approved):

kswang1029 (approved):
Confirmed the fix by testing rectangle and ellipse regions with different redundant position angles. Shared region also works.

kswang1029:
@acdo2002 could you check with histogram ICD tests too?

markccchiang (approved):

acdo2002:
Since the current test does not include this test (the current ICD test is that: CATYA.RegionHistogramData controls by the CARTASetHistogramRequirements instead of updated region), I wrote another quick test to test this issue. 

Yes, I see the improvement  on the pam's backend branch.

kswang1029:
@acdo2002 do you see existing tests related to histogram all pass?

acdo2002:
@kswang1029 
I think the most related is REGION_HISTOGRAM.test.ts, yes, it is passed.
Then I run all * HISTOGRAM *.test.ts, yes they are passed. 
So far I would say it looks ok.

Issue #696, opened 2021-01-21
Cleanup histogram code


veggiesaurus:
(possibly related to #680)

The histogram code needs to be cleaned up a bit. We have a `HistogramResults` struct and a `Histogram` class that could easily be combined. In its current state, it's probably leading to needless copies. We just need a `Histogram` struct with the following:
- a constructor that takes an existing set of bounds and bins (for use with HDF5 files)
- a constructor that takes a min, max, num_bins and a const ref to the data, which it uses to calculate the bins
- a join function for combining multiple histograms of the same size

This could be combined with some sensible histogram unit tests so we know when issues crop up

veggiesaurus:
@jolopezl has pointed out that the current code simply clamps a value's bin to between the first and last bins, even if the value is out of the range of the histogram. This hasn't been an issue yet, because our histograms are automatically calculated using the min and max bounds, but we should adjust this behaviour. @jolopezl suggests we just reject all values outside of the histogram range. I am inclined to agree. @kswang1029 is that ok?

In terms of implementation, it would just be replacing a single `std::clamp` call with an `if/else`

kswang1029:
@veggiesaurus @jolopezl Yes, in the future we should allow users to define 
- bin starting value
- bin width
- number of bins
- (or the alternative as bin starting and ending values, number of bins)
to construct a histogram from a region in the histogram widget.

veggiesaurus:
ok, then @jolopezl please go ahead with just rejecting out-of-bounds values. Could we combine this with the `isFinite` call and just use `if (val >= min_value && val <= max_value)`? (since `NaN >= val` and `NaN <= val` are both false for any finite `val`)

 Would we then not need `std::clamp` at all?

jolopezl:
For keeping the values right, I would change the condition to `if (std::isfinite(val) && _min_val <= val && val < _max_val)` and keep the `std::clamp` to get the bin number. It works fine because it's receiving the integer part so that the upper bound is never reached except for values above the range. Still, we could replace it with an if/else (`std::clamp` implementation is just that) to be sure that we keep an open-bound for the upper limit of each bin and ensure the behavior of `_min_val <= val < _max_val`. Note that I would add `_max_val` as a member of the class for simplicity.

veggiesaurus:
> For keeping the values right, I would change the condition to `if (std::isfinite(val) && _min_val <= val && val < _max_val)` 

[...]

PR #709, opened 2021-02-03
Histogram code cleanup

jolopezl:
Follow-up on #696:

- [x] More tests for histogram behavior, and internals
- [x] Reject values outside histogram boundaries, keeping semi-open bins (open bounded from above), except for the last bin, so that `lower_hist_limit <= val <= upper_hist_limit`.
- [x] Fixed join function to combine multiple histograms.
- [x] Add cross-checks for combining histograms.
- [x] Add a constructor that takes an existing set of bounds and bins (for use with HDF5 files)
- [x] Remove HistogramResults struct


src/ImageStats/Histogram.cc
```diff
@@ -13,33 +13,52 @@
 
 using namespace carta;
 
-Histogram::Histogram(int num_bins, float min_value, float max_value, const std::vector<float>& data)
-    : _num_bins(num_bins), _bin_width((max_value - min_value) / num_bins), _min_val(min_value), _hist(num_bins, 0), _data(data) {}
+Histogram::Histogram(int num_bins, float min_value, float max_value, const std::vector<float>& data) :_data(data) {
+    _histogram.num_bins = num_bins;
```
veggiesaurus: I'm wondering if we can just get rid of `num_bins`, since it's always equal to `histogram_bins.size()`

src/ImageStats/Histogram.cc
```diff
@@ -13,33 +13,52 @@
...
+    // ConsistencyCheck(this, h);
+    // if(!h)
+    const int num_bins = h.GetHistogramBins().size();
+    if (this->GetHistogramBins().size() != num_bins) {
+        return false;
+    }
 #pragma omp parallel for
     for (int i = 0; i < num_bins; i++) {
-        h._hist[i] += _hist[i];
+        _histogram.histogram_bins[i] += h.GetHistogramBins()[i];
```
veggiesaurus: I think it would be better to move `h.GetHistogramBins()` out of the parallel for (just using `auto& other_bins = h.GetHistogramBins()`

src/ImageStats/Histogram.h
```diff
@@ -18,25 +18,32 @@ struct HistogramResults {
...
     void setup_bins();
 
-    HistogramResults GetHistogram() const;
+    HistogramResults GetHistogram() const { return _histogram; }
+    float GetMinVal() const { return _min_val; }
+    float GetMaxVal() const { return _max_val; }
+    int GetNbins() const { return _histogram.num_bins; }
+    float GetBinWidth() const { return _histogram.bin_width; }
+    float GetBinCenter()  const { return _histogram.bin_center; }
+    std::vector<int> GetHistogramBins() const { return _histogram.histogram_bins; }
```
veggiesaurus: shouldn't this be a `const std::vector<int>&` return type? do we need to copy all the bins, rather than just getting a ref?

jolopezl: Yes, I agree. Do histograms ever change after filling? I am asking this to avoid having changed histogram stats that could be used elsewhere from a previous histogram setup.

src/ImageStats/Histogram.cc
jolopezl: Sure, good call.

src/ImageStats/Histogram.h
veggiesaurus: The only way to change them is to "add" another histogram

veggiesaurus: So the bin contents would change but not the actual bin count or Min/max

jolopezl: If you add another histogram, bin counts change so that the histogram results will change as well. I'll add the const and reference, but it's important to keep that in mind for the histogram results usage.

veggiesaurus: Sorry, I meant the number of bins will stay the same, not the bin counts ☺️

src/ImageStats/Histogram.cc
jolopezl: I prefer to have the variable cached there even when the calling size is O(1), but that's a bit of a personal taste. I'd say it makes the code more readable and, in theory, negligible faster. I'd rather change its type to `unsigned int`, but that will mess with `std::clamp`. So, better to have it as `int` and avoid using vector::size() which returns `unsigned long`.

veggiesaurus:
@jolopezl can you fix the merge conflicts?

src/ImageStats/Histogram.cc
```diff
@@ -6,40 +6,55 @@
...
+    _histogram.bin_width =  (max_value - min_value) / num_bins;
+    _min_val = min_value;
+    _max_val = max_value;
+    _histogram.histogram_bins.resize(num_bins);
+    _histogram.bin_center = _min_val + (_histogram.bin_width * 0.5);
+}
 
-Histogram::Histogram(Histogram& h, tbb::split)
-    : _num_bins(h._num_bins), _bin_width(h._bin_width), _min_val(h._min_val), _hist(h._hist.size(), 0), _data(h._data) {}
+Histogram::Histogram(Histogram& h, tbb::split) : _data(h.GetData()) {
```
veggiesaurus: we don't need to use tbb here anymore

src/ImageStats/Histogram.cc
```diff
@@ -6,40 +6,55 @@
...
+    _histogram.num_bins = h.GetNbins();
+    _histogram.bin_width = h.GetBinWidth();
+    _min_val = h.GetMinVal();
+    _max_val = h.GetMaxVal();
+    _histogram.histogram_bins = h.GetHistogramBins();
+}
 
 void Histogram::operator()(const tbb::blocked_range<size_t>& r) {
```
veggiesaurus: I don't think we need this function anymore

src/ImageStats/Histogram.h
```diff
@@ -18,25 +18,33 @@ struct HistogramResults {
...
 
-    HistogramResults GetHistogram() const;
+    HistogramResults GetHistogram() const { return _histogram; }
+    float GetMinVal() const { return _min_val; }
+    float GetMaxVal() const { return _max_val; }
+    int GetNbins() const { return _histogram.num_bins; }
+    float GetBinWidth() const { return _histogram.bin_width; }
+    float GetBinCenter()  const { return _histogram.bin_center; }
+    const std::vector<int>& GetHistogramBins() const { return _histogram.histogram_bins; }
+    const std::vector<float>& GetData() const { return _data; }
```
veggiesaurus: I don't think we need to hang on to the data after we've constructed the histogram

src/ImageStats/Histogram.h
```diff
@@ -18,25 +18,33 @@ struct HistogramResults {
...
     void operator()(const tbb::blocked_range<size_t>& r);
-    void join(Histogram& h); // NOLINT
+    bool join(Histogram& h); // NOLINT
     void setup_bins();
 
-    HistogramResults GetHistogram() const;
+    HistogramResults GetHistogram() const { return _histogram; }
+    float GetMinVal() const { return _min_val; }
+    float GetMaxVal() const { return _max_val; }
+    int GetNbins() const { return _histogram.num_bins; }
```
veggiesaurus: we should unify this, so that we don't need the `HistogramResults` struct. It doesn't really serve any purpose

jolopezl: @veggiesaurus, if you and others agree, I could refactor more code. I would remove HistogramResults and keep the histogram object and its content. That would change several parts of the codebase so that instead of passing HistogramResults around, we can pass Histogram. With the right method calls, it should be even easier to read.

src/ImageStats/Histogram.h
jolopezl: @veggiesaurus, I'd actually drop the data from the class. After calling setup_bins in the constructor, we don't need it anymore.

veggiesaurus: Exactly

src/ImageStats/Histogram.h
veggiesaurus: > I would remove HistogramResults and keep the histogram object and its content. That would change several parts of the codebase so that instead of passing HistogramResults around, we can pass Histogram. With the right method calls, it should be even easier to read.

@jolopezl please do so. I think it's just a needless complication at this point.

veggiesaurus (changes requested):
please refactor to remove the `HistogramResults` struct

veggiesaurus:
@jolopezl we've just merged in PR #717, which will ensure that running `reformat.sh` will also check test files. You may need to make some minor merge conflict changes after this :)

src/ImageStats/Histogram.cc
```diff
@@ -6,75 +6,89 @@
...
-    results.histogram_bins = _hist;
-    return results;
-}
+bool Histogram::ConsistencyCheck(const Histogram& a, const Histogram& b) {
+    if (a.GetNbins() != b.GetNbins()) {
+        std::printf("(debug) Histograms don't have the same number of bins: %d and %d\n", a.GetNbins(), b.GetNbins());
+        return false;
+    }
+    if (fabs(a.GetMinVal() - b.GetMinVal()) > std::numeric_limits<float>::epsilon()) {
+        std::printf("(debug) Lower histograms limits are not equal: %f and %f\n", a.GetMinVal(), b.GetMinVal());
```
veggiesaurus: please use the new logging approach (`spdlog::debug(...)`) now that #713 has been merged

veggiesaurus: this seems like something that should result in a `warn` level

src/ImageStats/Histogram.cc
```diff
@@ -6,75 +6,89 @@
...
+    _min_val = h.GetMinVal();
+    _max_val = h.GetMaxVal();
+    _histogram.histogram_bins = h.GetHistogramBins();
 }
 
-void Histogram::join(Histogram& h) { // NOLINT
-    auto num_bins = h._hist.size();
+bool Histogram::join(Histogram& h) { // NOLINT
+    if (!ConsistencyCheck(*this, h)) {
+        fmt::print("(debug) Consistency check failed to join histograms - won't join them\n");
```
veggiesaurus: Use `spdlog::warn`

src/Frame.cc
```diff
@@ -758,18 +758,18 @@ bool Frame::CalculateHistogram(int region_id, int channel, int stokes, int num_b
         }
         bool write_lock(false);
         tbb::queuing_rw_mutex::scoped_lock cache_lock(_cache_mutex, write_lock);
-        CalcHistogram(num_bins, stats, _image_cache, results);
+        hist = CalcHistogram(num_bins, stats, _image_cache);
```
jolopezl, line 761: I don't like this copy. CalcHistogram returns a histogram, that could actually be _moved_...

veggiesaurus: Won't it be implicitly moved?

jolopezl: Yes, good point, it should be implicitly moved; I think overthought that `=`

pford: Or pass hist by reference to CalcHistogram, which fills the values.

pford: or pass hist by reference

jolopezl: I think it's the same behavior. Either at this point or inside CalcHistogram the histogram is constructed and placed into hist.

jolopezl:
Given the number of changes, I ran a small manual test:

- First, I opened an image with backend@dev, and stored a histogram for the full image and a region. I kept the region for later use. 
- Then, I opened the same image with backend@thisPR, and did the same, using the previously stored region. 

Both exported histograms have the same length and number of counts.

src/ImageStats/Histogram.cc
```diff
@@ -6,46 +6,54 @@
...
+    _min_val = h.GetMinVal();
+    _max_val = h.GetMaxVal();
+    _histogram_bins = h.GetHistogramBins();
 }
 
-void Histogram::join(Histogram& h) { // NOLINT
-    auto num_bins = h._hist.size();
+bool Histogram::join(const Histogram& h) { // NOLINT
+    if (!ConsistencyCheck(*this, h)) {
+        spdlog::warn("A consistency check failed to join histograms");
```
confluence: I suggest "Could not join histograms: consistency check failed."

src/ImageStats/Histogram.cc
```diff
@@ -6,46 +6,54 @@
...
 
+#include "Logger/Logger.h"
 #include "Threading.h"
 
 using namespace carta;
 
-Histogram::Histogram(int num_bins, float min_value, float max_value, const std::vector<float>& data)
-    : _num_bins(num_bins), _bin_width((max_value - min_value) / num_bins), _min_val(min_value), _hist(num_bins, 0), _data(data) {}
+Histogram::Histogram(int num_bins, float min_value, float max_value, const std::vector<float>& data) {
+    _num_bins = num_bins;
```
confluence: Can we put these back in an initializer list, please?

src/ImageStats/Histogram.cc
```diff
@@ -55,26 +63,39 @@ void Histogram::setup_bins() {
...
-    results.bin_center = _min_val + (_bin_width / 2.0);
-    results.histogram_bins = _hist;
-    return results;
+bool Histogram::ConsistencyCheck(const Histogram& a, const Histogram& b) {
+    if (a.GetNbins() != b.GetNbins()) {
+        spdlog::warn("Histograms don't have the same number of bins: {} and {}", a.GetNbins(), b.GetNbins());
+        return false;
+    }
+    if (fabs(a.GetMinVal() - b.GetMinVal()) > std::numeric_limits<float>::epsilon()) {
+        spdlog::warn("Lower histograms limits are not equal: {} and {}", a.GetMinVal(), b.GetMinVal());
```
confluence: histograms -> histogram

src/ImageStats/Histogram.cc
```diff
@@ -55,26 +63,39 @@ void Histogram::setup_bins() {
...
+        return false;
+    }
+    if (fabs(a.GetMaxVal() - b.GetMaxVal()) > std::numeric_limits<float>::epsilon()) {
+        spdlog::warn("Upper histograms limits are not equal: {} and {}", a.GetMaxVal(), b.GetMaxVal());
```
confluence: histograms -> histogram

src/ImageStats/Histogram.cc
```diff
@@ -55,26 +63,39 @@ void Histogram::setup_bins() {
...
+        return false;
+    }
+    return true;
+}
+void Histogram::SetHistogramBins(const std::vector<int>& bins) {
+    if (bins.size() != _histogram_bins.size()) {
+        spdlog::error("Vector sizes are not equal, can't reset histogram counts");
```
confluence: I suggest "Could not reset histogram counts: vector sizes are not equal."

src/ImageStats/Histogram.h
```diff
@@ -14,29 +14,45 @@
...
-    void operator()(const tbb::blocked_range<size_t>& r);
-    void join(Histogram& h); // NOLINT
-    void setup_bins();
 
-    HistogramResults GetHistogram() const;
+    Histogram(const Histogram& h);
+
+    bool join(const Histogram& h); // NOLINT
+
+    float GetMinVal() const {
```
confluence, line 35: I'm not 100% convinced that we need to use setters and getters for these, rather than accessing the member attributes (as we did with the results struct), but I don't feel that strongly about it. Python instincts. :)

confluence (commented):
I think that this looks good overall; simplifying the classes was a good idea.

src/ImageStats/Histogram.cc
veggiesaurus: do we need `_num_bins` at all? Can't we use `bins.size()`? Less room for inconsistencies.

confluence: Good point -- I can't remember if this is a relic of the previous class structure.

src/ImageStats/Histogram.h
veggiesaurus: I think getters (with no setters) are the right approach here, because creating a histogram, and then simply changing the min or max values using `histogram.min_val = 20` will make the results incorrect.

confluence: Fair enough; it's safer to prevent this.

jolopezl: I didn't add setters for that reason; we don't want to corrupt the histogram behavior. I don't want to use direct access to member attributes because that makes it a bit harder to do changes in the class behavior without having an impact on the rest of the code, see for example, this PR :)

src/ImageStats/Histogram.cc
jolopezl: Changed.

test/TestHistogram.cc
```diff
@@ -43,22 +50,77 @@ class HistogramTest : public ::testing::Test {
...
+    data.push_back(9.1);  // should go to bin at pos. 9
+    data.push_back(10.0); // should go to bin at pos. 9 - last bin is closed from above
+    // values that will fall in the overflow and underflow range
+    data.push_back(-1.0);        // should not appear
+    data.push_back(00.0 - 1e-9); // should not appear
+    data.push_back(10.0 + 1e+9); // should not appear
+    data.push_back(11.0);        // should not appear
+    carta::Histogram hist(10, 0.0f, 10.0f, data);
+    auto counts = accumulate(hist.GetHistogramBins().begin(), hist.GetHistogramBins().end(), 0);
+    EXPECT_TRUE(counts == 12);
```
veggiesaurus: I think we should use `EXPECT_EQ` for all of these checks

test/TestHistogram.cc
```diff
@@ -43,22 +50,77 @@ class HistogramTest : public ::testing::Test {
...
+TEST_F(HistogramTest, TestHistogramJoin) {
+    std::vector<float> data(1024 * 1024);
+    std::for_each(data.begin(), data.end(), [&](float& v) { v = float_random(mt); });
+    carta::Histogram hist(1024, 0.0f, 1.0f, data);
+    const auto total_counts = accumulate(hist.GetHistogramBins().begin(), hist.GetHistogramBins().end(), 0);
+    carta::Histogram hist2(1024, 0.0f, 1.0f, data);
+    EXPECT_TRUE(CompareResults(hist, hist2)); // naive?
+    hist.join(hist2);
+    const auto total_counts2 = accumulate(hist.GetHistogramBins().begin(), hist.GetHistogramBins().end(), 0);
+    EXPECT_TRUE(total_counts * 2 == total_counts2);
```
veggiesaurus: Use `EXPECT_EQ` here as well

test/TestHistogram.cc
```diff
@@ -94,22 +151,17 @@ TEST_F(HistogramTest, TestMultithreadingPerformance) {
...
-    auto results_mt = mt.GetHistogram();
+    carta::Histogram hist_mt(1024, 0.0f, 1.0f, data);
     t.End("multi_threaded");
 
     auto st_time = t.GetMeasurement("single_threaded");
     auto mt_time = t.GetMeasurement("multi_threaded");
     double speedup = st_time / mt_time;
-    cout << speedup << endl;
-    EXPECT_GE(speedup, 1.5);
+    EXPECT_GE(speedup, 1.5) << "Speedup is: " << speedup;
```
veggiesaurus, line 164: I had no idea we could do this... Does it definitely work? Does it only print out if the test fails?

src/Session.cc
```diff
@@ -1183,18 +1183,18 @@ bool Session::CalculateCubeHistogram(int file_id, CARTA::RegionHistogramData& cu
...
+                carta::Histogram chan_histogram; // histogram for each channel using cube stats
                 for (size_t chan = 0; chan < num_channels; ++chan) {
                     if (!_frames.at(file_id)->CalculateHistogram(CUBE_REGION_ID, chan, stokes, num_bins, cube_stats, chan_histogram)) {
                         return calculated; // channel histogram failed
                     }
 
                     // add channel bins to cube bins
                     if (chan == 0) {
-                        cube_bins = {chan_histogram.histogram_bins.begin(), chan_histogram.histogram_bins.end()};
+                        cube_bins = {chan_histogram.GetHistogramBins().begin(), chan_histogram.GetHistogramBins().end()};
```
veggiesaurus, line 1194: can't we use `join` here? :thinking:

src/ImageStats/Histogram.h
```diff
@@ -14,29 +14,44 @@
...
-    Histogram(Histogram& h, tbb::split);
-
-    void operator()(const tbb::blocked_range<size_t>& r);
-    void join(Histogram& h); // NOLINT
-    void setup_bins();
 
-    HistogramResults GetHistogram() const;
+    Histogram(const Histogram& h);
+
+    bool join(const Histogram& h); // NOLINT
```
veggiesaurus: we no longer need the `// NOLINT` and can change this back to `Join` or `Add` or `Accumulate`

src/ImageStats/Histogram.cc
```diff
@@ -6,75 +6,93 @@
...
-
-void Histogram::join(Histogram& h) { // NOLINT
-    auto num_bins = h._hist.size();
+    const int num_bins = h.GetHistogramBins().size();
+    const auto& other_bins = h.GetHistogramBins();
     ThreadManager::ApplyThreadLimit();
 #pragma omp parallel for
     for (int i = 0; i < num_bins; i++) {
-        h._hist[i] += _hist[i];
+        _histogram_bins[i] += other_bins[i];
```
veggiesaurus: I wonder if there's actually any merit in doing this in parallel :thinking: I'd suspect not. What about using an `omp simd` loop instead? See here for details: https://stackoverflow.com/a/14717689/1727322

veggiesaurus: Have just done a quick test of calling `join` with a 10 000 bin histogram. Definitely slower when using `omp parallel`, and a little bit faster (than a sequential for-loop) using `omp simd`.

veggiesaurus (changes requested):

test/TestHistogram.cc
jolopezl: Definitely work. It will only print if the test fails. I think it's pretty cool, see https://github.com/google/googletest/blob/master/docs/advanced.md#teaching-googletest-how-to-print-your-values for more advanced examples. (TL;DR: Every macro can do that.)

test/TestHistogram.cc
jolopezl: Ok.

test/TestHistogram.cc
jolopezl: Ok.

src/ImageStats/Histogram.h
jolopezl: Ok, I will use `Add`

src/ImageStats/Histogram.cc
jolopezl: Keeping `#pragma omp simd` for that loop.

PR #710, opened 2021-02-03
Angus/omp threads check

veggiesaurus:
This PR makes a couple of changes:

- No longer enforces use of parallel variants of STL algorithms using the `_GLIBCXX_PARALLEL` flag. Instead, the parallel variants are manually called when sorting catalogs. The correct variant is determined using the `__has_includes` C++17 feature (clang doesn't have parallel STL support yet, and GCC can use the `__gnu_parallel::sort` function, which appears to be considerably faster than `std::sort`, even when `std::execution::par_unseq` is available
- Improves management of the number of OpenMP threads. We no longer assume that other libraries haven't changed the number of threads to be used. Instead, we apply the limit before each `#omp parallel`
- Default to using all available cores for OpenMP. This behaviour will also occur if the `omp_threads` argument is set to <= 0


CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ else ()
 endif ()
 
 # Enable parallel GCC algorithms (needed until there is relable support for C++17 from compilers)
```
confluence: I don't mind leaving this commented out instead of removing it, but I would edit the comment to reflect the current state and explain why this is disabled.

confluence (approved):
Apart from the comment in the cmake config, this looks fine.

CMakeLists.txt
veggiesaurus: I think it can be axed. fixing...

ajm-ska (approved):
I tried making a distributable package on MacOS. It seems to work fine. (Built and packaged on 10.15, running on 11.1).

I know it is not related to this PR, but while you are modifying src/Main.cc, how about also fixing the typo in the --help output here:
```
init_exit_after, Int, 
    number of seconds to stay alive at start if no clents connect
```

veggiesaurus:
> I know it is not related to this PR, but while you are modifying src/Main.cc, how about also fixing the typo in the --help output here:

ok, will do :)

Issue #574, opened 2020-08-04
Unify logging


kswang1029:
- [ ] provide info to the frontend if backend crashed
- [ ] provide backend log to the frontend

more detail follows later.

PR #713, opened 2021-02-04
Unify existing logging codes and output log messages into a log file

markccchiang:
This is related to #475 and #574. The first step is to unify existing logging codes with *spdlog*. There are five kinds of log messages: `DEBUG`, `INFO`, `WARN`, `ERROR`, and `PERF`. By default, only INFO, WARN, and ERROR messages will be shown on the console and saved in a log file. Setting `verbose=true`, DEBUG messages will be included. Setting `perflog=true`, PERF (performance) messages will be included. The output log file `carta.log` will be saved under the directory `/var/log/carta`. If this directory is not writable, then it will be saved under the user's home `$HOME/CARTA/log`. If don't want to output console messages into a log file, then setting `no_log=true`. There is also a little clean up of the code.


veggiesaurus:
> The output log file `carta.log` will be saved under the directory `/var/log/carta`

I think by default we should use user-level log files in `$HOME/.carta/log`. `/var/log` seems to be more commonly used for system log. @confluence what do you think?

I also think we should use something other than `carta.log`, can we not give it a timestamp instead?

src/Frame.cc
```diff
@@ -33,9 +33,8 @@ namespace fs = std::filesystem;
 
 using namespace carta;
 
-Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string& hdu, bool verbose, bool perflog, int default_channel)
+Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string& hdu, bool perflog, int default_channel)
```
veggiesaurus: do we still need `_perflog`?

veggiesaurus: can't we just rely on the logger to ignore non-performance output?

src/Logger.cc
```diff
@@ -0,0 +1,87 @@
+/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
+   Copyright 2018, 2019, 2020, 2021 Academia Sinica Institute of Astronomy and Astrophysics (ASIAA),
+   Associated Universities, Inc. (AUI) and the Inter-University Institute for Data Intensive Astronomy (IDIA)
+   SPDX-License-Identifier: GPL-3.0-or-later
+*/
+
+#include "Logger.h"
+
+#include <filesystem>
```
veggiesaurus: need to check whether we're building with boost filesystem or std filesystem

src/Logger.cc
```diff
@@ -0,0 +1,87 @@
+/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
+   Copyright 2018, 2019, 2020, 2021 Academia Sinica Institute of Astronomy and Astrophysics (ASIAA),
+   Associated Universities, Inc. (AUI) and the Inter-University Institute for Data Intensive Astronomy (IDIA)
+   SPDX-License-Identifier: GPL-3.0-or-later
+*/
+
+#include "Logger.h"
```
veggiesaurus: I think we should move these files into a `Logger` folder. I also think the function definitions in `Logger.h` should be moved into `Logger.tcc`

src/Logger.cc
```diff
@@ -0,0 +1,87 @@
...
+namespace fs = std::filesystem;
+
+void InitLoggers(bool no_log_file, bool debug_log, bool perf_log) {
+    std::string log_fullname;
+    if (!no_log_file) {
+        std::string first_choice_log_dir("/var/log");
+        if (!access(first_choice_log_dir.c_str(), W_OK)) {
+            log_fullname = first_choice_log_dir + "/carta/carta.log";
+        } else {
+            log_fullname = fs::path(getenv("HOME")).string() + "/CARTA/log/carta.log";
```
veggiesaurus: see comment about log location

src/Logger.h
```diff
@@ -0,0 +1,77 @@
...
+
+void InitLoggers(bool no_log_file, bool debug_log, bool perf_log);
+
+template <typename S, typename... Args>
+void SpdLog(const std::string& log_tag, const LogType& log_type, bool flush_now, const S& format, Args&&... args) {
+    std::shared_ptr<spdlog::logger> logger = spdlog::get(log_tag);
+    if (logger) {
+        switch (log_type) {
+            case LogType::DEBUG:
+                logger->debug(fmt::format(format, args...));
```
veggiesaurus: can you not just pass `format` and `args...` directly into these functions? That way we don't end up actually doing the format call unless that level of debugging is active

src/Main.cc
```diff
@@ -671,17 +674,17 @@ int main(int argc, const char* argv[]) {
...
                 }
             });
         }
 
         if (port_ok) {
-            fmt::print("Listening on port {} with root folder {}, base folder {}", port, root_folder, base_folder);
+            INFO("Listening on port {} with root folder {}, base folder {}", port, root_folder, base_folder);
             if (omp_thread_count > 0) {
-                fmt::print(", and {} OpenMP worker threads\n", omp_thread_count);
+                INFO(", and {} OpenMP worker threads", omp_thread_count);
```
veggiesaurus: this now uses two lines instead of one

confluence:
On a shared system, I think it would be useful to aggregate backend logs for all users in a shared log file in `/var/log/carta`, but we'd have to consider how to set up permissions to allow backend instances running as individual users to log there. But I think this should happen _in addition_ to logging to individual `~/.carta/log/` directories. The combined global log should probably be called something like `backend.log` (because we may want a separate controller log for authentication) and rotated (not by us by hand; with configuration added to `/etc/logrotate.d/`). We would want to keep a backlog of those for debugging purposes.

I'm not sure what the best practice is for preventing individual user logs from growing indefinitely. Manual rotation? Overwriting the log every time? Keeping N previous logs at a time?

Some of this should probably be configurable, so that e.g. system-level logs could be disabled on a single-user system.

src/Main.cc
```diff
@@ -705,10 +708,10 @@ int main(int argc, const char* argv[]) {
 #endif
                     auto open_result = system(fmt::format("{} {}", open_command, frontend_url).c_str());
                     if (open_result) {
-                        fmt::print("Failed to open the default browser automatically.\n");
+                        ERROR("Failed to open the default browser automatically.");
```
veggiesaurus: this can just be a warning

veggiesaurus (changes requested):
see comments

confluence:
@veggiesaurus pointed out that the controller can handle system-level logging of the backend processes (by redirecting their stdout) -- so I retract everything I said about `/var/log`, and agree that the backend process should log just to `.carta` and to stdout. But we should rotate the logs in `.carta`. How many logs we keep should be configurable, but what's a sensible default?

confluence (changes requested):
This generally looks good, apart from the logfile location issue (discussed in the comments).

markccchiang:
The location of a log file is now changed to `$HOME/.carta/log`. If we use the timestamp as the log file name and users don't aware need to delete them, it may increase the disk space. In the current setting, the number of the rotated log files is five (i.e., carta.log, carta.1.log, carta.2.log, ... carta.5.log). And the max size of each log file is 5 MB. Any suggestion that for a better setting?

confluence:
I think those are good defaults, and the numbered naming is fine. Do we log timestamps in the actual log messages?

markccchiang:
> I think those are good defaults, and the numbered naming is fine. Do we log timestamps in the actual log messages?

Yes, it does.

veggiesaurus:
@markccchiang the changes look good, I think we're mostly there. Couple of comments that I think still need to be addressed:

- You're now manually checking to see if the log level is <= `debug`. But that's exactly what spdlog does underneath. That's why you have the log levels. Can't you just call `logger->debug(format, args..)` if you've set the log level correctly? I've tested it, and it seems to work as expected. Look at `logger.h` line 79 *(line number might depend on the version of spdlog). It already checks whether the log should be printed: `bool log_enabled = should_log(lvl);`
- I wonder if we should use the spdlog `trace` level for performance logging? That would make things much cleaner, and we could have a single verbosity setting to control this. @kswang1029 what do you think? instead of needing to say `verbose=true` and `perflog=true`, we could just have a `verbosity=<N>` argument, where the default is 4 (critical, error, warn and info). To enable debug output you set it to 5, and to enable performance logging as well, you set it to 6.

kswang1029:
> @markccchiang the changes look good, I think we're mostly there. Couple of comments that I think still need to be addressed:
> 
> * You're now manually checking to see if the log level is <= `debug`. But that's exactly what spdlog does underneath. That's why you have the log levels. Can't you just call `logger->debug(format, args..)` if you've set the log level correctly? I've tested it, and it seems to work as expected. Look at `logger.h` line 79 *(line number might depend on the version of spdlog). It already checks whether the log should be printed: `bool log_enabled = should_log(lvl);`
> * I wonder if we should use the spdlog `trace` level for performance logging? That would make things much cleaner, and we could have a single verbosity setting to control this. @kswang1029 what do you think? instead of needing to say `verbose=true` and `perflog=true`, we could just have a `verbosity=<N>` argument, where the default is 4 (critical, error, warn and info). To enable debug output you set it to 5, and to enable performance logging as well, you set it to 6.

@veggiesaurus great idea

veggiesaurus:
> > @markccchiang the changes look good, I think we're mostly there. Couple of comments that I think still need to be addressed:
> > 
> > * You're now manually checking to see if the log level is <= `debug`. But that's exactly what spdlog does underneath. That's why you have the log levels. Can't you just call `logger->debug(format, args..)` if you've set the log level correctly? I've tested it, and it seems to work as expected. Look at `logger.h` line 79 *(line number might depend on the version of spdlog). It already checks whether the log should be printed: `bool log_enabled = should_log(lvl);`
> > * I wonder if we should use the spdlog `trace` level for performance logging? That would make things much cleaner, and we could have a single verbosity setting to control this. @kswang1029 what do you think? instead of needing to say `verbose=true` and `perflog=true`, we could just have a `verbosity=<N>` argument, where the default is 4 (critical, error, warn and info). To enable debug output you set it to 5, and to enable performance logging as well, you set it to 6.
> 
> @veggiesaurus great idea

Ok, @markccchiang please implement it this way. On that note, I don't see why we need those wrapper functions at all. Couldn't we just carefully create a spdlog object (like you're doing in `Logger::InitLoggers`), set that as the default logger,  and then use spdlog directly? What advantage does creating a light wrapper to spdlog serve?

markccchiang:
> > > @markccchiang the changes look good, I think we're mostly there. Couple of comments that I think still need to be addressed:
> > > 
> > > * You're now manually checking to see if the log level is <= `debug`. But that's exactly what spdlog does underneath. That's why you have the log levels. Can't you just call `logger->debug(format, args..)` if you've set the log level correctly? I've tested it, and it seems to work as expected. Look at `logger.h` line 79 *(line number might depend on the version of spdlog). It already checks whether the log should be printed: `bool log_enabled = should_log(lvl);`
> > > * I wonder if we should use the spdlog `trace` level for performance logging? That would make things much cleaner, and we could have a single verbosity setting to control this. @kswang1029 what do you think? instead of needing to say `verbose=true` and `perflog=true`, we could just have a `verbosity=<N>` argument, where the default is 4 (critical, error, warn and info). To enable debug output you set it to 5, and to enable performance logging as well, you set it to 6.
> > 
> > 
> > @veggiesaurus great idea
> 
> Ok, @markccchiang please implement it this way. On that note, I don't see why we need those wrapper functions at all. Couldn't we just carefully create a spdlog object (like you're doing in `Logger::InitLoggers`), set that as the default logger, and then use spdlog directly? What advantage does creating a light wrapper to spdlog serve?

Ok, will do that. The advantage of creating a light wrapper is that we don't need to write `spdlog::get(log_tag)->info(...)` every time while inserting a log message in the code.

veggiesaurus:
> The advantage of creating a light wrapper is that we don't need to write `spdlog::get(log_tag)->info(...)` every time while inserting a log message in the code.

After setting the default logger correctly, can't we just call `spdlog::info`? Isn't that the point of the default logger and logger registry?

markccchiang:
> After setting the default logger correctly, can't we just call `spdlog::info`? Isn't that the point of the default logger and logger registry?

Yes. Since there is only one logger, that way will not cause confusion.

veggiesaurus:
@markccchiang this is looking good. Two minor changes required:
- ` spdlog::get_level` doesn't seem to be available in the version of spdlog installed in 20.04. `spdlog::default_logger()->level()` seems to work fine though.
- We actually don't need to check if the level is `== spdlog::level::trace`, just remove those checks. Getting the clock time is pretty inexpensive, so there's no need to confirm that perflog is enabled. There's one exception: `TraceCountours` does some rather expensive checks of vertex and index data, so I suggest we still check there. But we should be checking if the level is `>= spdlog::level::trace` in case we add an additional custom level in future.

veggiesaurus:
one final change request from my side: there are a few places where we use `spdlog::error` instead of `spdlog::critical`. Any time we log something that will cause the backend to exit with a non-zero status, we should be using `spdlog::critical`. There are 3 or 4 of them that I've spotted so far.

src/Util.cc
```diff
@@ -40,20 +40,21 @@ bool CheckRootBaseFolders(string& root, string& base) {
...
     casacore::File base_folder(base);
     if (!(base_folder.exists() && base_folder.isDirectory(true) && base_folder.isReadable() && base_folder.isExecutable())) {
-        spdlog::critical("Invalid base directory, does not exist or is not a readable directory. Exiting carta.");
-        return false;
-    }
-    // absolute path: resolve symlinks, relative paths, env vars e.g. $HOME
-    try {
-        base = base_folder.path().resolvedName(); // fails on root folder /
-    } catch (casacore::AipsError& err) {
+        spdlog::warn("Invalid base directory, use root directory instead.");
```
veggiesaurus: Can you change this to `"Invalid base directory, using the provided root directory instead"`

src/Main.cc
```diff
@@ -540,12 +538,13 @@ int main(int argc, const char* argv[]) {
...
         bool no_browser = false;
+        bool no_log = false;
+        int verbosity = 4;
 
         { // get values then let Input go out of scope
             casacore::Input inp;
-            inp.version(VERSION_ID);
-            inp.create("verbose", "False", "display verbose logging", "Bool");
-            inp.create("perflog", "False", "display performance logging", "Bool");
+            inp.create("verbosity", to_string(verbosity), "display verbose logging from level (4: info, 5: debug, 6: trace)", "Int");
```
veggiesaurus: might want to also add the other levels (off, critical, error, warning)

src/Frame.cc
```diff
@@ -50,9 +48,7 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
       _moment_generator(nullptr) {
     if (!_loader) {
         _open_image_error = fmt::format("Problem loading image: image type not supported.");
-        if (_verbose) {
-            carta::Log(session_id, _open_image_error);
-        }
+        spdlog::debug("Session {}: {}", session_id, _open_image_error);
```
confluence: These log levels should be made more consistent. A failure to load an image entirely should produce an error. I will make suggestions for these cases; @veggiesaurus will also check them afterwards.

src/Frame.cc
```diff
@@ -61,9 +57,7 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
         _loader->OpenFile(hdu);
     } catch (casacore::AipsError& err) {
         _open_image_error = err.getMesg();
-        if (_verbose) {
-            carta::Log(session_id, _open_image_error);
-        }
+        spdlog::debug("Session {}: {}", session_id, _open_image_error);
```
confluence: This should also be an error.

src/Frame.cc
```diff
@@ -72,9 +66,7 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
     std::string log_message;
     if (!_loader->FindCoordinateAxes(_image_shape, _spectral_axis, _stokes_axis, log_message)) {
         _open_image_error = fmt::format("Problem determining file shape: {}", log_message);
-        if (_verbose) {
-            carta::Log(session_id, _open_image_error);
-        }
+        spdlog::debug("Session {}: {}", session_id, _open_image_error);
```
confluence: This should be a warning.

src/Frame.cc
```diff
@@ -103,9 +95,7 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
         _loader->LoadImageStats();
     } catch (casacore::AipsError& err) {
         _open_image_error = fmt::format("Problem loading statistics from file: {}", err.getMesg());
-        if (_verbose) {
-            carta::Log(session_id, _open_image_error);
-        }
+        spdlog::debug("Session {}: {}", session_id, _open_image_error);
```
confluence: Warning.

src/Frame.cc
```diff
@@ -514,10 +498,10 @@ bool Frame::ContourImage(ContourCallback& partial_contour_callback) {
             size_t dest_width = ceil(double(image_bounds.x_max()) / _contour_settings.smoothing_factor);
             size_t dest_height = ceil(double(image_bounds.y_max()) / _contour_settings.smoothing_factor);
             TraceContours(dest_vector.data(), dest_width, dest_height, scale, offset, _contour_settings.levels, vertex_data, index_data,
-                _contour_settings.chunk_size, partial_contour_callback, _perflog);
+                _contour_settings.chunk_size, partial_contour_callback);
             return true;
         }
-        fmt::print("Smoothing mode not implemented yet!\n");
+        spdlog::error("Smoothing mode not implemented yet!");
```
confluence: Warning?

src/Frame.cc
```diff
@@ -997,7 +978,7 @@ bool Frame::SetSpectralRequirements(int region_id, const std::vector<CARTA::SetS
         int axis, stokes;
         ConvertCoordinateToAxes(coordinate, axis, stokes);
         if (stokes >= nstokes) {
-            std::cerr << "Spectral requirement " << coordinate << " failed: invalid stokes axis for image." << std::endl;
+            spdlog::error("Spectral requirement {} failed: invalid stokes axis for image.", coordinate);
```
confluence: Warning?

src/Frame.cc
```diff
@@ -1238,7 +1217,7 @@ bool Frame::GetImageRegion(int file_id, const ChannelRange& chan_range, int stok
         image_region = this_region;
         return true;
     } catch (casacore::AipsError error) {
-        std::cerr << "Error converting full region to file " << file_id << ": " << error.getMesg() << std::endl;
+        spdlog::error("Error converting full region to file {}: {}", file_id, error.getMesg());
```
confluence, line 1220: This is probably fine as an error.

src/ImageData/Hdf5Loader.cc
```diff
@@ -188,7 +190,7 @@ bool Hdf5Loader::GetCursorSpectralData(
             LoadSwizzledData()->doGetSlice(tmp, slicer);
             data_ok = true;
         } catch (casacore::AipsError& err) {
-            std::cerr << "Could not load cursor spectral data from swizzled HDF5 dataset. AIPS ERROR: " << err.getMesg() << std::endl;
+            spdlog::error("Could not load cursor spectral data from swizzled HDF5 dataset. AIPS ERROR: {}", err.getMesg());
```
confluence: This should perhaps be a warning.

src/Main.cc
```diff
@@ -183,7 +180,7 @@ void OnMessage(uWS::WebSocket<false, true>* ws, std::string_view sv_message, uWS
                     if (message.ParseFromArray(event_buf, event_length)) {
                         session->OnRegisterViewer(message, head.icd_version, head.request_id);
                     } else {
-                        fmt::print("Bad REGISTER_VIEWER message!\n");
+                        spdlog::error("Bad REGISTER_VIEWER message!");
```
confluence: I think that these should all be warnings.

src/Main.cc
confluence: We should list all the levels in the description.

src/Main.cc
```diff
@@ -640,7 +641,7 @@ int main(int argc, const char* argv[]) {
                     }
                 });
             } else {
-                fmt::print("Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument\n");
+                spdlog::error("Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument");
```
confluence: This should be a warning.

src/Moment/MomentGenerator.cc
```diff
@@ -110,7 +112,7 @@ void MomentGenerator::SetMomentAxis(const CARTA::MomentRequest& moment_request)
     } else if (moment_request.axis() == CARTA::MomentAxis::STOKES) {
         _axis = _stokes_axis;
     } else {
-        std::cerr << "Do not support the moment axis: " << moment_request.axis() << std::endl;
+        spdlog::error("Do not support the moment axis: {}", moment_request.axis());
```
confluence: Description should be changed to `"Unsupported moment axis: "`.

src/OnMessageTask.cc
```diff
@@ -22,7 +20,7 @@ tbb::task* MultiMessageTask::execute() {
             if (message.ParseFromArray(_event_buffer, _event_length)) {
                 _session->OnSetSpatialRequirements(message);
             } else {
-                fmt::print("Bad SET_SPATIAL_REQUIREMENTS message!\n");
+                spdlog::error("Bad SET_SPATIAL_REQUIREMENTS message!");
```
confluence: These should also all be warnings.

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -22,9 +21,9 @@ SimpleFrontendServer::SimpleFrontendServer(fs::path root_folder) {
     _frontend_found = IsValidFrontendFolder(root_folder);
 
     if (_frontend_found) {
-        fmt::print("Serving CARTA frontend from {}\n", fs::canonical(_http_root_folder).string());
+        spdlog::info("Serving CARTA frontend from {}", fs::canonical(_http_root_folder).string());
     } else {
-        fmt::print("Could not find CARTA frontend files in directory {}.\n", _http_root_folder.string());
+        spdlog::error("Could not find CARTA frontend files in directory {}.", _http_root_folder.string());
```
confluence: This should be a warning.

confluence (commented):
I have made some suggestions for fixing inconsistent loglevels. I'm not sure what should be done about errors in subcomponents like regions, moments or tables.

src/Frame.cc
veggiesaurus: I agree. Not sure why they were only displayed under the `verbose` flag before

src/Frame.cc
veggiesaurus: :+1:

src/Frame.cc
veggiesaurus: actually just checked, this should be an error too, since it prevents the image from being loaded entirely (not just being loaded in pixel coords)

src/Frame.cc
veggiesaurus: :+1:

src/Frame.cc
veggiesaurus: :+1:

src/Frame.cc
veggiesaurus: :+1:

src/ImageData/Hdf5Loader.cc
veggiesaurus: :+1:

src/Main.cc
veggiesaurus: :+1:

src/Moment/MomentGenerator.cc
veggiesaurus: :+1:

src/OnMessageTask.cc
veggiesaurus: :+1:

src/SimpleFrontendServer/SimpleFrontendServer.cc
veggiesaurus: :+1:

veggiesaurus (approved):

src/Frame.cc
```diff
@@ -72,9 +66,7 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
     std::string log_message;
     if (!_loader->FindCoordinateAxes(_image_shape, _spectral_axis, _stokes_axis, log_message)) {
         _open_image_error = fmt::format("Problem determining file shape: {}", log_message);
-        if (_verbose) {
-            carta::Log(session_id, _open_image_error);
-        }
+        spdlog::error("Session {}: {}", session_id, _open_image_error);
```
confluence, line 69: This one should be a warning.

veggiesaurus: actually I think it should be an error, because it's returning straight after that, so the image will not be loaded successfully.

confluence (approved):
This is fine. I think that we should revisit the loglevels used in some of the components, but that will require a detailed analysis of the possible failure states, and is outside the scope of this PR.

Now that we have a more standardised logging system, this should be much easier to fix. :)

src/Frame.cc
confluence: Oh, good point. In that case this is fine.

PR #719, opened 2021-02-08
Add an install target so that the executable is copied correctly.

confluence:
This is needed for the executable to be installed correctly by Debian's packaging tools.


ajm-ska (approved):
I know this is for the Debian packaging. I just tried it with a normal 'make' build and the old desktop packaging process on Linux. It appears to have no affect on those other use cases. So it seems to be OK.

veggiesaurus:
@confluence do you think we need to add anything to the readme? I guess this is a pretty standard installation approach, so we can assume people will know what to do

confluence:
It should only have an effect on `make install` (which we seldom do). I don't think explicit documentation is required; installation into a system `bin` directory is the expected default behaviour.

veggiesaurus (approved):

PR #721, opened 2021-02-09
Added workarounds for determining the executable path.

### LINKED ISSUE #720: Executable path is not calculated correctly
`argv[0]` is not a reliable way to find the full path to the executable; it fails if the executable is in the path and is being called by a bare name (e.g. the way a desktop user will run it after installing the Debian package). This breaks finding the default location of the frontend to serve.

Coincidentally, the logger uses the current path as the path to the executable when it prints the executable location on startup.

There are ways to find the correct path both on Linux and on MacOS; for example the method documented here: https://stackoverflow.com/questions/22675457/what-is-the-equivalent-of-proc-self-exe-on-macintosh-os-x-mavericks

confluence:
Fixes #720. Needs further testing, especially on MacOS.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -17,6 +17,10 @@
 #include <tbb/task_scheduler_init.h>
 #include <uuid/uuid.h>
 
+#ifdef __APPLE__
+#include <mach-o/dyld.h>
+#endif
+
 #include <casacore/casa/Inputs/Input.h>
 #include <casacore/casa/OS/HostInfo.h>
 
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -4,6 +4,7 @@
    SPDX-License-Identifier: GPL-3.0-or-later
 */
 
+#include <climits>
 #include <iostream>
 #include <thread>
 #include <tuple>
@@ -519,6 +520,27 @@ void ExitBackend(int s) {
     exit(0);
 }
 
+bool FindExecutablePath(std::string& path) {
+    char path_buffer[PATH_MAX + 1];
+#ifdef __APPLE__
+    uint32_t len = sizeof(path_buffer);
+
+    if (_NSGetExecutablePath(path_buffer, &len) != 0) {
+        return false;
+    }
+#else
+    const int len = int(readlink("/proc/self/exe", path_buffer, PATH_MAX));
+
+    if (len == -1) {            
+        return false;
+    }
+        
+    path_buffer[len] = 0;
+#endif
+    path = path_buffer;
+    return true;
+}
+
 // Entry point. Parses command line arguments and starts server listening
 int main(int argc, const char* argv[]) {
     try {
@@ -586,6 +608,16 @@ int main(int argc, const char* argv[]) {
 
             InitLogger(no_log, verbosity);
         }
+        
+        std::string executable_path;
+        bool have_executable_path(FindExecutablePath(executable_path));
+        
+        if (!have_executable_path) {
+            spdlog::warn("Could not determine the full path to the backend executable.");
+            executable_path = "carta_backend";
+        }
+
+        spdlog::info("{}: Version {}", executable_path, VERSION_ID);
 
         if (!CheckRootBaseFolders(root_folder, base_folder)) {
             return 1;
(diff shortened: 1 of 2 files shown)


src/Main.cc
```diff
@@ -519,6 +520,27 @@ void ExitBackend(int s) {
     exit(0);
 }
 
+bool FindExecutablePath(std::string& path) {
+    char path_buffer[PATH_MAX + 1];
+#ifdef __APPLE__
+    uint32_t len = sizeof(path_buffer);
+
+    if (_NSGetExecutablePath(path_buffer, &len) != 0) {
```
ajm-ska, line 532: To get this to work on MacOS, you need to add the following (between lines 14 and 15):

```
#ifdef __APPLE__
#include <mach-o/dyld.h>
#endif
```

ajm-ska:
After doing `make install` on MacOS, the carta_backend executable is put in `/usr/local/bin/carta_backend`. I can start I by typing `carta_backend` in the terminal from any folder. It automatically opens the frontend from the default location:

>  [info] Serving CARTA frontend from /usr/local/share/carta/frontend

I see it won't work when directly running the carta_backend executable in its build folder i.e. `./carta_backend`. In this case it only seems to check `../share/carta/frontend`, and then stops:
> [warning] Could not find CARTA frontend files in directory /Users/acdc/carta-backend/build/./../share/carta/frontend.

Although I see that is similar with the `dev` branch behavior:
> [warning] Could not find CARTA frontend files in directory ./../share/carta/frontend.

If I run the confluence/fix-executable-path branche's built executable directly from the home folder i.e. `carta-backend/build/carta_backend` I get:
> [warning] Could not find CARTA frontend files in directory /Users/acdc/carta-backend/build/../share/carta/frontend.

I know directly executing `./carta_backend` is only for developers and the developer can easy specify the frontend location e.g. `./carta_backend -frontend_folder /usr/local/share/carta/frontend` so maybe it is a non-issue.
But at least this test proves that it now knows the proper location of the carta_backend executable and so can find the frontend when the carta_backend and frontend are installed properly on MacOS.

ajm-ska:
In relation to that, maybe I missed some previous discussion, but I thought the carta_backend was going to search for ./../share/carta/frontend as the first default, and if not there, the second default would be the absolute path `/usr/local/share/carta/frontend`?

Also, it looks like a single log file goes to `~/.carta/log/carta.log`. I thought it was going to go to `/var/log/carta/` plus have a time stamp and maybe keep only the 10 most recent log files?

src/Main.cc
confluence: I will add this.

confluence:
No, there was never a fallback to `/usr/local/share/carta/frontend`. There is no reason to do that; we don't know if the frontend will be installed in `/usr/share` or `/usr/local/share` or in a custom install directory in the user's home directory, and if it's the correct version of the frontend if it isn't in the expected location relative to the backend executable. The intention was always to specify a custom directory when running an uninstalled development version.

We discussed the log file location in the logging PR. The backend logs only to the user directory because it's a user-run application. On a shared system, the CARTA controller will redirect backend logs from stdout to per-user files in `/var/log/carta`.

veggiesaurus:
> In relation to that, maybe I missed some previous discussion, but I thought the carta_backend was going to search for ./../share/carta/frontend as the first default, and if not there, the second default would be the absolute path `/usr/local/share/carta/frontend`?

since the first approach (`../share/carta/frontend`) works for all the major approaches (installing via package manager, installing via `make install`, running from a local folder after extracting a tarball), there's no need to check the second default. In fact, we thought it would be better not to have multiple checks, because it would inevitably lead to frontend/backend version mismatches.

> Also, it looks like a single log file goes to `~/.carta/log/carta.log`. I thought it was going to go to `/var/log/carta/` plus have a time stamp and maybe keep only the 10 most recent log files?

For a user application, it makes more sense to log to the user's directory. The individual log entries have timestamps themselves, and we use a rotating set of 5 log files (up to 5 MB each). See https://github.com/CARTAvis/carta-backend/pull/713 for the details

ajm-ska:
OK. Thanks for the explanations.

veggiesaurus (approved):

ajm-ska (approved):

PR #724, opened 2021-02-10
Migration to cxxopts and extract program settings into a struct

### LINKED ISSUE #722: Allow users to pass a folder or a file as a positional argument
We should allow the user to pass a path to the backend, which will result in it being appended to the generated URL. For example:
- `carta_backend file.fits` would generate a URL with `?file=/path/to/pwd/file.fits`
- `carta_backend folder_name` would generate a URL with `?folder=/path/to/pwd/folder`
- `carta_backend /abs_path/file.fits` would generate a URL with `?file=/abs_path/file.fits`
- `carta_backend /abs_path/folder_name` would generate a URL with `?folder=/abs_path/folder_name`

We will need to adapt the code @ajm-asiaa has previously used for determining whether a folder is actually a CASA/Miriad image in disguise.

Bonus feature (requires minor frontend change): allow users to specify multiple files to open, will be handled in another issue

### LINKED ISSUE #723: Migrate to different command-line argument parsing library
#722 can't be implemented until `casacore::Input` is replaced with an alternative that can handle positional arguments correctly. This will also allow us to use a more conventional format for command-line arguments. 

Recommended library is the header-only [cxxopts](https://github.com/jarro2783/cxxopts). We can add this as a submodule.

veggiesaurus:
This PR moves us from `casacore::Input` to the header-only [cxxopts](https://github.com/jarro2783/cxxopts) library (added as a submodule with tag `v2.2.1` checked out), and extracts all the settings to a `ProgramSettings` struct, which can be constructed from `argc` and `argv`.

This PR also adjusts some of the available command-line arguments (details available via `--help`):
- The `threads` argument has been deprecated, and will be ignored. Previously, `threads` was used for the number of TBB task threads, but this is now fixed. `omp_threads` is still used for controlling the number of OpenMP threads used during `#pragma omp` sections.
- The `root` and `base` arguments have been deprecated, and have been replaced with `top_level_folder` and positional arguments respectively. The arguments will still work, but will be overwritten.
- Positional arguments are added. If an image is passed as the last argument, the generated URL will now append the correct `file=` query parameter. If a directory (other than a CASA or Miriad image directory) is passed, this will instead set the starting folder. Multiple files can be passed as arguments, but frontend support for multiple file arguments is not yet there, so only the first file is added to the URL.
- The `root` and `base` naming convention has also been replaced in the codebase by `top_level_folder` and `starting_folder` respectively.

Closes #723, closes #722.


src/Main.cc
```diff
@@ -113,14 +102,21 @@ void OnUpgrade(uWS::HttpResponse<false>* http_response, uWS::HttpRequest* http_r
 
 // Called on connection. Creates session objects and assigns UUID to it
 void OnConnect(uWS::WebSocket<false, true>* ws) {
-    uint32_t session_id = static_cast<PerSocketData*>(ws->getUserData())->session_id;
-    string address = static_cast<PerSocketData*>(ws->getUserData())->address;
+    auto socket_data = static_cast<PerSocketData*>(ws->getUserData());
+    if (!socket_data) {
+        spdlog::error("Error handling WebSocket connection");
```
jolopezl: The error message could be a bit more explicit. I suggest adding something like "Error handling WebSocket connection: can't get address, socket data doesn't exist"

src/Main.cc
```diff
@@ -688,67 +621,71 @@ int main(int argc, const char* argv[]) {
...
                 }
-                app.listen(host, port, LIBUS_LISTEN_EXCLUSIVE_PORT, [&](auto* token) {
+                app.listen(settings.host, settings.port, LIBUS_LISTEN_EXCLUSIVE_PORT, [&](auto* token) {
                     if (token) {
                         port_ok = true;
                     } else {
-                        spdlog::warn("Port {} is already in use. Trying next port.", port);
-                        ++port;
+                        spdlog::warn("Port {} is already in use. Trying next port.", settings.port);
+                        ++settings.port;
```
jolopezl, line 639: Just personal taste: I'd instead use `auto port = settings.port`, then update `port` if it is already in use. I like to keep the initial settings unchanged. Of course, you don't need to follow this comment :).

src/SessionManager/ProgramSettings.cc
```diff
@@ -0,0 +1,126 @@
...
+        ("base", "[Deprecated] Set starting folder for data files", cxxopts::value<string>(), "<path>")
+        ("root", "[Deprecated] Use 'top_level_folder' instead", cxxopts::value<string>(), "<path>");
+    // clang-format on
+
+    options.positional_help("<file(s) or folder to open>");
+    options.parse_positional("files");
+    auto result = options.parse(argc, argv);
+
+    if (result.count("version")) {
+        cout << VERSION_ID << endl;
```
jolopezl, line 71: Use spdlog instead?

src/SessionManager/ProgramSettings.cc
```diff
@@ -0,0 +1,126 @@
...
+        version = true;
+        return;
+    } else if (result.count("help")) {
+        cout << options.help() << endl;
```
jolopezl, line 75: Same as above, use spdlog?

test/TestHistogram.cc
```diff
@@ -15,9 +15,10 @@
 #include "Timer/Timer.h"
 
 using namespace std;
-random_device rd;
-mt19937 mt(rd());
-uniform_real_distribution<float> float_random(0, 1.0f);
+
+random_device hist_rd;
+mt19937 hist_mt(hist_rd());
+uniform_real_distribution<float> hist_float_random(0, 1.0f);
```
jolopezl: I don't get the purpose of this change, just naming?

jolopezl (commented):
Could you please write a more explicit description and purpose for your pull request? I understand that we can reference issues, but it's also nice to read something on the fly without opening each ticket.

This PR is relatively large and can be split into two unrelated PR to have a better history. One for program options and another one for improving the tests seems to be ideal. I am wondering if we would prefer to keep the test data in another repository.

I added some minor comments that can be considered just as little suggestions.

src/Main.cc
veggiesaurus: Yes, that's probably a good idea

src/SessionManager/ProgramSettings.cc
veggiesaurus: This and the help message are the two places where I suspect we shouldn't use spdlog. They're called before the logger is initialised, so they'll just end up in stdout anyway. Also, version and help text generally follows a convention, and adding logging prefixes would clutter that.

test/TestHistogram.cc
veggiesaurus: Yes, just because there are now multiple random generators in the different source files of the newly combined test target. Although I suspect we should make better use of test fixtures for this.

veggiesaurus:
> Could you please write a more explicit description and purpose for your pull request? I understand that we can reference issues, but it's also nice to read something on the fly without opening each ticket.


Will do

> This PR is relatively large and can be split into two unrelated PR to have a better history. One for program options and another one for improving the tests seems to be ideal. 

Yes, that might make life easier.

> I am wondering if we would prefer to keep the test data in another repository.

It could go in a separate repo which we include as a submodule. I don't have a strong opinion on that. It's currently pretty small, but I foresee it growing as we add more tests.

src/SessionManager/ProgramSettings.cc
jolopezl: > This and the help message are the two places where I suspect we shouldn't use spdlog. They're called before the logger is initialised, so they'll just end up in stdout anyway. Also, version and help text generally follows a convention, and adding logging prefixes would clutter that.

Good point!

confluence:
We can seamlessly transition to a submodule at any point -- we should consider doing that if the data gets large, but the data *shouldn't* get large for unit tests -- they should not be performance tests, and they should not be slow.

I suggest that we don't try to optimize this pre-emptively -- unit tests are likely to change rapidly, especially in the near future as we add more of them, and having to keep the data submodule in sync would add friction to running and maintaining them. Unless we updated it automatically when we built or ran the tests, but I'm not sure if that's a good idea.

test/TestHistogram.cc
jolopezl: Ok, I get it. Still, I liked letting my development environment generate just the test of interest with `cmake -Dtest=ON -DtestTileEncoding=OFF -DtestTable=OFF -DtestTimer=OFF =DtestBlockSmooth=OFF -DtestHistogram=ON ..`

veggiesaurus: I think the friction of having a separate target for each test suite was part of the reason tests weren't getting written. For now, I guess you can just use the command-line args to run a specific test.

src/Main.cc
veggiesaurus: Have just looked at the code again... introducing a `port` variable might be more trouble than it's worth, because `settings.port` is used elsewhere. Changing that would require more refactoring, and I think this PR has already ballooned a bit too much.

test/TestHistogram.cc
veggiesaurus: I've added test fixtures, so the names are back to how they were before.

jolopezl (approved):

markccchiang (approved):

PR #727, opened 2021-02-12
Angus/725 unit test cleanup

### LINKED ISSUE #725: Unit test clean-up
The unit tests should all be one single CMake target, with multiple test suits, rather than a separate target for each test suite. This will greatly simplify test implementation going forward

veggiesaurus:
As requested by @jolopezl. I've split out the unit test changes in #724 into a separate PR.

This PR makes the following changes:
- adds test images and tables to a `data` subdir, which is copied to the build dir when building with `-Dtest=On`.
- Combines all unit test suites into a single target, rather than having separate targets for each suite. This was getting mucky and complicated in terms of building.
- Adds test fixtures instead of polluting the global namespace with functions and consts needed by tests.
- some minor refactoring of the "base" and "root" names (in preparation for the next PR... was tricky to separate this part out)
- moves the `FindExecutablePath` function to `Util.cc`
- extracts the sub-directory check to a separate function `IsSubdirectory` (and adds some unit tests for it, some of which are currently failing, and we will need to fix in a separate PR).

Closes #725


src/Table/TableController.cc
```diff
@@ -362,7 +362,7 @@ bool TableController::FilterParamsChanged(const std::vector<CARTA::FilterConfig>
 fs::path TableController::GetPath(std::string directory, std::string name) {
     fs::path file_path(_root_folder);
     if (directory == "$BASE") {
-        // Replace $BASE macro with the base folder
+        // Replace $BASE macro with the starting folder
```
confluence: Missed a name change here. :)

src/Util.cc
```diff
@@ -5,29 +5,45 @@
...
-    if (base == "root")
-        base = root;
-
-    // check root
+    // TODO: Migrate to std::filesystem
+    // check top level
     casacore::File root_folder(root);
     if (!(root_folder.exists() && root_folder.isDirectory(true) && root_folder.isReadable() && root_folder.isExecutable())) {
-        spdlog::critical("Invalid root directory, does not exist or is not a readable directory. Exiting carta.");
+        spdlog::critical("Invalid top level directory, does not exist or is not a readable directory. Exiting carta.");
```
confluence: And a few others in here.

confluence (commented):
This is looking good; there are just a few stray renames.

confluence (approved):
OK, I think this looks good now.

jolopezl (approved):
Looks great to me. I'm sorry for bothering you with the split. Things are clearer this way. I was happy with having a separate PR for the tests only.

I have spotted something minor that we can consider for changes. Several files are missing a new empty line at the end of the file. Sometimes formatting programs will complain; we can consider adding a new line to all files with a missing blank line before EOF in another opportunity.

confluence:
I was hoping to add an EOF newline check to our `clang-format` script, but it turns out that there's no option for it. We can fix this as part of #729. In the meantime, there's no reason we can't do a once-off pass to try to add the currently missing newlines.

PR #731, opened 2021-02-15
Add icon files, desktop file and carta executable wrapper

confluence:
I am adding these here and not in the Debian packaging repo because they can be reused with any installation method.

The desktop file is compatible with any Linux distribution that uses the Freedesktop.org (XDG) specification.

Is there any other functionality that should go in the `carta` wrapper? Right now it automatically determines the host to listen on if no `host` parameter is provided.

The icon sizes were picked through the highly unscientific method of checking which sizes were the most widely used by applications installed on my desktop. We can easily add other sizes.


confluence:
I *just* noticed a bug in the wrapper (which would probably never actually come up, but should still be fixed). I'll convert this back from draft in a moment.

static/icons/scalable/cartaviewer.svg
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
```
veggiesaurus, line 1: wonder if you should add `static/icons/** -linguist-generated` to the `.gitattributes` file, so that they don't show up as text diffs?

confluence: I think we need to handle the SVG icons and the PNG icons differently. The SVGs should be marked as binary, so that they are excluded from text diffs. The PNGs are already detected as binary, but I suppose that we can mark them as generated, since they are programmatically generated from the full-colour SVG. I will add this.

confluence: That did not do what I expected. :)

confluence: Marking the SVGs as binary didn't seem to have any effect on the PR diff, but marking them as generated does work, so I guess we can go with that (and the SVGs themselves aren't hand-coded either, so I guess this makes sense).

veggiesaurus: github seems to explicitly ignore the binary attribute in `gitattributes` and instead rely on `linguist` :thinking:

scripts/carta
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+import argparse
+import subprocess
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--host', default=None)
+    known_args, ignored_args = parser.parse_known_args()
```
veggiesaurus: neat trick :)

veggiesaurus (approved):

ajm-ska:
I'm just back from the holidays, so please bear with me as I catch up on things.

So far I'm having trouble running it on ubuntu 16.04 (I know 16.04 is almost EOL, but I'm testing on available desktop computers in order to test the icons).

Two problems with the  `carta` wrapper script are:

> File "/usr/local/bin/carta", line 16    print(f"Could not set default host IP: {result.stderr}", file = sys.stderr)                                                          ^SyntaxError: invalid syntax

and

> Traceback (most recent call last):  File "/usr/local/bin/carta", line 12, in <module>    result = subprocess.run(["hostname", "-I"], capture_output=True)  File "/usr/lib/python3.5/subprocess.py", line 693, in run    with Popen(*popenargs, **kwargs) as process:TypeError: __init__() got an unexpected keyword argument 'capture_output'

The system does have python3. Could the problem be that it is on older python3; 3.5.2? 

> python3 --version
> Python 3.5.2

Hopefully the Python3 requirement would not be problematic for RedHat7/CentOS7 as those systems don't have Python3 by default. It needs to be installed from an extra repository.

ajm-ska:
Also, to make the wrapper script work on MacOS, we need to add: 
`import sys`

Otherwise:

> Traceback (most recent call last):
>   File "/usr/local/bin/carta", line 17, in <module>
>     print(f"Could not set default host IP: {result.stderr}", file = sys.stderr)
> NameError: name 'sys' is not defined

confluence:
Yes, f-strings don't work on older Python3, and I didn't realise that `sys` was still needed because it's only used in the error output. `capture_output` may also be too recent. I will look at this.

Making this Python 2/3-agnostic may be far more difficult, but Python2 is EOL and I don't think we should be supporting it. This could be a Bash script, but it would make the argument parsing a lot more untidy.

confluence:
I'm considering making a bash script to use on legacy systems, which should solve most of the problems. But if you're getting an error in MacOS that suggests that the `hostname` command is failing. Can you confirm how you run this in MacOS in your current wrapper?

ajm-ska:
Then maybe we should add to the README that at least Python version 3.6(?) is a requirement?

confluence:
I don't think we should add that requirement just for the wrapper script; I will see if we can provide a fallback.

veggiesaurus:
> Yes, f-strings don't work on older Python3, and I didn't realise that `sys` was still needed because it's only used in the error output. `capture_output` may also be too recent. I will look at this.
> 
> Making this Python 2/3-agnostic may be far more difficult, but Python2 is EOL and I don't think we should be supporting it. This could be a Bash script, but it would make the argument parsing a lot more untidy.

what about just supporting older python3? I agree 100% that we shouldn't be supporting python2

confluence:
If we're going to add a fallback bash script anyway, we could use it anywhere that doesn't have the minimum Python3 version.

veggiesaurus:
> Hopefully the Python3 requirement would not be problematic for RedHat7/CentOS7 as those systems don't have Python3 by default. It needs to be installed from an extra repository.



> If we're going to add a fallback bash script anyway, we could use it anywhere that doesn't have the minimum Python3 version.

if there's a fallback bash script, why do we need a python3 version at all? :thinking:

confluence:
We could just make Python3 a requirement for the wrapper script, which is optional, but if _we're_ going to end up having to create an alternative solution for legacy systems without Python3, then we may as well add it here, and not modify the Python3 script at all (except to fix the sys bug, which I'm about to fix).

confluence:
@ajm could you please check what the issue is with `hostname` on MacOS?

ajm-ska:
> @ajm could you please check what the issue is with `hostname` on MacOS?

OK. I'll try to check soon.

confluence:
> if there's a fallback bash script, why do we need a python3 version at all?

Good question. I think that the python3 script will be easier to maintain in the long term, and if we explicitly use a fallback option we can retire it when we no longer support those legacy systems and just use the Python version (rather than using the bash version forever because we need it now). But I could go either way. I don't think it would be very difficult in cmake to check for a minimum Python version and use that to decide which script to install.

I have a bash version (almost) ready to go, but we need to resolve the MacOS hostname issue first.

confluence:
Also, what do we want to fall back to if the hostname command fails (which shouldn't happen)? Setting to a default like 0.0.0.0 or localhost, or not setting it at all?

confluence:
What is the original rationale for enforcing a single hostname by default in the wrapper? Apparently this is not needed for the Electron app, and I don't think it's needed for a shared host system (where the controller would call the backend executable with specific parameters). So why change the default behaviour of listening on all ports? If we don't need this, we can replace the whole wrapper with a symbolic link.

ajm-ska:
Another thing is, can we cleanly quit from the wrapper? When I press ctrl+c on MacOS I see 

> ^C[2021-02-17 17:03:57.746] [info] Exiting backend.
> Traceback (most recent call last):
>   File "/usr/local/bin/carta", line 22, in <module>
>     subprocess.run(args)
>   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 503, in run
>     stdout, stderr = process.communicate(input, timeout=timeout)
>   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1122, in communicate
>     self.wait()
>   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1185, in wait
>     return self._wait(timeout=timeout)
>   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1915, in _wait
>     (pid, sts) = self._try_wait(0)
>   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1873, in _try_wait
>     (pid, sts) = os.waitpid(self.pid, wait_flags)
> KeyboardInterrupt

Does it terminate cleanly on Linux?

On the old carta-remote version, I had it listening for "q" to quit (using bash):

```
### Wait for the user to enter 'q' before exiting
while true; do
    echo -en "Press q to close CARTA:"
    read input
    if [[ $input = "q" ]] || [[ $input = "Q" ]]
        then break
    else
       echo "" 
    fi
done

### Stop the remote carta_backend process
kill $this_carta_pid

### Check if there are any other carta_pids active from the user (excluding Electron desktop carta_pid)
all_carta_pid=$(ps aux | grep $USER | grep '[c]arta_backend' | grep -v 54140 | awk '{print $2}' | wc -w)

### Only stop all HTTP servers if there are no other remote carta_processes running
### (This allows for multiple remote processes to be run)
if [ $all_carta_pid -eq 0 ]; then
       kill $httpserver_pid
fi

```

confluence:
We can suppress that traceback if necessary.

veggiesaurus:
> Another thing is, can we cleanly quit from the wrapper? When I press ctrl+c on MacOS I see 
> 
> > ^C[2021-02-17 17:03:57.746] [info] Exiting backend.
> > Traceback (most recent call last):
> >   File "/usr/local/bin/carta", line 22, in <module>
> >     subprocess.run(args)
> >   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 503, in run
> >     stdout, stderr = process.communicate(input, timeout=timeout)
> >   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1122, in communicate
> >     self.wait()
> >   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1185, in wait
> >     return self._wait(timeout=timeout)
> >   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1915, in _wait
> >     (pid, sts) = self._try_wait(0)
> >   File "/usr/local/Cellar/python@3.9/3.9.1/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 1873, in _try_wait
> >     (pid, sts) = os.waitpid(self.pid, wait_flags)
> > KeyboardInterrupt
> 
> Does it terminate cleanly on Linux?
> 
> On the old carta-remote version, I had it listening for "q" to quit (using bash):
> 
> ```
> ### Wait for the user to enter 'q' before exiting
> while true; do
>     echo -en "Press q to close CARTA:"
>     read input
>     if [[ $input = "q" ]] || [[ $input = "Q" ]]
>         then break
>     else
>        echo "" 
>     fi
> done
> 
> ### Stop the remote carta_backend process
> kill $this_carta_pid
> 
> ### Check if there are any other carta_pids active from the user (excluding Electron desktop carta_pid)
> all_carta_pid=$(ps aux | grep $USER | grep '[c]arta_backend' | grep -v 54140 | awk '{print $2}' | wc -w)
> 
> ### Only stop all HTTP servers if there are no other remote carta_processes running
> ### (This allows for multiple remote processes to be run)
> if [ $all_carta_pid -eq 0 ]; then
>        kill $httpserver_pid
> fi
> 
> ```

I think using ctrl + C is fine (we should probably print that somewhere). But we i want to make sure the backend output is visible in the terminal, so just waiting for input probably won't work

ajm-ska:
The new bash script eliminates the messy termination.

Now I'm trying the direct file opening with the `carta` script. Should that be working?

I run `carta CARTA/Images/aJ.fits`
The log shows: 

> [2021-02-17 18:27:25.726] [info] CARTA is accessible at http://140.109.176.28:3002/?token=72CD32FB-B9E6-457E-831E-544C85B9AD7B&file=Users%2Facdc%2FCARTA%2FImages%2FaJ.fits

and the URL in the browser is the same. The image does not load.

Doing the full path `carta /Users/acdc/CARTA/Images/aJ.fits` gives the same result.

I need to manually add a forward slash before Users in order to make it work: i.e `&file=/Users`.

confluence:
The direct image does load for me (without a leading slash before `home`). This shouldn't have anything to do with the wrapper; it just passes the parameters directly to the backend. Check if you get the same behaviour when calling the executable directly.

I think that perhaps the backend should be printing the URL *without* the image, but that's a separate issue.

ajm-ska:
I see. I get the same behaviour when calling the executable directly. It is probably a MacOS-only bug that went untested in the previous pull request.

ajm-ska:
@confluence How do we test the CARTA icons?

confluence:
There isn't really anything to test; I install these in the debian package, and they can be reused in other packages (or by a user doing a source install, like the desktop file). I haven't added anything to the cmake configuration; we'd have to check exactly which system icons are stored where on which operating system, and this seems like excessive overhead.

ajm-ska (approved):
The new bash-based wrapper script seems to work perfectly on MacOS and Linux now. I admire the simplicity of $@ to accept all the carta_backend flags!
I imagine all MacOS and Linux systems have bash installed by default, even though the user may not use it in their shell.

veggiesaurus (approved):

PR #732, opened 2021-02-15
fixed two more missing arguments

veggiesaurus:
a couple of rarely-used command-line opts were also missing after #724. I've also added a couple more unit tests that were used to discover and debug this issue.


confluence (approved):
It looks like that's all of them.

PR #733, opened 2021-02-15
Delete the idle session

### LINKED ISSUE #660: handle idle connections
Currently, a user can keep a CARTA tab open indefinitely, which would keep the backend session alive. This becomes a problem when they leave it in this idle fashion for a long time with a number of files open, because the process is just sitting there taking up memory.

We should add configurable options to either kill the process after it is idle for a long time, or just kill the idle session. We should also make this idle timeout configurable. If a user's session gets killed, their frontend will immediately reconnect, but will not open files until the user clicks on the "resume" button. Once that happens, the user's files will be reloaded into memory.

This will help the archive and server case quite a bit.

markccchiang:
Closes #660. Setting the argument `--idle_session_timeout` allows the backend to delete the session after it idles for the number of seconds. By default, there is no idle session timeout mechanism.


src/Main.cc
```diff
@@ -470,7 +476,16 @@ void OnMessage(uWS::WebSocket<false, true>* ws, std::string_view sv_message, uWS
         }
     } else if (op_code == uWS::OpCode::TEXT) {
         if (sv_message == "PING") {
-            ws->send("PONG", uWS::OpCode::TEXT);
+            auto t_session = session->GetLastMessageTimestamp();
+            auto t_now = std::chrono::high_resolution_clock::now();
+            auto dt = std::chrono::duration_cast<std::chrono::seconds>(t_now - t_session);
+            if ((settings.idle_session_timeout >= 0) && (dt.count() >= settings.idle_session_timeout)) {
+                spdlog::warn("Session {} idles for {} seconds, delete it now!", session_id, dt.count());
```
veggiesaurus: I think warning "delete it now" sounds like an instruction to the user. 
I think we should change it to something like `"Deleting session {} due to idle timeout of {} seconds"` or `"Session {} has been idle for {} seconds. Deleting"`

src/Main.cc
```diff
@@ -51,6 +51,27 @@ struct PerSocketData {
     string address;
 };
 
+void DeleteSession(int session_id) {
+    Session* session = sessions[session_id];
+    if (session) {
+        auto uuid = session->GetId();
+        auto address = session->GetAddress();
+        session->DisconnectCalled();
+        spdlog::info("Client {} [{}] Disconnected. Remaining sessions: {}", uuid, address, Session::NumberOfSessions());
```
veggiesaurus: This function name and warning doesn't make sense. It's misleading, because you can call `DeleteSession` on timeout, which will be called _before_ manually disconnecting the client.

src/SessionManager/ProgramSettings.cc
```diff
@@ -54,7 +54,8 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
         ("frontend_folder", "Set folder to serve frontend files from", cxxopts::value<string>(), "<path>")
         ("exit_after", "Number of seconds to stay alive after last sessions exists", cxxopts::value<int>(), "<duration>")
         ("init_exit_after", "Number of seconds to stay alive at start if no clients connect", cxxopts::value<int>(), "<duration>")
-        ("files", "Files to load", cxxopts::value<vector<string>>(positional_arguments));
+        ("files", "Files to load", cxxopts::value<vector<string>>(positional_arguments))
+        ("idle_session_timeout", "Delete the session if it idles for the number of seconds", cxxopts::value<int>()->default_value(to_string(idle_session_timeout)), "<duration>");
```
veggiesaurus: I don't think printing a default of -1 adds any information for the user. What does -1 mean? I suggest we follow the same approach as `exit_after` `init_exit_after` (using `applyOptionalArgument`).

src/Main.cc
```diff
@@ -470,7 +476,16 @@ void OnMessage(uWS::WebSocket<false, true>* ws, std::string_view sv_message, uWS
         }
     } else if (op_code == uWS::OpCode::TEXT) {
         if (sv_message == "PING") {
-            ws->send("PONG", uWS::OpCode::TEXT);
+            auto t_session = session->GetLastMessageTimestamp();
+            auto t_now = std::chrono::high_resolution_clock::now();
+            auto dt = std::chrono::duration_cast<std::chrono::seconds>(t_now - t_session);
+            if ((settings.idle_session_timeout >= 0) && (dt.count() >= settings.idle_session_timeout)) {
```
veggiesaurus: I think we should use > 0. Using `0` will cause the session to be disconnected at every `PING` message

veggiesaurus (changes requested):
Some minor changes required. Please also update the `ProgramSettingsTest` unit test to include the new argument.

src/Main.cc
```diff
@@ -51,6 +51,27 @@ struct PerSocketData {
...
+        session->DeleteCalled();
+        spdlog::info("Client {} [{}] Deleted. Remaining sessions: {}", uuid, address, Session::NumberOfSessions());
+        if (carta_grpc_service) {
+            carta_grpc_service->RemoveSession(session);
+        }
+        if (!session->DecreaseRefCount()) {
+            delete session;
+            sessions.erase(session_id);
+        } else {
+            spdlog::warn("Session {} reference count is not 0 ({}) while on deleting!", session_id, session->DecreaseRefCount());
```
confluence: I'd replace "while on deleting" with "on deletion" here.

src/Main.cc
```diff
@@ -51,6 +51,27 @@ struct PerSocketData {
...
+        }
+    } else {
+        spdlog::warn("Session {} not found to delete!", session_id);
```
confluence: Maybe "Could not delete session {}: not found!"

src/Frame.cc
```diff
@@ -207,7 +207,7 @@ bool Frame::ChannelsChanged(int channel, int stokes) {
     return (channel != _channel_index || stokes != _stokes_index);
 }
 
-void Frame::DisconnectCalled() {
+void Frame::DeleteCalled() {
```
confluence: Why not `Close`? Or at least just `Delete`?

src/Main.cc
```diff
@@ -51,6 +51,27 @@ struct PerSocketData {
     string address;
 };
 
+void DeleteSession(int session_id) {
+    Session* session = sessions[session_id];
+    if (session) {
+        auto uuid = session->GetId();
+        auto address = session->GetAddress();
```
confluence: If the only reason to retrieve these properties is to log the message, why not log this message inside the Session function?

src/Region/Region.cc
```diff
@@ -118,7 +118,7 @@ bool Region::IsConnected() {
     return _connected;
 }
 
-void Region::DisconnectCalled() { // to interrupt the running jobs in the Region
+void Region::DeleteCalled() { // to interrupt the running jobs in the Region
```
confluence: Again, I would call this `Delete`.

src/Session.cc
```diff
@@ -117,10 +118,10 @@ void Session::SetInitExitTimeout(int secs) {
     alarm(1);
 }
 
-void Session::DisconnectCalled() {
+void Session::DeleteCalled() {
```
confluence: And I would also call this `Delete`.

confluence (changes requested):

veggiesaurus:
@markccchiang I agree with the refactoring, and agree that we shouldn't call it `DeleteCalled`, but I disagree with @confluence about calling it `Delete`. `delete` is a reserved word in C++, and specifically implies memory management. I think we should provide a better description. What about `WaitForTaskCancellation`?

Another question (and here I can't remember if it was me or @pford who originally established this approach), but why aren't we simply putting this code in the destructor? It seems like `DisconnectCalled` was only ever called when removing a frame entirely, anyway.

confluence:
I'm happy with `WaitForTaskCancellation`. I was mostly objecting to `*Called` because it makes no sense as a function name; it sounds like the name of a boolean flag.

And that's a good point about the destructor.

markccchiang:
Have changed the name of the `Delete` function to `WaitForTaskCancellation`. I also made some code changes. `Session id` and `client id` are different. The session id is always increased when reconnection, but the client id is not changed if we resume the session after the reconnection. When a session timeout, we can just close the WebSockets and the `OnDisconnect` function will be called to destruct the session. The session's WaitForTaskCancellation is applied to its destruction and resuming. So I moved the log message before where the function is called.

markccchiang:
> Another question (and here I can't remember if it was me or @pford who originally established this approach), but why aren't we simply putting this code in the destructor? It seems like `DisconnectCalled` was only ever called when removing a frame entirely, anyway.

I think @robsimmonds had a comment on why it is not a good idea to put this function in the destructor in #183.

src/SessionManager/ProgramSettings.cc
```diff
@@ -54,7 +54,8 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
         ("frontend_folder", "Set folder to serve frontend files from", cxxopts::value<string>(), "<path>")
         ("exit_after", "Number of seconds to stay alive after last sessions exists", cxxopts::value<int>(), "<duration>")
         ("init_exit_after", "Number of seconds to stay alive at start if no clients connect", cxxopts::value<int>(), "<duration>")
-        ("files", "Files to load", cxxopts::value<vector<string>>(positional_arguments));
+        ("files", "Files to load", cxxopts::value<vector<string>>(positional_arguments))
+        ("idle_session_timeout", "Number of seconds to stay alive after receiving the last message from the frontend", cxxopts::value<int>(), "<duration>");
```
confluence: The description and argument name should be edited for compatibility with Angus' option PR, but I guess which PR should edit this depends on which PR is going to be merged first.

confluence (approved):
I think this is OK to merge. I made a note that the option needs to be adjusted to fit the changes we're making in the option PR.

confluence (changes requested):
Since the option text PR has been merged, this needs to be updated -- I will do it now.

confluence (approved):
I'm happy for this to be merged.

veggiesaurus (approved):

PR #734, opened 2021-02-17
updating Dockerfiles and README with spdlog

### LINKED ISSUE #726: spdlog dependency is missing in dockerfiles
#713 introduced a dependency in `spdlog` that wasn't propagated to the Dockerfiles. I develop using a Ubuntu container so that I noticed something was missing. I just installed `libspdlog-dev` using `apt`. We should update those files to keep them in sync with the new changes.

ajm-ska:
Here I updated the Dockerfiles and README with spdlog.

Unfortunately on ubuntu 18.04, we need a newer version of spdlog, otherwise some header file called `rotating_file_sink.h` is missing in the package manager's spdlog version 0.16.3.
Then to build spdlog from source, it requires fmt 5.3.0. 
I think Ubuntu's 18.04 package manager only supplies fmt version 4.0.0. So fmt is needed to be build from source too.

### DIFF
--- a/Dockerfiles/Dockerfile-CentOS6
+++ b/Dockerfiles/Dockerfile-CentOS6
@@ -46,25 +46,6 @@ RUN \
   make -j 2 && make install && \
   cd /root && rm -rf CMake
 
-# Install newer fmt
-RUN \
-  source /etc/bashrc && \
-  cd /root && \
-  git clone https://github.com/fmtlib/fmt && \
-  mkdir -p fmt/build && cd fmt/build && \
-  cmake .. && make && make install && \
-  cd /root && rm -rf fmt
-
-# Install spdlog with external 'fmt' library support
-RUN \
-  source /etc/bashrc && \
-  cd /root && \
-  git clone https://github.com/gabime/spdlog && \
-  mkdir -p spdlog/build && cd spdlog/build && \
-  cmake .. -DSPDLOG_FMT_EXTERNAL=ON && \
-  make && make install && \
-  cd /root && rm -rf spdlog
-
 # Install a newer pugixml from source
 RUN \
   source /etc/bashrc && \
--- a/Dockerfiles/Dockerfile-CentOS6
+++ b/Dockerfiles/Dockerfile-CentOS6
@@ -55,6 +55,16 @@ RUN \
   cmake .. && make && make install && \
   cd /root && rm -rf fmt
 
+# Install spdlog with external 'fmt' library support
+RUN \
+  source /etc/bashrc && \
+  cd /root && \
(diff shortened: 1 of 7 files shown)


veggiesaurus:
I think we should hold off on this PR until @markccchiang has investigated making spdlog a submodule (should also allow us to remove the `libfmt-dev` dependency, as spdlog has it included

veggiesaurus:
we can close this now, since we're using spdlog as a submodule

veggiesaurus:
@ajm-asiaa perhaps you can instead use this PR to _remove_ the dependencies on `fmt`. This is now bundled with the `spdlog` submodule

ajm-ska:
@veggiesaurus I have removed everything `fmt` and `spdlog` related from the Dockerfiles. But don't merge yet. I'm testing the Dockerfiles and seem to be having problems with the build in some of them.
```c++: fatal error: Killed signal terminated program cc1plus```
Maybe my Docker doesn't have enough resources.

ajm-ska:
@veggiesaurus Do these Dockerfiles work reliably for you? Take `Dockerfile-ubuntu-20.04` for example.
For me `make -j 4` and `make -j 2`fail. Only `make` works, but as you can imagine, it is quite slow.
I never had these problems previously. My Docker has 6 cores assigned and 2.25GB.
Could it be that fmt and spdlog are not getting built in time before they are needed?

veggiesaurus:
> @veggiesaurus Do these Dockerfiles work reliably for you? Take `Dockerfile-ubuntu-20.04` for example.
> For me `make -j 4` and `make -j 2`fail. Only `make` works, but as you can imagine, it is quite slow.
> I never had these problems previously. My Docker has 6 cores assigned and 2.25GB.
> Could it be that fmt and spdlog are not getting built in time before they are needed?

What's the exact docker command you're using?

ajm-ska:
> 
> What's the exact docker command you're using?

For example, `docker build -t ubuntu2004 -f Dockerfile-ubuntu-20.04 .` But maybe I wasn't clear. The Dockerfiles build fine. The problem is building the carta_backend inside the resulting Docker container.

veggiesaurus:
I meant the command for running the docker container

ajm-ska:
> I meant the command for running the docker container

I was running `docker run -ti ubuntu2004` then doing git clone to get carta-backend repository.

Thanks for checking on your system. So it seems carta_backend build in Docker needs a lot of RAM now. Originally I had 2.25GB assigned to Docker and `make -j 8` kept failing. I have increased Docker resources to 8GB and now `make -j 8` works. Feel free to go ahead and merge it.

veggiesaurus (approved):

PR #735, opened 2021-02-17
adjusted debug_no_auth text

veggiesaurus:
Minor change to the command-line argument text, but I think we should use this PR to take a closer look at the help text in general. This will be user-facing from 2.0 on for a number of use cases (basically any time a user opens CARTA from the commandline), so we may want to incorparate some of @ajm-asiaa's text from the electron app's command-line options.

### DIFF
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -57,7 +57,7 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
         ("files", "Files to load", cxxopts::value<vector<string>>(positional_arguments));
 
     options.add_options("deprecated and debug")
-        ("debug_no_auth", "Accept all incoming WebSocket connections (insecure, use with caution!)", cxxopts::value<bool>())
+        ("debug_no_auth", "Accept all incoming WebSocket connections on the specified port (not secure, use with caution!)", cxxopts::value<bool>())
         ("threads", "[Deprecated] No longer supported", cxxopts::value<int>(), "<thread count>")
         ("base", "[Deprecated] Set starting folder for data files", cxxopts::value<string>(), "<path>")
         ("root", "[Deprecated] Use 'top_level_folder' instead", cxxopts::value<string>(), "<path>");
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -49,7 +49,7 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
         ("host", "Only listen on the specified interface (IP address or hostname). If this is unset, the backend listens on all available interfaces.", cxxopts::value<string>()->default_value(host), "<interface>")
         ("p,port", "Manually set the port on which to host frontend files and accept WebSocket connections. If this is unset, the backend will automatically find an unused port, starting from the default.", cxxopts::value<int>()->default_value(to_string(DEFAULT_SOCKET_PORT)), "<port>")
         ("g,grpc_port", "Set gRPC server port. If this is unset, the gRPC service is disabled.", cxxopts::value<int>(), "<port>")
-        ("t,omp_threads", "Manually set OpenMP thread pool count. By default this is automatically set to the number of cores detected by the backend.", cxxopts::value<int>(), "<threads>")
+        ("t,omp_threads", "Manually set OpenMP thread pool count. By default this is automatically set to the number of logical cores detected by the backend.", cxxopts::value<int>(), "<threads>")
         ("top_level_folder", "Set top-level folder for data files. Files outside of this directory will not be accessible.", cxxopts::value<string>()->default_value(top_level_folder), "<dir>")
         ("frontend_folder", "Set folder from which frontend files are served. By default a path relative to the location of the backend executable is used: '../share/carta/frontend'.", cxxopts::value<string>(), "<dir>")
         ("exit_after", "Number of seconds to stay alive after last session exits. By default the backend will not shut down automatically after all sessions exit.", cxxopts::value<int>(), "<sec>")
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -10,6 +10,8 @@
 
 #include <cxxopts.hpp>
 
+#include <fmt/format.h>
+
 #include <images/Images/ImageOpener.h>
 
 #ifdef _BOOST_FILESYSTEM_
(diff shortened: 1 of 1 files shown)


ajm-ska (approved):

confluence (approved):
I agree that we should use this opportunity to check the other descriptions. Is the source of the wrapper published somewhere?

veggiesaurus:
> I agree that we should use this opportunity to check the other descriptions. Is the source of the wrapper published somewhere?

Here's the Electron app output from `--help`:

```
usage: carta [] CARTA file browser will default to the current path.
             [<path>] CARTA file browser will default to the specified
                      path <path> e.g. carta ~/CARTA/Images
             [<image>] CARTA will directly open the image named <image>
                       e.g. carta aJ.fits or carta ~/CARTA/Images/aJ.fits
             [--folder=<path>] Optional: An alternative way to define the
                               default CARTA file browser path.
                               Note: Not for directly opening an image.
             [--help] View this help output
Remote mode flags
             [--remote] start CARTA in 'remote' mode. For accessing CARTA's
                        frontend through your webrowser rather than the standard 
                        Electron interface. A free websocket port and a frontend
                        port will be chosen automatically.
             [--port=<number>] Optional: Manually choose a websocket port for the
                               backend. CARTA will check if the port is available
                               and issue a warning if not. A typical value is
                               between 1025-65535.
             [--fport=<number>] Optional: Manually choose a frontend port for the
                                CARTA web interface. CARTA will check if the port
                                is available and issue a warning if not. A typical
                                value is between 1025-65535.
             [--grpc_port=<number>] Optional: Manually choose a port to activate 
                                    and use the scripting functionality.
                                    Note: This feature is currently in development.
Advanced usage flags
             [--root=<path>] Define the lowest path the file browser can
                             navigate to. e.g. carta --root /home/bob means the 
                             the file browser can not access anything in /home
                             Note: --root can not be set inside --folder.
             [--threads=<number>] Set the number of threads. It controls how many
                                  tasks CARTA handles simultaneosuly. The default
                                  value is set as 4
             [--omp_threads=<number>] Set the number of OpenMP threads. It controls
                                      how trivially parallelisable tasks are split
                                      by CARTA. The default value is the
                                      automatically detected number of cores on
                                      your system; usually 4 or 8 on a typcial
             [--size=<width>x<height>] Manually define the dimensions of CARTA's
                                       Electron window in pixels.
             [--disable-gpu] May help if running CARTA Desktop through a VNC server
                             and images are not rendering properly.
             [--debug] Open the DevTools in the Electron window.
```

The last three may still be required for the electron app itself :thinking:

confluence:
* In the backend help for the positional parameters, we currently have the positional arguments called `files`, described as `files to load`, and also `<file(s) or folder to open>` in the positional help, but opening multiple files is not implemented, so this is misleading. I think we should rename these until this is implemented. Possibly `files` should be `path`, since it's the path to either a file or a directory.
* I think we should also mention that the current directory is used by default, as the wrapper does.
* In general we should incorporate more information about what the default behaviour is if an option is not set ~(unless it's super obvious because the existing description is a negation, e.g. for `no_browser`)~. So we should do this for `host`, `port`, `grpc_port`, `omp_threads`, `top_level_folder`, `frontend_folder`, `exit_after`, `init_exit_after`.
* There's a typo in the `exit_after` description; `exists` should be `exits`.
* ~We should mention the locations of files associated with CARTA, like the log file location -- but I think this should be in the body of the help rather than a particular option description.~
* I changed my mind about the format; I'll try to be consistent and mention the defaults in their associated option descriptions.

ajm-ska:
@veggiesaurus Here is a screen shot of the current help output:
Screen Shot 2021-02-18 at 5 02 25 PM

Last night I was a bit confused. For some reason, my eyes did not easily notice `--frontend_folder`! I think it was because every other flag has a space between them. Plus some of the descriptions around that area have spaces between them too (I know the spaces between descriptions are due to the length of the line of the flag text). In that area, lack of full stops also makes it feel like it is all part of one description, particularly the sentence `Set folder to serve frontend files from`

I feel it is hard to spot the `--frontend_folder` flag. Perhaps adding full stops to all descriptions and/or expanding on the frontend folder description so that it covers two lines could make everything easier to read.

kswang1029:
I wonder if we can use colors to make it more readable?
eg
Screen Shot 2021-02-18 at 8 17 08 PM

veggiesaurus:
> I wonder if we can use colors to make it more readable?
> eg
> Screen Shot 2021-02-18 at 8 17 08 PM

I guess this also depends somewhat on the terminal being used. For example, on my terminal, it looks like this:
image

kswang1029:
Oh sorry I was referring to the help of the carta_backend executable

veggiesaurus:
> Oh sorry I was referring to the help of the carta_backend executable

Simple white text for `--help` is pretty standard. We could probably put together a more comprehensive `man carta` manpage?

I think the issue @ajm-asiaa spotted was due to the custom args (such as `<port number>` or `<duration>`). When I use shorter args, it seems to behave nicely:

```
CARTA Backend
Usage:
  CARTA [OPTION...] <file(s) or folder to open>

  -h, --help                    Print usage
  -v, --version                 Print version
      --verbosity <level>       Display verbose logging from level (0: off,
                                1: critical, 2: error, 3: warning, 4: info, 5:
                                debug, 6: trace) (default: 4)
      --no_log                  Do not output to a log file
      --no_http                 Disable CARTA frontend HTTP server
      --no_browser              Prevent the frontend from automatically
                                opening in the default browser on startup
      --host <interface>        Only listen on the specified interface (IP
                                address or hostname) (default: 0.0.0.0)
  -p, --port <port>             Manually set port on which to host frontend
                                files and accept WebSocket connections
  -g, --grpc_port <port>        Set grpc server port
  -t, --omp_threads <threads>   Manually set OpenMP thread pool count
      --top_level_folder <dir>  Set top-level folder for data files. Files
                                outside of this directory will not be accessible
      --frontend_folder <dir>   Set folder to serve frontend files from
      --exit_after <sec>        Number of seconds to stay alive after last
                                sessions exists
      --init_exit_after <sec>   Number of seconds to stay alive at start if
                                no clients connect

 deprecated and debug options:
      --debug_no_auth      Accept all incoming WebSocket connections
                           (insecure, use with caution!)
      --threads <threads>  [Deprecated] No longer supported
      --base <dir>         [Deprecated] Set starting folder for data files
      --root <dir>         [Deprecated] Use 'top_level_folder' instead
```

confluence:
I should add a proper man page to the Ubuntu package (man pages don't use colours, but they do use bolding and underlining. `--help` pages don't normally have formatting.

ajm-ska:
Could we also add a full stop to the end of each description sentence?

confluence:
I've pushed some suggested changes. I think that some of these may be *too* verbose now. There isn't currently a good place to mention that the current directory is used as the default starting directory; perhaps there should be a top-level description paragraph.

veggiesaurus:
> I've pushed some suggested changes. I think that some of these may be _too_ verbose now. There isn't currently a good place to mention that the current directory is used as the default starting directory; perhaps there should be a top-level description paragraph.

@confluence We need to use shorter `<args>` text to prevent weird wrapping. Can you make the following switch:
- `port number` -> `port`
- thread number -> `threads` or `count` or `num`
- `path` -> `dir`
- `duration` -> `sec`

I think we should say "logical processors" or "logical cores" instead of "cores", because a 4-core hyperthreaded CPU would be allocated 8 threads.

ajm-ska:
Should CARTA in the line after Usage be lowercase? 
Or maybe it is tricky because the user could be either running `carta_backend` or the `carta` wrapper?
Screen Shot 2021-02-18 at 9 02 02 PM

veggiesaurus:
> Should CARTA in the line after Usage be lowercase?
> Or maybe it is tricky because the user could be either running `carta_backend` or the `carta` wrapper?
> Screen Shot 2021-02-18 at 9 02 02 PM

I suggest we use `carta` rather than `carta_backend`

confluence:
Will do. Since a `carta` wrapper is now part of the source and installed automatically, I think it's fine to use `carta`.

ajm-ska:
> I've pushed some suggested changes. I think that some of these may be _too_ verbose now. There isn't currently a good place to mention that the current directory is used as the default starting directory; perhaps there should be a top-level description paragraph.

I guess the old way of explaining the current directory will be used default starting directory can't fit into the new style?
```
usage: carta [] CARTA file browser will default to the current path.
             [<path>] CARTA file browser will default to the specified
                      path <path> e.g. carta ~/CARTA/Images
```

confluence:
If we add top-level information, we can make the specific options less verbose, and also add information about the default starting directory. I'm looking at the library documentation.

confluence:
Sorry; mouse finger slipped!

confluence:
I also think we should put the full name of the program in the description -- I will try that now.

veggiesaurus:
> I'm looking at the library documentation.

I think you want to cahnge the `help_string` parameter of the `Options` constructor

veggiesaurus:
I'm somewhat in two minds about the full stops at the end of each entry. Firstly, it looks weird when we have a default value, because we then have `. (default=3002)`. Secondly, I'm yet to find another program that has full stops at the end of each of their help entries

confluence:
For the description? Yes. I'm still looking for a way to add arbitrary top-level text at the bottom. Possibly that's outside the scope of the library and we should be adding it separately.

confluence:
I shoehorned it into the positional parameter information.

confluence:
I don't think that cxxopts has an option for adding top-level body text; I will open an issue for this.

(We could print additional information ourselves after the help output, but we would need to wrap it ourselves. We wouldn't easily be able to insert information in the middle of the help text without re-parsing it.)

confluence:
> I'm yet to find another program that has full stops at the end of each of their help entries

Firefox. :)

(Also Chromium.)

I see your point. We could remove them, but then we should also not start the entries with capital letters. The problem is that *that* looks very weird when there are multiple sentences in the same description. I guess we could rewrite all of them to be single sentences with semicolons.

Most utilities have much more brief descriptions which do not extend into multiple sentences.

veggiesaurus:
> I don't think that cxxopts has an option for adding top-level body text; I will open an issue for this.
> 
> (We could print additional information ourselves after the help output, but we would need to wrap it ourselves. We wouldn't easily be able to insert information in the middle of the help text without re-parsing it.)

isn't that exactly what the help_string constructor does? e.g.

```
angus@angus-ryzen:~/repos/carta-backend/cmake-build-relwithdebinfo (angus/help_text_improvements *)$ ./carta_backend --help
CARTA adopts a client-server architecture which is suitable for visualizing
images with large file sizes (GB to TB) easily obtained from ALMA, VLA, or SKA
pathfinder observations. It is impractical to process such a huge file with a
personal computer or laptop. In a client-server architecture, computation and
data storage are handled by remote enterprise-class servers or clusters with
high performance storage, while processed products are sent to clients only
for visualization with modern web features, such as GPU-accelerated rendering.
This architecture also enables users to interact with the ALMA and VLA science
archives by using CARTA as an interface.
Usage:
  carta [OPTION...] <file or folder to open>

  -h, --help                    Print usage.
  -v, --version                 Print version.
      --verbosity <level>       Display verbose logging from this level (0:
                                off, 1: critical, 2: error, 3: warning, 4:
                                info, 5: debug, 6: trace). (default: 4)
      --no_log                  Do not log output to a log file. By default
                                output is logged both to the terminal and to a
                                log file, '~/.carta/log/carta.log'.
      --no_http                 Disable frontend HTTP server. By default the
                                backend serves the frontend files.
      --no_browser              Don't open the frontend URL in a browser on
                                startup. By default the backend prints the
                                frontend URL in the terminal and also opens it in
                                the default browser.
      --host <interface>        Only listen on the specified interface (IP
                                address or hostname). If this is unset, the
                                backend listens on all available interfaces.
                                (default: 0.0.0.0)
  -p, --port <port>             Manually set the port on which to host
                                frontend files and accept WebSocket connections. If
                                this is unset, the backend will automatically
                                find an unused port, starting from the
                                default. (default: 3002)
  -g, --grpc_port <port>        Set gRPC server port. If this is unset, the
                                gRPC service is disabled.
  -t, --omp_threads <threads>   Manually set OpenMP thread pool count. By
                                default this is automatically set to the number
                                of cores detected by the backend.
      --top_level_folder <dir>  Set top-level folder for data files. Files
                                outside of this directory will not be
                                accessible. (default: /)
      --frontend_folder <dir>   Set folder from which frontend files are
                                served. By default a path relative to the location
                                of the backend executable is used:
                                '../share/carta/frontend'.
      --exit_after <sec>        Number of seconds to stay alive after last
                                session exits. By default the backend will not
                                shut down automatically after all sessions
                                exit.
      --init_exit_after <sec>   Number of seconds to stay alive at start if
                                no clients connect. By default the backend will
                                not shut down automatically if no clients
                                connect.

 Deprecated and debug options:
      --debug_no_auth      Accept all incoming WebSocket connections on the
                           specified port (not secure; use with caution!)
      --threads <threads>  [Deprecated] No longer supported.
      --base <dir>         [Deprecated] Set starting folder for data files.
                           Use the positional parameter instead.
      --root <dir>         [Deprecated] Use 'top_level_folder' instead.

```

confluence:
No, that's in the wrong place -- that should be used for a brief name or description. I don't think it's a good place for a larger body of text because it's above the usage.

Other utilities sometimes insert brief extra text between the usage and the options list, or longer extra text at the end, after the options list.

confluence:
The usual location is at the end; I suppose that we could easily do this ourselves if we wrap our extra text to the same width as the generated help.

veggiesaurus:
> The usual location is at the end; I suppose that we could easily do this ourselves if we wrap our extra text to the same width as the generated help.

I see what you mean. Yes, let's do that. Just need to manually wrap our strings to 80 characters

confluence:
To be clear, are you suggesting wrapping the literal strings in the source? I was considering this or a word wrap function, and writing a word wrap function just for this seems like overkill. Although we could probably modify the one used by cxxopts.

confluence:
I've set up a literal wrapped string in a way that should be easy to maintain (a literal multiline string which can easily be wrapped by editors). I'm going to try to move most of the information about default behaviour into this extra text, which may make the option descriptions much shorter.

src/SessionManager/ProgramSettings.cc
```diff
@@ -38,41 +40,73 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
...
+
+The gRPC service is disabled unless a gRPC port is set. By default the number of 
+OpenMP threads is automatically set to the detected number of logical cores.
+
+Logs are written both to the terminal and to a log file, '.carta/log/carta.log' 
+in the user's home directory. Possible log levels are:{}
+
+Options are provided to shut the backend down automatically if it is idle.
+)",
+        DEFAULT_SOCKET_PORT, log_levels);
```
veggiesaurus, line 102: isn't that list of log levels going to break your wrapping?

confluence: No, they're all very short lines. It's in a separate variable so that the wrapping doesn't break it.

veggiesaurus:
thanks for the changes @confluence. Looks good. I'm merging them in.

PR #736, opened 2021-02-17
fixed build issues when using boost fs

veggiesaurus:
Minor PR to fix build issues when using boost::filesystem on MacOS


ajm-ska (approved):
Thank you @veggiesaurus. This seems to build and run fine now on MacOS 10.14 with AppleClang 11.0 now using`cmake .. -DUseBoostFilesystem=True`
It also still works without boost on MacOS 11.0 for example.

Issue #680, opened 2020-12-23
performance degradation of cube histogram calculations from v1.3.1 to v1.4


kswang1029:
With same test image and same machine (mac), v1.3.1 takes 57s and v1.4 takes 80s. 

In addition, in v1.4, the progress updating cadence is still a bit too high (1s). Maybe every 2s/5s is better (need some tests).

I guess this is related to https://github.com/CARTAvis/carta-backend/issues/642

kswang1029:
update: the progress update cadence is addressed in #681

kswang1029:
I tried different progress update cadences as 1s, 2s, and 10s, and don't see any changes in the performance. So I guess we can probably rule out the possibility due to the mechanism of progress update.

veggiesaurus:
@kswang1029 any chance this is due to having a different number of threads active?

kswang1029:
with same data and same laptop and same OMP setup (all cores), we see different memory footprints. Left is with v1.3.1 and right is v1.4.0. v1.3.1 is faster in computing the histogram overall!
![Screen Shot 2021-01-21 at 7 06 24 PM](https://user-images.githubusercontent.com/20819712/105343226-4e740c80-5c1c-11eb-9e79-a8803e4edcd5.png)

kswang1029:
@acdo2002 could you test the cube histogram performance using v1.3.1 (electron app), v1.4 (electron app) and `angus/use_histogram_add` built in release mode with a linux box? Please make sure there is no cache affecting the test.

acdo2002:
@kswang1029 I have the backend `v1.4` and `angus/use_histogram_add` on my Ubuntu laptop, does `v1.3.1` refer to the backend `v1.3` release?

veggiesaurus:
> @kswang1029 I have the backend `v1.4` and `angus/use_histogram_add` on my Ubuntu laptop, does `v1.3.1` refer to the backend `v1.3` release?

Yes, the 1.3.1 patch release only had frontend changes, so you can just `git checkout release/1.3` on the backend (and protobuf)

acdo2002:
I run the histogram per cube and per channel with `dev`, `v1.4`, & `v1.3` on Ubuntu linux.
All run with omp_threads/threads=4
per cube:
`dev`: 42.86+-1.1 sec
`1.4`: 44.5+-0.4 sec
`1.3`: 40.14+-0.5 sec
per channel:
`dev`: 4.8+-0.1 sec
[...]

PR #737, opened 2021-02-18
use Histogram::Add when building the cube histogram

veggiesaurus:
Minor PR to make use of @jolopezl's `Histogram::Add` function when calculating a cube histogram. 
@kswang1029 can you check that this seems to behave correctly? I've done a simple comparison between HDF5 and FITS cube histograms, seems to be fine. Might be good to also check if #680 is still an issue.

### DIFF
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -1178,23 +1178,21 @@ bool Session::CalculateCubeHistogram(int file_id, CARTA::RegionHistogramData& cu
                 float progress = 0.50;
                 CARTA::RegionHistogramData half_progress;
                 CreateCubeHistogramMessage(half_progress, file_id, stokes, progress);
-                auto message_histogram = half_progress.add_histograms();
+                half_progress.add_histograms();
                 SendFileEvent(file_id, CARTA::EventType::REGION_HISTOGRAM_DATA, request_id, half_progress);
 
                 // get histogram bins for each channel and accumulate bin counts in cube_bins
-                std::vector<int> cube_bins;
                 carta::Histogram chan_histogram; // histogram for each channel using cube stats
+                carta::Histogram cube_histogram;
                 for (size_t chan = 0; chan < num_channels; ++chan) {
                     if (!_frames.at(file_id)->CalculateHistogram(CUBE_REGION_ID, chan, stokes, num_bins, cube_stats, chan_histogram)) {
                         return calculated; // channel histogram failed
                     }
 
-                    // add channel bins to cube bins
                     if (chan == 0) {
-                        cube_bins = {chan_histogram.GetHistogramBins().begin(), chan_histogram.GetHistogramBins().end()};
-                    } else { // add chan histogram bins to cube histogram bins
-                        std::transform(chan_histogram.GetHistogramBins().begin(), chan_histogram.GetHistogramBins().end(),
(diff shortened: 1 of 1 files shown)


kswang1029:
@acdo2002 could you run ICD tests to see if histogram is calculated correctly?

acdo2002:
Sure will do soon.

acdo2002:
I check CentOS and MacOS with `angus/use_histogram_add` branch.
The ICD test of CubeHistogram either per cube or per channel(channel:116) are all identical(pick several sequence to compare) with previous [PR695](https://github.com/CARTAvis/carta-backend/pull/695) with 3 different --omp_threads:4,5,&8.

Thus I would say the histogram calculation is correct.

kswang1029:
@veggiesaurus regarding the performance, I tested again with v1.3.1, v1.4 and the local Mac build of this branch in release mode. The test image is 28G roughly 2k x 2k x 2k. Each test is with a clean OS reboot. The results are
v1.3.1: 60.73s
v1.4: 68.37s
local Mac build in release mode: 70.98.

veggiesaurus:
> @veggiesaurus regarding the performance, I tested again with v1.3.1, v1.4 and the local Mac build of this branch in release mode. The test image is 28G roughly 2k x 2k x 2k. Each test is with a clean OS reboot. The results are
> v1.3.1: 60.73s
> v1.4: 68.37s
> local Mac build in release mode: 70.98.

Ok, so it seems this is roughly on par with 1.4. still a mystery as to where the regression cropped up

kswang1029:
Is there a way to check the build is in which mode? What’s the expected performance difference between release and relwithdebinfo?

veggiesaurus:
> Is there a way to check the build is in which mode? What’s the expected performance difference between release and relwithdebinfo?

It should be reasonably small. I'd like to test on Linux and see if there's any difference.

I guess you can check between release and relwithdebinfo by looking at the binary size (relwithdebinfo is pretty big, around 100mb)

kswang1029:
Hmm... I just checked "carta_backend".
v1.3.1: 2.3 MB
v1.4: 10 MB
this branch in release: 25 MB. 
So I guess none of the above are built in relwithdebinfo. 

UPDATE: local build in relwithdebinfo is 27 MB (is that right?)
UPDATE2: local build in debug is 67 MB

veggiesaurus:
> Hmm... I just checked "carta_backend".
> v1.3.1: 2.3 MB
> v1.4: 10 MB
> this branch in release: 25 MB.
> So I guess none of the above are built in relwithdebinfo.

wonder why we've grown so much since 1.3 :thinking:

kswang1029:
> It should be reasonably small. I'd like to test on Linux and see if there's any difference.

@acdo2002 could you test the cube histogram performance using v1.3.1, v1.4 and this branch built in release mode with a linux box? Please make sure there is no cache affecting the test.

veggiesaurus:
> > It should be reasonably small. I'd like to test on Linux and see if there's any difference.
> 
> @acdo2002 could you test the cube histogram performance using v1.3.1, v1.4 and this branch built in release mode with a linux box? Please make sure there is no cache affecting the test.

I think we should continue this discussion in #680

confluence (approved):

PR #738, opened 2021-02-19
Create the third party folder

markccchiang:
Create the `third-party` folder and move `uWebSockets` and `cxxopts` modules in. Add a new module `spdlog` with the version `1.8.2`. Use the `fmt` lib that bundled with `spdlog` instead of an external one.


veggiesaurus:
can we remove the fmt requirement in the readme as well?

src/FileList/FileExtInfoLoader.cc
```diff
@@ -8,8 +8,10 @@
 
 #include "FileExtInfoLoader.h"
 
-#include <fmt/format.h>
-#include <fmt/ostream.h> // not needed here but *must* include before miriad.h
+#include <spdlog/spdlog.h>
```
confluence: Is this necessary? It looks like only the fmt headers are needed.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -8,8 +8,10 @@
...
+
+#include <spdlog/fmt/fmt.h>
+#include <spdlog/fmt/ostr.h> // not needed here but *must* include before miriad.h
```
confluence, line 14: Should we not be using the original fmtlib headers from `spdlog/fmt/bundled/format.h` and `spdlog/fmt/bundled/ostream.h`?

src/FileList/FileListHandler.cc
```diff
@@ -12,6 +12,7 @@
 #include <casacore/casa/OS/File.h>
 
 #include "FileInfoLoader.h"
+#include "Logger/Logger.h"
```
confluence, line 15: Is the intention here to replace the fmt include? See previous comment about the location of the fmtlib headers bundled with spdlog.

confluence: This also applies to the other include changes.

confluence (changes requested):
I don't have a strong opinion about whether we should be importing the bundled fmt headers directly or using the wrappers -- the wrappers aren't adding anything, but the path is shorter and cleaner.

I do think that we should be importing *only* the fmt header(s), and not the logger header, in places where we only want to use the fmt library and not the logger.

What may be preferable to either option is adding a symbolic link from `third-party/spdlog/fmt/bundled` to `third-party/fmt`, and then keeping the `fmt` imports exactly as they are. At least I think that should work, and it would better reflect what we're doing here.

src/FileList/FileListHandler.cc
confluence: In places like this I don't think we should be importing the logger if we only want to use fmt.

veggiesaurus:
> What may be preferable to either option is adding a symbolic link from `third-party/spdlog/fmt/bundled` to `third-party/fmt`, and then keeping the `fmt` imports exactly as they are. At least I think that should work, and it would better reflect what we're doing here.

why not just add `third-party/spdlog/fmt/bundled` to the include path?

confluence:
> why not just add `third-party/spdlog/fmt/bundled` to the include path?

Then we'd have to use paths relative to that path, so e.g. `format.h` or `ostream.h`, which would be less clear. The  symlink would allow us to use include paths consistent with global include paths.

I think the later include location would override the system location, so we should get the bundled version even if a system version is installed.

confluence:
> why not just add `third-party/spdlog/fmt/bundled` to the include path?

Then we'd have to use paths relative to that path, so e.g. `format.h` or `ostream.h`, which would be less clear. The  symlink would allow us to use include paths consistent with global include paths.

I think the later include location would override the system location, so we should get the bundled version even if a system version is installed.

confluence:
Mostly I don't think that we should be hardcoding a custom dependency location in include paths in the source code. This is the domain of the build configuration.

veggiesaurus:
@confluence I've cleaned up the includes a bit, please re-review

confluence (approved):
OK, I'm happy with these changes.

Issue #618, opened 2020-08-31
support exporting a subimage (backend)


kswang1029:
The feature is to be able to export a subimage from the original cube by specifying a rectangle region and a range of channels.

relavent frontend issue: https://github.com/CARTAvis/carta-frontend/issues/1175

Requirement [doc](https://docs.google.com/document/d/1Jpju5UUzyFD98-iTgXomtmmHlfbWSHHokEVPCiA2t3I/edit#)

zarda:
Using the FileLoader::GetSubImage() to chop image or cube, they could works individually. Trying to apply them together but there is a problem.
As we run a new function:
```
bool FileLoader::GetSubImage(
    const casacore::Slicer& slicer, const casacore::LattRegionHolder& region, casacore::SubImage<float>& sub_image) {
    ImageRef image = GetImage();
    if (!image) {
        return false;
    }

    auto temp_image = casacore::SubImage<float>(*image, region);
    sub_image = casacore::SubImage<float>(temp_image, slicer);  // <- stop here
    return true;
}
```
and got
`Error: Slicer error:infer: endResult>=shape`
, which means the given `region` needs to be reshaped.
Otherwise, we interchange the proceeding order:
```
bool FileLoader::GetSubImage(
    const casacore::Slicer& slicer, const casacore::LattRegionHolder& region, casacore::SubImage<float>& sub_image) {
    ImageRef image = GetImage();
    if (!image) {
        return false;
    }

    auto temp_image = casacore::SubImage<float>(*image, slice);
    sub_image = casacore::SubImage<float>(temp_image, region);  // <- stop here
    return true;
}
```
and got 
`Error: Exception: shape of lattice [640, 800, 10, 1] mismatches lattice shape in region [640, 800, 25, 1].`
, which seem to report similar error about channel size.

veggiesaurus:
@zarda perhaps check with @pford?

pford:
Check out RegionHandler::ApplyRegionToFile: 

1. Use Frame::GetImageRegion to apply the region definition to the image to get the region (limit x and y ranges, all channels/stokes)
2. Create a box using Frame::GetImageSlicer (all x/y, limit channels/stokes)
3. Get the intersection of (1) and (2) to create an ImageRegion with all limits (x, y, chan, stokes)
4. Apply the ImageRegion (3) to the image to get the SubImage as in Frame::GetRegionData or Frame::GetRegionStats

[...]

PR #739, opened 2021-02-19
Hj/#618 sub image

zarda:
Support the feature of exporting subimage: (co-work with frontend [implement](https://github.com/CARTAvis/carta-frontend/issues/1175)) in #618.

- Export original image
- Export subimage in a spectral range of original image
- Export subimage in a combo of stokes of original image
- Export a region of chopped original image
- Export a shared region of chopped original image
- Drop degenerated axes

[Reference](https://docs.google.com/document/d/1Jpju5UUzyFD98-iTgXomtmmHlfbWSHHokEVPCiA2t3I/edit?pli=1#)


zarda:
Will merge #678 to reduce possible conflicts.

src/Frame.cc
```diff
@@ -1385,57 +1379,257 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+            return;
         }
 
-        // Get a copy of the original pixel data
-        casacore::IPosition start(image->shape().size(), 0);
-        casacore::IPosition count(image->shape());
-        casacore::Slicer slice(start, count);
-        casacore::Array<casacore::Float> temp_array;
-        image->doGetSlice(temp_array, slice);
+        bool is_keep_degenerate = false;
```
confluence: Rename this to `keep_degenerate`.

src/Frame.cc
```diff
@@ -1385,57 +1379,257 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
-    if (output_file_type == CARTA::FileType::CASA) {
-        // Remove the old image file if it has a same file name
-        if (fs::exists(output_filename)) {
-            fs::remove_all(output_filename);
+    auto image = GetImage();
+
+    auto image_shape = image->shape();
+    if (image_shape.size() > 2) {
+        // Validate channels & stokes
+        ssize_t channels_max = _spectral_axis > -1 ? image_shape[_spectral_axis] : 1;
```
confluence: Why use `ssize_t` here? Can any of these values ever be negative?

~(I have also commented on the corresponding protocol buffer PR.)~ The protocol buffer types are fine the way they are; we use the same signed integer type for all integer values. But we only use `ssize_t` in two existing places in the backend code, and I'm not convinced that we need to be using it at all.

If not `size_t`, then maybe just `int`?

src/Frame.cc
```diff
@@ -1385,57 +1379,257 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+        ssize_t channels_start = 0;
+        ssize_t channels_stride = 1;
+        ssize_t channels_end = channels_max - 1;
+        if (save_file_msg.channels().size() > 0) {
+            channels_start = std::min<ssize_t>(std::max<ssize_t>(save_file_msg.channels(0), 0), channels_max - 1);
```
confluence: Is it necessary to use templated `std::min` and `std::max` here?

src/Frame.cc
```diff
@@ -1385,57 +1379,257 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+        std::unique_lock<std::mutex> ulock(_image_mutex); // Lock the image while saving the file
+        switch (output_file_type) {
+            case CARTA::FileType::CASA:
+                success = ExportCASAImage(*image, output_filename, message);
+                break;
+            case CARTA::FileType::FITS:
+                success = ExportFITSImage(*image, output_filename, message);
+                break;
+            default:
+                message = "No saving file action!";
```
confluence: This should be something like "Could not export file. Unknown file type", with the type appended.

src/Frame.cc
```diff
@@ -1385,57 +1379,257 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+        if (found != std::string::npos) {
+            message.replace(found, root_folder.size(), "");
         }
+    }
+
+    save_file_ack.set_success(success);
+    save_file_ack.set_message(message);
+}
+
+void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save_file_msg, CARTA::SaveFileAck& save_file_ack,
```
confluence: There's a lot of code repetition in this function which could be refactored.

src/Frame.h
```diff
@@ -149,6 +157,7 @@ class Frame {
 
     // Apply Region/Slicer to image (Frame manages image mutex) and get shape, data, or stats
     casacore::LCRegion* GetImageRegion(int file_id, std::shared_ptr<carta::Region> region);
+    casacore::LCRegion* GetImageRegion(int file_id, std::shared_ptr<carta::Region> region, casacore::IPosition shape);
```
confluence: This does not appear to be defined or used in `Frame.cc`.

confluence (changes requested):

src/Frame.cc
zarda: Fix

src/Frame.cc
zarda: Use int

src/Frame.cc
zarda: Correcte it

src/Frame.cc
zarda: Replaced

src/Frame.h
zarda: Removed it

src/Frame.cc
zarda: Refactored the repetition codes. The remaining code should adapt the PR #678, suppose to going to refactor code after merging it.

zarda: Refactored code

src/Frame.cc
```diff
@@ -1385,57 +1380,68 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+        std::unique_lock<std::mutex> ulock(_image_mutex); // Lock the image while saving the file
+        switch (output_file_type) {
+            case CARTA::FileType::CASA:
+                success = ExportCASAImage(*image, output_filename, message);
+                break;
+            case CARTA::FileType::FITS:
+                success = ExportFITSImage(*image, output_filename, message);
+                break;
+            default:
+                message = "Could not export file. Unknown file type";
```
confluence: We should print out the file type here. Because it's an enum value, we probably need to add a function or  just an `unordered_map` to `Util` to map it to a string name. This may come in useful in other messages.

zarda: Good idea. Will log a message "Export a CASA image" for example.

confluence: I meant that you should log the type of file when the file **could not** be exported:

    "Could not export file. Unknown file type {}", ...

It's OK also to log it when exporting was successful.

src/Util.h
```diff
@@ -116,4 +117,9 @@ struct PointXy {
     }
 };
 
+// Map for enmu CARTA:FileType to string
+static std::unordered_map<CARTA::FileType, string> ExportTypeToString{{CARTA::FileType::CASA, "CASA"}, {CARTA::FileType::CRTF, "CRTF"},
```
confluence: Please rename this to something like `FileTypeToString` or maybe `FileTypeString` -- it's being used in export-related code, but it should be named more generically.

zarda: Rename it

confluence (approved):
This looks good to me.

src/ImageData/FileLoader.h
```diff
@@ -166,6 +166,10 @@ class FileLoader {
     // SubImage
     bool GetSubImage(const casacore::Slicer& slicer, casacore::SubImage<float>& sub_image);
     bool GetSubImage(const casacore::LattRegionHolder& region, casacore::SubImage<float>& sub_image);
+    bool GetSubImage(const casacore::LattRegionHolder& region, casacore::SubImage<float>& sub_image, bool keep_degenerate);
```
pford: `keep_degenerate=true` is the default in AxesSpecifier, so the SubImage defaults are basically `bool keep_degenerate=true, bool preserve_axes_order=false`.  If you add these parameters with defaults to the FileLoader::GetSubImage functions you would get the same behavior without duplicate functions.  Then in the Frame::SaveFile set these parameters according to the SaveFile message fields.

src/Frame.cc
```diff
@@ -1448,3 +1457,182 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+void Frame::ValidateChannelStokes(std::vector<int>& channels, std::vector<int>& stokes, const CARTA::SaveFile& save_file_msg) {
+    auto image = GetImage();
+    auto image_shape = image->shape();
+
+    int channels_max = _spectral_axis > -1 ? image_shape[_spectral_axis] : 1;
+    int channels_start = 0;
+    int channels_stride = 1;
+    int channels_end = channels_max - 1;
+    int channels_length = channels_max;
+    if (save_file_msg.channels().size() > 0) {
```
pford, line 1594: Please fix `repeated` typos for channels and stokes in carta-protobuf messages.rst.txt

src/Frame.cc
```diff
@@ -1448,3 +1457,182 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+    // Ouput results
+    channels.push_back(channels_start);
+    channels.push_back(channels_end);
+    channels.push_back(channels_stride);
+    stokes.push_back(stokes_start);
+    stokes.push_back(stokes_end);
+    stokes.push_back(stokes_stride);
+}
+
+casacore::Slicer Frame::GetExportImageSlicer(const CARTA::SaveFile& save_file_msg, casacore::IPosition image_shape) {
```
pford, line 1625: I think all of these Slicer calculations could be simplified similar to Frame::GetImageSlicer by indexing into the IPositions.  Start with a start/end/stride for the entire image, then: if you have a region, modify start/end/stride [0] and [1]; if you have a _spectral_axis, modify start/end/stride [_spectral_axis] according to the channels vector; and if you have a _stokes_axis, modify start/end/stride [_stokes_axis] according to the stokes vector.  But we could refactor this later; I have to do a bit of refactoring for supporting pV images anyway.

pford (changes requested):
Typo fix and simplify FileLoader::GetSubImage

src/Frame.cc
zarda: Correct it.

src/Frame.cc
zarda: Got it

src/ImageData/FileLoader.h
zarda: Well spot. Simplify functions.

pford (changes requested):
Code changes look good; however, there are no comments.  Please explain the purpose and parameters for the new Frame methods, particularly if parameters are return values.  GetExportRegionSlicer is a bit complicated but has no explanation what is going on.

pford (approved):

kswang1029:
> [2021-03-15 14:02:17.917] [warning] Session 14: Image has no beam information.
[2021-03-15 14:02:42.173] [error] Cannot apply region to file 3: LCPolygon - entire polygon is outside the lattice
[2021-03-15 14:03:08.556] [info] Exported a CASA file '/Users/kswang/Desktop/carta_test_subimage/test2.image'.

should this be an error or a warning?🤔

kswang1029:
Tested the case that when animation playback is on or region spectral profile is streaming, saving image works properly and gets higher priority than those tasks. Those will be paused while saving is in progress and resumed once image is saved. 👍

zarda:
> > [2021-03-15 14:02:17.917] [warning] Session 14: Image has no beam information.
> > [2021-03-15 14:02:42.173] [error] Cannot apply region to file 3: LCPolygon - entire polygon is outside the lattice
> > [2021-03-15 14:03:08.556] [info] Exported a CASA file '/Users/kswang/Desktop/carta_test_subimage/test2.image'.
> 
> should this be an error or a warning?🤔

This error message is claimed from casacore, which tells an error. Though carta tells not proceed outside region.

confluence:
I think that should be a warning in our code.

zarda:
Does this denote the error message in `Region::GetAppliedPolygonRegion` should be changed to warning? Might be in case of a region from another matching image is not inside.

kswang1029:
There is an edge case (but might happen accidentally on matched images) that the requested subimage only contains one pixel (1x1) in the spatial domain. The backend can still generate it successfully but when loading it back the frontend will crash due to tile calculations. I think we should prevent users making a 1x1 subimage (either intentionally or accidentally). I also tested the case by requesting a 1xn subimage, both backend and the frontend work fine. So only the 1x1 case need to be dealt with.

veggiesaurus:
> There is an edge case (but might happen accidentally on matched images) that the requested subimage only contains one pixel (1x1) in the spatial domain. The backend can still generate it successfully but when loading it back the frontend will crash due to tile calculations. I think we should prevent users making a 1x1 subimage (either intentionally or accidentally). I also tested the case by requesting a 1xn subimage, both backend and the frontend work fine. So only the 1x1 case need to be dealt with. 

I don't think we should prevent users from generating them, we should rather fix it on the front-end, or change the compression type when the image is less than 4x4

kswang1029:
> > There is an edge case (but might happen accidentally on matched images) that the requested subimage only contains one pixel (1x1) in the spatial domain. The backend can still generate it successfully but when loading it back the frontend will crash due to tile calculations. I think we should prevent users making a 1x1 subimage (either intentionally or accidentally). I also tested the case by requesting a 1xn subimage, both backend and the frontend work fine. So only the 1x1 case need to be dealt with.
> 
> I don't think we should prevent users from generating them, we should rather fix it on the front-end, or change the compression type when the image is less than 4x4

that works too.

kswang1029 (approved):
Feature-wise this PR is ready 👍

kswang1029:
could you update the branch with the latest dev to incorporate the new carta-casacore? I will need the backend to have a final check of the frontend branch. thanks.

veggiesaurus (approved):

PR #743, opened 2021-02-23
Logging ICD messages

### LINKED ISSUE #475: [question] logging ICD message history in verbose mode

markccchiang:
Closes #475. Logging ICD messages when setting `--verbosity 6` (trace mode).


veggiesaurus (changes requested):
I think we need a separate log level for this. Performance logging shouldn't be conflated with ICD logging. Rather introduce a separate level (7?) And use that

src/SessionManager/ProgramSettings.cc
```diff
@@ -76,7 +76,8 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
  3   warning
  4   info
  5   debug
- 6   trace)");
+ 6   trace
```
confluence: Should we rename this to `performance` (just here, in the user-facing information)?

It's also apparently possible to create a custom log formatter to change the level name(s) in the logs, but I'm not sure if that's absolutely necessary.

Maybe `trace (performance)`?

src/SessionManager/ProgramSettings.cc
```diff
@@ -76,7 +76,8 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
...
+ 7   trace + ICD)");
```
confluence: Similarly, I think that we should rename this to something like `protocol` or `trace (protocol)`.

It's already implied that each level includes the levels above it.

(ICD is our internal term for a document.)

src/Logger/Logger.cc
```diff
@@ -111,17 +117,21 @@ std::unordered_map<CET, std::string> event_type_map = {{CET::EMPTY_EVENT, "EMPTY
...
 void LogReceivedEventType(const CARTA::EventType& event_type) {
-    if (event_type_map.count(event_type)) {
-        spdlog::trace("[<==] {}", event_type_map[event_type]);
-    } else {
-        spdlog::error("Unknown event type: {}!", event_type);
+    if (log_icd_messages) {
+        if (event_type_map.count(event_type)) {
+            spdlog::trace("[<==] {}", event_type_map[event_type]);
+        } else {
+            spdlog::error("Unknown event type: {}!", event_type);
```
confluence: I think we should always log these errors, and only log the non-error messages if `log_icd_messages` is set.

src/Logger/Logger.cc
```diff
@@ -111,17 +117,21 @@ std::unordered_map<CET, std::string> event_type_map = {{CET::EMPTY_EVENT, "EMPTY
...
 void LogSentEventType(const CARTA::EventType& event_type) {
-    if (event_type_map.count(event_type)) {
-        spdlog::trace("[==>] {}", event_type_map[event_type]);
-    } else {
-        spdlog::error("Unknown event type: {}!", event_type);
+    if (log_icd_messages) {
+        if (event_type_map.count(event_type)) {
+            spdlog::trace("[==>] {}", event_type_map[event_type]);
+        } else {
+            spdlog::error("Unknown event type: {}!", event_type);
```
confluence: As above.

confluence (changes requested):
I think we should change the level names in the help, and always log errors caused by unrecognised messages.

I was hoping that we could define custom levels in spdlog, but it doesn't look like it's possible, and there's a closed issue which suggests that it will never be implemented. So I think this custom handling is OK.

src/SessionManager/ProgramSettings.cc
veggiesaurus: each level implies all the previous levels. I suggest you just say "ICD". I also think we should rename "trace" to `performance`. We can use `#defines` to customise the `spdlog::trace` prefix as well

veggiesaurus: the one problem with that is that we're still calling `spdlog::trace` when logging performance. We can add a function alias though:
```
// Adds an alias to trace for consistency between function name and prefix
namespace spdlog {
constexpr auto performance = [](auto&& ...args) {
    return trace(std::forward<decltype(args)>(args)...);
};
```

src/SessionManager/ProgramSettings.cc
veggiesaurus: Agreed. See below for example of how to move from "trace" to "performance"

src/SessionManager/ProgramSettings.cc
veggiesaurus: I agree, protocol makes more sense than ICD

markccchiang: So should we rename all the `spdlog::trace` as `spdlog::performance` in the original code. And add a new customize `spdlog::icd` as well?

veggiesaurus: I think we should rename the existing `spdlog::trace`, yes. Not sure about adding a new ICD, since we only use it once.

confluence: This alias isn't reflected in the logged level names, right?

markccchiang: Another way is that we can create a new logger and customize its format. I guess.

veggiesaurus: we need to do that too. Something like 
```#define SPDLOG_LEVEL_NAMES {"performance", "debug", "info", "warning", "error", "critical", "off" }```

_before_ including `spdlog.h` would do it

File: src/SessionManager/ProgramSettings.cc [hunk 5]
```diff
@@ -76,7 +76,8 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
  3   warning
  4   info
  5   debug
- 6   trace)");
+ 6   performance
+ 7   ICD)");
```
confluence: Please change this to `protocol`.

veggiesaurus (approved):
I'm happy with these changes. @kswang1029 please test the output to see if we can merge and close #475

kswang1029:
@veggiesaurus will do

kswang1029:
> [2021-02-25 11:04:16.385] [performance] [==>] SPATIAL_PROFILE_DATA
[2021-02-25 11:04:16.468] [performance] [<==] SET_CURSOR
[2021-02-25 11:04:16.468] [performance] Fill spatial profile in 0.014 ms
[2021-02-25 11:04:16.468] [performance] [==>] SPATIAL_PROFILE_DATA
[2021-02-25 11:04:16.669] [performance] [<==] SET_CURSOR
[2021-02-25 11:04:16.669] [performance] Fill spatial profile in 0.013000000000000001 ms
[2021-02-25 11:04:16.669] [performance] [==>] SPATIAL_PROFILE_DATA
[2021-02-25 11:04:17.491] [performance] [<==] SET_SPECTRAL_REQUIREMENTS
[2021-02-25 11:04:17.612] [performance] [==>] SPECTRAL_PROFILE_DATA
[2021-02-25 11:04:17.612] [performance] Fill spectral profile in 120.863 ms
[2021-02-25 11:04:25.222] [info] Client 1 [127.0.0.1] Deleted. Remaining sessions: 1
[2021-02-25 11:04:25.222] [debug] 0x7fb2fe809a00 ~Session 0
[2021-02-25 11:04:25.222] [info] No remaining sessions.
[2021-02-25 11:04:26.575] [debug] 0x7fb2fd83ac00 ::Session (1)
[2021-02-25 11:04:26.575] [info] Session 2 [127.0.0.1] Connected. Num sessions: 1
[2021-02-25 11:04:26.647] [performance] [<==] REGISTER_VIEWER
[2021-02-25 11:04:26.647] [performance] [==>] REGISTER_VIEWER_ACK

I see with verbosity=7 protocol messages and performance info are all with [performance]. Is this expected? For protocol messages, shouldn't it be [protocol] (much easier for log post-processing when needed)?

veggiesaurus:
> I see with verbosity=7 protocol messages and performance info are all with [performance]. Is this expected? For protocol messages, shouldn't it be [protocol] (much easier for log post-processing when needed)?

I think this is one of the issues with spdlog, and adding another level would complicate matters quite a bit. I suggest we replace it with `[info]`, since its less misleading than `[performance]`. And then in terms of post-processing, you can just grep for `==>` or `<==`

@markccchiang I think you can use `spdlog::info` for the ICD messages, it's also already behind your `log_icd_messages` flag

confluence:
If we're going to do that, I'd really prefer to use `debug` -- I think that's much more suitable; this is far too verbose for `info`.

veggiesaurus:
> If we're going to do that, I'd really prefer to use `debug` -- I think that's much more suitable; this is far too verbose for `info`.

good point

veggiesaurus (approved):

veggiesaurus:
@kswang1029 are you happy with `[2021-02-25 11:04:26.647] [debug] [<==] REGISTER_VIEWER`

kswang1029:
Yes using debug is sensible 👍

src/Logger/Logger.cc
```diff
@@ -55,6 +57,10 @@ void InitLogger(bool no_log_file, int verbosity) {
         case 6:
             stdout_logger->set_level(spdlog::level::trace);
             break;
+        case 7:
+            stdout_logger->set_level(spdlog::level::trace);
```
confluence: This should be `debug` now, and we should switch 6 and 7 around.

src/SessionManager/ProgramSettings.cc
```diff
@@ -76,7 +76,8 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
  3   warning
  4   info
  5   debug
- 6   trace)");
+ 6   performance
+ 7   protocol)");
```
confluence: Swap 6 and 7, and I suggest changing "protocol" to something like "+ protocol messages", otherwise users will expect a level called protocol in the logs, which we don't have.

confluence (changes requested):

src/Logger/Logger.cc
veggiesaurus: I don't know about that actually. The ICD logging is the most verbose logging we'd want, so it should still be 7.

src/SessionManager/ProgramSettings.cc
veggiesaurus: I think the only way to make this consistent is to move it to a separate input argument `debug_protocol_messages` or something like that, and have that displayed using `spdlog::debug` with an additional prefix `[protocol]`

veggiesaurus: That would leave us with something like 
`[2021-02-25 11:04:26.647] [debug] [protocol] <== REGISTER_VIEWER`

confluence: I think that we should use separate flags for protocol messages *and* performance messages, and log performance messages to `info`, instead of using the `trace` level for them.

So

* `log_performance` -> performance logged to `info` with added `[performance]` prefix
* `log_protocol_messages` -> ICD messages logged to `debug` with added `[protocol]` prefix
* remove `trace` / `performance` level completely

src/Logger/Logger.cc
```diff
@@ -78,3 +84,25 @@ void InitLogger(bool no_log_file, int verbosity) {
...
+
+void LogReceivedEventType(const CARTA::EventType& event_type) {
+    auto event_name = CARTA::EventType_Name(CARTA::EventType(event_type));
+    if (!event_name.empty()) {
+        if (log_icd_messages) {
+            spdlog::debug("<== {}", event_name);
+        }
+    } else {
+        spdlog::error("Unknown event type: {}!", event_type);
+    }
```
confluence: Please disregard my previous remark about logging this even when we aren't logging ICD messages -- we already log an error for an unrecognised received message in `OnMessageTask`. And I don't think it makes sense to check for unrecognised outgoing messages. So we can put the `log_icd_messages` check at the top level, and also log these unknown event type messages to `debug` like the other protocol messages.

confluence (changes requested):

src/SessionManager/ProgramSettings.cc
veggiesaurus: I'm happy with that suggestion. However, I do think elegantly adding a `[performance]` prefix might be a little tricky, since we need to handle variadic arguments. 

@markccchiang please give this a try

src/Logger/Logger.cc
```diff
@@ -50,31 +103,46 @@ void InitLogger(bool no_log_file, int verbosity) {
...
-    // Set flush policy on severity
-    stdout_logger->flush_on(spdlog::level::err);
-
-    // Register the stdout logger
-    if (!spdlog::get(STDOUT_TAG)) {
-        spdlog::register_logger(stdout_logger);
+void LogReceivedEventType(const CARTA::EventType& event_type) {
+    auto event_name = CARTA::EventType_Name(CARTA::EventType(event_type));
+    if (!event_name.empty()) {
+        if (log_protocol_messages) {
```
confluence: Please put this check on the top level, and don't log anything if `log_protocol_messages` is unset.

src/Logger/Logger.cc
```diff
@@ -50,31 +103,46 @@ void InitLogger(bool no_log_file, int verbosity) {
...
+            spdlog::debug("[protocol] <== {}", event_name);
+        }
     } else {
-        spdlog::critical("Duplicate registration of the logger: {}!", STDOUT_TAG);
+        spdlog::debug("Unknown event type: {}!", event_type);
```
confluence: Please make this message format match the known event protocol message -- maybe something like `"[protocol] <== unknown event type: {}!"`

src/Logger/Logger.cc
```diff
@@ -50,31 +103,46 @@ void InitLogger(bool no_log_file, int verbosity) {
...
-    if (!no_log_file) {
-        spdlog::info("Writing to the log file: {}", log_fullname);
+void LogSentEventType(const CARTA::EventType& event_type) {
+    auto event_name = CARTA::EventType_Name(CARTA::EventType(event_type));
+    if (!event_name.empty()) {
+        if (log_protocol_messages) {
+            spdlog::debug("[protocol] ==> {}", event_name);
+        }
+    } else {
+        spdlog::debug("Unknown event type: {}!", event_type);
```
confluence: As above.

src/Logger/Logger.cc
```diff
@@ -50,31 +103,46 @@ void InitLogger(bool no_log_file, int verbosity) {
             stdout_logger->set_level(spdlog::level::warn);
             break;
         case 5:
```
confluence: Has 4 (info) disappeared?

In the discussion above we realised that the cleanest way to do this is with separate flags both for performance and for protocol logging (please see my comments below).

src/SessionManager/ProgramSettings.cc
```diff
@@ -75,8 +75,9 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
  2   error
  3   warning
  4   info
- 5   debug
- 6   trace)");
+ 5   info + performance messages
+ 6   debug
+ 7   debug + protocol messages)");
```
confluence: We should restore this to just `5 debug`, and use explicit optional flags called `log-performance` and `log-protocol-messages` to turn these options on.

If you add the flags, I will update the help messages.

confluence (changes requested):
I think this is almost done, but we decided that the approach of a single log level number wasn't working for the performance and protocol messages -- there should be optional boolean flags for both of these.

src/SessionManager/ProgramSettings.cc
markccchiang: So there are `6` verbosity levels, and we add `--log-performance` and `--log-protocol-messages` options?

veggiesaurus: Yes, that should work. Although we should be consistent with our use of _ in our arguments

confluence: Yes, that's right. But only 5 verbosity levels (excluding `off`) -- we currently aren't actually using `trace` for anything, so we should remove it completely.

confluence: My bad -- as @veggiesaurus points out, we use underscores in our option names, so we should use `log_performance` and `log_protocol_messages`.

markccchiang: OK. Will do it later.

src/Logger/Logger.cc
```diff
@@ -14,7 +14,11 @@ namespace fs = boost::filesystem;
 namespace fs = std::filesystem;
 #endif
 
-void InitLogger(bool no_log_file, int verbosity) {
+static bool log_protocol_messages(false);
+
+void InitLogger(bool no_log_file, int verbosity, bool log_performance, bool log_protocol_messages_) {
+    log_protocol_messages = log_protocol_messages_;
```
confluence, line 20: Can we use `this` here to avoid using a different parameter name?

src/Logger/Logger.cc
```diff
@@ -35,11 +39,11 @@ void InitLogger(bool no_log_file, int verbosity) {
...
     auto stdout_logger = std::make_shared<spdlog::logger>(STDOUT_TAG, std::begin(stdout_sinks), std::end(stdout_sinks));
 
+    // Set flush policy on severity
+    stdout_logger->flush_on(spdlog::level::err);
+
     // Set logger's level according to the verbosity number
     switch (verbosity) {
-        case 0:
-            stdout_logger->set_level(spdlog::level::off);
-            break;
```
confluence: This should still be an option -- please put it back.

src/SessionManager/ProgramSettings.cc
```diff
@@ -70,13 +72,11 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
     auto result = options.parse(argc, argv);
 
     std::string log_levels(R"(
- 0   off
```
confluence: See previous comment; please put this back.

src/Logger/Logger.cc
```diff
@@ -49,32 +53,78 @@ void InitLogger(bool no_log_file, int verbosity) {
...
             stdout_logger->set_level(spdlog::level::debug);
             break;
-        case 6:
-            stdout_logger->set_level(spdlog::level::trace);
-            break;
         default: {
             stdout_logger->set_level(spdlog::level::info);
             break;
         }
     }
```
confluence, line 70: Please also set the level of the performance logger to the same level as the default logger.

At the moment there is an inconsistency between the two flags: if log_protocol_messages is enabled, but the log level is < debug, no protocol messages are printed -- which is the behaviour I would expect, because the protocol messages are logged at the debug level. But because the performance messages are implemented as a separate logger, if log_performance is enabled they are always logged, even if the log level is < info, even though they are now info level messages. We should set the perf logger level to the same level, so that the verbosity flag always takes precedence.

confluence (changes requested):
This is good -- but the missing "off" level needs to come back, and the perf log level should match the default log level.

I have noticed that some casacore messages are printed even when our logs are silenced -- I will see if I can track these down. I'm not sure if we can suppress them easily; it would be nice, but it's not necessary for this PR.

src/Logger/Logger.cc
```diff
@@ -14,7 +14,11 @@ namespace fs = boost::filesystem;
 namespace fs = std::filesystem;
 #endif
 
-void InitLogger(bool no_log_file, int verbosity) {
+static bool log_protocol_messages(false);
+
+void InitLogger(bool no_log_file, int verbosity, bool log_performance, bool this_log_protocol_messages) {
+    log_protocol_messages = this_log_protocol_messages;
```
confluence: ~I meant that you could use the `this` pointer to refer to the member variable, and call the parameter the same thing.~

src/Logger/Logger.cc
```diff
@@ -49,32 +56,104 @@ void InitLogger(bool no_log_file, int verbosity) {
...
+                perf_logger->set_level(spdlog::level::info);
+                break;
+            case 5:
+                perf_logger->set_level(spdlog::level::debug);
+                break;
+            default: {
+                perf_logger->set_level(spdlog::level::info);
+                break;
+            }
+        }
```
confluence: Instead of replicating this, read the level from the default logger.

confluence (changes requested):

src/Logger/Logger.cc
markccchiang: But this syntax would result in a compile error. Or just create a Logger class?

confluence: Sorry; I didn't notice that the function is not a class member. The solution you had before (with the underscore at the end) was fine!

confluence (changes requested):

confluence (approved):
I have updated the help descriptions for the new flags. Everything now seems to be working as expected.

@veggiesaurus and @kswang1029 , are we *sure* that we want the performance messages to be logged at the `info` level? They're verbose enough that `debug` may be more appropriate. This would be a very small change.

kswang1029:
> I have updated the help descriptions for the new flags. Everything now seems to be working as expected.
> 
> @veggiesaurus and @kswang1029 , are we _sure_ that we want the performance messages to be logged at the `info` level? They're verbose enough that `debug` may be more appropriate. This would be a very small change.

debug sounds proper.

PR #746, opened 2021-02-25
Angus/745 preference layout endpoints

### LINKED ISSUE #745: Handle preference and layout REST API calls
The CARTA controller implements a number of REST endpoints for handling user preferences and layouts. In order to get around the issues with storing preferences in local storage (not easily user-accessible, non-standard preference location, settings don't persist when running on different cluster nodes or different ports), we should implement these same endpoints on the backend, using the existing HTTP server capabilities. 

The following endpoints are required:
- [x] `GET /api/database/preferences`: returns the current preferences in JSON format
- [x] `DELETE /api/database/preferences`: clears the preferences specified by `req.body.keys: string[]`
- [ ] `PUT /api/database/preferences`: merges the existing preferences object with the supplied object in `req.body.update`
- [ ] `GET /api/database/layouts`: returns the current layouts in JSON format
- [ ] `DELETE /api/database/layout`: clears the layout specifed by `req.body.layoutName: string`
- [ ] `PUT /api/database/preferences`: Inserts or replaces the layout specified by `req.body.layoutName` with the supplied object in `req.body.layout`

Authentication can be handled by the same auth key used in the websocket connection

veggiesaurus:
Closes #745 by implementing the REST endpoints specified in the issue description. Preferences will be stored in `~/.carta/config/preferences.json`, and layouts will be stored in individual files under the `~/.carta/config/layouts` folder, in the form `<layoutname>.json`. JSON files will have their schema URLs specified, which will provide a pleasant UX when editing the preference file with a sensible text editor that supports JSON. For example, editing the CARTA controller config:

https://user-images.githubusercontent.com/592504/109165748-9559a280-7784-11eb-9832-3e755d7136ef.mp4

@kswang1029 and I will work on improving the preferences schema with examples and descriptions. 

I realised that it would be wise to not always assume the user folder is `~/.carta`, for the following reasons:
1) The user may want to install a beta version of CARTA side-by-side, and those preferences should probably not overwrite their existing preferences.
2) Unit testing these functions requires writing user preferences. I'd rather not overwrite existing preferences with unit-test generated stuff. 

I've therefore added a bit of flexibility through the `CARTA_USER_FOLDER_PREFIX` define. When building tests, this gets set to `".carta-unit-tests"`, and when building the development builds, we can just run `cmake -DCARTA_USER_FOLDER_PREFIX=".carta-beta"`. This will also affect the log files once this PR is merged.


confluence:
When this is merged I will update the `carta-ng` package to use a custom user directory. (I've realised that it should also get a custom name in the desktop launcher.)

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -12,9 +12,25 @@
 #include "Logger/Logger.h"
 #include "MimeTypes.h"
 
+#ifndef CARTA_USER_FOLDER_PREFIX
+#define CARTA_USER_FOLDER_PREFIX ".carta"
+#endif
```
confluence: This is duplicated across multiple files -- should it go in `Constants.h`?

test/CMakeLists.txt
```diff
@@ -31,10 +33,10 @@ set(TEST_SOURCES
         TestTileEncoding.cc
         TestTimer.cc
         TestUtil.cc
+        TestRestApi
```
confluence: Missing a .cc?

veggiesaurus:
PR now ready for review. @kswang1029 can you also have a look on your local dev environment? It should now modify `~/carta/config/preferences.json` and `~/carta/config/layouts`.

Upgrading users' existing preferences will have to be done by the frontend, I'll make an issue for it once this PR has been merged

src/SimpleFrontendServer/SimpleFrontendServer.h
```diff
@@ -21,21 +22,51 @@ namespace fs = std::filesystem;
...
 public:
-    SimpleFrontendServer(fs::path root_folder);
+    SimpleFrontendServer(fs::path root_folder, std::string auth_token);
     bool CanServeFrontend() {
         return _frontend_found;
     }
-    void HandleRequest(uWS::HttpResponse<false>* res, uWS::HttpRequest* req);
+
+    void RegisterRoutes(uWS::App& app);
+    nlohmann::json GetExistingPreferences();
```
veggiesaurus, line 44: These methods are public because they're being tested, rather than the handlers (which would require mocking the `uWS::App`, `uWS::HttpRequest` and `uWS::HttpResponse` classes.

I'm open to alternative suggestions... I have tried adding `RestApiTest` as a friend class, but that didn't seem to work

veggiesaurus: One option is to do the following:

```
// Allows testing of protected methods in SimpleFrontendServer without polluting the original class
class TestSimpleFrontendServer : public carta::SimpleFrontendServer {
public:
    TestSimpleFrontendServer(fs::path root_folder, std::string auth_token):
        carta::SimpleFrontendServer(root_folder, auth_token)
    {

    }
    FRIEND_TEST(RestApiTest, UpdatePreferencesFromString);
    FRIEND_TEST(RestApiTest, EmptyStartingPrefs);
    FRIEND_TEST(RestApiTest, GetExistingPrefs);
    FRIEND_TEST(RestApiTest, DeletePrefsEmpty);
    FRIEND_TEST(RestApiTest, DeletePrefsInvalid);
    FRIEND_TEST(RestApiTest, DeletePrefsIgnoresInvalidKeys);
    FRIEND_TEST(RestApiTest, DeletePrefsHandlesMissingKeys);
    FRIEND_TEST(RestApiTest, DeletePrefsSingleKey);
    FRIEND_TEST(RestApiTest, DeletePrefsKeyList);
    FRIEND_TEST(RestApiTest, EmptyStartingLayouts);
    FRIEND_TEST(RestApiTest, GetExistingLayouts);
    FRIEND_TEST(RestApiTest, DeleteLayout);
    FRIEND_TEST(RestApiTest, DeleteLayoutEmpty);
    FRIEND_TEST(RestApiTest, DeleteLayoutInvalid);
    FRIEND_TEST(RestApiTest, DeleteLayoutIgnoresInvalidKeys);
    FRIEND_TEST(RestApiTest, DeleteLayoutMissingName);
    FRIEND_TEST(RestApiTest, SetLayout);
};
```

The methods can then be made `protected` instead of `public`

kswang1029:
> PR now ready for review. @kswang1029 can you also have a look on your local dev environment? It should now modify `~/carta/config/preferences.json` and `~/carta/config/layouts`.
> 
> Upgrading users' existing preferences will have to be done by the frontend, I'll make an issue for it once this PR has been merged

does this work with --no_http? I don't see the new folder "config" in .carta folder.

veggiesaurus:
> > PR now ready for review. @kswang1029 can you also have a look on your local dev environment? It should now modify `~/carta/config/preferences.json` and `~/carta/config/layouts`.
> > Upgrading users' existing preferences will have to be done by the frontend, I'll make an issue for it once this PR has been merged
> 
> does this work with --no_http? I don't see the new folder "config" in .carta folder.

No, the HTTP server is required for this. I'm not sure what the use case would be for supporting a REST API for preferences/layouts but not serving the frontend :thinking:

kswang1029:
if I manually edit a wrong layout name in preferences.json, the frontend will be initiated in an embarrassing state.
Screen Shot 2021-03-03 at 3 15 07 PM
Screen Shot 2021-03-03 at 3 15 18 PM
 

UPDATE: same issue if we manually delete the layout file but the same layout is used as the default layout in preferences.json.

kswang1029:
If a wrong color map is defined in preferences.json, CARTA uses a colormap that is not part of the default list.
Screen Shot 2021-03-03 at 3 19 38 PM

veggiesaurus:
> if I manually edit a wrong layout name in preferences.json, the frontend will be initiated in an embarrassing state.
> Screen Shot 2021-03-03 at 3 15 07 PM
> Screen Shot 2021-03-03 at 3 15 18 PM
> 
> UPDATE: same issue if we manually delete the layout file but the same layout is used as the default layout in preferences.json.

that seems out of the scope of this PR. I guess you could make a frontend issue for it though? The same should occur if you manually edit the `localStorage` incorrectly. I guess we should revert to the default layout if the specified layout doesn't exist?

veggiesaurus:
> If a wrong color map is defined in preferences.json, CARTA uses a colormap that is not part of the default list.
> Screen Shot 2021-03-03 at 3 19 38 PM

could also be handled in a frontend issue. I think it's just defaulting to the first colourmap

confluence:
If we encounter an incorrect preference, we could attempt to recover by doing something like 1) backing up the user's file, 2) writing out a fresh default file, and 3) notifying the user of the issue. I'm not sure if this would have to be an up-front check or if we can add this recovery procedure to anywhere in the frontend code that the preferences are used and already being checked for correctness. Not sure if that's within the scope of this PR, though -- it seems like more of a frontend thing.

veggiesaurus:
> If we encounter an incorrect preference, we could attempt to recover by doing something like 1) backing up the user's file, 2) writing out a fresh default file, and 3) notifying the user of the issue. I'm not sure if this would have to be an up-front check or if we can add this recovery procedure to anywhere in the frontend code that the preferences are used and already being checked for correctness. Not sure if that's within the scope of this PR, though -- it seems like more of a frontend thing.

yes, this might be more of a frontend thing. It can easily validate the received data against the relevant schema.

confluence:
Also not sure if we need to back up and overwrite the file ourselves -- if we just notify the user that the preferences file has an error (presumably because they edited it by hand and made a mistake), we can put the onus on them either to fix it or to move it out of the way themselves (at which point we should detect that there's no file and write a new one). I think that's standard procedure.

src/SimpleFrontendServer/SimpleFrontendServer.h
confluence: I think it's OK to add the `FRIEND_TEST`s for now -- if we can figure out a cleaner way later, we can change it.

kswang1029:
https://github.com/CARTAvis/carta-frontend/issues/1335 @veggiesaurus please help to edit the issue if that is not clear.

veggiesaurus:
> Also not sure if we need to back up and overwrite the file ourselves -- if we just notify the user that the preferences file has an error (presumably because they edited it by hand and made a mistake), we can put the onus on them either to fix it or to move it out of the way themselves (at which point we should detect that there's no file and write a new one). I think that's standard procedure.

I'd rather leave it up to them to fix. As long as we supply a decent schema file, this should be easy for users to do.

confluence:
Do we already write out a new file if no file is found?

veggiesaurus:
> Do we already write out a new file if no file is found?

yes, if you set preferences where there's no existing file, it creates a new one

kswang1029:
Tested the "migration" capability by using the same URL with different web browsers. Working nicely 👍

confluence (approved):
Looks good now.

veggiesaurus:
have confirmed this works fine with the carta-controller (which continues to manage preferences and layouts in a db, as before)

veggiesaurus:
just fixing some build issues with macos, then will merge this in

PR #748, opened 2021-02-26
fix carta/44 rectangle region export import bug 

kswang1029:
The root cause is the calculation of rectangle width and height from the four corners. A rectangle is actually treated as a four-point polygon with rotation considered. So the width and height calculations was not correct, resulting the issue.


kswang1029:
close for now and I will make a cleaner PR.

Issue #541, opened 2020-06-18
position offset seen in exported region


kswang1029:
sorry I did not catch this in the PR #539 

When we create a rectangle region in CARTA, we see
<img width="1177" alt="Screen Shot 2020-06-18 at 10 06 03 AM" src="https://user-images.githubusercontent.com/20819712/84969650-b248cb00-b14b-11ea-8e2e-a891c391152f.png">

Then we export this region as a casa region text file and import it back. We see
<img width="1175" alt="Screen Shot 2020-06-18 at 10 06 17 AM" src="https://user-images.githubusercontent.com/20819712/84969678-c1c81400-b14b-11ea-90ee-78de57713400.png">

The region file reads as

> #CRTFv0 CASA Region Text Format version 0
centerbox [[00:00:00.23094, +029.59.56.0000], [4.0647arcsec, 1.9307arcsec]] coord=J2000, linewidth=1, linestyle=-, symsize=1, symthick=1, color=green, font=Helvetica, fontsize=10, fontstyle=bold, usetex=false

The offset appears to come from rounding of the center position in pixel. It should be (2.7274940773070684, 1.8217011035674615), but when export-then-import, it becomes (1.9999679327011108, 1.000012755393982). I suspect it could be due to rounding (2.7274940773070684, 1.8217011035674615) to (2, 1) before converting to world coordinates. 

This only happens to rectangle region. Point and ellipse regions are fine. 
Same issue happens when exporting to ds9 region format. 

@pford could you have a look?

pford:
I noticed this while testing, and tracked it down to the world->pixel conversion of the center coordinates using casacore::DirectionCoordinate::toPixel which ultimately calls wcslib wcss2p.  Although the pixel coordinates returned are double values, they are not fractional pixels.

kswang1029:
Is this a casacore/wcslib bug or “feature”?

pford:
Possible solution is to use AST instead of WCSLIB for pixel/world conversion

veggiesaurus:
> Possible solution is to use AST instead of WCSLIB for pixel/world conversion

I think switching to AST might help quite a bit for a number of situations

kswang1029:
To highlight the issue a bit more I have done a controlled test.
Initially we make a set of regions with _easy-to-read_ coordinates via the GUI like the following.
[...]

PR #749, opened 2021-02-26
fix rectangle width and height calculations

### LINKED ISSUE #44: Rectangular regions change between export and import
When I export rectangular regions (default CRTF and World) and then import them onto the same image, they have changed size and shape. Regions that are not rotated appear to shift slightly in position. Rectangular regions that are rotated will change size and shape. I do not see this problem with elliptical or freeform regions.

I am using CARTA v1.4 on MacOS 10.15.7

kswang1029:
This PR fixes https://github.com/CARTAvis/carta/issues/44. The root cause is the calculations of width and height were valid only when rotation=0. If rotation!=0, we need to use corner coordinates to calculate distance properly.

### DIFF
--- a/src/Region/RegionImportExport.cc
+++ b/src/Region/RegionImportExport.cc
@@ -332,14 +332,19 @@ bool RegionImportExport::ConvertRecordToRectangle(
 
     double cx, cy, width, height;
     casacore::Double blc_x = x[0];
+    casacore::Double brc_x = x[1];
     casacore::Double trc_x = x[2];
+    casacore::Double tlc_x = x[3];
     casacore::Double blc_y = y[0];
+    casacore::Double brc_y = y[1];
     casacore::Double trc_y = y[2];
+    casacore::Double tlc_y = y[3];
+
     // Control points: center point, width/height
     cx = (blc_x + trc_x) / 2.0;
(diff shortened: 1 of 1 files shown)


kswang1029:
@acdo2002 please help to verify the fix. thanks.

acdo2002:
@kswang1029 
I checkout to your branch `ksw/fix_carta_44` and rebuild carta_backend (i.e. rm build/mkdir build/cd build/cmake ../make).
Although the offset is still there, but size(width & height) & rotation looks almost the same (difference ~ 3e-3%) once export->import again.
rect-test1
Region 1 & 2 are the original set (cyan color).
Region 3 & 4 are the export(region 1 & 2)->import result (yellow color)
Region 5 & 6 are the import again result (yellow color) => center, size & PA are consistent with Region 3 & 4

Another test:
I test 0-180 deg with 20 deg interval, they are exported and imported again, 
rect2

although they have an offset in the center, but the sizes are roughly same and PAs are same. They can be moved by cursor and overlap with its original region.
rect3

kswang1029:
Thanks! That one-pixel offset is another bug #541.

veggiesaurus (approved):

PR #750, opened 2021-02-26
Modify check for boolean attribute

### LINKED ISSUE #747: cannot read one line of header in hdf5
When we check the image "M17_SWex.hdf5" Header, we realize there is a line in fileInfoResponse.fileInfoExtended['0'].headerEntries shows
{ value: "1", entryType: 2, numericValue: 1 }.

We expect to see every attribute has its own name (like other lines in the Header):
{ name:"XXX", value: "1", entryType: 2, numericValue: 1 }

from carta header, you can see there is one attribute cannot render:
hdf5-header-name-missing

I check the result with executing `h5dump -H M17_SWex.hdf5`, it constrains the missing attribute could be 

>       ATTRIBUTE "EXTEND" {
>          DATATYPE  H5T_STD_U8LE
>          DATASPACE  SCALAR
>       }
(I match each attribute, only "DATE", "EXTEND", and "ORIGIN" has no match with carta Header. However only "EXTEND" has no string data inside the attribute, which is consistent to what we see on the missing attribute "1")

I put "M17_SWex.hdf5"  on google [drive ](https://drive.google.com/drive/folders/1qj53HcB_6KNjQcxWM_gKhwh0ybqNXmFI?usp=sharing). Or going to [carta-backend-ICD-test](https://github.com/CARTAvis/carta-backend-ICD-test) code:public/set_QA/M17_SWex.hdf5.zip, there also provides the image we tested (hdf5 converter version 0.1.13).

confluence:
Possibly fixes #747, which is caused by incorrect detection of boolean attributes in HDF5 files.

In addition to the symptoms described in the issue (a header entry in the file browser with no name and a value of 1), viewing affected files in the browser causes a casacore error to be logged to the console, for example:

    Keyword Error:  name = EXTEND of type  LONG  and value 1                                                                                                                     
            Keyword value has wrong data type.                                                                                   

Any boolean attribute is affected, including `SIMPLE` -- I'm assuming that the reason the `SIMPLE` attribute is always displayed correctly is that it is handled differently in casacore.

In `Hdf5Attributes.cc`, we currently check whether an integer attribute is a boolean by looking for a signed 4-byte integer. However, when I test this (on 64-bit Linux), these values are actually stored as unsigned 1-byte integers (displayed in `h5dump` as `H5T_STD_U8LE`).

@pford, can you recall how this check was originally tested -- on what operating system(s), and with what files?

~This is not documented online very well, but it appears that the underlying type is not consistent across all architectures, and this may affect both the converter (which maps booleans to the HDF5 integer type using the C++ API's `NATIVE_HBOOL`, which should be the same as `hbool_t`) and the backend (which tries to map them back). I have found reports that on some architectures `hbool_t` is a 4-byte _unsigned_ integer -- but possibly only on Windows, which we don't officially support.~

~The change in this PR checks specifically for an unsigned 1-byte integer, but if necessary I can make the check more permissive, and attempt to cover both options. Could we test this in as many different environments as possible? We may also need to test whether files produced by the converter in different environments are the same.~

Edit: it turns out that this is not that mysterious: at least one old version of the converter wrote booleans as signed 4-byte integers. Pam had an old test file; this caused the discrepancy. So there shouldn't be any complications to merging this.

We are planning to preserve the original FITS header in the HDF5 file; once that is implemented in the converter, we will be able to bypass this entire parsing round trip.

### DIFF
--- a/src/ImageData/Hdf5Attributes.cc
+++ b/src/ImageData/Hdf5Attributes.cc
@@ -72,10 +72,7 @@ std::string Hdf5Attributes::ReadScalar(hid_t attr_id, hid_t data_type_id, const 
             casacore::Double value;
             casacore::HDF5DataType data_type((casacore::Double*)0);
             H5Aread(attr_id, data_type.getHidMem(), &value);
-            std::ostringstream ostream;
-            ostream.precision(13);
-            ostream << value;
-            std::string key_value = fmt::format("{:<8}= {}", name, ostream.str());
+            std::string key_value = fmt::format("{:<8}= {:#.13G}", name, value);
             return fmt::format("{:<80}", key_value);
         } break;
         case H5T_STRING: {
--- a/src/ImageData/Hdf5Attributes.cc
+++ b/src/ImageData/Hdf5Attributes.cc
@@ -53,7 +53,7 @@ std::string Hdf5Attributes::ReadScalar(hid_t attr_id, hid_t data_type_id, const 
     int sz = H5Tget_size(data_type_id);
     switch (H5Tget_class(data_type_id)) {
         case H5T_INTEGER: {
-            if ((sz == 4) && (H5Tget_sign(data_type_id) == H5T_SGN_2)) {
+            if ((sz == 1) && (H5Tget_sign(data_type_id) == H5T_SGN_NONE)) {
                 casacore::Bool value;
                 casacore::HDF5DataType data_type((casacore::Bool*)0);
                 H5Aread(attr_id, data_type.getHidMem(), &value);


pford:
> @pford, can you recall how this check was originally tested -- on what operating system(s), and with what files?

No, but I initially started carta development on RH6 and MacOSX 10.11 or 12 (?).  When my computer was replaced with RH7 I started using Docker with ubuntu:latest. The bool check works on Ubuntu 20.04 (my current dev Docker os).  I probably tested with M17_SWex.hdf5, perhaps a few others, since it was one of my earliest test HDF5 images.

In CartaHdf5Image::SetUpImage(), the HDF5Attributes are read into a Vector of Strings and then used in the casacore Image where appropriate, such as CoordinateSystem and ImageInfo, and the leftover headers go into MiscInfo.  These leftover headers include "simple", which is removed in SetUpImage because converting the image to FITS headers adds the SIMPLE header so it would be there twice.  That is why SIMPLE works.

confluence:
> The bool check works on Ubuntu 20.04

Do you mean that the check which is currently in `dev` works for you in Ubuntu 20.04?

pford:
Sorry, yes the M17_SWex.hdf5 headers work with dev in Docker with Ubuntu 20.04. I built your branch and got the casacore wrong data type error "EXTEND of type LONG and value 1201805313"!

confluence:
Hmm. What version of the `M17_SWex.hdf5` file are you using? What output do you get from:

    h5dump M17_SWex.hdf5 | grep -A6 EXTEND

and also

    h5dump M17_SWex.hdf5 | grep -A11 HDF5_CONVERTER_VERSION

?

confluence:
Could you please also test [this tiny image](https://github.com/CARTAvis/carta-backend/files/6052653/image-10-10-10.zip) -- both how `h5dump` reports the boolean type, and what happens when you open it in CARTA? (You will get a `passing empty or nonexistant spectral Coordinate axis` error; that is expected and unrelated.)

pford:
I downloaded Ming-Yi's version of M17_SWex.hdf5 and it now works with no errors with your branch.  As you suspected, my image must have been from an old version of the converter.

pford:
The small .zip image works too in your branch on Ubuntu 20.04.

pford (approved):
Approved as long as we no longer have to support "legacy" images produced by the old hdf5 converter.

confluence:
I don't think this is an issue with converter versions (unless I'm overlooking something; I should check some of the old versions). The converter has always written these attributes out in the same way (once we added types; way back at the start it used strings for everything). I think it must have been built and used to create that file on a system where the boolean type is a 4-byte signed integer. But I don't know what that system could have been -- as far as I know, on Windows it's unsigned, and on Linux and MacOS it's an unsigned 1-byte integer.

@pford do you still have your old M17_SWex.hdf5, or did you overwrite it?

I'd like to know exactly what happened, but it's probably not that critical, and this will become moot anyway when we start using the original header and stop re-parsing the attributes.

confluence:
Never mind! I just tried 0.1.7 and it uses signed 4-byte integers for booleans. The mystery is solved.

pford:
I overwrote M17_SWex.hdf5.  The version on my Mac has EXTEND = "T" (string).

[This image](https://drive.google.com/file/d/1gycnGXIHy0asKQfJcGKubAIIvmTbvjfy/view?usp=sharing) gets the EXTEND of type LONG with large value error.  h5dump output:
     ATTRIBUTE "EXTEND" {
         DATATYPE  H5T_STD_I32LE
         DATASPACE  SCALAR
         DATA {
         (0): 721553409
         }
      }
      ATTRIBUTE "HDF5_CONVERTER_VERSION" {
         DATATYPE  H5T_STRING {
            STRSIZE 256;
            STRPAD H5T_STR_NULLTERM;
            CSET H5T_CSET_ASCII;
            CTYPE H5T_C_S1;
         }
         DATASPACE  SCALAR
         DATA {
         (0): "0.1.7"
         }
      }

confluence:
Yeah, it looks like there was a change between 0.1.7 and 0.1.8. Still trying to figure out what that change was (or if it was a change in the hdf5 library we bundled with the converter).

confluence:
I rebuilt that old version with the latest HDF5 on Ubuntu 20.04, and it doesn't use the weird type -- so I think it must have been a change in the libraries. (I found information in the 1.10.0 changelog about the change to boolean types, but given the overflow errors in the old version, possibly a bug was also fixed.)

acdo2002 (approved):
Thanks Adrianna & Pam!!!
Now the carta Header shows the EXTEND = "T" attribute properly.
fix-hdf5-header

kswang1029:
@confluence one more exception spotted with an optical image
Screen Shot 2021-02-27 at 10 01 20 PM
file attached
[tmp.zip](https://github.com/CARTAvis/carta-backend/files/6054575/tmp.zip)

the header order in fits seems not the same as in the hdf5.

confluence:
@kswang1029 good find. But in this file the issue is with the `DATAMIN` and `DATAMAX` entries not being parsed correctly (as integers rather than as floats). I'll try to figure out why; they have integer zero values, but there's no obvious difference between them and other float entries which also have integer zero values and are being parsed correctly.

This is unrelated to the boolean value parsing; the same issue exists both in the current development branch and here.

I checked if the converter version matters, but the file is identical (except for the version entry) when converted with the latest version.

confluence:
OK, what's happening is that when we convert the HDF5 attributes back to strings to input to casacore's FITS header parser, all floats with integer values are formatted as integers. Some of these values (e.g. `BZERO`) must be converted to floats by casacore, and some (e.g. `ALLG-MAX`) stay as integers but casacore doesn't care (because they're not common entries, maybe). There is apparently no special handling for `DATAMIN` or `DATAMAX`, but casacore _does_ care that they're the wrong type.

The correct way to fix this is to make sure that when we stringify floats with integer values we output them in the correct notation.

confluence:
This should now be fixed. This file still triggers an error, but that's a problem with a date in the original FITS header.

kswang1029 (approved):

PR #753, opened 2021-03-02
Adding ARM support

ajm-ska:
We have finally received the M1 Mac Mini, so I would like to add the ARM support changes now.

I decided to the do the `CMakeLists.txt` like this:
> if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm" AND ${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
>     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_ARM_ -march=armv8-a")

It will automatically use the ARM changes if it detects the system has ARM architecture and is a Mac.

I tried to look for suitable optimisation flags for the M1, but `armv8-a` is the only one I could find. 
Or maybe the optimization flag is not necessary? 
One side effect is, for example, it won't be suitable for a Raspberry Pi (previously I used -mcpu=cortex-a72.cortex-a53" for Raspberry Pi). 
However, in this case it is OK because the condition is ARM **and** Mac. 
I don't mean to limit things, but I assume there will be more M1 Mac users than Raspberry Pi users.

@veggiesaurus Is this the best way to find the .h files?

> #include "../third-party/sse2neon/sse2neon.h"

I can't find any other way that works.


CMakeLists.txt
```diff
@@ -25,7 +25,13 @@ endif ()
 #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
 option(EnableAvx "Enable AVX codepaths instead of SSE4" OFF)
 
-if (EnableAvx)
+# Automatically detect if building on an ARM based system such as the Apple M1.
+# It will replace SSE functions with ARM NEON functions using sse2neon.h from 
+# https://github.com/DLTcollab/sse2neon, redistributable under the MIT License.
+
+if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm" AND ${CMAKE_SYSTEM_NAME} MATCHES "Darwin") 
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_ARM_ -march=armv8-a")
```
veggiesaurus: sse2neon [suggests](https://github.com/DLTcollab/sse2neon#usage) using `-march=armv8-a+fp+simd+crypto+crc`, should we not follow this suggestion?

I guess cross-compiling for ARM M1 will be an edge case that is not worth sorting out yet? If that's so, we might be able to use `-march=native` :thinking:

src/DataStream/Compression.cc
```diff
@@ -9,7 +9,11 @@
 #include <array>
 #include <cmath>
 
+#if _ARM_ // defined in cmake files
+#include "../third-party/sse2neon/sse2neon.h"
```
veggiesaurus: Please make a symbolic link in `thirdparty/include` to `sse2neon`, so that we can simply use `#include <sse2neon/sse2neon.h>`. This is what we do with spdlog, cxxopts and other submodule dependencies

src/DataStream/Smoothing.h
```diff
@@ -11,7 +11,11 @@
 #ifndef CARTA_BACKEND__SMOOTHING_H_
 #define CARTA_BACKEND__SMOOTHING_H_
 
+#if _ARM_ // defined in cmake files
+#include "../third-party/sse2neon/sse2neon.h"
+#else
 #include <x86intrin.h>
+#endif // _ARM_
```
veggiesaurus: our general approach with conditional `#include` entries is to move them below all other includes (for example, the boost FS stuff and the mach-o include

veggiesaurus (changes requested):

ajm-ska:
Thank you for the comments @veggiesaurus !

Looks like we need to use `-march=armv8-a+fp+simd+crypto+crc` because it doesn't like `-march=native`:

> clang: error: the clang compiler does not support '-march=native'

`-march=armv8-a+fp+simd+crypto+crc` seems to work fine.

I incorporated your other suggestions to make it more in the form of the _BOOST_FILESYSTEM_ conditional.

PR #754, opened 2021-03-03
611 Concatenate individual stokes images

### LINKED ISSUE #611: backend support loading individual Stokes images for visualization with the Stokes analysis widget
With the current implementation, the Stokes analysis widget is functional only if the loaded image is 4D with Stokes IQU or IQUV.

However, it is common that Stokes images are stored as separate files. We will need to enhance the Stokes analysis widget so that it can still display polarization info properly from separate Stokes image files. 

related to https://github.com/CARTAvis/carta-frontend/issues/1153

markccchiang:
Closes #611. Allows users to select multiple files with individual stokes and concatenate them as one. It works with the frontend branch [qi/issue_1153](https://github.com/CARTAvis/carta-frontend/tree/qi/issue_1153) by Qi. This step will be done on the file browser.


src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+
+        for (int i = 1; i <= 4; ++i) { // concatenate stokes file in the order I, Q, U, V (i.e., 1, 2, 3 ,4)
+            auto stokes_type = static_cast<CARTA::StokesType>(i);
+            if (extended_images.count(stokes_type)) {
+                try {
+                    concatenate_image->setImage(*extended_images[stokes_type], casacore::False);
+                } catch (const casacore::AipsError& error) {
+                    err = "Fail to concatenate images: " + error.getMesg() + "\n";
+                    success = false;
+                    break;
```
confluence: Is there any reason to use this flag here and break, instead of returning false as above?

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+                }
+            }
+        }
+    }
+
+    if (success) { // set concatenated image name
+        concatenate_name = _concatenate_name;
+    }
```
confluence: If we exit above instead of breaking we won't need a check here.

src/ImageData/StokesFilesConnector.cc
confluence: If we always exit in the same way we can simplify the exit code with a lambda, e.g.

    auto fail_exit = [&](std::str err) {
        response.set_success(false);
        response.set_message(err);
        return false;
    };

...and then we could `fail_exit("...write the error inline...")` wherever we want to exit with a false status.

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+
+        // modify the original image and extend the image coordinate system with a stokes coordinate
+        for (auto& loader : _loaders) {
+            auto stokes_type = loader.first;
+            auto* image = loader.second->GetImage();
+            try {
+                extended_images[stokes_type] =
+                    std::make_shared<casacore::ExtendImage<float>>(*image, new_image_shape, coord_sys[stokes_type]);
+            } catch (const casacore::AipsError& error) {
+                err = "Fail to extend the image: " + error.getMesg() + "\n";
```
confluence: We should try to be consistent and use `fmt::format` for things like this rather than manual string concatenation.

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+                casacore::Vector<casacore::Int> vec(1);
+                casacore::Stokes::StokesTypes stokes_type;
+
+                if (GetStokesType(carta_stokes_type, stokes_type)) {
+                    vec(0) = stokes_type;                         // set stokes type
+                    casacore::StokesCoordinate stokes_coord(vec); // set stokes coordinate
+                    tmp_coord_sys.addCoordinate(stokes_coord);    // add stokes coordinate to the coordinate system
+                    coord_sys[carta_stokes_type] = tmp_coord_sys; // fill the new coordinate system map
+                } else {
+                    err = "Fail to set the stokes coordinate system!\n";
```
confluence: Minor issue: in all these error messages "fail" should be changed to "failed".

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+    } else { // concatenate images along the stokes axis
+        concatenate_image = std::make_shared<casacore::ImageConcat<float>>(stokes_axis);
+
+        for (int i = 1; i <= 4; ++i) { // concatenate stokes file in the order I, Q, U, V (i.e., 1, 2, 3 ,4)
+            auto stokes_type = static_cast<CARTA::StokesType>(i);
+            if (_loaders.count(stokes_type)) {
+                casacore::StokesCoordinate& stokes_coord =
+                    const_cast<casacore::StokesCoordinate&>(_loaders[stokes_type]->GetImage()->coordinates().stokesCoordinate());
+                if (stokes_coord.stokes().size() != 1) {
+                    err = "Stokes coordinate has non or multiple stokes types!\n";
```
confluence: "non" should be "no"

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+                    err = "Fail to concatenate images: " + error.getMesg() + "\n";
+                    success = false;
+                    break;
+                }
+            }
+        }
+    }
+
+    if (success) { // set concatenated image name
+        concatenate_name = _concatenate_name;
```
confluence: "concatenate" should be "concatenated"

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+    ImageTypes image_types;                         // used to check whether the file type is the same
+
+    for (int i = 0; i < message.stokes_files_size(); ++i) {
+        auto stokes_file = message.stokes_files(i);
+        auto stokes_type = message.stokes_files(i).stokes_type();
+        casacore::String hdu(stokes_file.hdu());
+        casacore::String full_name(GetResolvedFilename(_top_level_folder, stokes_file.directory(), stokes_file.file()));
+
+        if (_loaders.count(stokes_type)) {
+            err = "Stokes type for is duplicate!\n";
```
confluence: Should this be something like "Duplicate Stokes type found!" ?

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+        casacore::IPosition shape;
+        int spectral_axis;
+
+        if (ref_index == 0) {
+            loader.second->FindCoordinateAxes(ref_shape, ref_spectral_axis, ref_stokes_axis, err);
+        } else {
+            loader.second->FindCoordinateAxes(shape, spectral_axis, stokes_axis, err);
+            if ((ref_shape.nelements() != shape.nelements()) || (ref_shape != shape) || (ref_spectral_axis != spectral_axis) ||
+                (ref_stokes_axis != stokes_axis)) {
+                err = "Images shapes or axes are not consistent!\n";
```
confluence: "Images" should be "Image"

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+            prefix_file_name = prefix_file_name.substr(0, pos_head);
+
+            // get the common file name start from the tail
+            std::string tmp_reverse_filename = stokes_file.file();
+            std::reverse(tmp_reverse_filename.begin(), tmp_reverse_filename.end());
+            int tmp_pos_tail = StrCmp(postfix_file_name, tmp_reverse_filename);
+            if (tmp_pos_tail < pos_tail) {
+                pos_tail = tmp_pos_tail;
+            }
+            postfix_file_name = postfix_file_name.substr(0, pos_tail);
```
confluence: This can be done a lot more concisely with `std::mismatch`:

        auto common_prefix = [](std::string a, std::string b) {
                if(a.size() > b.size()) {
                        std::swap(a, b);
                }
                return std::string(a.begin(), std::mismatch(a.begin(), a.end(), b.begin()).first);
        };

What you are doing with the suffix (reversing, accumulating the prefix, reversing back at the end) is fine, and you can still do that using this lambda.

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+            break;
+        case CARTA::StokesType::V:
+            out_stokes_type = casacore::Stokes::V;
+            success = true;
+            break;
+        default:
+            break;
+    }
+    return success;
+}
```
confluence: This can also be simplified with an inline `unordered_map` to map the CARTA types to the casacore types.

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+                if (str1.at(pos) == str2.at(pos)) {
+                    ++pos;
+                } else {
+                    break;
+                }
+            }
+        }
+    }
+    return pos;
+}
```
confluence: See comments above for turning this into a lambda and using `std::mismatch`.

confluence (changes requested):

src/ImageData/StokesFilesConnector.cc
veggiesaurus: `return false` will exit the lamda, so you'd need to `return fail_exit()`

confluence: Ah, good point.

src/ImageData/StokesFilesConnector.cc
veggiesaurus: I don't think we should use a custom function called `StrCmp`, as it will cause confusion with [`strcmp`](http://www.cplusplus.com/reference/cstring/strcmp/)

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+        ++ref_index;
+    }
+    return true;
+}
+
+bool StokesFilesConnector::GetStokesType(const CARTA::StokesType& in_stokes_type, casacore::Stokes::StokesTypes& out_stokes_type) {
+    bool success(false);
+    switch (in_stokes_type) {
+        case CARTA::StokesType::I:
+            out_stokes_type = casacore::Stokes::I;
```
veggiesaurus: `return true;` instead of `break`;

src/ImageData/StokesFilesConnector.cc
veggiesaurus: I think just simplifying to remove `success` and use `return true` or `return false` would make it nice and compact

src/ImageData/StokesFilesConnector.cc
```diff
@@ -0,0 +1,324 @@
...
+    }
+    return pos;
+}
+
+void StokesFilesConnector::ClearCache() {
+    for (auto& loader : _loaders) {
+        loader.second.reset();
+    }
+    _loaders.clear();
+    _concatenate_name = "";
```
veggiesaurus: should it be `_concatenated_name`?

veggiesaurus (changes requested):

kswang1029:
@markccchiang if per-plane-beam presents, after forming a hypercube, beam table is not sent to the frontend so that beam element at the bottom-left corner is not rendered.

markccchiang:
> @markccchiang if per-plane-beam presents, after forming a hypercube, beam table is not sent to the frontend so that beam element at the bottom-left corner is not rendered.

Do we have sample images that I can test?

kswang1029:
The one I shared to you is

kswang1029:
The hypercube contains repeated header entries.

Screen Shot 2021-03-08 at 12 54 58 PM

kswang1029:
@markccchiang when I load Q and U (two images) as a hypercube, the stokes widget does not show profiles. If we launch a spectral profile widget, Stokes="Active" and "Q" have profiles but "U" is empty. 

> [2021-03-08 13:24:28.512] [warning] Spectral requirement Uz failed: invalid stokes axis for image.

kswang1029:
@markccchiang from the icd message I see the image id of the hypercube is 0, shall we do the same as the in-memory moment images where we use id from 1000 onward (for hypercube we start from 2000?)? On a related issue, shall we support hypercube when a session is resumed? Technically it is unlike moment images which take time to generate so resuming a hypercube seems feasible. @veggiesaurus thoughts?

kswang1029:
@markccchiang @panchyo0 If we select I, Q, V to form a hypercube, the backend will accept the request and form a IQ"U" hypercube. We should forbid such kind of request both in the frontend and the backend as there is no way in the image header that we can define such combination.

We have CRVAL4, CDELT4, CRPIX4 so the delta indices in adjacent Stokes need to be identical. In the IQV case, the delta indices are 1 (Q-I = 1) and 2 (V-Q = 2) but we only can define one CDELT4 in the header. To support practical use cases, I suggest we only allow
-IQUV
-IQU
-QU
-QUV 
-IV

Theoretically IU, QV can be possibly defined in the header but practically the use case rare and unclear so it is not included in the above list.

veggiesaurus:
> @markccchiang from the icd message I see the image id of the hypercube is 0, shall we do the same as the in-memory moment images where we use id from 1000 onward (for hypercube we start from 2000?)? On a related issue, shall we support hypercube when a session is resumed? Technically it is unlike moment images which take time to generate so resuming a hypercube seems feasible. @veggiesaurus thoughts?

We should definitely handle these hypercube images properly when resuming a session. As @kswang1029 says, there's no technical reason why we shouldn't be able to. 

@kswang1029 why do we need to change the file ID? The ID is just a counter that increments as you open new images. Starting the moment images from 1000 is currently useful in terms of knowing which images to request on resume.

veggiesaurus:
> @markccchiang @panchyo0 If we select I, Q, V to form a hypercube, the backend will accept the request and form a IQ"U" hypercube. We should forbid such kind of request both in the frontend and the backend as there is no way in the image header that we can define such combination.

I think technically we can support a list of Stokes values by using the non-linear FITS header standards. But that gets ugly.

veggiesaurus:
> > @markccchiang @panchyo0 If we select I, Q, V to form a hypercube, the backend will accept the request and form a IQ"U" hypercube. We should forbid such kind of request both in the frontend and the backend as there is no way in the image header that we can define such combination.
> 
> I think technically we can support a list of Stokes values by using the non-linear FITS header standards. But that gets ugly.

Section 6.1 of " E. W. Greisen et al.: Representations of spectral coordinates in FITS" has details on the `xxxx-TAB` type. We could use `STOK-TAB` and an ancillary table, but I do wonder if its more effort than its worth

markccchiang:
> The hypercube contains repeated header entries.
> 
> Screen Shot 2021-03-08 at 12 54 58 PM

Do we have images that their header names end with the underscore `_`? If not, I can remove the header name ends with the `_`.

Hmmm I am not 100% sure. Do you know the root cause of those repeated headers?

kswang1029:
> > @markccchiang from the icd message I see the image id of the hypercube is 0, shall we do the same as the in-memory moment images where we use id from 1000 onward (for hypercube we start from 2000?)? On a related issue, shall we support hypercube when a session is resumed? Technically it is unlike moment images which take time to generate so resuming a hypercube seems feasible. @veggiesaurus thoughts?
> 
> We should definitely handle these hypercube images properly when resuming a session. As @kswang1029 says, there's no technical reason why we shouldn't be able to.
> 
> @kswang1029 why do we need to change the file ID? The ID is just a counter that increments as you open new images. Starting the moment images from 1000 is currently useful in terms of knowing which images to request on resume.

If resume is no problem at all, usual ID counter is fine. 👍

markccchiang:
> 
> Hmmm I am not 100% sure. Do you know the root cause of those repeated headers?

While concatenating, the repeated header name will be saved as the name with an underscore:
`2021-03-08 08:05:06	INFO	ImageFitsConverter::ImageHeaderToFITS 	Truncating miscinfo field casambm_3 to casambm_`

markccchiang:
> We have CRVAL4, CDELT4, CRPIX4 so the delta indices in adjacent Stokes need to be identical. In the IQV case, the delta indices are 1 (Q-I = 1) and 2 (V-Q = 2) but we only can define one CDELT4 in the header. To support practical use cases, I suggest we only allow
> -IQUV
> -IQU
> -QU
> -QUV
> -IV
> 
> Theoretically IU, QV can be possibly defined in the header but practically the use case rare and unclear so it is not included in the above list.

So do we allow the `IQ` case?

kswang1029:
> > We have CRVAL4, CDELT4, CRPIX4 so the delta indices in adjacent Stokes need to be identical. In the IQV case, the delta indices are 1 (Q-I = 1) and 2 (V-Q = 2) but we only can define one CDELT4 in the header. To support practical use cases, I suggest we only allow
> > -IQUV
> > -IQU
> > -QU
> > -QUV
> > -IV
> > Theoretically IU, QV can be possibly defined in the header but practically the use case rare and unclear so it is not included in the above list.
> 
> So do we allow the `IQ` case?

I suggest we don't.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -151,9 +152,14 @@ bool FileExtInfoLoader::FillFileInfoFromImage(CARTA::FileInfoExtended& extended_
                 // Create header entry for each FitsKeyword
                 fhi.kw.first(); // go to first card
                 casacore::FitsKeyword* fkw = fhi.kw.next();
+                std::set<casacore::String> name_set; // used to check the repetition of name
+                casacore::Regex underscore_suffix(casacore::Regex::fromPattern("*_"));
```
veggiesaurus: why can't we use `std::regex` here? We know there are issues with `casacore::Regex`

markccchiang:
I have updated the protocol branch [mark/concatenate_images](https://github.com/CARTAvis/carta-protobuf/tree/mark/concatenate_images) and add a `stokes_files` in the `ImageProperties` message: 

截圖 2021-03-12 上午11 26 14

While resuming a session, if the size of `stokes_files` is greater than 1, then the backend will open a concatenated stokes image. @panchyo0 could you please update the frontend and inform me if there are problems? Thank you!

markccchiang:
I have fixed bugs and enabled to resume hypercube images. @kswang1029 could you please verify them again? Thank you!

kswang1029:
@markccchiang If individual I, Q and U images have per-plane-beam, when forming a hypercube IQU, we only see a beam element rendered when selecting I, not Q nor U. Could you check if beam table for Q and U are handled properly?

Once the fix is in place, this PR is ready (will approve then).

markccchiang:
> @markccchiang If individual I, Q and U images have per-plane-beam, when forming a hypercube IQU, we only see a beam element rendered when selecting I, not Q nor U. Could you check if beam table for Q and U are handled properly?
> 
> Once the fix is in place, this PR is ready (will approve then).

It seems the concatenation function only uses the beam information from the first sub-image. Will try to modify the ImageInfo and insert the other stokes files beam information.

veggiesaurus:
> @markccchiang If individual I, Q and U images have per-plane-beam, when forming a hypercube IQU, we only see a beam element rendered when selecting I, not Q nor U. Could you check if beam table for Q and U are handled properly?
> 
> Once the fix is in place, this PR is ready (will approve then).

Shouldn't the been table be the same between Stokes values?

markccchiang:
@kswang1029 please check does the new code solve the beam information problem for the other stokes Q, U, and V. Thank you!

kswang1029:
> @kswang1029 please check does the new code solve the beam information problem for the other stokes Q, U, and V. Thank you!

Yes, confirmed that beam info is available and beam element is rendered properly. 👍

kswang1029 (approved):

veggiesaurus:
@markccchiang can you solve the merge conflicts here?

PR #763, opened 2021-03-04
Set explicit cmake build type

confluence:
Currently the build type is set to `Release` as a side effect of the spdlog cmake configuration. I propose that we explicitly set it to `RelWithDebInfo`.

### DIFF
--- a/third-party/install/uWebSockets.cmake
+++ b/third-party/install/uWebSockets.cmake
@@ -1,6 +1,6 @@
 macro(install_uWebSockets)
 
-    set(CMAKE_C_FLAGS "-DLIBUS_NO_SSL")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DLIBUS_NO_SSL")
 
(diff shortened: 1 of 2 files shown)


third-party/install/uWebSockets.cmake
```diff
@@ -1,6 +1,6 @@
 macro(install_uWebSockets)
 
-    set(CMAKE_C_FLAGS "-O3 -DLIBUS_NO_SSL")
```
veggiesaurus, line 3: should we not also be including the existing `CMAKE_C_FLAGS`?

confluence: Probably; I will add that.

confluence: Done. It currently has no effect, but I have tested that it works.

veggiesaurus: it will have an effect if you manually specify flags when calling cmake though

veggiesaurus (approved):

PR #766, opened 2021-03-05
prevents caching of rest API routes

veggiesaurus:
Adds cache control headers to prevent preferences and layout endpoints from being cached


confluence (approved):
Tested in the package.

PR #773, opened 2021-03-14
769 Replace the z profile count with a shared mutex

### LINKED ISSUE #769: Frame::WaitForTaskCancellation infinite loop
A difficult-to-reproduce bug: Sometimes `Frame::WaitForTaskCancellation` gets stuck in an infinite loop (sleeping most of the time), because `_z_profile_count` is stuck at 1. I'm not really sure why this is happening.  

The problem is that this prevents new sessions from being created, as the main thread is entirely taken up by this infinite loop.

markccchiang:
This intends to solve #769. Based on @veggiesaurus and @pford's suggestions that we replace the `_z_profile_count` with a shared lock to avoid the stuck problem happened on the server. The moment calculations are also applied in the same way. The exclusive lock is needed before doing the Frame/Region destruction. That means all calculations are finished and we can delete the object safely.


veggiesaurus:
:thinking: I don't think we can assume this fixes #769 until we can reliably reproduce it

src/Region/RegionHandler.h
```diff
@@ -42,7 +42,7 @@ namespace carta {
 
 class RegionHandler {
 public:
-    RegionHandler();
+    RegionHandler(){};
```
veggiesaurus: `= default` would work there, right?

src/Frame.h
```diff
@@ -170,12 +169,13 @@ class Frame {
...
         std::vector<carta::CollapseResult>& collapse_results);
     void StopMomentCalc();
-    void IncreaseMomentsCount();
-    void DecreaseMomentsCount();
 
     // Save as a new file or convert it between CASA/FITS formats
     void SaveFile(const std::string& root_folder, const CARTA::SaveFile& save_file_msg, CARTA::SaveFileAck& save_file_ack);
 
+    // Use a shared lock for long time calculations, use an exclusive lock for the object destruction
+    mutable std::shared_mutex _life_mutex;
```
veggiesaurus: if this is public, it shouldn't be `_life_mutex`. Either make it public, or (probably neater) encapsulate the shared and exclusive locks in functions.

src/Frame.h
```diff
@@ -254,11 +253,8 @@ class Frame {
     tbb::queuing_rw_mutex _cache_mutex; // allow concurrent reads but lock for write
     std::mutex _image_mutex;            // only one disk access at a time
 
-    // Z profile counter, so Frame is not destroyed until finished
-    std::atomic<int> _z_profile_count;
-
-    // Moments counter, so Frame is not destroyed until finished
-    std::atomic<int> _moments_count;
+    // Use a shared lock for long time calculations, use an exclusive lock for the object destruction
+    mutable std::shared_mutex _life_mutex;
```
confluence: I would suggest renaming "life mutex" to something like "calculation mutex" or "active task mutex".

confluence (commented):
I would prefer to rename the mutex to something that more clearly describes its function, but apart from that everything looks good.

src/Frame.h
veggiesaurus: > I would prefer to rename the mutex to something that more clearly describes its function, but apart from that everything looks good.

agreed.

kswang1029 (approved):
I cannot reproduce the deadlock with a more reliable reproducible procedure from v1.4. So I think the fix works. I cannot reproduce the crash event so that might be something else. Good to merge this for now. 👍

PR #776, opened 2021-03-15
663 pv uv images

### LINKED ISSUE #663: Support UV and pV images
These image types occur in all supported CARTA image types (CASA, FITS, HDF5, Miriad).

Adding "requiring_frontend" tag as

    Images with UV coordinates
    pV images

Frontend issue [#1150](https://github.com/CARTAvis/carta-frontend/issues/1150)

pford:
Closes #663.

Support for pV images, UV image support required on frontend only.  Draft PR pending frontend development.  In particular, frontend cannot display pVimage.image with pos-stokes-vel axes.

As part of pV support, replaced spectral/channel variables with z and depth.  The spectral axis and channels were used to set an image plane or create a Slicer, and this terminology became confusing when the spectral axis was one being rendered.  The loader now determines not only the spectral and stokes axes (used for extended file info) but also the z axis.  Frame stores _width, _height, _depth, and _num_stokes for clarity and to avoid indexing into _image_shape in many places.


veggiesaurus:
@pford is this ready for a review? The frontend branch is ready

src/ImageData/FileLoader.cc
```diff
@@ -83,118 +85,155 @@ bool FileLoader::GetCoordinateSystem(casacore::CoordinateSystem& coord_sys) {
...
-            spectral_axis = 2;
-            _num_channels = shape(spectral_axis);
-            _num_stokes = 1;
-        }
-
-        // save axes
-        _spectral_axis = spectral_axis;
-        _stokes_axis = stokes_axis;
-        return true;
+    // Cope with incomplete/invalid headers for 3D, 4D images
```
confluence, line 146: I would suggest adding debugging messages to this guessing code, so that if something weird happens we can turn on debug logging and follow the logic.

src/Util.h
```diff
@@ -57,23 +57,23 @@ std::string GetAuthToken(uWS::HttpRequest* http_request);
 
 // ************ structs *************
 //
-// Usage of the ChannelRange:
+// Usage of the AxisRange:
```
confluence, line 60: It's a little confusing that this is called `AxisRange` and not `ZAxisRange` or `ZRange`, if it is used only for the Z axis (the problem is the hardcoded end in the argument-less constructor).

confluence: OK, I see now that the end isn't actually hardcoded to the depth; it's hardcoded to the constant for using the full axis which is later translated to the depth. So I guess my issue is the slightly inconsistent naming of `ALL_Z` and `AxisRange`, both of which are actually completely axis-agnostic, but are only used for the Z axis. But this is not a huge problem.

confluence (approved):
I think this is fine.

veggiesaurus (approved):

src/Util.h
pford: I made the struct name generic since it could be used for any axis including stokes, e.g. refactor Frame::GetImageSlicer with a stokes range. ALL_Z was an incomplete refactor of this struct; the axis constants should be generic instead of redundant DEFAULT, CURRENT, and ALL constants for Z and STOKES.  I will do that in a new branch since this one has been merged.

Issue #431, opened 2020-01-30
CARTA hangs in file info gathering step


jott3077:
I have a directory with manny CASA images. When I browse through, look at some, then go back and load others, at some point while just checking the file info, CARTA 1.2.2 hangs. When I then go back to 'load' The 'loading file' message also keeps on spinning.

veggiesaurus:
@arusstaylor does this seem similar to what you experienced on carta-testing?

arusstaylor:
The load file message and spinning is the same, but it happened to me directly on startup when the file load tool comes up automatically.   So not related to any specific action with the file load tool.

veggiesaurus:
> The load file message and spinning is the same, but it happened to me directly on startup when the file load tool comes up automatically. So not related to any specific action with the file load tool.

Your backend log file seemed to indicate that there were other sessions still running, so perhaps one of those other sessions was causing an issue :thinking:

keflavich:
I'm seeing the same issue on CARTA 1.3.1.  Is there any way to load a file _without_ using the file browser?

jott3077:
try carta <imagename>


On June 27, 2020 09:09:30 Adam Ginsburg <notifications@github.com> wrote:
>
> I'm seeing the same issue on CARTA 1.3.1. Is there any way to load a file 
> without using the file browser?
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.


Sent with AquaMail for Android
https://www.mobisystems.com/aqua-mail

keflavich:
That had no effect - I still see "No file loaded".

However, running carta & specifying the file name does work if I start from a directory that contains no .image files.

I suggest *not* loading the file browser at startup.  In folders containing large numbers of .images - which as you can see from both @jott3077 and my use cases, is common if not the default - CARTA simply fails to load.

kswang1029:
I tried 
`carta carta_lots_files` 
to launch CARTA with a custom folder "carta_lots_files" which contains 1000 fits images, 1000 miriad images (folder image), 1000 casa images (folder image), 1000 CRTF files, and 1000 plain text files. It works although it takes some time to load (with MBP SSD). 

@keflavich in your case, what does your folder look like?

veggiesaurus:
> I suggest _not_ loading the file browser at startup. In folders containing large numbers of .images - which as you can see from both @jott3077 and my use cases, is common if not the default - CARTA simply fails to load.

[...]

Issue #761, opened 2021-03-04
refactor Constants.h


veggiesaurus:
`Constants.h` contains `#define` values for a whole host of different unrelated things. The problem with this approach is that changing anything in this file leads to huge recompiles, because loads of files include `Constants.h` (directly or indirectly). We should rather separate out these values into the appropriate headers (`Logger.h`, `Frame.h` etc)

confluence:
I suggest that at the same time we refactor `Util.h` and `Util.cc`, for similar reasons. It's also a collection of unrelated functions, many of which are only used in one place. In some cases the presence of a particular function in `Util` rather than a more specific place is also complicating the refactoring of the constants.

confluence:
I have bumped the estimate, since this is a bit more involved than previously thought. I will handle #744 in the same PR, since they are related tasks.

PR #777, opened 2021-03-16
431 File list interruption

markccchiang:
For #431, it seems there is no way to increase the speed of gathering a file list. So we apply a cancellation mechanism for it. Also, the backend will report the progress of gathering a file list after 5 secs and update the progress every 2 secs. This branch works with the protobuf branch [mark/stop_getting_file_list](https://github.com/CARTAvis/carta-protobuf/tree/mark/stop_getting_file_list). It also needs the frontend's change.


src/Constants.h
```diff
@@ -69,6 +69,10 @@
 // Shared region polygon approximation
 #define DEFAULT_VERTEX_COUNT 1000
 
+// file list
+#define REPORT_FIRST_PROGRESS_AFTER_SECS 5
```
veggiesaurus: what files are these used in? As specified in #761 , I'd like to refactor this file

veggiesaurus: to avoid rebuilds

markccchiang: They are used in files `src/FileList/FileListHandler.cc` and `src/Table/TableController.cc`.

kswang1029 (approved):
feature-wise this looks good.

src/Table/TableController.cc
```diff
@@ -202,7 +202,24 @@ void TableController::OnFileListRequest(
     file_list_response.set_parent(parent_path.string());
 
     try {
+        // get total number of files in the directory
+        using std::filesystem::directory_iterator;
```
veggiesaurus: use `fs::` and make sure it works with boost FS

src/Table/TableController.cc
```diff
@@ -202,7 +202,24 @@ void TableController::OnFileListRequest(
...
+        using func_ptr = bool (*)(const std::filesystem::path&);
+        auto total_files = std::count_if(directory_iterator(file_path), directory_iterator{}, (func_ptr)std::filesystem::is_regular_file);
```
veggiesaurus: not sure why you need to use `using` above. It doesn't really help anything. Why not just provide the types in the `count_if`?

src/FileList/FileListHandler.h
```diff
@@ -37,11 +38,18 @@ class FileListHandler {
     void OnRegionFileInfoRequest(
         const CARTA::RegionFileInfoRequest& request, CARTA::RegionFileInfoResponse& response, ResultMsg& result_msg);
 
+    void StopGettingFileList() {
+        _stop_getting_file_list = true;
+    }
+    void SetProgressCallBack(const std::function<void(CARTA::Progress)>& progress_callback) {
```
veggiesaurus: if we use `callback` as one word, it should be `SetProgressCallback`, shouldn't it?

src/Session.cc
```diff
@@ -301,10 +301,14 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
 }
 
 void Session::OnFileListRequest(const CARTA::FileListRequest& request, uint32_t request_id) {
+    auto progress_callback = [&](CARTA::Progress progress) { SendEvent(CARTA::EventType::FILE_LIST_PROGRESS, request_id, progress); };
+    _file_list_handler->SetProgressCallBack(progress_callback);
     CARTA::FileListResponse response;
     FileListHandler::ResultMsg result_msg;
     _file_list_handler->OnFileListRequest(request, response, result_msg);
-    SendEvent(CARTA::EventType::FILE_LIST_RESPONSE, request_id, response);
+    if (!response.cancel()) {
```
veggiesaurus, line 309: @kswang1029 I wonder if we should still return the _partial_ results, even after cancellation? At least then users can see some of the files or directories. In fact, if we do this, we can also move the directories first in order to prioritise them

src/Table/TableController.cc
veggiesaurus: i.e. `auto total_files = std::count_if(fs::directory_iterator(file_path), directory_iterator{}, (bool (*)(const fs::path&))fs::is_regular_file);;`

veggiesaurus: CASA images are folders, so I do wonder whether we should actually include _all_ directory entries, not just regular files

src/Table/TableController.cc
```diff
@@ -234,6 +251,28 @@ void TableController::OnFileListRequest(
                 stat(entry.path().c_str(), &file_stats);
                 file_info->set_date(file_stats.st_mtim.tv_sec);
             }
+
+            ++num_of_files_done;
+            percentage = (float)num_of_files_done / (float)total_files;
+            auto current_time = std::chrono::high_resolution_clock::now();
+            auto dt = std::chrono::duration<double>(current_time - start_time).count();
+
+            auto report_progress = [&]() {
```
veggiesaurus: I don't like the idea of redefining a lambda inside the loop. Why does it need to be in the loop, instead of outside?

veggiesaurus: `current_time` could easily be a parameter

veggiesaurus (changes requested):

src/Session.cc
kswang1029: we could do that as long as we warn user well that what they see is not complete. The warning text should be persistent somewhere on UI I think.

veggiesaurus:
@markccchiang please also see the comments in the protobuf repo. I'd like to keep things one-to-one in terms of event types and messages. Please refactor so that you have a single cancel event and a single file list progress event, each with a field indicating what type of list it is

src/Session.cc
veggiesaurus: right, it could be shown in the folder name. e.g scratch -> images -> lots _(partial results)_.

I think it would be particularly helpful for navigating to a subdirectory

kswang1029: If we loaded an image from the partial list then try to append a new one, when we launch the file browser will it request the full list again?

veggiesaurus: good question...

kswang1029: I remembered the frontend does request the file list again (and it should instead of using cached list). So I think we are fine.

markccchiang: So should the backend return the partial file list after the cancelation? If the frontend caches a file list, next time it can tell the backend to get a life list started from the *n-th* file or folder.

veggiesaurus: No, next time it should still start from scratch because the file list may change

src/FileList/FileListHandler.cc
```diff
@@ -157,24 +178,58 @@ void FileListHandler::GetFileList(CARTA::FileListResponse& file_list, std::strin
...
+                progress.set_file_list_type(CARTA::FileListType::Image);
+                progress.set_percentage(percentage);
+                progress.set_checked_count(num_of_files_done);
+                progress.set_total_count(total_files);
+                _progress_callback(progress);
+                start_time = current_time;
+            };
+
+            // report the progress if it fits the conditions
+            if (!_first_report && dt > REPORT_FIRST_PROGRESS_AFTER_SECS) {
```
confluence: The name `_first_report` is misleading -- I suggest changing it to something like `_first_report_made`.

`REPORT_FIRST_PROGRESS_AFTER_SECS` is inconsistently named, and it's not clear that this has anything to do with the file list. There are also some unclearly named constants used for image moments (including one called `REPORT_FIRST_PROGRESS_AFTER_MILLI_SECS`!). But I think that the correct way to fix this is to refactor the constants file and move these into the scopes where they are actually used, which should happen in a separate PR.

confluence: I see now that these are used in two different places, so the name should probably be fixed in the short term. I would change it to `FILE_LIST_FIRST_PROGRESS_AFTER_SECS`.

src/Table/TableController.cc
```diff
@@ -202,7 +202,36 @@ void TableController::OnFileListRequest(
...
 
     try {
+        // get total number of files or directories under the directory
+        auto total_regular_files =
+            std::count_if(fs::directory_iterator(file_path), fs::directory_iterator{}, (bool (*)(const fs::path&))fs::is_regular_file);
+        auto total_directories =
+            std::count_if(fs::directory_iterator(file_path), fs::directory_iterator{}, (bool (*)(const fs::path&))fs::is_directory);
+        auto total_files = total_regular_files + total_directories;
+
+        // initialize variables for the progress report and the interruption option
```
confluence, line 213: A lot of this code is duplicated between this function and the one in `FileListHandler.cc`. Would it make sense to factor out the common parts to a generic function which could be called by both?

confluence (changes requested):

src/FileList/FileListHandler.cc
markccchiang: I will refactor the constants for image moments in a separate PR.

src/Constants.h
```diff
@@ -69,6 +69,10 @@
 // Shared region polygon approximation
 #define DEFAULT_VERTEX_COUNT 1000
 
+// file list
+#define FILE_LIST_FIRST_PROGRESS_AFTER_SECS 5
+#define UPDATE_FILE_LIST_PROGRESS_PER_SECS 2
```
confluence: How about `FILE_LIST_PROGRESS_INTERVAL_SECS`, similar to what you did for the image moments constant in the other PR?

src/Timer/ProgressReporter.cc
```diff
@@ -0,0 +1,31 @@
+/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
```
confluence: I think that this should be called `ListProgressReporter`, since it is specific to list progress reports.

confluence (changes requested):
I have two more name suggestions, but apart from that, I think this looks fine.

confluence (approved):

PR #780, opened 2021-03-20
607 matched region distortion check

### LINKED ISSUE #607: apply region distortion check before applying polygon approximation
If we have two matched identical images, slightly different analytics might be derived from an ellipse region due to polygon approximation on the secondary image. We should examine region distortion (by two reference vectors to see if dot product is zero (or smaller than epsilon) and length ratio is the same) and only apply polygon approximation when distortion is not negligible.

pford:
Closes #607 

Checks the region distortion before converting it to an approximate polygon.  If there is no distortion, the region is converted to the same region type in the matched image.

If the backend is started in debug mode (--verbosity 5) the log contains the results of the distortion checks, including the difference in the vector lengths ratio between the reference and matched images, and the dot product of the vectors in the matched image.

The vectors used for the checks are center->p0 and center->p1.  The ellipse uses existing code to make a 4-point polygon region in the reference ellipse for the vector points.  The rectangle uses existing code to get the reference rectangle corners (applying rotation if necessary) then calculates the midpoints for the vector points.  These points and the center point are converted to pixel coordinates in the matched image. 
ellipse_check  rectangle_check


kswang1029:
@pford do we apply the same check for rectangle region too? I will check the debug log to see what epsilon we should adopt to enable polygon approximation.

pford:
Rectangle distortion check added.

src/Region/Region.cc
```diff
@@ -388,50 +389,94 @@ casacore::LCRegion* Region::GetImageRegion(
...
+
+                // Compare converted to reference length ratio
+                double length_ratio_difference = fabs(ref_length_ratio - converted_length_ratio);
+                spdlog::debug("{} distortion check: length ratio difference={:.3e}", RegionName(region_type), length_ratio_difference);
+                if (length_ratio_difference < 1e-4) {
+                    // Passed ratio check; check dot product
+                    double converted_dot_product = (v0_delta_x * v1_delta_x) + (v0_delta_y * v1_delta_y);
+                    spdlog::debug("{} distortion check: dot product={:.3e}", RegionName(region_type), converted_dot_product);
+
+                    if (fabs(converted_dot_product) < 1e-4) {
```
kswang1029: @pford Thanks for preparing the handy debug log for testing. I have done some tests with rectangle and ellipse both unrotated and rotated and focused on the threshold of the dot product check. By experimenting the different threshold, **1e-2** (no idea why we need a larger threshold here🙈 Maybe this is due to imageA_pixel->imageA-world->imageB_world->imageB_pixel conversion errors?) works properly without any inconsistent statistics as far as I can try and see.

src/Region/Region.cc
```diff
@@ -388,50 +389,94 @@ casacore::LCRegion* Region::GetImageRegion(
...
+                auto v1_delta_y = y[1] - y[4];
+
+                auto v0_length = sqrt((v0_delta_x * v0_delta_x) + (v0_delta_y * v0_delta_y));
+                auto v1_length = sqrt((v1_delta_x * v1_delta_x) + (v1_delta_y * v1_delta_y));
+                double converted_length_ratio = v1_length / v0_length;
+
+                // Compare converted to reference length ratio
+                double length_ratio_difference = fabs(ref_length_ratio - converted_length_ratio);
+                spdlog::debug("{} distortion check: length ratio difference={:.3e}", RegionName(region_type), length_ratio_difference);
+                if (length_ratio_difference < 1e-4) {
```
kswang1029, line 440: Regarding this arbitrary 1e-4 threshold for the length ratio difference, it works fine so far. 👍  It may need quite some case studies to verify our current choice but for now I suggest we just adopt it.

kswang1029:
@pford once we update the threshold for the dot product check, I think this draft PR can turn into a formal PR. 👍

src/Region/Region.cc
pford: I suspect it is due to world->pixel conversion using wcslib wcss2p().  Working on using ast for this conversion for issue #541 .

markccchiang (approved):

confluence (approved):

PR #781, opened 2021-03-22
1271 Overwrite a default compression quality with higher precision

markccchiang:
According to the suggestion by @kswang1029 from [#1271](https://github.com/CARTAvis/carta-frontend/issues/1271), the backend will overwrite a default ZFP compression quality with higher precision. If there is a dominant bin in the pixel histograms, i.e., >= 90% of pixels concentrate in a bin.

### DIFF
--- a/src/Constants.h
+++ b/src/Constants.h
@@ -98,7 +98,4 @@
 #define CARTA_PREFERENCES_SCHEMA_URL "https://cartavis.github.io/schemas/preference_schema_1.json"
 #define CARTA_LAYOUT_SCHEMA_URL "https://cartavis.github.io/schemas/layout_schema_2.json"
 
-// ZFP compression
-#define HIGH_COMPRESSION_QUALITY 20
-
 #endif // CARTA_BACKEND__INTERFACECONSTANTS_H_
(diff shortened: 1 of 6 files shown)


src/Constants.h
```diff
@@ -98,4 +98,7 @@
 #define CARTA_PREFERENCES_SCHEMA_URL "https://cartavis.github.io/schemas/preference_schema_1.json"
 #define CARTA_LAYOUT_SCHEMA_URL "https://cartavis.github.io/schemas/layout_schema_2.json"
 
+// ZFP compression
+#define HIGH_COMPRESSION_QUALITY 20
```
veggiesaurus: please move this to `Frame.cc` if we only need it there

src/ImageStats/Histogram.cc
```diff
@@ -94,3 +94,18 @@ void Histogram::SetHistogramBins(const std::vector<int>& bins) {
     }
     _histogram_bins = bins;
 }
+
+bool Histogram::HasDominantBin() {
```
veggiesaurus: I think `HasDominantBin` sounds like it should be a `const` function, but it isn't. Perhaps `CheckForDominantBin`?

PR #782, opened 2021-03-23
ajm/778 cmakelists changes

### LINKED ISSUE #778: curl-dev, tbb-dev and uuid-dev dependencies are not specified in CMakeLists.txt
The dependencies curl-dev, tbb-dev and uuid-dev are needed to compile carta-backend, however cmake command doesn't complain about those dependencies when they are missing, but instead one gets an error at compilation time.
It would be useful that the CMakeLists.txt file contains the proper FIND_PACKAGE directives so that those dependencies are found at configuration time.

ajm-ska:
I have added cmake checks for:
`curl`
`pugixml`
`tbb`
`uuid`
`zfp`
`zstd`
Plus a check for `gtest` if the user wishes to build with `-Dtest=on`

For some reason we needed to use `PKG_SEARCH_MODULE` instead of `FIND_PACKAGE` for UUID.
Also to be both Mac and Linux compatible, we needed to do the same trick as gRPC for both ZSTD and TBB. i.e. the using both PKG_SEARCH_MODULE and FIND_PACKAGE.

I have also temporarily modified the CircleCI config.yml to exclude the Timer related tests.

> ./carta_backend_tests --gtest_output=xml --gtest_filter=-"TimerTest*"

I have done this because otherwise the TimerTests keep fail by taking an unusual amount of time. I suspect this due to a limitation on our free CircleCI plan. We are currently well over our free weekly quota of 2500 credits. Although we are over the limit, the Circle CI build and tests still run, they just appear to be very slow.

This should close #778


confluence:
`libcurl4-openssl-dev` or `libcurl4-gnutls-dev` is required to build on Ubuntu. We include `curl/curl.h`.

ajm-ska:
> `libcurl4-openssl-dev` or `libcurl4-gnutls-dev` is required to build on Ubuntu. We include `curl/curl.h`.

I know that now. 
I have already modified the CMakeLists.txt file so that cmake can correctly search for curl:
`FIND_PACKAGE(CURL REQUIRED)`

CMakeLists.txt
```diff
@@ -44,6 +44,37 @@ endif ()
...
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
+FIND_PACKAGE(CURL REQUIRED)
+FIND_PACKAGE(PUGIXML REQUIRED)
+FIND_PACKAGE(ZFP REQUIRED)
+FIND_PACKAGE(PkgConfig REQUIRED)
+
+PKG_SEARCH_MODULE(UUID REQUIRED uuid)
+if (UUID_FOUND)
+    message(STATUS "Found uuid using pkg-config")
```
veggiesaurus, line 78: shouldn't we actually use the found paths in our include and lib variables?

ajm-ska:
@veggiesaurus Do you mean like the above commit? I think it makes the CMakeLists.txt unnecessarily long and complicated. The found paths are used anyway further down in the file in the `set(LINK_LIBS` block.
But if you like it that way, I can extend all the other FIND_PACKAGE lines too:

FIND_PACKAGE(CURL REQUIRED)
FIND_PACKAGE(PUGIXML REQUIRED)
FIND_PACKAGE(ZFP REQUIRED)
FIND_PACKAGE(PkgConfig REQUIRED)
FIND_PACKAGE(HDF5)
FIND_PACKAGE(Protobuf REQUIRED)
FIND_PACKAGE(Threads)

veggiesaurus:
> @veggiesaurus Do you mean like the above commit? I think it makes the CMakeLists.txt unnecessarily long and complicated. The found paths are used anyway further down in the file in the `set(LINK_LIBS` block.
> But if you like it that way, I can extend all the other FIND_PACKAGE lines too:
> 
> FIND_PACKAGE(CURL REQUIRED)
> FIND_PACKAGE(PUGIXML REQUIRED)
> FIND_PACKAGE(ZFP REQUIRED)
> FIND_PACKAGE(PkgConfig REQUIRED)
> FIND_PACKAGE(HDF5)
> FIND_PACKAGE(Protobuf REQUIRED)
> FIND_PACKAGE(Threads)

Sorry, I meant doing something like `INCLUDE_DIRECTORIES(${UUID_INCLUDE_DIR})`, in case the include files are in a non-standard location. Or is that not required? Perhaps we should be using the "config" approach, as suggested here? https://stackoverflow.com/a/20857070/1727322

ajm-ska:
@veggiesaurus Maybe I'm doing it wrong, but I don't think the "config" approach works so well in our case.

If I do, for example, `FIND_PACKAGE(CURL CONFIG REQUIRED)` it tries to look for `CURLConfig.cmake`.

> CMake Error at CMakeLists.txt:47 (FIND_PACKAGE):
>   Could not find a package configuration file provided by "CURL" with any of
>   the following names:
> 
>    CURLConfig.cmake
>    curl-config.cmake
> 
>   Add the installation prefix of "CURL" to CMAKE_PREFIX_PATH or set
>   "CURL_DIR" to a directory containing one of the above files.  If "CURL"
>   provides a separate development package or SDK, be sure it has been
>   installed.

Such "Config.cmake" files do not seem to exist for libraries simply installed by the package managers (apt, yum, brew). Almost all our libraries are most easily installed by the package manager.

The previous `FIND_PACKAGE(CURL REQUIRED)` seems to work fine:

> -- Found CURL: /usr/lib/x86_64-linux-gnu/libcurl.so (found version "7.47.0") 

The "Config.cmake" files only seem to exist for libraries that we must install from source e.g. `FIND_PACKAGE(ZFP CONFIG REQUIRED)` works: 

> -- Found ZFP: /usr/local/lib/cmake/zfp/zfp-config.cmake (found version "0.5.5") 

At this stage, zfp is the only package (besides carta-casacore) that we need to install from source (Although not anymore on ubuntu thanks to the PPAs). However, it looks like it works with the PPA installed version too. So I have only left in "config" for zfp:
> FIND_PACKAGE(ZFP CONFIG REQUIRED)

ajm-ska:
Thanks to @kswang1029 for noticing that we should search for wcslib and cfitsio too.

veggiesaurus (approved):
After doing a bit of a deep-dive into cmake over the last week, it looks like the appropriate thing to do is to call `find_package`, and then use `target_link_libraries(${TBB_LIBRARIES}` and so on to automatically add all the required libraries _and_ include dirs, but I think this is a separate issue outside of the scope of this PR.

Issue #716, opened 2021-02-08
JSON-based config support


veggiesaurus:
We need to add support for system- and user-based config files for the backend. We already use JSON for the frontend and carta-node-server configurations, and JSON schema allow us to easily validate things.

The backend should look for a system config file in `/etc/carta/backend.json`, and a user config file in `~/.carta/backend.json`. We should also allow a command-line parameter `no_user_config` parameter to disable the user config (useful for the server use case). This will leave us with 4 sources of settings:

0) default settings
1) system config file
2) user config file
3) command-line arguments

In terms of precedence, anything specified in (3) should override (2). Anything specified in (2, 3) should override (1). The only exception being the `no_user_config` argument, which will always disable (2). One nice way to handle this is by storing the resultant settings as a JSON object, formed by merging multiple JSON sources. We'd need a JSON object for the defaults and the command-line arguments.

I suggest we use the modern and widely used [nlohmann/json](https://github.com/nlohmann/json) header-only package for this. It's easily installable on ubuntu, centOS (as `json-devel`) and homebrew. I think we'd need to support both versions 2 and 3, though, since ubuntu bionic has version 2.1..1, rather than 3.7.3 in focal.

We should have a well-defined JSON schema for the backend configuration. This will make it easier to ensure that our config files are valid, and we can also use it to generate documentation (we've done this for carta-node-server).

This feature would really benefit from extensive unit tests to ensure that the correct options get set when using various config sources.

confluence:
I have previously expressed my concerns about users hand-editing JSON. I think that this can be mitigated if we have an option to generate a default / example config file, perhaps accessible through a flag to the backend, and include this when we package the backend (not in the location which activates it; as `/etc/carta/backend.json-example`, or something like that).

JSON doesn't natively support comments (although there are dialects / parsers that do). `nlohmann/json` has them disabled by default, but it's possible to enable C-like comments. It may be useful to allow this for hand-editable config files, but it may affect interoperability if they ever need to be parsed or written by another library (e.g. in future it may be useful to have a UI for editing this file in the frontend). Also, trailing commas are not allowed in JSON, which can make selective commenting tricky.

Despite my reservations, I think that if we're already using JSON for all our other configuration files, it would be good to be consistent.

I think that the override hierarchy that you have described makes sense.

veggiesaurus:
[...]

PR #783, opened 2021-03-26
ProgramSettings: JSON based user and system settings

jolopezl:
Currently, this is just a rough implementation of JSON-based user and system settings. Follows what's prescribed in #716.

Needs:

- [x] Full list of parameters.
- [x] Check input file format, prevent crashes.
- [x] Add `no_user_config`.
- [x] Fix program settings tests.
- [x] Add tests for JSON based configs
- [x] Add a complete set of tests

The implementation is simple. It allows to check four sources of settings (quoting #716):

1. default settings
2. system config file
3. user config file
4. command-line arguments

Anything specified in (3) should override (2). Anything specified in (2, 3) should override (1). The only exception being the `no_user_config` argument, which will always disable (2).


jolopezl:
A bit of explanation: 
- Command-line settings are parsed into a JSON object. System and user settings are parsed into two JSON objects, with the user settings merged on top of the system systems (user > system). Next, command-line settings' JSON is merged on top of the result of system and user settings (command-line > user > system), if there's any.
- System and user settings can be overridden with `--no_system_config` and `--no_user_config` command-line options respectively.

test/TestProgramSettings.cc
```diff
@@ -214,4 +216,42 @@ TEST_F(ProgramSettingsTest, MultipleImagesFromPositional) {
     EXPECT_EQ(settings.files[0], casa_image.string().substr(1));
     EXPECT_EQ(settings.files[1], fits_image.string().substr(1));
     EXPECT_EQ(settings.files[2], hdf5_image.string().substr(1));
+}
+
+TEST_F(ProgramSettingsTest, ExpectedValuesLongJSON) {
+    // auto settings = SettingsFromString(
```
veggiesaurus: can we remove these commented out lines?

src/SessionManager/ProgramSettings.h
```diff
@@ -36,8 +38,18 @@ struct ProgramSettings {
...
+    bool no_user_config = false;
+    bool no_system_config = false;
+
+    nlohmann::json command_line_settings;
+    bool system_settings_json_exists = false;
+    bool user_settings_json_exists = false;
+
     ProgramSettings() = default;
     ProgramSettings(int argc, char** argv);
+    void CommandLineSettings(int argc, char** argv);
```
veggiesaurus: `ApplyCommandLineSettings` perhaps?

veggiesaurus: or `GetJSONConfigFromCommandLine` :shrug: `CommandLineSettings` by itself seems a bit vague.

test/TestProgramSettings.cc
jolopezl: Done

src/SessionManager/ProgramSettings.h
jolopezl: Done

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,99 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    }
+    if (j.contains("exit_timeout")) {
+        wait_time = j.at("exit_timeout");
+    }
+    if (j.contains("initial_timeout")) {
+        init_wait_time = j.at("initial_timeout");
+    }
+    if (j.contains("idle_timeout")) {
+        idle_session_wait_time = j.at("idle_timeout");
+    }
```
confluence: This is quite verbose. We're limited in how we can simplify this because we can't easily iterate over the destination variables or map them to the JSON values (although... would it make sense to store these values permanently *as* a JSON object?), but we could use [`value`](https://nlohmann.github.io/json/doxygen/classnlohmann_1_1basic__json_a11641b35219676b225d9bd15c7677659.html#a11641b35219676b225d9bd15c7677659):

    verbosity = j.value("verbosity", verbosity);
    no_log = j.value("no_log", no_log);
    (...)

Does this make sense? Does it look better?

src/SessionManager/ProgramSettings.cc
```diff
@@ -182,6 +283,55 @@ sending messages to the backend).
             files.push_back(relative_path.string());
         }
     }
+
+    // produce JSON for over-ridding system and user configuration;
```
confluence: over-ridding -> overriding

src/SessionManager/ProgramSettings.cc
```diff
@@ -182,6 +283,55 @@ sending messages to the backend).
...
+    // Options here need to match all options available for system and user settings
+    command_line_settings = json({}); // needs to have empty JSON at least in case of no command line options
+    if (result.count("verbosity")) {
```
confluence: This is also a lot of boilerplate to check if the keys exist. But I'm not sure if we can easily do anything about this because of the need to cast to different types.

test/TestProgramSettings.cc
```diff
@@ -214,4 +216,39 @@ TEST_F(ProgramSettingsTest, MultipleImagesFromPositional) {
...
+    EXPECT_EQ(settings.no_http, true);
+    EXPECT_EQ(settings.no_browser, true);
+    EXPECT_EQ(settings.host, "helloworld");
+    EXPECT_EQ(settings.port, 1234);
+    EXPECT_EQ(settings.grpc_port, 5678);
+    EXPECT_EQ(settings.omp_thread_count, 10);
+    EXPECT_EQ(settings.top_level_folder, "/tmp");
+    EXPECT_EQ(settings.frontend_folder, "/var");
+    EXPECT_EQ(settings.wait_time, 10);
+    EXPECT_EQ(settings.init_wait_time, 11);
```
confluence, line 261: We should also check the values of options which weren't set (because they weren't present in the config). If this is intended to be a user-editable configuration file, should we also test config with invalid values, and make sure that we can recover from them?

src/SessionManager/ProgramSettings.cc
veggiesaurus: that looks much neater

test/TestProgramSettings.cc
veggiesaurus: yes, we need these additional checks in the unit tests

src/SessionManager/ProgramSettings.cc
veggiesaurus: The only solution I can think of is something like 

```
vector<string> bool_keys = {"verbosity", "no_log", "log_performance", "log_protocol_messages", ...};

for(auto& key : bool_keys)) {
    if (result.count(key)) {
        command_line_settings[key] = result[key].as<bool>();
    }
}

// Repeat for int and string

```

src/SessionManager/ProgramSettings.cc
jolopezl: I agree with both of you. I implemented it both ways, and of course, using `value` is shorter, but it becomes harder to guess what's happening when you read it for the first time. I think it is more evident to know what's going on when you read:
```
    if (j.contains("verbosity")) {
        verbosity = j.at("verbosity");
    }
```
rather than:
```
verbosity = j.value("verbosity", verbosity);
```
without going to the reference/documentation.

I will use `value`.

src/SessionManager/ProgramSettings.cc
jolopezl: I agree with both of you.

test/TestProgramSettings.cc
jolopezl: I totally agree with you. I was, and I think I am still simplistic at this point, but I think it's better now.

jolopezl:
@confluence, @veggiesaurus, could you please recheck this PR. My apologies that this fell between the cracks. I had changes done without a push.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +38,141 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+
+    const fs::path user_settings_path = fs::path(getenv("HOME")) / CARTA_USER_FOLDER_PREFIX / "backend.json";
+    const fs::path system_settings_path = "/etc/carta/backend.json";
+
+    json settings;
+    if (fs::exists(system_settings_path) && !no_system_config) {
+        settings = JSONConfigSettings(system_settings_path.string());
+        system_settings_json_exists = true;
+        spdlog::info("System settings found. Command-line arguments and will override system settings");
+        spdlog::info("If you want to ignore system settings, use --no_system_config as command-line argument.");
```
confluence: These messages seem much too verbose for `info`. I would 

1. add a paragraph for the detailed precedence and option explanation to the program help message
2. change these just to print "Reading [system/user] settings from [path to file]"
3. change them to debug.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +38,141 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    }
+
+    if (fs::exists(user_settings_path) && !no_user_config) {
+        auto user_settings = JSONConfigSettings(user_settings_path.string());
+        user_settings_json_exists = true;
+        spdlog::info(
+            "User settings found, applying on top of system settings. Note that Command-line arguments will override system and user "
+            "settings");
+        spdlog::info("If you want to ignore system settings, use --no_user_config as command-line argument.");
```
confluence: As above.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +38,141 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    return j;
+}
+
+void ProgramSettings::SetSettingsFromJSON(const json& j) {
+    ValidateJSON(j); // check for invalid keys, store them in std::set for easy counting
+    // logic here is
+    // value = check invalid, if invalid then default value, otherwise check json and assign its value, otherwise default
+    if (!invalid_keys.count("verbosity")) {
+        verbosity = j.value("verbosity", verbosity);
+    }
```
confluence: As I recall we changed this from a slightly different syntax to be less verbose, but these invalid checks reintroduce the same verbosity.

I'm also not sure that `ValidateJSON` should be a separate function to this and use `invalid keys` to store intermediate state. Ideally validation and setting the value should be done once in the same loop(s).

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +38,141 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    if (!invalid_keys.count("initial_timeout")) {
+        init_wait_time = j.value("initial_timeout", init_wait_time);
+    }
+    if (!invalid_keys.count("idle_timeout")) {
+        idle_session_wait_time = j.value("idle_timeout", idle_session_wait_time);
+    }
+}
+
+void ProgramSettings::ValidateJSON(const json& j) {
```
confluence: Can we not turn the sets of keys into unordered maps, mapping the key values to pointers to the attributes, and set the attribute values here instead of doing it in a separate function?

~These collections could also be in this function's scope, since they are only used here. But maybe it's clearer to have them declared in the class scope.~ Sorry; I see they are also used in another place.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +38,141 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    for (const auto& key : int_keys) {
+        if (!j.count(key)) {
+            continue;
+        }
+        if (!j[key].is_number_integer()) {
+            spdlog::warn("Config file has problems, please check key with name {}, it's current value is {}", key, j[key]);
```
confluence: In all these messages "it's" should be "its", and the commas should be full stops.

confluence (changes requested):

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +38,141 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+        }
+    }
+    if (invalid_keys.size()) {
+        std::stringstream ss;
+        for (const auto& key : invalid_keys) {
+            ss << key << ", ";
+        }
+        auto res = ss.str();
+        res = res.substr(0, res.size() - 2); // remove traling space and comma
+        spdlog::warn("List of invalid keys in settings = {}", res);
```
confluence: This is redundant, since we already have individual warnings for each invalid entry.

src/SessionManager/ProgramSettings.cc
confluence: We should also mention the expected type in these messages.

confluence (changes requested):

src/SessionManager/ProgramSettings.cc
jolopezl: Ok.

src/SessionManager/ProgramSettings.cc
jolopezl: Ok, good point removing.

src/SessionManager/ProgramSettings.cc
jolopezl: We changed the code to another place to reduce verbosity. It was possible because the verbosity overhead came from string values but not from variables with different types. I think this validation check is important, and it's good to have it in a separate method. Otherwise, we would need to add another library to check against the JSON schema. If we mix the validation part into the settings assignment, I think the verbosity will only increase.

src/SessionManager/ProgramSettings.cc
jolopezl: Wouldn't that require the same amount of verbosity to define all map elements pointing to each different setting? Also, three different maps, one for each value type, same as with the keys. I agree that the code can be simplified, but it will look less clear.

src/SessionManager/ProgramSettings.cc
confluence: We changed to `.value` to avoid repeated `.contains` checks, but we have now gone back to the same problem but with a different membership check. I'll elaborate on my suggestion in my other comment.

src/SessionManager/ProgramSettings.cc
confluence: I think it will look much clearer to define the maps and iterate over the keys and values in a single set of loops both to validate the type and to set the value. We are halfway towards this solution anyway with the three sets of keys grouped by type. This will allow us to remove the additional pass over all the key values and attributes.

confluence: To clarify, I am suggesting that we convert the sets of keys to maps of keys to attribute pointers and use them once, not that we create an additional set of maps and use them in a separate set of three loops. _(Edited for grammar.)_

src/SessionManager/ProgramSettings.cc
jolopezl: I do remember that, but this adds another level of check for some type of safety. So, it's in principle a new change.

confluence: Yes, I understand -- I'm not objecting to the check, but to the verbosity of the explicit repetition (which is the kind of verbosity the original change removed). Hence my suggestion to combine this and the validation function into a single pass which uses maps of keys to attributes.

src/SessionManager/ProgramSettings.cc
jolopezl: Thanks for explaining your points in detail. I implemented the changes as you suggest.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,80 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    } catch (json::exception err) {
+        spdlog::warn("Config file {} has problems, please check", json_file_path);
+        spdlog::warn(err.what());
+    }
+    return j;
+}
+
+void ProgramSettings::SetSettingsFromJSON(const json& j) {
+    for (const auto& key : int_keys_map) {
+        if (!j[key.first].is_number_integer()) {
```
confluence: One last thing: shouldn't we check if `j.contains(key.first)` *before* the type validation here?

confluence (changes requested):

src/SessionManager/ProgramSettings.cc
jolopezl: Of course!

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,83 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    return j;
+}
+
+void ProgramSettings::SetSettingsFromJSON(const json& j) {
+    for (const auto& key : int_keys_map) {
+        if (!j.contains(key.first)) {
+            continue;
+        }
+        if (!j[key.first].is_number_integer()) {
+            spdlog::warn("Config file has problems, please check key with name {}. Its current value is {}, and a number is expected",
```
confluence: Very last comment: I would change these messages to something like:

"Problem in config file at key {}: current value is {}, ~and~but a number is expected."

It's slightly inconvenient that we can't tell the user what the file is because we've merged them, but I'm not sure if it's enough of a problem for us to re-run this function on the two files separately instead of merging them. What do you think?

(Edited text suggestion: I think "but" is clearer.)

confluence (changes requested):
Last comment about the warning message text.

src/SessionManager/ProgramSettings.cc
jolopezl: > Very last comment: I would change these messages to something like:

Feel free to add as many comments as needed. No worries.

> 
> "Problem in config file at key {}: current value is {}, and a number is expected."
> 

Thanks for the suggestion. I will change the message accordingly; also fix the number/boolean/string part such that the message is correct.

> It's slightly inconvenient that we can't tell the user what the file is because we've merged them, but I'm not sure if it's enough of a problem for us to re-run this function on the two files separately instead of merging them. What do you think?

I think it's our responsibility to give the best possible indication of what can be going wrong. I will adapt the code a little bit. It isn't a huge change. Just extending the JSONConfigSettings method to something like:
```cpp
json ProgramSettings::JSONConfigSettings(const std::string& json_file_path) {
    std::ifstream ifs(json_file_path, std::ifstream::in);
    json j;
    try {
        j = json::parse(ifs);
    } catch (json::exception err) {
        spdlog::warn("Config file {} has problems, please check", json_file_path);
        spdlog::warn(err.what());
    }
    for (const auto& key : int_keys_map) {
        if (j.contains(key.first) && !j[key.first].is_number_integer()) {
            spdlog::warn("Problem in config file {}, at key {}: current value is {}, and a number is expected.", json_file_path, key.first, j[key.first]);
        }
    }
    for (const auto& key : bool_keys_map) {
        if (j.contains(key.first) && !j[key.first].is_boolean()) {
            spdlog::warn("Problem in config file {}, at key {}: current value is {}, and a boolean is expected.", json_file_path, key.first, j[key.first]);
        }
    }
    for (const auto& key : strings_keys_map) {
        if (j.contains(key.first) && !j[key.first].is_string()) {
            spdlog::warn("Problem in config file {}, at key {}: current value is {}, and a string is expected.", json_file_path, key.first, j[key.first]);
        }
    }
    return j;
}
```
I would like to do something more elegant and get rid of those three loops and keep one with one fancy function, but I guess it's ok the way it is now.

Instead of storing valid keys elsewhere, as I attempted before with the std::set, I would rather test for their validity after merging all JSON settings again; what do you think?

```cpp
void ProgramSettings::SetSettingsFromJSON(const json& j) {
    for (const auto& key : int_keys_map) {
        if (!j.contains(key.first) || !j[key.first].is_number_integer()) {
            continue;
        }
        *key.second = j[key.first];
    }
    for (const auto& key : bool_keys_map) {
        if (!j.contains(key.first) || !j[key.first].is_boolean()) {
            continue;
        }
        *key.second = j[key.first];
    }
    for (const auto& key : strings_keys_map) {
        if (!j.contains(key.first) || !j[key.first].is_string()) {
            continue;
        }
        *key.second = j[key.first];
    }
}
```

confluence: I don't love looping over the keys twice, but this probably makes more sense conceptually than fully processing each file instead of merging (especially since the last thing we're merging is commandline parameters, not a file. I do have some specific remarks, which I will add in annotations.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,86 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    ApplyCommandLineSettings(argc, argv);
+
+    const fs::path user_settings_path = fs::path(getenv("HOME")) / CARTA_USER_FOLDER_PREFIX / "backend.json";
+    const fs::path system_settings_path = "/etc/carta/backend.json";
+
+    json settings;
+    if (fs::exists(system_settings_path) && !no_system_config) {
+        settings = JSONConfigSettings(system_settings_path.string());
+        system_settings_json_exists = true;
+        spdlog::info("Reading system settings from {}", system_settings_path.string());
```
confluence: Please add full stops to the ends of these messages.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,86 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    json j;
+    try {
+        j = json::parse(ifs);
+    } catch (json::exception& err) {
+        spdlog::warn("Config file {} has problems, please check:", json_file_path);
+        spdlog::warn(err.what());
+    }
+    for (const auto& key : int_keys_map) {
+        if (j.contains(key.first) && !j[key.first].is_number_integer()) {
+            spdlog::warn("Problem in config file {}, at key {}: current value is {}, and a number is expected.", json_file_path, key.first,
```
confluence: Please remove the comma before "at key". Maybe also change "and" to "but", but I don't feel that strongly about it.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,86 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+            spdlog::warn("Problem in config file {}, at key {}: current value is {}, and a string is expected.", json_file_path, key.first,
+                j[key.first]);
+        }
+    }
+    return j;
+}
+
+void ProgramSettings::SetSettingsFromJSON(const json& j) {
+    for (const auto& key : int_keys_map) {
+        if (!j.contains(key.first) || !j[key.first].is_number_integer()) {
```
confluence: If we're going to split this up like this, I think that we should *remove* the entries with bad types from the json object in `JSONConfigSettings`, instead of repeating the `is_type` checks here. Then we would only need to use `contains` here.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,86 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    std::ifstream ifs(json_file_path, std::ifstream::in);
+    json j;
+    try {
+        j = json::parse(ifs);
+    } catch (json::exception& err) {
+        spdlog::warn("Config file {} has problems, please check:", json_file_path);
+        spdlog::warn(err.what());
+    }
+    for (const auto& key : int_keys_map) {
+        if (j.contains(key.first) && !j[key.first].is_number_integer()) {
```
confluence, line 77: I think we should also remove these bad keys/values from the json object here, so that we don't have to re-check the same thing later.

src/SessionManager/ProgramSettings.cc
```diff
@@ -182,6 +270,25 @@ sending messages to the backend).
             files.push_back(relative_path.string());
         }
     }
+
+    // produce JSON for overridding system and user configuration;
+    // Options here need to match all options available for system and user settings
+    command_line_settings = json({}); // needs to have empty JSON at least in case of no command line options
+    for (const auto& key : int_keys_map) {
+        if (result.count(key.first)) {
+            command_line_settings[key.first] = result[key.first].as<int>();
```
confluence, line 286: What happens here if the value can't be converted to a number? And what string values are valid for booleans?

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,86 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+        settings.merge_patch(user_settings); // user on top of system
+    }
+
+    if (system_settings_json_exists || user_settings_json_exists) {
+        settings.merge_patch(command_line_settings); // force command-line on top of user and sytem
+        SetSettingsFromJSON(settings);
+    }
+}
+
+json ProgramSettings::JSONConfigSettings(const std::string& json_file_path) {
```
confluence: I would suggest changing this function name to something like `JSONFromConfigFile` or `JSONSettingsFromFile`.

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,86 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
...
+    std::ifstream ifs(json_file_path, std::ifstream::in);
+    json j;
+    try {
+        j = json::parse(ifs);
+    } catch (json::exception& err) {
+        spdlog::warn("Config file {} has problems, please check:", json_file_path);
```
confluence: I would suggest "Error parsing config file {}.".

confluence (changes requested):
Some text changes, a suggestion for avoiding a repeated check, and a question about commandline parameter types.

src/SessionManager/ProgramSettings.cc
jolopezl: I thought about this, and I kept believing that cxxopts is taking care of it. I took a fast look at https://git.io/JOW9B and other functions and didn't find a cross-check for type safety. I bet I am missing it. Maybe @veggiesaurus  can confirm?

veggiesaurus: cxxopts takes care of that. If you use the wrong type of argument it should just exit

src/SessionManager/ProgramSettings.cc
```diff
@@ -34,6 +37,89 @@ void applyOptionalArgument(T& val, const string& argument_name, const cxxopts::P
 }
 
 ProgramSettings::ProgramSettings(int argc, char** argv) {
+    if (argc > 1) {
+        spdlog::info("Using command-line settings");
```
confluence: I think this and the other two messages should be `debug`.

confluence (approved):
I think that the messages about the sources of the settings should be logged at `debug` level, but apart from that I'm happy with this.

src/SessionManager/ProgramSettings.cc
jolopezl: I attempted to change this to debug, but it won't print since the Logger is not yet initialized at that point. I would suggest removing that message line. It's quite obvious if you are starting the backend with or without command line arguments unless you are using an alias or something else. What do you think? Maybe @veggiesaurus wants to comment too.

This is how it usually looks like:
```
root@bc0736f1a5a3:~/carta-backend/build# ./carta_backend --verbosity 5
[2021-04-19 09:38:36.589] [info] Using command-line settings
[2021-04-19 09:38:36.593] [info] Writing to the log file: /root/.carta/log/carta.log
[2021-04-19 09:38:36.593] [info] /root/carta-backend/build/carta_backend: Version 2.0.0-dev.21.04.13
[2021-04-19 09:38:36.601] [warning] Could not find CARTA frontend files in directory /root/carta-backend/build/../share/carta/frontend.
[2021-04-19 09:38:36.601] [warning] Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument.
[2021-04-19 09:38:36.604] [info] Listening on port 3002 with top level folder /, starting folder /root/carta-backend/build. The number of OpenMP worker threads will be handled automatically.
^C[2021-04-19 09:38:37.963] [info] Exiting backend.
```

And here's proof that it won't appear if you set it to `spdlog::debug`, even if asking for verbosity level 5 (debug):
```
root@bc0736f1a5a3:~/carta-backend/build# ./carta_backend --verbosity 5
[2021-04-19 09:34:01.189] [info] Writing to the log file: /root/.carta/log/carta.log
[2021-04-19 09:34:01.189] [info] /root/carta-backend/build/carta_backend: Version 2.0.0-dev.21.04.13
[2021-04-19 09:34:01.191] [warning] Could not find CARTA frontend files in directory /root/carta-backend/build/../share/carta/frontend.
[2021-04-19 09:34:01.191] [warning] Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument.
[2021-04-19 09:34:01.191] [info] Listening on port 3002 with top level folder /, starting folder /root/carta-backend/build. The number of OpenMP worker threads will be handled automatically.
^C[2021-04-19 09:34:02.219] [info] Exiting backend.
```

Maybe, we should initialize the logger before just to show this message from the command-line settings if they are available. It can also provide further debugging messaging for the program settings processing. But, at the same time, it requires processing a few options ahead, see for instance:
```cpp
InitLogger(settings.no_log, settings.verbosity, settings.log_performance, settings.log_protocol_messages);
```

jolopezl:
I am not sure why, now after a few changes, the following test is failing:
```cpp
    auto json_string = R"(
    {
        "verbosity": "debug",
        "no_log": 1,
        "no_http": 1,
        "no_browser": 1,
        "host": true,
        "port": "1234",
        "grpc_port": "5678",
        "omp_threads": true,
        "top_level_folder": 1,
        "frontend_folder": false,
        "exit_timeout": "",
        "initial_timeout": ""
    })";
    nlohmann::json j;
    try {
        j = nlohmann::json::parse(json_string);
    } catch (nlohmann::json::exception& err) {
        GTEST_COUT << err.what() << "\n";
    }
```
with:
```
unknown file: Failure
C++ exception with description "[json.exception.type_error.302] type must be number, but is string" thrown in the test body.
[  FAILED  ] ProgramSettingsTest.ValidateJSON_Singlefield (2 ms)
```
At this point, I don't understand why `json::parse` would expect a given type, no reason for believing so.

src/SessionManager/ProgramSettings.cc
confluence: Why not defer the printing until after the logger is initialised in `Main.cc`? You could save a message summarising the source(s) of the configuration in its own field on the settings object, or pass a separate output parameter when calling `ProgramSettings` -- but that's a bit C-ish and may be kind of gross to do in a constructor, so maybe not. I think a single summarised message (at debug level) would be fine.

confluence:
That's weird. Can you do a binary search and narrow it down to a specific line?

veggiesaurus:
> I am not sure why, now after a few changes, the following test is failing:
> 
> ```c++
>     auto json_string = R"(
>     {
>         "verbosity": "debug",
>         "no_log": 1,
>         "no_http": 1,
>         "no_browser": 1,
>         "host": true,
>         "port": "1234",
>         "grpc_port": "5678",
>         "omp_threads": true,
>         "top_level_folder": 1,
>         "frontend_folder": false,
>         "exit_timeout": "",
>         "initial_timeout": ""
>     })";
>     nlohmann::json j;
>     try {
>         j = nlohmann::json::parse(json_string);
>     } catch (nlohmann::json::exception& err) {
>         GTEST_COUT << err.what() << "\n";
>     }
> ```
> 
> with:
> 
> ```
> unknown file: Failure
> C++ exception with description "[json.exception.type_error.302] type must be number, but is string" thrown in the test body.
> [  FAILED  ] ProgramSettingsTest.ValidateJSON_Singlefield (2 ms)
> ```
> 
> At this point, I don't understand why `json::parse` would expect a given type, no reason for believing so.

Isn't it because `port`, `grpc_port`, `exit_timeout` and `initial_timeout` should all be integers?

jolopezl:
> > I am not sure why, now after a few changes, the following test is failing:
> > ```c++
> >     auto json_string = R"(
> >     {
> >         "verbosity": "debug",
> >         "no_log": 1,
> >         "no_http": 1,
> >         "no_browser": 1,
> >         "host": true,
> >         "port": "1234",
> >         "grpc_port": "5678",
> >         "omp_threads": true,
> >         "top_level_folder": 1,
> >         "frontend_folder": false,
> >         "exit_timeout": "",
> >         "initial_timeout": ""
> >     })";
> >     nlohmann::json j;
> >     try {
> >         j = nlohmann::json::parse(json_string);
> >     } catch (nlohmann::json::exception& err) {
> >         GTEST_COUT << err.what() << "\n";
> >     }
> > ```
> > 
> > 
> > with:
> > ```
> > unknown file: Failure
> > C++ exception with description "[json.exception.type_error.302] type must be number, but is string" thrown in the test body.
> > [  FAILED  ] ProgramSettingsTest.ValidateJSON_Singlefield (2 ms)
> > ```
> > 
> > 
> > At this point, I don't understand why `json::parse` would expect a given type, no reason for believing so.
> 
> Isn't it because `port`, `grpc_port`, `exit_timeout` and `initial_timeout` should all be integers?

There's no type validation at this point. So, every field should be parsed to whatever it contains. It's funny, this part of the test is attempting to give the wrong value, and the program should reject those, but it's failing at the parsing level, so the test is not even run.

jolopezl:
> That's weird. Can you do a binary search and narrow it down to a specific line?

Of course, I am debugging the problem. I pointed it out here since it was pretty surprising to get this failure.

veggiesaurus:
> There's no type validation at this point. So, every field should be parsed to whatever it contains. It's funny, this part of the test is attempting to give the wrong value, and the program should reject those, but it's failing at the parsing level, so the test is not even run.

It's not failing in the `parse` call, it's failing in the `SetSettingsFromJSON` call, alt least on my machine.  Why do you check for the correct types in `JSONSettingsFromFile` but not in `SetSettingsFromJSON`? 

It's failing at `key.second = j[key.first];`
Stack:

```
__cxa_throw 0x00007f3114f0a750
void nlohmann::detail::from_json<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > >, int, 0>(nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&, int&) json.hpp:3722
decltype((from_json(fp, fp0)) , ((void)())) nlohmann::detail::from_json_fn::operator()<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > >, int>(nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&, int&) const json.hpp:3791
decltype((nlohmann::(anonymous namespace)::from_json(forward<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&>(fp), fp0)) , ((void)())) nlohmann::adl_serializer<int, void>::from_json<nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&, int>(nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&, int&) json.hpp:4386
int nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > >::get<int, int, 0>() const json.hpp:19421
nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > >::operator int<int, 0>() const json.hpp:19723
carta::ProgramSettings::SetSettingsFromJSON(nlohmann::basic_json<std::map, std::vector, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, long, unsigned long, double, std::allocator, nlohmann::adl_serializer, std::vector<unsigned char, std::allocator<unsigned char> > > const&) ProgramSettings.cc:106
ProgramSettingsTest_ValidateJSON_Allfields_Test::TestBody() TestProgramSettings.cc:286
void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) 0x00005585d6124ee1
void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) 0x00005585d6124eda
testing::Test::Run() 0x00005585d6119ba6
testing::Test::Run() 0x00005585d6119b66
testing::TestInfo::Run() 0x00005585d6119d05
testing::TestInfo::Run() 0x00005585d6119cf0
testing::TestSuite::Run() 0x00005585d6119ded
testing::TestSuite::Run() 0x00005585d6119d52
testing::internal::UnitTestImpl::RunAllTests() 0x00005585d611a30c
bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) 0x00005585d6125451
bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) 0x00005585d612544a
testing::UnitTest::Run() 0x00005585d611a53c
RUN_ALL_TESTS() gtest.h:2473
main TestMain.cc:11
__libc_start_main 0x00007f3114c780b3
_start 0x00005585d5bfe1be
```

veggiesaurus:
> > That's weird. Can you do a binary search and narrow it down to a specific line?
> 
> Of course, I am debugging the problem. I pointed it out here since it was pretty surprising to get this failure.

Are you building and debugging in release mode by any chance? that might be why you're getting errors that point back to the incorrect location (`parse`)

src/SessionManager/ProgramSettings.cc
veggiesaurus: I don't see why we need to print this line in the first place. If you put command-line options in and they are incorrect, we'll get an error. If you put correct ones in, I'm pretty sure you'd expect them to be used.

veggiesaurus: The other issue here is that this message will _always_ go to stdout, because the log hasn't been initialised. So your actual log file will be missing this line anyway. That could get rather confusing.

confluence:
The keys with bad types should be removed after failing the type check, so that we don't have to redo the same type check later. But perhaps that isn't doing what we expect.

jolopezl:
> > > That's weird. Can you do a binary search and narrow it down to a specific line?
> > 
> > 
> > Of course, I am debugging the problem. I pointed it out here since it was pretty surprising to get this failure.
> 
> Are you building and debugging in release mode by any chance? that might be why you're getting errors that point back to the incorrect location (`parse`)

Exactly because of that. Thanks for pointing it out. That solves my silly puzzle. I'll need some additional changes then. This is a good example of why test-driven development is good.

confluence:
Oh, of course; the tests bypass the function which does the type check and removes the bad entries. So perhaps that function should accept a string instead, so that we can use it from the tests.

jolopezl:
> The keys with bad types should be removed after failing the type check, so that we don't have to redo the same type check later. But perhaps that isn't doing what we expect.

I agree with you. I was thinking in the same direction, but I guess this needs another level of checks since one of the settings files could be wrong, and we might want just to skip that one or the other. The command-line settings are always ok thanks to `cxxopts`.

confluence:
We already apply the type check to each file individually -- the problem is that we don't use this function in the tests, because we use a hardcoded string instead of reading a string from a file.

jolopezl:
@confluence, @veggiesaurus, I finally addressed this problem:

> 
> ```
> unknown file: Failure
> C++ exception with description "[json.exception.type_error.302] type must be number, but is string" thrown in the test body.
> [  FAILED  ] ProgramSettingsTest.ValidateJSON_Singlefield (2 ms)
> ```
> 
> At this point, I don't understand why `json::parse` would expect a given type, no reason for believing so.

and it has nothing to do with the type safety we implemented to cross-check user input but actually using the JSON object correctly. The exception arises when printing information:

```cpp
            spdlog::warn("Problem in config file {} at key {}: current value is {}, but a number is expected.", json_file_path, key.first,
                j[key.first]);
```

Here, `j[key.first]` needs to be a string, so it fails when reading something else (as we attempt through testing). The following fixes the problem:

```cpp
            spdlog::warn("Problem in config file {} at key {}: current value is {}, but a number is expected.", json_file_path, key.first,
                j[key.first].dump());
```

I am adding adequate changes so that we get it right. At the same time, I have adapted the testing to more appropriately test the code. Special thanks to @confluence for the offline discussion on this part that made this take the right direction.

test/TestProgramSettings.cc
```diff
@@ -214,4 +221,82 @@ TEST_F(ProgramSettingsTest, MultipleImagesFromPositional) {
...
+}
+
+TEST_F(ProgramSettingsTest, ValidateJSONFromFile_BadFields) {
+    const std::string input = "./data/settings-bad-fields.json";
+
+    carta::ProgramSettings settings;
+    auto j = settings.JSONSettingsFromFile(input);
+    std::cout << "Object: " << j.dump() << std::endl;
```
confluence: Left in by mistake?

test/TestProgramSettings.cc
```diff
@@ -214,4 +221,82 @@ TEST_F(ProgramSettingsTest, MultipleImagesFromPositional) {
...
+    EXPECT_EQ(settings.wait_time, 10);
+    EXPECT_EQ(settings.init_wait_time, 11);
+}
+
+TEST_F(ProgramSettingsTest, ValidateJSONFromFile) {
+    const std::string input = "./data/settings-good-fields.json";
+
+    carta::ProgramSettings settings;
+    auto j = settings.JSONSettingsFromFile(input);
+    settings.SetSettingsFromJSON(j);
```
confluence, line 289: Because we're testing `JSONSettingsFromFile` in these tests, maybe we should be checking the JSON object output by that function, rather than the final settings object. There could be a separate test to make sure that `SetSettingsFromJSON` correctly leaves some or all values set to default if provided with a JSON object which is empty or has some entries missing. Right now these tests are end-to-end tests of both functions together.

test/TestProgramSettings.cc
jolopezl: Ouh...

src/SessionManager/ProgramSettings.cc
confluence: The message about the commandline options is the least relevant, but the message(s) about the locations of the config files could be useful (at debug level).

Those messages should either be logged at debug level after the logger is initialised or not printed at all -- I think this is too verbose for info.

test/TestProgramSettings.cc
```diff
@@ -214,4 +221,104 @@ TEST_F(ProgramSettingsTest, MultipleImagesFromPositional) {
...
+
+TEST_F(ProgramSettingsTest, ValidateJSONFromFileWithBadFields) {
+    const std::string input = "./data/settings-bad-fields.json";
+    carta::ProgramSettings settings;
+    auto j = settings.JSONSettingsFromFile(input);
+    settings.SetSettingsFromJSON(j);
+    EXPECT_EQ(j.size(), 0);
+}
+
+TEST_F(ProgramSettingsTest, TestValuesFromGoodSettings) {
```
confluence, line 293: I meant that in these tests we should parse a JSON string. We're still testing two functions at once here. The JSON string should contain the cleaned JSON with all bad values removed.

confluence (changes requested):

test/TestProgramSettings.cc
jolopezl: Hi @confluence, I think that's what's done in `ValidateJSONFromFileWithGoodFields`:
```cpp
TEST_F(ProgramSettingsTest, ValidateJSONFromFileWithGoodFields) {
    const std::string input = "./data/settings-good-fields.json";
    carta::ProgramSettings settings;
    auto j = settings.JSONSettingsFromFile(input);
    EXPECT_EQ(j.size(), 12);
    EXPECT_EQ(j["verbosity"], 6);
    EXPECT_EQ(j["port"], 1234);
    EXPECT_EQ(j["grpc_port"], 5678);
    EXPECT_EQ(j["omp_threads"], 10);
    EXPECT_EQ(j["exit_timeout"], 10);
    EXPECT_EQ(j["initial_timeout"], 11);
    EXPECT_EQ(j["no_log"], true);
    EXPECT_EQ(j["no_http"], true);
    EXPECT_EQ(j["no_browser"], true);
    EXPECT_EQ(j["host"], "helloworld");
    EXPECT_EQ(j["top_level_folder"], "/tmp");
    EXPECT_EQ(j["frontend_folder"], "/var");
}
```

jolopezl: >The message about the commandline options is the least relevant, but the message(s) about the locations of the config files could be useful (at debug level).

>Those messages should either be logged at debug level after the logger is initialised or not printed at all -- I think this is too verbose for info.

I agree with you. What I can do for the moment is to set them as debug. Still, the messages will never be printed since the logger defaults haven't changed at this point. We can later iterate on this behavior. What do you think? For the moment, I think that reducing this additional verbosity is important.

jolopezl: I am tagging @veggiesaurus since he developed this part of the code, I think. 

What do you think if we initialize the logger immediately after parsing the first command-line settings if these include a higher level of verbosity? That would help to debug problems heere. If something else needs to be done in the current position of `InitLogger` in `Main.cpp`. We can reset call the function again. Am I missing something important there?

confluence: 1. `ValidateJSONFromFileWithGoodFields` and `ValidateJSONFromFileWithBadFields` test that `JSONSettingsFromFile` loads settings from a file correctly. `TestValuesFromGoodSettings` and `TestDefaultsFallbackFromBadSettings` should be testing **only** `SetSettingsFromJSON`. They should not be calling `JSONSettingsFromFile` at all. They should be constructing JSON objects directly from strings and passing those JSON objects to `SetSettingsFromJSON`. These objects should not contain any values with incorrect types, because those would have been removed by `JSONSettingsFromFile` in the actual code.
2. I would prefer not to leave in code that we know isn't working. If we aren't going to defer printing the config file location messages now, we should remove them entirely.

jolopezl: Ok, I understand your point now. Thanks for explaining it further. I don't find that repeating the operation and reading from a file, then passing it to `SetSettingsFromJSON` is wrong. We know from the first test that the output from `JSONSettingsFromFile` is the expected (a good JSON and an empty JSON), then we want to check that `SetSettingsFromJSON` works properly. IMHO the first part is not being tested; at this point, it's only preparing the input for the test. Of course, this requires that the JSON files are kept consistent.

confluence: Now that I'm thinking about this, I don't think we should print those warnings about bad keys immediately either. If the user wants to set the loglevel to > warning, we don't want to print warnings before the log is configured.

I would prefer to save any messages that arise during parsing of options, and print them later once the logger is initialised, than to move the initialisation of the logger earlier. I don't see how we could do the latter correctly -- we won't know what the verbosity is until we have finished parsing the options, so we shouldn't log messages before then.

If we're going to save the messages, we'll also need to save the loglevels -- so perhaps a list of level / message tuples. I'm not sure if spdlog has a generic function which takes a string level parameter; if not I guess we could save the actual functions by reference.

confluence:
Note that #800 is about to get merged -- that will add another preference.

test/TestProgramSettings.cc
jolopezl: That's a good point, indeed. I changed that code in the direction of your suggestion, but being a bit more empirical, with a vector storing warnings and debug messages that are flushed after the logger is initialized. I prefer this over a list of tuples that requires conditionals for deciding the print level.

jolopezl:
> Note that #800 is about to get merged -- that will add another preference.

I prefer to open another short pull request for adding additional preferences. That would serve as a simplistic example for that task.

jolopezl:
> > Note that #800 is about to get merged -- that will add another preference.
> 
> I prefer to open another short pull request for adding additional preferences. That would serve as a simplistic example for that task.

maybe not anymore since it got merged already...

src/SessionManager/ProgramSettings.h
```diff
@@ -47,6 +88,15 @@ struct ProgramSettings {
     }
     bool operator!=(const ProgramSettings& rhs) const;
     bool operator==(const ProgramSettings& rhs) const;
+
+    std::vector<std::string> warning_msgs;
+    std::vector<std::string> debug_msgs;
+    void flushMessages() {
```
confluence: This needs to be called FlushMessages to fit the naming convention.

confluence (changes requested):
Apart from the function name, this looks good.

confluence (approved):

veggiesaurus (approved):

PR #784, opened 2021-03-26
541 Fix rectangle export bug

### LINKED ISSUE #541: position offset seen in exported region

pford:
Closes #541 .

Fixed error which assigned double pixel coordinates to int vector before being converted to world coordinates for export.  Was not a wcslib error after all.

### DIFF
--- a/src/Region/RegionImportExport.cc
+++ b/src/Region/RegionImportExport.cc
@@ -358,7 +358,7 @@ bool RegionImportExport::ConvertRecordToRectangle(
     try {
         // Convert center position to world coords
         casacore::Vector<casacore::Double> world_center;
-        casacore::IPosition pixel_center(_coord_sys->nPixelAxes(), 0);
+        casacore::Vector<casacore::Double> pixel_center(_coord_sys->nPixelAxes(), 0.0);
         pixel_center(0) = cx;
         pixel_center(1) = cy;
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
This fix works! 👍

PR #786, opened 2021-03-30
Added a default FITS image, mostly for use by the controller

confluence:
This is a default image for the Ubuntu package to install to a system data directory, and for the controller to use in its default unconfigured state.


kswang1029:
Nice

default fits-image-2021-3-30-20-44-52

veggiesaurus:
@confluence don't you need to edit the `CMakeLists.txt` to ensure this gets copied to `/usr/local/share/carta` when installing?

confluence:
For the Ubuntu packages it would be `/usr/share/carta`. I was planning to leave this to the packages, like the icons. I'm not sure how easily we can find an appropriate data directory in the cmake config which will work with a source install on any platform.

veggiesaurus:
> For the Ubuntu packages it would be `/usr/share/carta`. I was planning to leave this to the packages, like the icons. I'm not sure how easily we can find an appropriate data directory in the cmake config which will work with a source install on any platform.

Wouldn't it be `${CMAKE_INSTALL_PREFIX}/share/carta`?

confluence:
Hmm, maybe that's fine. I guess this is simpler than icons. I will add it.

Issue #628, opened 2020-09-09
support stokes for the statistics and histogram widgets


kswang1029:
related to the frontend issue: https://github.com/CARTAvis/carta-frontend/issues/1171

This requires ICD changes and backend implementation.

PR #790, opened 2021-04-07
628 Support stokes for statistics and histogram widgets

markccchiang:
For #628, the backend supports the stokes option for statistics and histogram widgets. It works with the modified protobuf branch [mark/stokes_for_stat_and_hist_calc](https://github.com/CARTAvis/carta-protobuf/tree/mark/stokes_for_stat_and_hist_calc) and frontend branch [thchang/stokes_dropdown_in_histogram_and_statistics](https://github.com/CARTAvis/carta-frontend/tree/thchang/stokes_dropdown_in_histogram_and_statistics).


pford (approved):

confluence (approved):

kswang1029:
@TienHao any chance to have the frontend part ready for testing soon?

TienHao:
> @TienHao any chance to have the frontend part ready for testing soon?

@kswang1029 a draft PR(https://github.com/CARTAvis/carta-frontend/pull/1570) for the frontend part has been opened. It is ready for testing.

kswang1029:
@markccchiang The following screenshot is with your branch. If we request a histogram from region 1 and stokes Q explicitly, we see the histogram is not identical to the one generated with the current dev backend with the same setup.
Screen Shot 2021-07-20 at 2 05 44 PM

from my initial investigation, it appears that the histogram bins are not identical. Left is this branch and right is dev.
Screen Shot 2021-07-20 at 2 14 45 PM

could you check if the histogram calculation is correct?

kswang1029:
@acdo2002 please help on testing this, especially the correctness.

acdo2002:
@kswang1029 Ok I am checking them (backend + frontend) together now.

acdo2002:
I use `HH211_IQU.fits` and set two regions **manually**, at lease the export data of Statistics widget and Histogram widget are identical to the current dev branch (I use Animator widget to change the Stokes I Q U, and use `diff xx1.tsv xx2.tsv` to check the exported data difference).

Region 1 (exclude NaN region):
Center 250 700
Size 10 10
Bottom left 245 695
Top right 255 705
PA 0

Region 2 (include part of NaN region):
Center 245 200
Size 400 300
PA 45

markccchiang:
@kswang1029 If I set a region that is larger than the image and then checks its histogram. The results are shown below:

my branch

Compared the histogram with the same setting on the dev branch:

dev branch

 They looks identical. Not sure are there some non-NAN pixels outside the circle from the `IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.image`?

kswang1029:
> @kswang1029 If I set a region that is larger than the image and then checks its histogram. The results are shown below:
> 
> my branch
> 
> Compared the histogram with the same setting on the dev branch:
> 
> dev branch
> 
> They looks identical. Not sure are there some non-NAN pixels outside the circle from the `IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.image`?

there should be no non-NAN inside the square and outside the circle. Are you able to reproduce the issue if you define a region between the square and the circle like what I did before?

acdo2002:
I perhaps can reproduce this issue (discrepancy between regions that all cover the data).
However this issue already exists in the current dev branch.

I set two regions:
Region 1 Square:
Center 121 120
Size 190 189
Bottom left 26 25.5
Top right 216 214.5
PA 0

Region 2 Circle: 
Center 128 136
Size 100 100

regions_discrepancy

The exported data of statistics widget are identical between two regions (both dev and this branch).
However the exported data of histogram widget are different between two regions (both dev and this branch).

The exported data of statistics widget & histogram widget (given the same stoke & channel), this branch is identical to the current dev.

markccchiang:
> I perhaps can reproduce this issue (discrepancy between regions that all cover the data).
> However this issue already exists in the current dev branch.
> 
> I set two regions:
> Region 1 Square:
> Center 121 120
> Size 190 189
> Bottom left 26 25.5
> Top right 216 214.5
> PA 0
> 
> Region 2 Circle:
> Center 128 136
> Size 100 100
> 
> regions_discrepancy
> 
> The exported data of statistics widget are identical between two regions (both dev and this branch).
> However the exported data of histogram widget are different between two regions (both dev and this branch).
> 
> The exported data of statistics widget & histogram widget (given the same stoke & channel), this branch is identical to the current dev.

I think this problem is due to the number of bins setting in the histogram calculation. When the frontend chooses `AUTO_BIN_SIZE,` the backend will calculate the number of bins based on the region shape. So if you change the region shape, the number of bins changed, and so does the region histogram plot. Also, if you move the region without changing its shape. Since the region shape is converted from the coordinate system, it may have a slight difference in different positions. So the region histogram plot changed too. @kswang1029 is it OK if we just set the number of bins equal to that in the `Render Configuration` widget (on the left button panel) when `AUTO_BIN_SIZE`?. I think that way would not cause misleading to users.

kswang1029:
> > I perhaps can reproduce this issue (discrepancy between regions that all cover the data).
> > However this issue already exists in the current dev branch.
> > I set two regions:
> > Region 1 Square:
> > Center 121 120
> > Size 190 189
> > Bottom left 26 25.5
> > Top right 216 214.5
> > PA 0
> > Region 2 Circle:
> > Center 128 136
> > Size 100 100
> > regions_discrepancy
> > The exported data of statistics widget are identical between two regions (both dev and this branch).
> > However the exported data of histogram widget are different between two regions (both dev and this branch).
> > The exported data of statistics widget & histogram widget (given the same stoke & channel), this branch is identical to the current dev.
> 
> I think this problem is due to the number of bins setting in the histogram calculation. When the frontend chooses `AUTO_BIN_SIZE,` the backend will calculate the number of bins based on the region shape. So if you change the region shape, the number of bins changed, and so does the region histogram plot. Also, if you move the region without changing its shape. Since the region shape is converted from the coordinate system, it may have a slight difference in different positions. So the region histogram plot changed too. @kswang1029 is it OK if we just set the number of bins equal to that in the `Render Configuration` widget (on the left button panel) when `AUTO_BIN_SIZE`?. I think that way would not cause misleading to users.

@markccchiang uh right. It is region-shape dependent. We have a plan to enhance the histogram widget by allowing users to specify a custom bin count and range etc. It is likely to be scheduled for the final v3 release. So let’s leave it for now. Thanks for checking.

kswang1029 (approved):

veggiesaurus:
@markccchiang can you make a PR for the protobuf changes? (including a major version bump and changelog entries). We will need to also bump the ICD version in the frontend and backend repos, since this is a breaking change.

Issue #652, opened 2020-10-27
backend crash when cancelling moment calculations with per-plane-beam cube


kswang1029:
Before generating moment images, the backend will perform beam convolutions first if the cube has per-plane-beam. When it is convolving and if we hit the cancel button, the backend will crash.

markccchiang:
There is no cancellation mechanism for the beam convolution. To solve it we may need to modify the casa object `casa::Image2DConvolver`.

Issue #602, opened 2020-08-21
improving HDF5 moment generation with ZYX cube 


kswang1029:
as a performance enhancement of the moment generator, when the input image is in the HDF5 format, ZXY cube should be used instead of the XYZ cube.

confluence:
We may need a complete reimplementation, similar to the Z profiles and the flux density stat -- it's difficult to reuse casa's functions with custom data. But I have a general idea for creating an HDF5 image object which reads from the rotated dataset and using it with casacore (somehow) by specifying custom axes, slicers, etc.. If this is possible, we may be able to do the same for the Z profiles.

kswang1029:
this overlaps withhttps://github.com/CARTAvis/carta-backend/issues/892

PR #791, opened 2021-04-07
652 Cancel beam convolutions in the moments calculation

markccchiang:
This branch is to solve #652 by adding a cancellation mechanism for `carta::Image2DConvolver`. The `carta::Image2DConvolver` is exported and modified from the `casa::Image2DConvolver`. Just like the `carta::ImageMoments` is exported and modified from the `casa::ImageMoments`. A tester is added to check the consistency between `carta::ImageMoments/carta::Image2DConvolver` and `casa::ImageMoments/casa::Image2DConvolver` for moments calculations. There is also a little clean-up of the code for image moments.


kswang1029:
@markccchiang the integrated progress (first 50% from convolution and the rest for moment calculations) works properly. 👍 When I cancel the request while the backend is busy on convolution, I see 
Screen Shot 2021-04-25 at 2 01 59 PM
Any idea? The backend and frontend are still functional though.

markccchiang:
> Any idea? The backend and frontend are still functional though.

Thanks for the reminding! I think have fixed it by ending the moments calculation process immediately if the beam convolution is canceled. Please check it again.

kswang1029 (approved):
feature-wise this is good! 👍

confluence:
Is there a long-term plan to submit the modified casa code as a patch to casa?

As an alternative strategy, we could avoid keeping these kinds of modified casa / casacore files in the CARTA codebase by explicitly forking casacore and casa and basing our custom carta-casacore build on those forked versions, rather than directly on casacore and casa.

Now that we have committed more explicitly to installing our casacore+casa build in a custom location, so that it doesn't clash with more standard installed versions, I think we have more leeway to incorporate these kinds of modifications. But that would require discussion and reorganisation which we should probably defer until after the 2.0 release.

There isn't really an ideal solution for maintaining all of these modifications ourselves, but doing it that way would at least maintain clearer separation between casa and CARTA code. It would also make changes much easier to review, since they would be shown as diffs against the original casa code.

kswang1029:
We could consider for post v2.0 to implement our own moment generator with beam convolution. The actual procedures are not that complicated. We will just need to make sure they are implemented in an efficient way. (same idea for pv image generator)

veggiesaurus:
> We could consider for post v2.0 to implement our own moment generator with beam convolution. The actual procedures are not that complicated. We will just need to make sure they are implemented in an efficient way. (same idea for pv image generator)

I suspect #602 would require our own implementation as well.

kswang1029:
> > We could consider for post v2.0 to implement our own moment generator with beam convolution. The actual procedures are not that complicated. We will just need to make sure they are implemented in an efficient way. (same idea for pv image generator)
> 
> I suspect #602 would require our own implementation as well.

Yes I guess so. So we might get it done at the same time.

kswang1029:
If the current approach as a temporary solution to fix the backend crash is not healthy to the codebase even for a short term, maybe we can consider to do the following:

we still use what imageanalysis provides and block the cancellation function at the frontend _only_ when beam convolution is ongoing. Once it is done and calculating moment images, we reenable the cancellation function at the frontend. This will require a bit frontend change too though.

confluence (changes requested):
Please merge the latest changes from dev. Apart from that, I think it's OK to merge this for now and revisit the issue of using modified casa code here ASAP after the release.

test/TestMoment.cc
```diff
@@ -0,0 +1,204 @@
...
+            auto carta_moment_image = dynamic_pointer_cast<casacore::ImageInterface<casacore::Float>>(carta_results[i]);
+
+            EXPECT_EQ(casa_moment_image->shape().size(), carta_moment_image->shape().size());
+            CompareImageData(casa_moment_image, carta_moment_image);
+        }
+    }
+};
+
+TEST_F(MomentTest, CheckConsistency) {
+    string file_name = "data/images/fits/M17_SWex_unittest.fits";
```
veggiesaurus, line 161: perhaps we could use our existing 10x10x10 noise image, instead of adding this file?

kswang1029: That is with NaN pixel

test/TestMoment.cc
```diff
@@ -0,0 +1,204 @@
...
+        casacore::Slicer section(start, end, casacore::Slicer::endIsLast);
+        data.resize(section.length().product());
+        casacore::Array<float> tmp(section.length(), data.data(), casacore::StorageInitPolicy::SHARE);
+        casacore::SubImage<float> subimage(*image, section);
+        casacore::RO_MaskedLatticeIterator<float> lattice_iter(subimage);
+
+        for (lattice_iter.reset(); !lattice_iter.atEnd(); ++lattice_iter) {
+            casacore::Array<float> cursor_data = lattice_iter.cursor();
+
+            if (image->isMasked()) {
```
veggiesaurus: are you actually testing any data with a mask? is this code used?

test/TestMoment.cc
veggiesaurus: ok. Then I suggest we add a similar image with minimal headers / trimmed history etc. That M17 image has a huge number of headers that are not needed

kswang1029: I pushed an updated FITS file without those lengthy history in header. Please check 73f110e.

confluence (approved):

test/TestMoment.cc
veggiesaurus: Can you do the same thing with the other image as well?

kswang1029: done with 7318425

PR #792, opened 2021-04-08
1271 Change ZFP compression quality with respect to the compressibility

markccchiang:
This branch is trying to solve [#1271](https://github.com/CARTAvis/carta-frontend/issues/1271) in another way that suggested by @kswang1029. We check ZFP compressibility for each tile of an image. If the compressibility is smaller than `0.05` then we compress the data again with the highest precision (32). If the compressibility for a tile with the highest precision is smaller than `0.1`, then we use such compression data. Otherwise, we use the compression data with the default precision.

### DIFF
--- a/src/Frame.cc
+++ b/src/Frame.cc
@@ -433,7 +433,7 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
                         tile_ptr->set_image_data(compression_buffer_hq.data(), compressed_size_hq);
 
                         spdlog::debug("Using high compression quality. Previous compression ratio: {:.3f}", compression_ratio);
-                        compression_ratio = compressed_size_hq;
+                        compression_ratio = compression_ratio_hq;
                         use_high_precision = true;
                     }
                 }
--- a/src/Frame.cc
+++ b/src/Frame.cc
@@ -32,6 +32,8 @@ namespace fs = boost::filesystem;
 namespace fs = std::filesystem;
 #endif
 
+static const int HIGH_COMPRESSION_QUALITY(32);
+
 using namespace carta;
 
 Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string& hdu, int default_z)
@@ -392,6 +394,8 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
     int tile_width;
     int tile_height;
     if (GetRasterTileData(tile_image_data, tile, tile_width, tile_height)) {
+        size_t tile_image_data_size = sizeof(float) * tile_image_data.size(); // tile image data size in bytes
+
         if (ZStokesChanged(z, stokes)) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


src/Frame.cc
```diff
@@ -402,18 +403,62 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
-            int precision = lround(compression_quality);
-            Compress(tile_image_data, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
-            tile_ptr->set_image_data(compression_buffer.data(), compressed_size);
+            {
+                // compress the data with a default precision
+                std::vector<char> compression_buffer;
+                size_t compressed_size;
+                int precision = lround(compression_quality);
+                Compress(tile_image_data, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
+                float compressibility = ((float)nan_encodings_size + (float)compressed_size) / (float)tile_image_data_size;
```
veggiesaurus: I think we should exclude the `nan_encodings` from the compressibility calculations. @kswang1029's investigations were all based on images without any nan encoding changes.

src/Frame.cc
```diff
@@ -402,18 +403,62 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
+
+                if (precision < HIGH_COMPRESSION_QUALITY && compressibility < 0.05) {
+                    // re-get the tile image data
+                    std::vector<float> tile_image_data_2;
+                    int tile_width_2;
```
veggiesaurus: why do we need to get the tile data again?

kswang1029:
@markccchiang @veggiesaurus I briefly tested with clean component images and it works nicely as we hope so! 

@markccchiang could you refine the log message to "info" and add the compression ratios before and after? At the moment it is not clear it is before or after upgrade.

> [2021-04-12 10:59:26.020] [warning] Change to higher compression quality: 11->32. The compressed ratio for tile (layer:3, x:7, y:6) is 0.002.

veggiesaurus:
> @markccchiang @veggiesaurus I briefly tested with clean component images and it works nicely as we hope so! 
> 
> @markccchiang could you refine the log message to "info" and add the compression ratios before and after? At the moment it is not clear it is before or after upgrade.
> 
> > [2021-04-12 10:59:26.020] [warning] Change to higher compression quality: 11->32. The compressed ratio for tile (layer:3, x:7, y:6) is 0.002.
> 

This seems like something that should be a debug message to me. It's of no real interest to the user is it?

kswang1029:
> > @markccchiang @veggiesaurus I briefly tested with clean component images and it works nicely as we hope so!
> > @markccchiang could you refine the log message to "info" and add the compression ratios before and after? At the moment it is not clear it is before or after upgrade.
> > > [2021-04-12 10:59:26.020] [warning] Change to higher compression quality: 11->32. The compressed ratio for tile (layer:3, x:7, y:6) is 0.002.
> 
> This seems like something that should be a debug message to me. It's of no real interest to the user is it?

ok debug is sensible. @markccchiang sorry for the confusion.

src/Frame.cc
```diff
@@ -402,18 +403,55 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
+                        spdlog::debug("The compression ratio for tile (layer:{}, x:{}, y:{}) is {:.3f}.", tile.layer, tile.x, tile.y,
+                            compression_ratio);
+                    }
+                } else {
+                    // set compression data with default precision
+                    raster_tile_data.set_compression_quality(compression_quality);
+                    tile_ptr->set_nan_encodings(nan_encodings.data(), sizeof(int32_t) * nan_encodings.size());
+                    tile_ptr->set_image_data(compression_buffer.data(), compressed_size);
+                    spdlog::debug(
+                        "The compression ratio for tile (layer:{}, x:{}, y:{}) is {:.3f}.", tile.layer, tile.x, tile.y, compression_ratio);
```
confluence: There is a lot of code repetition here which could be eliminated. The message about the ratio change could be a separate message.

src/Frame.cc
```diff
@@ -402,18 +403,55 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
+                std::vector<char> compression_buffer;
+                size_t compressed_size;
+                int precision = lround(compression_quality);
+                Compress(tile_image_data, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
+                float compression_ratio = (float)compressed_size / (float)tile_image_data_size;
+
+                if (precision < HIGH_COMPRESSION_QUALITY && compression_ratio < 0.05) {
+                    // re-compress the data with a higher precision
+                    std::vector<char> compression_buffer_2;
+                    size_t compressed_size_2;
```
confluence: The `_2` suffix is meaningless. If we want something short, what about `_hq`?

src/Frame.cc
```diff
@@ -402,18 +403,55 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
+                        tile_image_data, 0, compression_buffer_2, compressed_size_2, tile_width, tile_height, HIGH_COMPRESSION_QUALITY);
+                    float compression_ratio_2 = (float)compressed_size_2 / (float)tile_image_data_size;
+
+                    if (compression_ratio_2 < 0.1) {
+                        // set compression data with high precision
+                        raster_tile_data.set_compression_quality(HIGH_COMPRESSION_QUALITY);
+                        tile_ptr->set_nan_encodings(nan_encodings.data(), sizeof(int32_t) * nan_encodings.size());
+                        tile_ptr->set_image_data(compression_buffer_2.data(), compressed_size_2);
+                        spdlog::debug(
+                            "Change to high compression quality: {}->{}. The change of compression ratio for tile (layer:{}, x:{}, y:{}) "
```
confluence: I suggest splitting this into the normal message about the ratio and an additional message which says something like:

"Using high compression quality. Previous compression ratio: {:.3f}"

~I don't think it's necessary to mention the low precision because we don't normally, or the high precision because it's a constant.~

confluence (changes requested):

src/Frame.cc
kswang1029: The zfp precision here can be even greater than 32. 32 here is our choice of the upper limit we provide to users. It would be still good to mention we change from which lower precision to what we selected as high precision in the debug message just in case in the future we might need it for debugging purpose. Same for the compression rate as for now we manually set 0.05 and 0.1 as thresholds to determine if we can upgrade the zfp precision.

confluence: Fair enough; I'll edit my comment.

src/Frame.cc
```diff
@@ -409,11 +410,44 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
             }
 
             auto t_start_compress_tile_data = std::chrono::high_resolution_clock::now();
-            std::vector<char> compression_buffer;
-            size_t compressed_size;
-            int precision = lround(compression_quality);
-            Compress(tile_image_data, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
-            tile_ptr->set_image_data(compression_buffer.data(), compressed_size);
+            {
```
confluence: Is there a reason for this new scope block?

confluence (commented):
The refactoring looks good; thank you! I just have a question about the block (which isn't present in the existing code).

confluence (approved):
Looks good to me.

src/Frame.cc
```diff
@@ -409,11 +410,42 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
-            tile_ptr->set_image_data(compression_buffer.data(), compressed_size);
+            float compression_ratio = (float)compressed_size / (float)tile_image_data_size;
+            bool use_high_precision(false);
+
+            if (precision < HIGH_COMPRESSION_QUALITY && compression_ratio < 0.05) {
+                // re-compress the data with a higher precision
+                std::vector<char> compression_buffer_hq;
+                size_t compressed_size_hq;
+                Compress(tile_image_data, 0, compression_buffer_hq, compressed_size_hq, tile_width, tile_height, HIGH_COMPRESSION_QUALITY);
+                float compression_ratio_hq = (float)compressed_size_hq / (float)tile_image_data_size;
```
veggiesaurus: can we change to the more conventional definition of compression ratio? (uncompressed / comppressed). This will mean our ratio checks change to `> 20` and `> 10`

veggiesaurus (changes requested):

veggiesaurus (approved):

kswang1029 (approved):
Good to go 👍

PR #793, opened 2021-04-09
ajm/dockerfile updates

ajm-ska:
I have updated the Dockerfiles and tested that each one can build with `-Dtest=on` and runs the `carta_backend` and `carta_backend_tests`. Here is a brief description of some of the changes:

**Ubuntu 20.04:** 
Switched to @confluence's cartavis-team PPA repository for carta-casacore, casacore-data, and zfp.

**Ubuntu 18.04:**
Switched to @confluence's cartavis-team PPA repository for carta-casacore, casacore-data, zfp, grpc, googletest, and protobuf. 
The only problem is we need to install a more recent cmake from source because the CMakeLists.txt in the carta_backend's Unit Tests gives a: 
```
Unknown CMake command "add_compile_definitions".
```
with 18.04's default cmake version 3.10.2. It needs to be cmake > 3.12. 
Or perhaps @confluence could create a cmake PPA if you think it is worthwhile?

**Ubuntu 16.04**
Looks like no changes are needed.

**CentOS8**
Discovered the OpenStack repo which has recent protobuffer, grpc and zfp RPMs. That simplifies the Dockerfile somewhat. We also have the old cmake version problem here. Surprisingly the default cmake version in RHEL 8/CentOS 8 is 3.11.4. Apparently a newer cmake will come with the RHEL 8.4 release.

Maybe next time we need to update the Dockerfiles, we can convert this `CentOS8` one to `AlmaLinux` seeing that CentOS 8 will be EOL in Dec 2021 as they are changing to CentOS Stream.

**CentOS7**
Similarly, the OpenStack repo can be used to simplify the Dockerfile a lot. CentOS7 has a cmake 3.17.5 package so that is convenient. I installed the newer TBB from source (seemingly not in OpenStack, unfortunately) so the carta_backend's CMakeLists.txt can detect it properly.

**CentOS6**
I propose to remove CentOS6. I'm having some trouble getting it to build with `-Dtest=on`. Although it _should_ be possible as our CircleCI CentOS6 container is working fine. I can't figure out the difference. Although initially built from a Dockerfile, I have been manually updating the CircleCI containers and pushing them to Dockerhub whenever updates were needed. But anyway, RHEL6/CentOS6 has been EOL since 11/30/2020, so maybe we should remove it rather than spending more time on it.

Finally, the **CMakeLists.txt** is modified again to work with the unusual installed location of pugixml in Ubuntu 18.04.


confluence:
The minimum CMake version specified in the main `CMakeLists.txt` is 3.10, and in the test `CMakeLists.txt` is 3.5, which is clearly incorrect. We need to decide whether we _need_ to use this 3.12 feature or if we can work around it in 3.10.

If we're bumping the minimum version to 3.12, we should provide a Bionic package in our repo. Someone else has already backported 3.16 (the Focal version) in a PPA; we can just copy that package to our PPA (the way I've done with the Bionic gRPC dependencies).

Dockerfiles/Dockerfile-ubuntu-18.04
```diff
@@ -4,38 +4,28 @@ FROM ubuntu:18.04
...
+    software-properties-common unzip vim wcslib-dev wget uuid-dev && \
   update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 80 --slave /usr/bin/g++ g++ /usr/bin/g++-8 --slave /usr/bin/gcov gcov /usr/bin/gcov-8
 
-# Get carta dependencies
-# casacore data from Kernsuite PPA
+# carta_backend unit tests require cmake >3.12 but ubuntu 18.04 provides 3.10.2
 RUN \
-  apt-add-repository -y -s ppa:kernsuite/kern-5 && \
-  apt-get update && \
-  apt-get -y install casacore-data
```
confluence: Our PPA does not package casacore-data. We can get this either from KERN or from the main Ubuntu repository. I assume that there's a reason that we've been using the KERN version, so we should probably keep using it for consistency.

Dockerfiles/Dockerfile-ubuntu-18.04
```diff
@@ -4,38 +4,28 @@ FROM ubuntu:18.04
...
-
-# grpc from webispy ppa
-RUN \
-  add-apt-repository -y ppa:webispy/grpc && \
-  apt-get update && \
-  apt-get install -y libprotobuf-dev protobuf-compiler libgrpc++-dev libgrpc-dev protobuf-compiler-grpc googletest
+  ln -snf /usr/share/zoneinfo/$CONTAINER_TIMEZONE /etc/localtime && echo $CONTAINER_TIMEZONE > /etc/timezone && \
+  add-apt-repository -y ppa:cartavis-team/carta && \
+  apt-get -y update && \
+  apt-get -y install casacore-dev googletest libgrpc++-dev libprotobuf-dev libzfp-dev protobuf-compiler-grpc
```
confluence: This is unnecessary. We only have to install `casacore-dev`; the other dependencies will be installed automatically. When I have finished my current changes to the carta-casacore package, this will be called `carta-casacore-dev` (so we should use that name and only merge this PR once the new packages are in place).

Dockerfiles/Dockerfile-ubuntu-18.04
confluence: It may be unnecessary to install casacore-data explicitly here; it should be installed automatically later as a recommended dependency of our casacore package. But we should check this; I've seen that in some environments apt is configured not to install recommended dependencies by default.

Dockerfiles/Dockerfile-ubuntu-20.04
```diff
@@ -13,23 +13,11 @@ RUN \
     software-properties-common unzip wcslib-dev wget uuid-dev
 
 # Get carta dependencies
-# casacore data from Kernsuite PPA
+# carta-casacore and zfp from cartavis-team PPA (includes casacore-data)
 RUN \
-  apt-add-repository -y -s ppa:kernsuite/kern-7 && \
+  add-apt-repository -y ppa:cartavis-team/carta && \
   apt-get update && \
-  apt-get -y install casacore-data
```
confluence: As above.

Dockerfiles/Dockerfile-ubuntu-20.04
```diff
@@ -13,23 +13,11 @@ RUN \
...
-  add-apt-repository -y ppa:cartavis/carta-casacore && \
-  apt-get update && \
-  apt-get install carta-casacore
-
-# zfp from cartavis PPA
-RUN \
-  add-apt-repository -y ppa:cartavis/zfp && \
-  apt-get update && \
-  apt-get install zfp
+  apt-get -y install casacore-dev libzfp-dev
```
confluence: As above.

confluence (changes requested):
I'm in the process of modifying our packages to install / use casacore libraries in a private location in `/opt/`. This will allow our packages to be installed in parallel to the casacore packages from KERN without interfering with them.

Once this is done I will add the required cmake config update to this PR (so that a source install of the backend will be compatible with the packaged install of carta-casacore) and make a corresponding carta-casacore PR recommending the same install prefix (so that a source install of carta-casacore will be compatible with the updated backend config).

We will also need to make that change to the Xenial dockerfile, where we build carta-casacore from source.

Dockerfiles/Dockerfile-ubuntu-18.04
confluence: Sorry, my bad -- I forgot that these were build dependencies, not runtime dependencies.

veggiesaurus:
we can just revert to `add_definitions` instead of `add_compile_definitions`, rather than requiring a newer version of cmake.

ajm-ska:
> we can just revert to `add_definitions` instead of `add_compile_definitions`, rather than requiring a newer version of cmake.

OK. I will try it.

ajm-ska:
Using `add_definitions` instead in `../test/CMakeLists.txt` does indeed work!

It seems it is not necessary to explicitly install `casacore-data` because it is installed automatically as a dependency by `casacore-dev`. 

No hurry with the new PPAs. We can sit on this PR for a while.

confluence:
Here's the corresponding [carta-casacore PR](https://github.com/CARTAvis/carta-casacore/pull/12). I'm going to add my config changes into this branch now.

confluence:
I have added my `CMakeLists.txt` change.

Something that I haven't checked yet is that if you have another version of casacore installed in the standard location and you build the backend, it will use all of the correct files from the custom location. Obviously if you don't have our version installed at all it will fail to find imageanalysis, but I want to make 100% sure that you can't get a mixture of versions if you have both.

confluence:
I have edited the Dockerfiles to match.

Can we check if `ENV CASAPATH` is actually needed? It isn't even showing the correct path to the data for the latest Ubuntu versions. I have never set this.

Also, do we still need to set `LD_LIBRARY_PATH`? For which libraries? The custom casacore library location is now being added to the runpath in the executable; can we do this instead for any other custom locations if necessary? I think that cmake may be doing this automatically for linked directories.

Dockerfiles/Dockerfile-ubuntu-18.04
confluence: I've re-added the KERN PPAs to these two docker files. We can also use the standard Ubuntu version of casacore-data, but that has to pull in a bunch of other casacore dependencies because of how it's structured.

confluence:
I have updated the README. I'm not sure if we should include information about both the release and the development package in the development readme.

Some of these changes need to be backported to 1.4 (the new PPA and the new casacore location), but some are only relevant to the development version.

Should we create a new branch for a 1.4 backport, so that we can reproduce the necessary changes there and merge them in parallel?

ajm-ska:
@confluence  Indeed the two ENV lines:

> ENV LD_LIBRARY_PATH "/usr/local/lib64:/usr/local/lib:$LD_LIBRARY_PATH"
> ENV CASAPATH "/usr/local/share/casacore linux local `hostname`"

Seem to be unnecessary. I tested their removal in each of the five Dockerfiles, going as far as opening a FITS image in the resulting carta_backends.

There were only two problems:
1. libzfp in the ubuntu 16.04 Dockerfile installed to /usr/local/lib and then the carta_backend can not find it at run time. However, that is fixed now by building ZFP with the `-DCMAKE_INSTALL_PREFIX=/usr` flag.

2. libtbb in the CentOS7 Dockerfile where it also gets installed to /usr/local/lib. Unfortunately doing the same `-DCMAKE_INSTALL_PREFIX=/usr` trick didn't work because CentOS doesn't seem to have /usr/lib in LD_LIBRARY_PATH by default. Most library files seem to be searched through `/lib64` which is a symlink to `/usr/lib64`. Building TBB with `-DCMAKE_INSTALL_PREFIX=/` so that it installs to /usr/lib doesn't seem to work either. So I just left it to install to /usr/local/lib and put in `ENV LD_LIBRARY_PATH "/usr/local/lib"` for the CentOS7 Dockerfile only.

ajm-ska:
> Should we create a new branch for a 1.4 backport, so that we can reproduce the necessary changes there and merge them in parallel?

OK. The `Jenkinsfile` and the `.circleci/config.yml` file would also need to be updated if we would want it to be able to pass the CI tests.

confluence:
>     1. libzfp in the ubuntu 16.04 Dockerfile installed to /usr/local/lib and then the carta_backend can not find it at run time. However, that is fixed now by building ZFP with the `-DCMAKE_INSTALL_PREFIX=/usr` flag.

We should not be installing things built from source into `/usr`. It would be better to put the `LD_LIBRARY_PATH` back for Xenial (unless we want to modify the config file just for Xenial, which I don't think we should bother doing at this point).

confluence:
I will start a backport branch today.

veggiesaurus:
> We should not be installing things built from source into `/usr`. It would be better to put the `LD_LIBRARY_PATH` back for Xenial (unless we want to modify the config file just for Xenial, which I don't think we should bother doing at this point).

I agree. We really shouldn't be concerned about Xenial at this point. It's ending standard support [this month](https://wiki.ubuntu.com/Releases), and we haven't officially supported it for CARTA since 20.04 was released.

ajm-ska:
> > We should not be installing things built from source into `/usr`. It would be better to put the `LD_LIBRARY_PATH` back for Xenial (unless we want to modify the config file just for Xenial, which I don't think we should bother doing at this point).
> 
> I agree. We really shouldn't be concerned about Xenial at this point. It's ending standard support [this month](https://wiki.ubuntu.com/Releases), and we haven't officially supported it for CARTA since 20.04 was released.

OK. I have added `ENV LD_LIBRARY_PATH "/usr/local/lib"` and tested it to see that it works.

ajm-ska:
@confluence I made an RPM for carta-casacore using the new install location: `-DCMAKE_INSTALL_PREFIX=/opt/carta-casacore`

However, for RPMs it looks like the lib files go to a lib64 directory by default; `/opt/carta-casacore/lib64`
So would you be OK about adding that to the CMakeLists.txt file?
```
LINK_DIRECTORIES(/opt/carta-casacore/lib /opt/carta-casacore/lib64)
```

confluence:
I would prefer to fix this in the RPM rather than adding a new edge case to the config. But if there isn't an easy way to do it in the short term I guess it should be fine (we already have corner cases for MacOS).

ajm-ska:
> I would prefer to fix this in the RPM rather than adding a new edge case to the config. But if there isn't an easy way to do it in the short term I guess it should be fine (we already have corner cases for MacOS).

Looks like I can override the RPM macros. By default `%{_lib}` is `lib64` because I am building 64-bit libraries on a 64-bit platform. But I can force rename it to anything, in this case `lib` (even though lib is supposed to be for 32-bit libraries)
```
%define _lib /lib
```

ajm-ska:
The many commits to the `Jenkinsfile` were my attempt to add a wait and retry feature on the assumption that some ICD tests may fail due to occasional delays in the network or hard drive access. The feature worked when tested in one stage; if a test in the stage failed, it would wait 30 seconds, restart the backend, and try to run through the tests again. It would only attempt this 3 times in a row. 

Unfortunately when added to all stages, the Jenkinsfile became too large to be interpreted by Jenkins. This is a known problem when using `declarative pipelines` in Jenkins. So for now I have just added the feature to the definitions at the end. It doesn't work as well in that location because it doesn't restart the backend. It just tries 3 times on the same backend process. So if the backend has crashed or hung for some reason, then the test will fail on each retry. I guess we should be made aware of those cases anyway, but some other time I will attempt to write the Jenkinsfile in a different way, maybe using the 'shared libraries' method. But that can be another Pull Request.

I know the `session` and `animator` tests are currently failing. I believe that is due to the ICD upgrade. The ICD tests are now using version 20, but this branch is still using version 19. This is another situation that should be improved in the ICD tests.

ajm-ska:
I'm currently just testing Homebrew with carta-casacore installed to `/opt/carta-casacore`. Hopefully I will be finished soon.

ajm-ska:
On MacOS, although I have added `-DCMAKE_INSTALL_PREFIX=/opt/carta-casacore` to the carta-casacore Homebrew Formula, it looks like Homebrew automatically prefixes everything with the `HOMEBREW_PREFIX` of `/usr/local`. 
So it can not install directly in `/opt/carta-casacore`. 
Instead it can only go to `/usr/local/opt`, where it is actually a symlink back to the Cellar (with a version number):
`carta-casacore -> ../Cellar/carta-casacore/2021.2.4`

The carta-backend formula can be set to use a specific version of carta-casacore. For example,
Adding:
`path = HOMEBREW_PREFIX/"Cellar/carta-casacore/2021.2.4/include"`
Then putting this in the carta-backend's cmake:
`"-DCMAKE_CXX_FLAGS=-I#{path}/casacode -I#{path}/casacore",`
So there should be no conflict with other versions of carta-casacore installed on the same system.

Anyway, Homebrew can build this carta-backend branch fine using the carta-casacore installed in `/usr/local/opt`, and it works.

confluence:
Shall we merge this first and finish the 1.4 backport branch separately? I don't want to spend a huge amount of time on the backport; just to replace all references to the old PPA to avoid confusion. I'm not even 100% sure if it's necessary, given how close we are to the 2.0 release.

ajm-ska:
OK. 

I have just modified the Dockerfile to re-enable the Timer unit tests for MacOS now that their reliability has been improved in the other PR (adjust block_for_millis function).

jolopezl:
@ajm-asiaa, after rebuilding my ubuntu 20.04 docker container from this update I had to manually install `libfmt-dev` to build carta-backend. Maybe something went missing?

veggiesaurus:
> @ajm-asiaa, after rebuilding my ubuntu 20.04 docker container from this update I had to manually install `libfmt-dev` to build carta-backend. Maybe something went missing?

that's odd. Don't we use fmt from spdlog?

ajm-ska:
> > @ajm-asiaa, after rebuilding my ubuntu 20.04 docker container from this update I had to manually install `libfmt-dev` to build carta-backend. Maybe something went missing?
> 
> that's odd. Don't we use fmt from spdlog?

That's right. That is why we took `libfmt-dev` out.
Do you run `git submodule update --init --recursive` first?

jolopezl:
Now I see the problem: `TestTimer.cc` is still using `libfmt` rather than `spdlog/fmt`.

jolopezl:
See: https://github.com/CARTAvis/carta-backend/pull/805.

PR #799, opened 2021-04-19
auth adjustments: fallback to query token

veggiesaurus:
Attempts to authenticate using URL query token if cookies or auth headers aren't available. This solves the issue of authenticating when the frontend and backend are hosted in separate locations.

### DIFF
--- a/src/Util.cc
+++ b/src/Util.cc
@@ -327,7 +327,6 @@ string IPAsText(string_view binary) {
 }
 
 string GetAuthToken(uWS::HttpRequest* http_request) {
-    string req_token;
     // First try the cookie auth token
     string cookie_header = string(http_request->getHeader("cookie"));
     if (!cookie_header.empty()) {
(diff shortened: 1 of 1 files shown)


veggiesaurus:
@ajm-asiaa has requested this for the electron app and archive

confluence (approved):
Looks fine to me.

veggiesaurus:
Will merge in once @ajm-asiaa has tested with his archive build

markccchiang (approved):

ajm-ska:
@veggiesaurus I have tried it in my local ALMA-archive test implementation. It accepts the token and connects with the backend now.

However, I have noticed a new problem. It looks like the automatic file opening feature no longer works (i.e. adding &file=<image name> to the end of URL). It doesn't work with --debug_no_auth either, so it is probably related to some other change in the backend or frontend, or the way to use it has changed? 
Anyway, for now this Pull Request solves the token problem. 
I'll raise the automatic file opening issue on Slack.

PR #801, opened 2021-04-21
adjust block_for_millis function

veggiesaurus:
This should fix the MacOS timing issues with the `block_for_millis` function in the TestTimer unit tests

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -17,7 +17,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 # Needed by clang-tidy and other clang tools
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-set(TEST_COMMON_LIBS gtest gtest_main)
+set(TEST_COMMON_LIBS gtest gtest_main fmt)
 
 enable_testing()
 find_package(GTest)
--- a/test/TestTimer.cc
+++ b/test/TestTimer.cc
@@ -7,8 +7,8 @@
 #include <chrono>
 #include <thread>
 
-#include <gtest/gtest.h>
 #include <fmt/format.h>
+#include <gtest/gtest.h>
 
 #include "../src/Timer/Timer.h"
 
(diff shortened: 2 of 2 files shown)


ajm-ska:
I have run `./carta_backend_tests  --gtest_filter="Timer*" --gtest_repeat=100` multiple times for this branch on MacOS, but also on other older branches but could not reproduce a failure.

Then I tried running full set of tests multiple times `./carta_backend_tests --gtest_repeat=100`
I noticed that when Jenkins was still in the process of running the other ICD tests then I got multiple TimerTest failures, unfortunately both in older branches and this branch...

So it seems TimerTests fail if the system is busy. Perhaps the Mac doesn't have enough threads? What should we do?

ajm-ska:
Maybe we might as well go ahead and merge it? It might help reduce the chances of failure somewhat?

veggiesaurus:
@ajm-asiaa I have another commit that should _really_ solve this :facepalm: Could you give it a go again?

ajm-ska:
I set`./carta_backend_tests --gtest_repeat=100` running while Jenkins was busy running the ICD tests.

Unfortunately I got one failed case while the animator ICD test were running, two failed cases while the contour ICD tests were running, one failed case when the region statistics were running, one failed case while region manipulation ICD tests were running. In all cases, it was the same test:

> [  PASSED  ] 123 tests.
> [  FAILED  ] 1 test, listed below:
> [  FAILED  ] TimerTest.MeasurementStringWorks

In fact, by having the Unit Tests running continuously at the same time as the ICD tests, it probably caused the Animator ICD tests to fail (Jenkins run 6).

I think it is less 'flaky' than before at least.

veggiesaurus:
@ajm-asiaa ok, I've updated the measurement string test as well

ajm-ska:
I have noticed two different tests failed this time:

> [  FAILED  ] 1 test, listed below:
> [  FAILED  ] TimerTest.AccurateAverage
 

> [  FAILED  ] 1 test, listed below:
> [  FAILED  ] TileEncodingTest.RoundTrip

veggiesaurus:
@ajm-asiaa can I merge these in as they are "less flaky" than before?

ajm-ska (approved):
Yes. A lot better than before.

PR #802, opened 2021-04-22
Make default image 3D

confluence:
Generate a 3D default fits file from the original RGB image to work around [frontend issue #1418](https://github.com/CARTAvis/carta-frontend/issues/1418).

Casacore generates the correct default headers for a 3D image with minimal headers, but not for a 2D image with minimal headers.

Is 3.1M excessive for this image? I could export it at a lower resolution.


veggiesaurus:
can't you just make the 3rd axis degenerate?

confluence:
Good point -- let me try that.

confluence:
That works. I have updated the file.

veggiesaurus:
that's fine, but we should probably fix https://github.com/CARTAvis/carta-frontend/issues/1418 anyway :) 

But this at least gives us breathing room in case the issue is a CASA one, rather than CARTA

Issue #788, opened 2021-04-05
CARTA FITS support


pford:
Add support for FITS images which result in casacore::FITSImage errors.  These include:

- Spectral axis defined in velocity without a rest frequency so unable to convert to frequency.
- FITS images with BITPIX = 64 (int64).
- Compressed FITS including .fz (UPDATE: removed .fits.gz).

cfitsio will be used to determine the HDU list for the file browser (for all FITS files) and to read the headers to create a new CartaFitsImage class derived from base class casacore::ImageInterface (similar to CartaHdf5Image).

Handles issues #360 , #373 , #635

pford:
Removing fits.gz support after finding last-minute non-trivial issue.

Issue #648, opened 2020-06-22
feature request: handle fits.gz files


teuben:
Some archives deliver their data in fits.gz format, in the case I was using, they were able to get the sizes down to 50%.
Both fv and ds9 handle this transparently, and we're going to miss the boat of community acceptance if carta doesn't handle this transparently.
Yes, this means I had a case where I think carta does a better job than ds9, but I resorted to ds9 because I had too many files and didn't want to uncompress them.
In ds9 it's 8 steps (clicking/selecting) to get a spectrum on the screen.
In carta this is 1 step (arguably I would love to see that go to 0, if you are loading a cube, why give people an X and Y scan on the right, but not the more obvious Z scan, or give all 3)

pford:
Moved issue from carta to carta-backend

pford:
Giving an estimate of 0 story points since it is a duplicate of #384, but keeping this ticket for test images.

veggiesaurus:
@pford I'm not too clued up on all the FITS conventions, but as far as I know, there are actually two conventions here: one is literally a FITS file that has been gzipped, and another is a fits.gz file, which is slightly different, based on the examples that we have been provided by users.

teuben:
@veggiesaurus they really better be exactly the same.  I cannot imagine there being two types of those.

pford:
@veggiesaurus @teuben the CASA image opener recognizes .fz files created with fpack as FITS, but .gz is UNKNOWN type so CARTA will have to handle these file types differently. I do not see anything in the casacore code which recognizes and supports gzipped files.  Boost supports this but is not currently a required library.  I will investigate C++ STL next, perhaps std::filesystem.

Kechil:
Removed duplicate to prevent the issue being closed and because it has some use.

kswang1029:
test images are available via #682

kswang1029:
Based on @pford it appears that we will need to decompress the fits.gz file into RAM before we can access the image data. The implies there is no scalability for large images. Also based on @veggiesaurus, there appears a way to get the decompressed size of a fits.gz _without_ an actual decompression. The info may be available from the file header of the fits.gz. So a possible solution is we set a threshold for the _decompressed_ size so that only those fits.gz with decompressed size less than the threshold can be loaded in CARTA. This should still support such use case to some extent without messing up the resource management of a server.

This might be useful

```
$ gunzip -l syslog.1.gz
     compressed        uncompressed  ratio uncompressed_name
[...]

Issue #635, opened 2020-07-28
Support for 64-bit integer FITS files


drtobybrown:
CARTA is having trouble finding and loading a 3D cube in fits format that python (and DS9) handles fine. The error message is less than helpful.

```
$ ls testfile.fits 
testfile.fits
$ carta testfile.fits 
```
<img width="407" alt="Screenshot 2020-07-28 at 12 44 16" src="https://user-images.githubusercontent.com/6336385/88714740-11e2b080-d0d2-11ea-8a86-bf6aa4e4cac0.png">

Happy to provide the fits file privately on request.

Issue #384, opened 2019-11-27
recognize and display gzipped fits files


jott3077:
fits files are frequently stored in a gzipped format. It would be great if CARTA could recognize them and  display them, so they can be viewed without the user having to gunzip (CARTA would gunzip on the fly - cf kvis has this feature).

veggiesaurus:
somewhat related to #360. This would probably require changes to casacore

kswang1029:
What happens to a read-only file system? And security? performance?

veggiesaurus:
> What happens to a read-only file system? And security? performance?

There are ways to open gzip'd files without writing to disk, but that would obviously require more memory. No major security issues, other than the common security issue with compressed files: it's possible to create a small compressed file that decompresses to an infinitely large (at least from a practical perspective) file.

pford:
Test files in #648

kswang1029:
test images are available via #682

PR #803, opened 2021-04-23
fits support (casacore workaround)

pford:
Epic #788 covers several reported issues with FITS images not supported by casacore, including: a velocity axis which cannot be converted to frequencies (#373 TabularCoordinate error), int64 FITS images (#635 BITPIX = 64), and compressed FITS files which were ignored or failed to load (#360 .fz).

Removed from this PR: #384 and #648 .fits.gz (also bz) support.

The workaround uses the CFITSIO library to read the headers for inclusion in the file browser for all FITS images, replacing the former casacore::FitsInput implementation.  If the FITS file is selected and casacore::FITSImage fails, the FitsLoader creates a CartaFitsImage derived from casacore::ImageInterface.  The WCSLIB library is used to set up a casacore::CoordinateSystem in this class.  The code is basically identical to casacore::FITSImage but if the SpectralCoordinate fails then a LinearCoordinate is created instead.  The image information can be converted to FITS headers for the file browser.  When the image is selected to be loaded, CFITSIO is used to read the (possibly compressed) data and mask directly. The fz images use tiled compression, so CFITSIO can decompress data as needed on the fly.

The FITS headers have been corrected to read the "internal" data type.  This change breaks the check in the FILEINFO_FITS_MULTIHDU ICD test.


kswang1029:
@pford quick question: in this branch, do we show the header as it is in the FITS file, or do we still "process" it via casacore?

kswang1029:
Regarding `BITPIX`, with the images I have I see two issues with inconsistencies in value and its comment
Screen Shot 2021-04-23 at 10 36 30 AM
Screen Shot 2021-04-23 at 10 36 14 AM

kswang1029:
@pford We are adding a new flag (`--read_only_mode`) to disable write permission (including saving region, saving image, and saving preferences/layouts) when launching the carta_backend. So in this case, what would happen if we try to load a tar.gz image with size larger than available free memory?

pford:
> 
> 
> @pford quick question: in this branch, do we show the header as it is in the FITS file, or do we still "process" it via casacore?

Yes, using ImageHeaderToFITS

pford:
> 
> 
> @pford We are adding a new flag (`--read_only_mode`) to disable write permission (including saving region, saving image, and saving preferences/layouts) when launching the carta_backend. So in this case, what would happen if we try to load a tar.gz image with size larger than available free memory?

The backend would unzip the file to a temporary directory if it exists (returned by std/boost filesystem::temp_directory_path()) and delete the decompressed file on FitsLoader exit, so this file would not require later cleanup.  But I can disable this if necessary once I see where the readonly flag is stored.

kswang1029:
> > @pford We are adding a new flag (`--read_only_mode`) to disable write permission (including saving region, saving image, and saving preferences/layouts) when launching the carta_backend. So in this case, what would happen if we try to load a tar.gz image with size larger than available free memory?
> 
> The backend would unzip the file to a temporary directory if it exists (returned by std/boost filesystem::temp_directory_path()) and delete the decompressed file on FitsLoader exit, so this file would not require later cleanup. But I can disable this if necessary once I see where the readonly flag is stored.

ok I see. So when in read-only mode and we disable it, we will not be able to load tar.gz that is larger than free memory, right? This might sound overkill. @veggiesaurus what would you think? Maybe we should still allow users to load tar.gz even when in the read only mode?

pford:
I removed the BITPIX comment.  The original comment is lost and the "Floating point" comment was created by casacore, so rather than modify it I just removed it.

kswang1029:
@pford When we just request a file list from a folder and that folder contains some fits.gz files, does the backend try to decompress them first before returning a file list? I see this issue with a 19GB fits.gz in my folder so that the file list does not return quickly as usual.

veggiesaurus:
> @pford When we just request a file list from a folder and that folder contains some fits.gz files, does the backend try to decompress them first before returning a file list? I see this issue with a 19GB fits.gz in my folder so that the file list does not return quickly as usual.

Ideally we should be able to get the file list without decompressing anything, and we should also be able to get the file _info_ (at least for single-HDU images) by decompressing only a little bit of the file

kswang1029:
> @pford When we just request a file list from a folder and that folder contains some fits.gz files, does the backend try to decompress them first before returning a file list? I see this issue with a 19GB fits.gz in my folder so that the file list does not return quickly as usual.

UPDATE: also after waiting for a while, the returned file list _does not_ contain my 19GB fits.gz file. The other smaller (7MB) fits.gz file does show up.

pford:
Since the magic number is for a gz file, the backend tries to open the file with cfitsio (fits_open_file); if it opens successfully, the file is closed and added to the file list (name/type/size) as a FITS file.

For file info, the backend uses cfitsio to open the .gz file, advance through the hdus, and determine whether each hdu is an image and naxis >= 2.  Now it skips the memory check -> decompression step and creates the CartaFitsImage from the compressed file for ImageHeaderToFITS.

pford:
@kswang1029 I am not sure why your large fits.gz file does not show in the list. Is it on the mix 'n match server?

kswang1029:
> @kswang1029 I am not sure why your large fits.gz file does not show in the list. Is it on the mix 'n match server?

https://drive.google.com/file/d/1lim2Rx9DazLGqNHEVjtd5mt8ncl6v1CD/view?usp=sharing
the gzipped size is 19GB. It is an image cube from the ALMA archive.

src/ImageData/CartaFitsImage.h
```diff
@@ -0,0 +1,105 @@
...
+    void CloseFileIfError(const int& status, const std::string& error);
+
+    void SetUpImage();
+    void GetFitsHeaders(int& nkeys, std::string& hdrstr);
+
+    // casacore ImageFITSConverter workaround
+    casacore::CoordinateSystem SetCoordinateSystem(
+        int nheaders, const std::string& header_str, casacore::RecordInterface& unused_headers, int& stokes_fits_value);
+    bool AddDirectionCoordinate(casacore::CoordinateSystem& coord_sys, const ::wcsprm& wcs, std::vector<int>& direction_axes);
+    bool AddStokesCoordinate(casacore::CoordinateSystem& coord_sys, const ::wcsprm& wcs, const casacore::IPosition& shape, int& stokes_axis,
```
markccchiang: The `stokes_fits_value` and `stokes_axis` may be swapped in the header function definition?

veggiesaurus:
@pford this seems like an ideal case for some unit tests (loading small FITS files of various dimensions and types)

src/ImageData/CartaFitsImage.h
pford: Fixed it, thanks for catching this.

pford:
Removed fits.gz and fits.bz support in this PR.  CFITSIO decompresses these compressed files into memory **when the file is opened**, then returns an error status if they do not fit.  The file was opened to determine image HDUs, dimensions, etc. even before reading headers.

This work will be done in a separate branch/PR after investigating (1) how to show HDUs/headers without a complete file decompression and (2) how to determine whether decompression to file is necessary before opening with CFITSIO.  I tested the .fits.bz2 file created by KSW and found that CFITSIO supports this format as well; it is not mentioned in any issues but will be handled in the same way as gz.

The compressed FITS files show in the file list but then have the error "Compressed FITS gz/bz not supported yet" when selected (cannot just say "Compressed FITS" because fz is supported).

src/ImageData/CartaFitsImage.cc
```diff
@@ -0,0 +1,1271 @@
...
+            return freq_type;
+        } else {
+            int velref = wcs_spectral.velref;
+            if (velref > 256) {
+                velref -= 256;
+            }
+
+            std::vector<casacore::MFrequency::Types> velref_freq_types = {casacore::MFrequency::LSRK, casacore::MFrequency::BARY,
+                casacore::MFrequency::TOPO, casacore::MFrequency::LSRD, casacore::MFrequency::GEO, casacore::MFrequency::REST,
+                casacore::MFrequency::GALACTO};
```
jolopezl, line 998: If this vector never changes, could you add it as a member of the class?

src/ImageData/CartaFitsImage.cc
```diff
@@ -0,0 +1,1271 @@
...
+    }
+
+    // Use SPECSYS
+    casacore::String specsys(wcs_spectral.specsys);
+    specsys.upcase();
+
+    std::unordered_map<std::string, casacore::MFrequency::Types> specsys_freq_types = {{"TOPOCENT", casacore::MFrequency::TOPO},
+        {"GEOCENTR", casacore::MFrequency::GEO}, {"BARYCENT", casacore::MFrequency::BARY}, {"HELIOCEN", casacore::MFrequency::BARY},
+        {"LSRK", casacore::MFrequency::LSRK}, {"LSRD", casacore::MFrequency::LSRD}, {"GALACTOC", casacore::MFrequency::GALACTO},
+        {"LOCALGRP", casacore::MFrequency::LGROUP}, {"CMBDIPOL", casacore::MFrequency::CMB}, {"SOURCE", casacore::MFrequency::REST}};
```
jolopezl, line 1017: Same here: if this map never changes, could you add it as a member of the class?

src/ImageData/CartaFitsImage.cc
```diff
@@ -0,0 +1,1271 @@
...
+                sub_record.define("unit", unit);
+            } else {
+                sub_record.define("comment", comment);
+            }
+        }
+
+        // Add to headers Record unless already defined
+        if (!header_rec.isDefined(name)) {
+            header_rec.defineRecord(name, sub_record);
+        }
```
jolopezl: This conditional comes at the end of the for-loop. Does it make more sense to use it at the beginning, right after doing `name.downcase()`, and continuing to the following key if the header record is already defined? That would skip all the other conditional tests.

src/ImageData/CartaFitsImage.tcc
```diff
@@ -0,0 +1,125 @@
...
+    // Get section components (convert to 1-based)
+    std::vector<long> start, end, inc;
+    casacore::IPosition slicer_start = section.start();
+    casacore::IPosition slicer_end = section.end();
+    casacore::IPosition slicer_stride = section.stride();
+    for (int i = 0; i < slicer_start.size(); ++i) {
+        start.push_back(slicer_start[i] + 1);
+        end.push_back(slicer_end[i] + 1);
+        inc.push_back(slicer_stride[i]);
+    }
```
jolopezl, line 29: I don't know how big slice_start.size() is, but if you know the size of the vector ahead of the for-loop, it might be good to reserve its size in memory beforehand. It's a free lunch for some speed gain. So, I  just add `start.reserve(slicer.size())`, `end.reserve(slicer.size())`, `inc.reserve(slicer.size())`. Also, better, use a simple assignment instead of `push_back` if the size is known and the vector initialized. Of course, this is a less relevant suggestion that you can ignore. The performance gain might be negligible.

jolopezl (commented):
Looks good to me. This is a nice casacore workaround. I only have minor comments that @pford can consider. I miss testing, especially for adding a new class `CartaFitsImage` that should have some expected behavior.  For instance, and out of curiosity, does `casacore::FitsImage` and `carta::CartaFitsImage` share some functionality?

src/ImageData/CartaFitsImage.tcc
veggiesaurus: I think its only 3 or 4 elements long, so there's no need to do so

jolopezl: Ok, then I don't think it's necessary to change this at all.

src/ImageData/CartaFitsImage.cc
pford: Since this is used only once during image setup, I thought it would be better to create the vector temporarily rather than storing it for the lifetime of the object.

src/ImageData/CartaFitsImage.cc
pford: Same here; only needed temporarily

src/ImageData/CartaFitsImage.tcc
pford: Yes, it is as large as the number of dimensions in the image.  For example, to display a raster image with shape [80, 77, 30], you need data for channel 0.  The slicer is start=[0, 0, 0], end=[79, 76, 0], stride=[1, 1, 1].  Slicers are also used for spectral profiles (e.g. for a cursor at position [57, 54, 0] then start=[57, 54, 0], end=[57, 54, 29], stride=[1, 1, 1]) and regions (e.g. stats for a rectangle with bottom left corner blc=[52, 53] top right corner trc=[59, 57] requests data with slicer start=[52, 53, 0] end=[59, 57, 29] stride=[1, 1, 1]).

src/ImageData/CartaFitsImage.cc
pford: Fixed

confluence (approved):

confluence (changes requested):
Looks like this needs another merge with dev, but apart from that I think it's fine.

ajm-ska (approved):
We are aware that the file-browser ICD tests have been failing. We have made modifications to the ICD tests to take the FITS header changes into account. Once this branch is merged with the dev branch, we will also merge the updated ICD tests and then all tests will pass again.

confluence (approved):

markccchiang (approved):

kswang1029 (approved):
Verified with some test images I have so far. 👍

Issue #806, opened 2021-04-28
Allow the user to customise the browser command


confluence:
At the moment we always use `xdg-open` / `open` to open the frontend in a browser. We should convert the boolean `--no-browser` flag to a `--browser` flag which can be set to an arbitrary string. If it is unset, we use the default. If it is set to `none`, we disable the browser. The string should allow an optional wildcard in an arbitrary position to be replaced by the URL, in case the user wants to do something complicated with quotes, etc..

This will allow users to open CARTA in a specific browser and to pass additional custom parameters to the browser, which will allow for a smoother transition away from the bundled electron app.

veggiesaurus:
@jolopezl can you take care of this? We'd need to make (minor) changes to the config files as well.

confluence:
A small clarification of what I mean about the wildcard substitution: the user should be able to insert a placeholder like `%u` (as in the FreeDesktop spec) or `CARTA_URL` in the middle of the command like this: `'firefox %u --foo --bar '`, and then we should run `firefox url_goes_here --foo --bar`. If we can't find the placeholder, we default to appending the URL at the end: if the user sets the browser to `'firefox --foo --bar'`  we should run `firefox --foo --bar url_goes_here`.

I don't have a strong opinion about the format of the placeholder.

confluence:
(I may be overcomplicating this preemptively; I don't have a specific use case for this in mind. We could leave this out until someone complains.)

veggiesaurus:
using the freedesktop spec'd placeholder (`%u`) sounds entirely sensible to me 👍

confluence:
Wait, I've just realised a problem -- we _can't_ use the FreeDesktop placeholder, because we want to be able to put this in a desktop file, and if we do, it will be interpreted by FreeDesktop. So we should use our own custom placeholder.

PR #810, opened 2021-05-06
Program Settings: Custom browser command as command-line option (#806)

jolopezl:
Adds a command-line option to specify a browser and arguments for the browser. Follows #806.

The usage is simple:
```
./carta_backend --browser <browser>
```
where `<browser>` contains the entire system call to open the front end URL. The user can use the placeholder `%url` to add a custom position for the URL, for example, before or after browser arguments. If the `%url` wildcard is not given, the URL default position is at the end. For instance, a general case looks like this:
```
./carta_backend --browser "firefox -CreateProfile CARTAuser %url"
```
The `browser` option can also be passed using system and user JSON configuration files. As with the other command-line settings, just add:
```
{
    "browser" : "open -a firefox"
}
```
Note that for macOS, Google Chrome needs to be spelled `Google\ Chrome`, which means `Google\\ Chrome` in the JSON file; otherwise, the parsing will fail.

**Working/tested**: The following are examples tested and working on macOS:
```
./carta_backend --verbosity 4 --frontend_folder ~/temp/CARTAvis/carta-frontend/build \
 --root ~/temp/CARTAvis/data --base ~/temp/CARTAvis/data --omp_threads 4 \
 --browser "open -a safari"
```
```
./carta_backend --verbosity 4 --frontend_folder ~/temp/CARTAvis/carta-frontend/build \
 --root ~/temp/CARTAvis/data --base ~/temp/CARTAvis/data --omp_threads 4 \
 --browser "open -a firefox"
```
```
./carta_backend --verbosity 4 --frontend_folder ~/temp/CARTAvis/carta-frontend/build \
 --root ~/temp/CARTAvis/data --base ~/temp/CARTAvis/data --omp_threads 4 \
 --browser "open -a Google\ Chrome"
```
**TO DO:**
- [ ] Test functionality on linux.
- [ ] Need a test?

**Open question:**
- [ ] The issued command is the full responsibility of the user. Do we want some minimal checking?

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -700,14 +700,23 @@ int main(int argc, char* argv[]) {
                     frontend_url += query_url;
                 }
                 if (!settings.no_browser) {
+                    if (settings.browser.size() > 0) {
+                        auto cmd = settings.ParseBrowserCommand(frontend_url);
+                        spdlog::info("open with {}", cmd);
+                        auto cmd_result = system(cmd.c_str());
+                        if (cmd_result) {
+                            spdlog::warn("Failed to open the browser. Check the custom input at --browser.");
+                        }
+                    } else {
 #if defined(__APPLE__)
-                    string open_command = "open";
+                        string open_command = "open";
 #else
-                    string open_command = "xdg-open";
+                        string open_command = "xdg-open";
 #endif
-                    auto open_result = system(fmt::format("{} \"{}\"", open_command, frontend_url).c_str());
-                    if (open_result) {
-                        spdlog::warn("Failed to open the default browser automatically.");
+                        auto open_result = system(fmt::format("{} \"{}\"", open_command, frontend_url).c_str());
+                        if (open_result) {
+                            spdlog::warn("Failed to open the default browser automatically.");
+                        }
                     }
                 }
                 spdlog::info("CARTA is accessible at {}", frontend_url);
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -701,8 +701,7 @@ int main(int argc, char* argv[]) {
                 }
                 if (!settings.no_browser) {
                     if (settings.browser.size() > 0) {
-                        auto cmd = settings.ParseBrowserCommand(frontend_url);
-                        spdlog::info("open with {}", cmd);
+                        auto cmd = settings.GenerateBrowserCommand(frontend_url);
                         auto cmd_result = system(cmd.c_str());
                         if (cmd_result) {
                             spdlog::warn("Failed to open the browser. Check the custom input at --browser.");
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -137,7 +137,7 @@ void ProgramSettings::ApplyCommandLineSettings(int argc, char** argv) {
         ("log_protocol_messages", "enable protocol message debug logs", cxxopts::value<bool>())
         ("no_http", "disable frontend HTTP server", cxxopts::value<bool>())
         ("no_browser", "don't open the frontend URL in a browser on startup", cxxopts::value<bool>())
-        ("browser", "user custom browser + args", cxxopts::value<string>(), "<browser>")
+        ("browser", "custom browser command", cxxopts::value<string>(), "<browser>")
         ("host", "only listen on the specified interface (IP address or hostname)", cxxopts::value<string>(), "<interface>")
         ("p,port", fmt::format("manually set the HTTP and WebSocket port (default: {} or nearest available port)", DEFAULT_SOCKET_PORT), cxxopts::value<int>(), "<port>")
         ("g,grpc_port", "set gRPC service port", cxxopts::value<int>(), "<port>")
@@ -198,6 +198,11 @@ written to a separate log file, '{}/log/performance.log'.
 Options are provided to shut the backend down automatically if it is idle (if no 
 clients are connected), and to kill frontend sessions that are idle (no longer 
 sending messages to the backend).
+
+Disabling the browser takes precedence over a custom browser command. The custom 
+browser command may contain the placeholder CARTA_URL, which will be replaced by 
+the frontend URL. If the placeholder is omitted, the URL will be appended to the 
+end.
 )",
         CARTA_DEFAULT_FRONTEND_FOLDER, DEFAULT_SOCKET_PORT, CARTA_USER_FOLDER_PREFIX, log_levels, CARTA_USER_FOLDER_PREFIX);
 
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -137,6 +137,7 @@ void ProgramSettings::ApplyCommandLineSettings(int argc, char** argv) {
         ("log_protocol_messages", "enable protocol message debug logs", cxxopts::value<bool>())
         ("no_http", "disable frontend HTTP server", cxxopts::value<bool>())
         ("no_browser", "don't open the frontend URL in a browser on startup", cxxopts::value<bool>())
+        ("browser", "user custom browser + args", cxxopts::value<string>(), "<browser>")
         ("host", "only listen on the specified interface (IP address or hostname)", cxxopts::value<string>(), "<interface>")
         ("p,port", fmt::format("manually set the HTTP and WebSocket port (default: {} or nearest available port)", DEFAULT_SOCKET_PORT), cxxopts::value<int>(), "<port>")
         ("g,grpc_port", "set gRPC service port", cxxopts::value<int>(), "<port>")
@@ -237,6 +238,8 @@ sending messages to the backend).
     applyOptionalArgument(init_wait_time, "initial_timeout", result);
 
     applyOptionalArgument(idle_session_wait_time, "idle_timeout", result);
+
+    applyOptionalArgument(browser, "browser", result);
 
     // base will be overridden by the positional argument if it exists and is a folder
     applyOptionalArgument(starting_folder, "base", result);
--- a/src/SessionManager/ProgramSettings.h
+++ b/src/SessionManager/ProgramSettings.h
@@ -102,8 +102,8 @@ struct ProgramSettings {
         debug_msgs.clear();
     }
 
-    std::string ParseBrowserCommand(const std::string& url) {
-        const std::string wildcard = "%url";
+    std::string GenerateBrowserCommand(const std::string& url) {
+        const std::string wildcard = "CARTA_URL";
         auto wildcard_pos = browser.find(wildcard);
         std::string cmd;
         if (wildcard_pos != std::string::npos) {
--- a/src/SessionManager/ProgramSettings.h
+++ b/src/SessionManager/ProgramSettings.h
@@ -42,6 +42,8 @@ struct ProgramSettings {
     int idle_session_wait_time = -1;
     bool read_only_mode = false;
 
+    std::string browser;
+
     bool no_user_config = false;
     bool no_system_config = false;
 
@@ -72,7 +74,8 @@ struct ProgramSettings {
     std::unordered_map<std::string, std::string*> strings_keys_map{
         {"host", &host},
         {"top_level_folder", &top_level_folder},
-        {"frontend_folder", &frontend_folder}
+        {"frontend_folder", &frontend_folder},
+        {"browser", &browser}
     };
     // clang-format on
 
@@ -97,6 +100,19 @@ struct ProgramSettings {
         std::for_each(debug_msgs.begin(), debug_msgs.end(), [](const std::string& msg) { spdlog::debug(msg); });
         warning_msgs.clear();
         debug_msgs.clear();
+    }
+
+    std::string ParseBrowserCommand(const std::string& url) {
+        const std::string wildcard = "%url";
+        auto wildcard_pos = browser.find(wildcard);
+        std::string cmd;
+        if (wildcard_pos != std::string::npos) {
+            cmd = fmt::format(
+                "{0}{2}{1}", browser.substr(0, wildcard_pos), browser.substr(wildcard_pos + wildcard.size(), browser.size()), url);
+        } else {
+            cmd = fmt::format("{} {}", browser, url);
+        }
+        return cmd;
     }
 };
 } // namespace carta


kswang1029:
@jolopezl can the following be simplified from
```
--browser "open -a safari"
--browser "open -a firefox"
--browser "open -a Google\ Chrome"
```
to
```
--browser safari
--browser firefox
--browser chrome
```
?

this might be more friendly to users?

ajm-ska:
It looks like if I have Chrome already open, and Chrome is set as the default browser, then doing `./carta_backend --browser "open -a safari`, opens CARTA on both Chrome and Safari.
If Safari is set as the default browser (and Chrome is already open), then CARTA only opens in Safari with the above command. 
Although if Chrome is the default browser and already open, then doing `./carta_backend --browser "open -a firefox` only opens CARTA in Firefox.

jolopezl:
> @jolopezl can the following be simplified from
> 
> ```
> --browser "open -a safari"
> --browser "open -a firefox"
> --browser "open -a Google\ Chrome"
> ```
> 
> to
> 
> ```
> --browser safari
> --browser firefox
> --browser chrome
> ```
> 
> ?
> 
> this might be more friendly to users?

We discussed this briefly and thought it's the user's responsibility to give the correct browser command and its arguments. The appropriate command depends on the system and can change with different versions, etc.

kswang1029:
> > @jolopezl can the following be simplified from
> > ```
> > --browser "open -a safari"
> > --browser "open -a firefox"
> > --browser "open -a Google\ Chrome"
> > ```
> > 
> > 
> > to
> > ```
> > --browser safari
> > --browser firefox
> > --browser chrome
> > ```
> > 
> > 
> > ?
> > this might be more friendly to users?
> 
> We discussed this briefly and thought that it's the user's responsibility to give the right browser command and its arguments. The appropriate command depends on the system and can change with different versions, etc.

ok. I was thinking if we can make the backend smarter on this aspect. The backend should be able to tell what OS it is then trigger the corresponding command under the scene.

jolopezl:
> > > @jolopezl can the following be simplified from
> > > ```
> > > --browser "open -a safari"
> > > --browser "open -a firefox"
> > > --browser "open -a Google\ Chrome"
> > > ```
> > > 
> > > 
> > > to
> > > ```
> > > --browser safari
> > > --browser firefox
> > > --browser chrome
> > > ```
> > > 
> > > 
> > > ?
> > > this might be more friendly to users?
> > 
> > 
> > We discussed this briefly and thought that it's the user's responsibility to give the right browser command and its arguments. The appropriate command depends on the system and can change with different versions, etc.
> 
> ok. I was thinking if we can make the backend smarter on this aspect. The backend should be able to tell what OS it is then trigger the corresponding command under the scene.

Of course, we can, but it could break anytime in the future between releases if firefox decides to start calling himself firefox-browser. We can discuss adding support now for Ubuntu 18.04 and 20.04, Centos 7 and 8, and macOS, but I foresee hotfixes to patch for system or browser changes.

veggiesaurus:
> @jolopezl can the following be simplified from
> 
> ```
> --browser "open -a safari"
> --browser "open -a firefox"
> --browser "open -a Google\ Chrome"
> ```
> 
> to
> 
> ```
> --browser safari
> --browser firefox
> --browser chrome
> ```
> 
> ?
> 
> this might be more friendly to users?

I'd rather leave this up to the users. We give them the option to provide whatever command they like, and it's up to them to get that command right. Jupyter notebook takes a similar approach to this. There are just too many variables and headaches that could occur if we try to be smart about it. They might have an alias to the applications in their bash profile, for example. Or have Chromium instead of Google Chrome, or have Safari Tech Preview, etc etc.

jolopezl:
> It looks like if I have Chrome already open, and Chrome is set as the default browser, then doing `./carta_backend --browser "open -a safari`, opens CARTA on both Chrome and Safari.
> If Safari is set as the default browser (and Chrome is already open), then CARTA only opens in Safari with the above command.
> Although if Chrome is the default browser and already open, then doing `./carta_backend --browser "open -a firefox` only opens CARTA in Firefox.

Thanks for looking at it. Although, I can't reproduce the first case:

> It looks like if I have Chrome already open, and Chrome is set as the default browser, then doing `./carta_backend --browser "open -a safari`, opens CARTA on both Chrome and Safari.

Could you please, confirm? I tried:
- setting Google Chrome as the default browser,
- tested default browser by doing: `open https://www.gmail.com`; it opens Gmail on Chrome,
- running `./carta_backend --browser "open -a safari`, and it pops up Safari, without opening Chrome.

src/Main.cc
```diff
@@ -700,14 +700,23 @@ int main(int argc, char* argv[]) {
                     frontend_url += query_url;
                 }
                 if (!settings.no_browser) {
+                    if (settings.browser.size() > 0) {
+                        auto cmd = settings.ParseBrowserCommand(frontend_url);
+                        spdlog::info("open with {}", cmd);
```
veggiesaurus: I don't think we need this output, or at least we can leave it as debug

src/SessionManager/ProgramSettings.h
```diff
@@ -98,6 +101,19 @@ struct ProgramSettings {
         warning_msgs.clear();
         debug_msgs.clear();
     }
+
+    std::string ParseBrowserCommand(const std::string& url) {
```
veggiesaurus: I think `GenerateBrowserCommand` makes more sense here. It's parsing the browser _argument_ to generate a command string

veggiesaurus (changes requested):

veggiesaurus:
> > It looks like if I have Chrome already open, and Chrome is set as the default browser, then doing `./carta_backend --browser "open -a safari`, opens CARTA on both Chrome and Safari.
> > If Safari is set as the default browser (and Chrome is already open), then CARTA only opens in Safari with the above command.
> > Although if Chrome is the default browser and already open, then doing `./carta_backend --browser "open -a firefox` only opens CARTA in Firefox.
> 
> Thanks for looking at it. Although, I can't reproduce the first case:
> 
> > It looks like if I have Chrome already open, and Chrome is set as the default browser, then doing `./carta_backend --browser "open -a safari`, opens CARTA on both Chrome and Safari.
> 
> Could you please, confirm? I tried:
> 
> * setting Google Chrome as the default browser,
> * tested default browser by doing: `open https://www.gmail.com`; it opens Gmail on Chrome,
> * running `./carta_backend --browser "open -a safari`, and it pops up Safari, without opening Chrome.

I can't reproduce this either. Since this is up to the user to get right, something like `--browser="/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome"` would call the application directly, instead of using `open`

ajm-ska:
Here is a video of it happening to me:
https://user-images.githubusercontent.com/20802551/117416976-7cbbf480-af4c-11eb-9c2f-0f291e2bb63f.mov

src/SessionManager/ProgramSettings.cc
```diff
@@ -137,6 +137,7 @@ void ProgramSettings::ApplyCommandLineSettings(int argc, char** argv) {
         ("log_protocol_messages", "enable protocol message debug logs", cxxopts::value<bool>())
         ("no_http", "disable frontend HTTP server", cxxopts::value<bool>())
         ("no_browser", "don't open the frontend URL in a browser on startup", cxxopts::value<bool>())
+        ("browser", "user custom browser + args", cxxopts::value<string>(), "<browser>")
```
veggiesaurus: shouldn't it be `"use custom browser + args"`? I prefer something along the lines of 

```
Specify command to use to invoke a web browser when opening the CARTA frontend. If not specified, the default browser is used.
```

confluence: We should not be putting long explanations in the command descriptions. Any long explanation should go in the body of the help.

confluence: I thought the intention was to merge these into a single option, and use `--browser=none`, but maybe it's fine to leave this as two options. On the one hand, I would prefer not to have a proliferation of related and overlapping options. On the other hand, checking explicitly for a "none" keyword is a bit hacky.

We should add in the body of the help that `no_browser` takes precedence over `browser`.

veggiesaurus: fair enough. But `user -> use` makes sense, right?

kswang1029: can we have _one_ example for args so that users can be inspired for the alternative options?

src/SessionManager/ProgramSettings.h
```diff
@@ -98,6 +101,19 @@ struct ProgramSettings {
...
+        const std::string wildcard = "%url";
```
confluence: We can't use `%url`; this will be interpreted by FreeDesktop as `%u` in a desktop file on Linux. I would prefer something simpler like `CARTA_URL`.

src/SessionManager/ProgramSettings.cc
confluence: I suggest  `custom browser command` here, with an explanation later; something like:

`Disabling the browser takes precedence over a custom browser command. The custom browser command may contain the placeholder CARTA_URL, which will be replaced by the frontend URL. If the placeholder is omitted, the URL will be appended to the end.`

(Or whatever we use for the placeholder, which can't be `%url`, as I explained in the other comment.)

Edit: I prefer `custom browser command` to `use custom browser command` because the latter makes it sound like a boolean.

confluence (changes requested):

src/SessionManager/ProgramSettings.cc
veggiesaurus: initially I'd wanted that too but @jolopezl convinced me otherwise :smile:

FWIW jupyter also uses two separate args

confluence:
General comment: I think that we should absolutely leave this to users and make no attempt to check for correctness, which is highly environment-specific. We can build on this to create custom launchers (and suggest specific recipes for using the option in the documentation) to help users transition away from the bundled electron app, but the option should not be restricted to those use cases.

src/SessionManager/ProgramSettings.cc
confluence: Makes sense.

confluence: > can we have _one_ example for args so that users can be inspired for the alternative options?

I would really prefer not to put that in the executable help. It's much too specific for this context. I do think that we should add some information about this to the _manual_, and we can give lots of examples there.

src/SessionManager/ProgramSettings.h
jolopezl: Fixed.

src/SessionManager/ProgramSettings.h
jolopezl: Changed.

confluence:
This looks good; I'm just building this so that I can test the Linux desktop file use case.

jolopezl:
> Here is a video of it happening to me:
> https://user-images.githubusercontent.com/20802551/117416976-7cbbf480-af4c-11eb-9c2f-0f291e2bb63f.mov

I still can't reproduce. I also tested with zsh. What terminal are you using? Maybe some underlying instruction to open links as they appear?

confluence:
I'm seeing some weird behaviour trying to use Chrome (actually Chromium) on Linux.

If I run the backend with `--no-browser`, and then manually run `chromium-browser --new-window --app=URL_WITH_TOKEN_GOES_HERE`, I get the expected result.

However, if I run the backend with `--browser="chromium-browser --new-window --app=CARTA_URL"`, I get a blank window and CARTA never loads. I also don't see `CARTA is accessible at...` in the backend output until after I close the Chromium window.

I tested this with `firefox --new-window` as the browser, and that works fine, but chromium doesn't work at all, even with no parameters. However, if I explicitly run the chromium process in the background, by making the browser command `--browser="chromium-browser --app=CARTA_URL --new-window&"`, it works as expected (and I have confirmed that I can make a custom launcher with a desktop file using this command).

So it looks like the chromium process is blocking the backend process from continuing. But why (when Firefox doesn't behave like this), is there anything we can do about this, and is it our problem? Is explicitly launching it in the background an acceptable workaround?

@veggiesaurus and anyone else with a Linux desktop, can you reproduce this, or is it just me?

confluence:
OK, I think this is a non-mystery mystery -- I checked again and when I compare launching browsers on the commandline, Chromium stays in the foreground and Firefox does not. So I think this is not our problem, and we just need to remember to background the process explicitly in the command.

confluence:
@veggiesaurus has pointed out that it may be a good idea for us _always_ to fork the browser process.

confluence:
> @veggiesaurus and anyone else with a Linux desktop, can you reproduce this, or is it just me?

Apparently this doesn't happen with Chrome, so it may be behaviour specific to Chromium.

veggiesaurus:
> > @veggiesaurus and anyone else with a Linux desktop, can you reproduce this, or is it just me?
> 
> Apparently this doesn't happen with Chrome, so it may be behaviour specific to Chromium.

wait, no. It _does_ happen with Chrome. It happens with `google-chrome`, `chromium` and `firefox` for me.

I wonder if the easiest solution is to just append a ` &` to the command string _if_ it doesn't already end with it (and to not change anything if there's no `--browser` argument.

confluence:
> I wonder if the easiest solution is to just append a ` &` to the command string

How portable is that? And should we also suppress the output?

jolopezl:
@confluence, @veggiesaurus, I just checked on Ubuntu 18.04, and 20.04 (VMs), and the wrong behavior happens with Chrome and Chromium, but it doesn't happen with Firefox. But, as @confluence suggested, if I add `&` at the end of the command, Chrome or Chromium works perfectly.

So, we see all kinds of mixed behaviors. May I suggest that we leave this low-level setting to the end-user? We should describe this subtle behavior in the documentation rather than trying to get it right for all.

We can force the process to the background, of course. Do we anticipate side effects? As @confluence says, we would need to redirect the output as well.

What's your preference? I prefer to let the user get the command right.

confluence:
I think that we should ensure that the command is always run as an independent process which will remain alive if the terminal in which the backend is running is closed -- because this is consistent with the behaviour of `open` / `xdg-open`, and because if we don't do this and the process stays in the foreground it causes very counter-intuitive and user-unfriendly behaviour.

But I would prefer to do this in a cleaner C++ way than to tack on a `&` to the shell command. Just the `&` is not sufficient for the correct behaviour -- that will not disown the process, so it will be closed when the terminal is closed (there are various ways to make that not happen, e.g. the `disown` shell command). From a cursory look at Jupyter notebook (which uses `webbrowser`, which uses `subprocess`) we could use `setsid` to run the process independently. I think we should also put some thought into what happens to the output.

I think @veggiesaurus is currently suggesting that we merge this as-is and refine it during the beta in time for the final 2.0 release -- I would prefer to do this than to rush in a hacky fix. In the meantime, we can document this limitation in the information we give to beta testers, and suggest temporary shell workarounds.

jolopezl:
> I think @veggiesaurus is currently suggesting that we merge this as-is and refine it during the beta in time for the final 2.0 release -- I would prefer to do this than to rush in a hacky fix. In the meantime, we can document this limitation in the information we give to beta testers, and suggest temporary shell workarounds.

Ok, I agree. Let's merge this – if you all approve – and provide a complete solution for the final 2.0 release.

confluence (approved):

ajm-ska:
> > Here is a video of it happening to me:
> > https://user-images.githubusercontent.com/20802551/117416976-7cbbf480-af4c-11eb-9c2f-0f291e2bb63f.mov
> 
> I still can't reproduce. I also tested with zsh. What terminal are you using? Maybe some underlying instruction to open links as they appear?

I'm using zsh. I'm not sure if the version makes a difference, but it is Chrome 90.0.4430.93.

If I try to open Safari directly with ` --browser "/Applications/Safari.app/Contents/MacOS/Safari"` then Chrome doesn't open, the Safari window opens, but it hangs because there is an error in the terminal:

> objc[35741]: Class AMSupportURLConnectionDelegate is implemented in both /usr/lib/libauthinstall.dylib (0x7fff83050ac0) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x11158a910). One of the two will be used. Which one is undefined.
> objc[35741]: Class AMSupportURLSession is implemented in both /usr/lib/libauthinstall.dylib (0x7fff83050b10) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x11158a960). One of the two will be used. Which one is undefined.

Does that work for everyone else?

Another thing I notice on MacOS is if Firefox is closed and I use  `--browser "open -a firefox"` then two Firefox windows are actually opened; A blank window and a window with CARTA.

Screenshot 2021-05-10 at 08 34 24

I can also report things work fine with Edge. Doing ` --browser "open -a Microsoft\ Edge" ` only opens CARTA in Edge.

veggiesaurus (approved):

veggiesaurus:
I'm going to merge this in before code freeze. I suggest we make an issue for improving the behaviour of the `--browser` flag, and see if we can fix it before the 2.0 release itself.

jolopezl:
@ajm-asiaa, as a minor follow up to your finding:

> Another thing I notice on MacOS is if Firefox is closed and I use `--browser "open -a firefox"` then two Firefox windows are actually opened; A blank window and a window with CARTA.

This seems to be an ancient bug: https://bugzilla.mozilla.org/show_bug.cgi?id=531552, which is 12 years old with some recent reports of the same happening, and new bugs filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1627403, https://bugzilla.mozilla.org/show_bug.cgi?id=1600153. It seems to be something different, but the effect is the same. I am reporting this so that we don't worry too much about it for now.

PR #811, opened 2021-05-07
Fix the problem of catching stokes information form the header

markccchiang:
Some image files, like `mini_stokes_cube_noise.fits`, have a `Stokes` keyword from the header that its letters are not all capital.  Therefore the backend can not catch stokes information so that spectral profile of a given Stokes cannot be generated. This PR is to solve such a problem.


kswang1029 (approved):
issue fixed. Now spectral profile of a selected Stokes can be generated successfully.

veggiesaurus (approved):

PR #816, opened 2021-05-12
attempt to match required token to cookie, bearer auth header, custom header or query parameter

### LINKED ISSUE #815: cannot save layout
thanks for the help from @acdo2002 @YuHsuan-Hwang and @ajm-asiaa 

The issue here is: if we have a global installation of CARTA on a server and every user just types carta to launch it, it appears that a layout cannot be saved which might depend on the permission of CARTA's "pwd". 

Can anyone try and see if it is also reproducible from your side? If it is confirmed, we will need to fix this before v2.0 release. 

edit by @veggiesaurus: This is an issue with the older backend rejecting auth tokens, as they have been overwritten with the token from the newer backend

veggiesaurus:
Closes #815 by allowing token matching to any of the following:
- The `carta-auth-token` cookie
-The standard `Authorization` header
-The `token` URL query parameter
-The non-standard `carta-auth-token` header

Previously, the first non-empty option of those above was used for token matching. Now auth succeeds if _any_ of the above match the required token.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -80,21 +80,10 @@ void OnUpgrade(uWS::HttpResponse<false>* http_response, uWS::HttpRequest* http_r
         address = IPAsText(http_response->getRemoteAddress());
     }
 
-    // Check if there's a token to be matched
-    if (!auth_token.empty()) {
-        string req_token = GetAuthToken(http_request);
-
-        if (!req_token.empty()) {
-            if (req_token != auth_token) {
-                spdlog::error("Incorrect auth token supplied! Closing WebSocket connection");
-                http_response->close();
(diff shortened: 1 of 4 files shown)


veggiesaurus:
@ajm-asiaa and @kswang1029 please also give this a try to see if it works fine. I've briefly tested it with both SDM and UDM approaches.

confluence (approved):
Looks OK.

ajm-ska:
@kswang1029 This can be tested directly on almat8 now. Double check that you are using `/usr/bin/carta`.
I tested it myself and the issue appears to be gone.

ajm-ska (approved):

PR #818, opened 2021-05-12
WebBrowser: Create web browser class to improve --browser setting

jolopezl:
This pull request implements better handling of the `--browser` settings previously added. It launches a new process and detaches it from the parent. The implementation is different for Linux and macOS. In principle, forking a new process and detaching it in macOS also works, but this is better handled by the native functionality offered by the program `open`, which handles launching a new process easily.


jolopezl:
@ajm-asiaa, you tested this implementation in https://github.com/CARTAvis/carta-backend/pull/810. Could you please have a look at how it runs now? It should be exactly the same as before.

ajm-ska (approved):
I tested it on MacOS. It works great! Now on MacOS I see we can do, for example, `--browser firefox`, so no need for Mac-specific commands using 'open' (i.e. `--browser="open -a firefox"`) anymore.

confluence:
@jolopezl could you please merge in latest dev, to get the protobuf changes (so that this can be tested with the dev frontend)?

jolopezl:
> @jolopezl could you please merge in latest dev, to get the protobuf changes (so that this can be tested with the dev frontend)?

@confluence, thanks. I think it's ready for review now.

confluence:
This is not working for me on Ubuntu 20.04. Here is my debug log output:

```
adrianna@sergyar:~/repos/other-carta-backend/build$ ./carta_backend --browser="chromium --app=CARTA_URL --new-window" --frontend_folder ../../carta-frontend/build --verbosity 5
[2021-07-07 15:07:48.198] [info] Writing to the log file: /home/adrianna/.carta/log/carta.log
[2021-07-07 15:07:48.198] [debug] Using command-line settings
[2021-07-07 15:07:48.198] [info] /home/adrianna/repos/other-carta-backend/build/carta_backend: Version 2.0.0-dev.21.04.27
[2021-07-07 15:07:48.201] [info] Serving CARTA frontend from /home/adrianna/repos/carta-frontend/build
[2021-07-07 15:07:48.201] [info] Listening on port 3002 with top level folder /, starting folder /home/adrianna/repos/other-carta-backend/build. The number of OpenMP worker threads will be handled automatically.
[2021-07-07 15:07:48.201] [debug] WebBrowser: custom command is chromium --app=http://localhost:3002/?token=17c7aa02-1b21-4073-930d-4610ca77837a  --new-window, attempting to open the browser now.
[2021-07-07 15:07:48.201] [debug] WebBrowser: Trying to launch CARTA with user provided browser command: chromium --app=http://localhost:3002/?token=17c7aa02-1b21-4073-930d-4610ca77837a  --new-window
[2021-07-07 15:07:48.202] [info] CARTA is accessible at http://localhost:3002/?token=17c7aa02-1b21-4073-930d-4610ca77837a
```

The browser is not launched. When I manually execute `chromium --app=http://localhost:3002/?token=17c7aa02-1b21-4073-930d-4610ca77837a  --new-window` from a terminal, it works.

confluence:
I have also reproduced this with a Firefox command, so it's not anything specific to Chrome or Chromium.

jolopezl:
@confluence, in Ubuntu 20.04 it works with `--browser "/usr/bin/chromium-browser --app=CARTA_URL --new-window"` for Chromium, and `--browser "/usr/bin/firefox CARTA_URL"` for Firefox. We can make it more flexible and guess if some executable is in the path, then handle that properly. I kept it simple following our discussion in the previous related pull request to leaving the specific command to the user and not attempting to do something else.

veggiesaurus:
> @confluence, in Ubuntu 20.04 it works with `--browser "/usr/bin/chromium-browser --app=CARTA_URL --new-window"` for Chromium, and `--browser "/usr/bin/firefox CARTA_URL"` for Firefox. We can make it more flexible and guess if some executable is in the path, then handle that properly. I kept it simple following our discussion in the previous related pull request to leaving the specific command to the user and not attempting to do something else.

but then we have completely different behaviour on MacOS versus linux, which will lead to confusion, surely?

confluence:
We should not require specifying the absolute path to the executable if the executable is on the system path. This is very counterintuitive -- a command that the user can use in the shell should also work as the browser parameter.

It's also not clear to me why this isn't failing with an error if the executable cannot be found.

confluence:
If we're going to delegate to `open` on the Mac (which is sensible), then it may make more sense to assume the presence of Bash on Linux and just rewrite the command to use the appropriate bash-isms to put the process in the background and disown it.

This wasn't the approach that we initially discussed, but I don't think it makes sense to have such drastically different mechanisms on the two platforms.

jolopezl:
> > @confluence, in Ubuntu 20.04 it works with `--browser "/usr/bin/chromium-browser --app=CARTA_URL --new-window"` for Chromium, and `--browser "/usr/bin/firefox CARTA_URL"` for Firefox. We can make it more flexible and guess if some executable is in the path, then handle that properly. I kept it simple following our discussion in the previous related pull request to leaving the specific command to the user and not attempting to do something else.
> 
> but then we have completely different behavior on macOS versus Linux, which will lead to confusion, surely?

@veggiesaurus, we don't have completely different behavior on macOS versus linux. It's the case that macOS has a register for firefox, but it doesn't for chrome, for instance. So, if you want to launch other applications such as Chrome, you have to give the full path: `/Applications/Google\ Chrome.app`, then `open -a /Applications/Google\ Chrome.app https://...` will do the right job starting the application and detaching it from the current process. I don't know if one can append applications to the macOS `open` application; surely possible otherwise, I would find it very restrictive, but I think that's out of scope.

> We should not require specifying the absolute path to the executable if the executable is on the system path. This is very counterintuitive -- a command that the user can use in the shell should also work as the browser parameter.
> 

@confluence, sure, we can improve that direction, but it adds a couple of challenges. Certainly, that implementation is easy using the Boost library, but I understand that we don't want to add more dependencies. I am not sure if one can implement a find-in-path for all Linux systems working equally. That's what a quick look gives me. So, I would try to stick to the idea that this is a highly customizable command where the user can take full responsibility.

EDIT: After searching a bit more, there are ways, but these are all tricky. For instance, one could ask the bash `which program` and get the output or something like `command -v program`.

> It's also not clear to me why this isn't failing with an error if the executable cannot be found.

My bad. I will fix that.

> If we're going to delegate to `open` on the Mac (which is sensible), then it may make more sense to assume the presence of Bash on Linux and just rewrite the command to use the appropriate bash-isms to put the process in the background and disown it.
> 
> This wasn't the approach that we initially discussed, but I don't think it makes sense to have such drastically different mechanisms on the two platforms.

Both systems are pretty similar, but since macOS offers `open`, it simplifies the implementation a lot, and using it makes it less prone to error. So having a different implementation in Linux is completely fine, I think. However, this made it more tricky to implement this change, and I guess it isn't bullet-proof. If we want to use bash-isms, aren't we susceptible to the environment each user can have defined? We could make all the implementation easier with a simpler system call, for example, using `nohup firefox args CARTA_URL >/dev/null 2>1 &`, with another set of pros and cons.

confluence:
> this is a highly customizable command where the user can take full responsibility

It does not make sense to make the user responsible for constructing a command which bypasses the normal expected features of the shell.

I think that a solution on Linux which uses the shell and behaves like the shell is more consistent and predictable. Bash is so ubiquitous that we can assume that it is installed even if it isn't the user's shell; if necessary we can call it explicitly.

I would say that respecting the shell path is a requirement, and it makes more sense to reuse the shell in order to meet it than to make a more complicated low-level implementation.

jolopezl:
> > this is a highly customizable command where the user can take full responsibility
> 
> It does not make sense to make the user responsible for constructing a command which bypasses the normal expected features of the shell.
> 
> I think that a solution on Linux which uses the shell and behaves like the shell is more consistent and predictable. Bash is so ubiquitous that we can assume that it is installed even if it isn't the user's shell; if necessary we can call it explicitly.
> 
> I would say that respecting the shell path is a requirement, and it makes more sense to reuse the shell in order to meet it than to make a more complicated low-level implementation.

This goes against the conclusions of the previous related pull request. Honestly, I don't see anything highly complicated in typing `/usr/bin/google-chrome args` to get the desired browser up and running. Using this program setting is an advanced feature per se. I agree that getting the PATH of the executable is a must, and it much better. However, I don't understand why this bypasses the expected features of the shell; this is just using what the system gives you to work on it.

We can get the PATH of an executable with a couple of additional C-like functions, but it isn't as nice since it will probably be less cross-platform. If we were using boost, it would be more straightforward.

I will reimplement something using the shell in another pull request, and then we can compare and find out what better suits our needs. In brief, I will be proposing to relying on having `bash` on the host system and just making a system call of the style: `bash -c "google-chrome --arg=CARTA_URL --new-window" >/dev/null 2>1 &`.

ajm-ska (approved):
Seems to work fine on MacOS and CentOS7.

src/Util.cc
```diff
@@ -443,3 +443,29 @@ int GetNumItems(const string& path) {
         return -1;
     }
 }
+
+// https://stackoverflow.com/a/46931770/1727322
+std::vector<std::string> split_string(const std::string& s, char delim) {
```
veggiesaurus: we already have a `SplitString` function

src/Util.cc
```diff
@@ -443,3 +443,29 @@ int GetNumItems(const string& path) {
...
+    while (std::getline(ss, item, delim)) {
+        result.push_back(item);
+    }
+
+    return result;
+}
+
+// quick alternative to bp::search_path that allows us to remove
+// boost:filesystem dependency
+fs::path search_path(std::string filename) {
```
veggiesaurus: `SearchPath`

src/SessionManager/WebBrowser.cc
```diff
@@ -0,0 +1,135 @@
...
+            close(fd);
+
+            auto result = execv(args[0], args.data());
+            if (result == -1) {
+                spdlog::debug("WebBrowser: execv failed. CARTA can't start with the requiered settings in --browser.", result);
+            }
+            _exit(1);
+        } else if (pid2 == -1) {
+            spdlog::debug("WebBrowser: Failed to fork a new process. CARTA can't start with the requiered settings in --browser.");
+            if (chdir("/") == -1) {
```
veggiesaurus, line 126: why do we need to `chdir` to `/`?

src/SessionManager/WebBrowser.cc
```diff
@@ -0,0 +1,135 @@
...
+        spdlog::debug("WebBrowser: custom command is {}, attempting to open the browser now.", _cmd);
+        OpenBrowser();
+    } else {
+        spdlog::debug("WebBrowser: using default browser.");
+        OpenSystemBrowser();
+    }
+}
+
+void WebBrowser::ParseCmd() {
+    if (_cmd[_cmd.size() - 1] == '&') {
```
veggiesaurus, line 46: do we still need to do this?

src/SessionManager/WebBrowser.h
```diff
@@ -0,0 +1,39 @@
...
+    std::string Error() const {
+        return _error;
+    }
+
+private:
+    void ParseCmd();
+    void OpenSystemBrowser();
+    void OpenBrowser();
+    bool _status;
+    bool _path_exists;
```
veggiesaurus: I can only see `_path_exists = true`. Where are you initialising this?

src/SessionManager/WebBrowser.cc
```diff
@@ -0,0 +1,135 @@
...
+                spdlog::debug("WebBrowser: Failed to change dir to \"/\"");
+            }
+            _exit(0);
+        } else if (pid2 != 0) {
+            // kill this child parent so that we don't endup with a zoombie process
+            _exit(1);
+        }
+    }
+    if (pid == -1) {
+        spdlog::debug("WebBrowser: Failed to fork a new process. CARTA can't start with the requiered settings in --browser.");
```
veggiesaurus: if `pid` is -1, it means we couldn't fork a new process. Why should we exist under these circumstances? Under all other circumstances we continue even if the browser fails to start

veggiesaurus: typo, and this should be at least a `warning`.

src/SessionManager/WebBrowser.cc
```diff
@@ -0,0 +1,135 @@
...
+            dup2(fd, STDERR_FILENO);
+            close(fd);
+
+            auto result = execv(args[0], args.data());
+            if (result == -1) {
+                spdlog::debug("WebBrowser: execv failed. CARTA can't start with the requiered settings in --browser.", result);
+            }
+            _exit(1);
+        } else if (pid2 == -1) {
+            spdlog::debug("WebBrowser: Failed to fork a new process. CARTA can't start with the requiered settings in --browser.");
```
veggiesaurus: `spdlog::warning` and typo. A warning is fine here because you're exiting the forked process

src/SessionManager/WebBrowser.cc
```diff
@@ -0,0 +1,135 @@
...
+            int fd = open("/dev/null", O_WRONLY);
+            dup2(fd, STDOUT_FILENO);
+            dup2(fd, STDERR_FILENO);
+            close(fd);
+
+            auto result = execv(args[0], args.data());
+            if (result == -1) {
+                spdlog::debug("WebBrowser: execv failed. CARTA can't start with the requiered settings in --browser.", result);
+            }
+            _exit(1);
```
veggiesaurus, line 123: `execv` only returns if an error occurs, so no need to check `result`

src/Util.h
```diff
@@ -24,6 +25,8 @@
 #include "ImageStats/BasicStatsCalculator.h"
 #include "ImageStats/Histogram.h"
 
+namespace fs = std::filesystem;
```
veggiesaurus, line 33: we still need to support boost::filesystem as well, at least for now

veggiesaurus (changes requested):
some minor comments

confluence:
I have retested in Ubuntu (with Chromium) and it seems to work as expected. The executable is found on the path, the browser window is launched, and the browser stays alive if the backend is shut down or if the terminal the backend is running in is closed. This is consistent with the default (`xdg-open`) behaviour. 👍

src/SessionManager/WebBrowser.cc
jolopezl: > do we still need to do this?

I don't think it's fully needed, but I also think it isn't of any harm. The new process is launched and detached from the parent process, and output is ignored, so there is no need to send it to the background. So I am open to keeping it or leaving it, not sure if it can be harmful or not.

src/SessionManager/WebBrowser.cc
jolopezl: What do you mean? No checking for error?

jolopezl: What do you mean? No checking for error?

src/SessionManager/WebBrowser.cc
jolopezl: > `spdlog::warning` and typo. A warning is fine here because you're exiting the forked process

Ok, changing it to: `WebBrowser: Failed to fork a new process. CARTA can't start with the required settings in --browser.`

src/SessionManager/WebBrowser.h
jolopezl: good catch, thanks

src/SessionManager/WebBrowser.cc
jolopezl: I am sorry, I left this one without a reply. This is mostly a good practice. I found this [article](http://www.microhowto.info/howto/cause_a_process_to_become_a_daemon.html) pretty good explaining how to start a daemon process properly, together with some implementation details [here](http://www.microhowto.info/howto/cause_a_process_to_become_a_daemon_in_c.html#id2407077). So I am just going to quote it here for completeness:

> If a daemon were to leave its current working directory unchanged then this would prevent the filesystem containing that directory from being unmounted while the daemon was running. It is therefore good practice for daemons to change their working directory to a safe location. This could be:
>- the root directory, or
>- a directory containing files that are needed by the daemon.
>(The latter option turns what would have been a drawback into a feature. If files are needed by the daemon then the filesystem containing them presumably should not be unmounted while the daemon is running. Moving the current working directory to that filesystem prevents this from happening.)

So, I believe this isn't a problem for any running browser, and we could leave the working path unchanged. But thinking further, if more low-level settings are passed to the browser in execution, it might be a good idea to get rid of potential problems.

PR #821, opened 2021-05-18
PV ellipse stats release 2.0

pford:
When an LCRegion is needed for stats/histograms/profiles, a WCRegion is created then converted to an LCRegion in a given image's coordinate system and shape.  However, WCEllipsoid fails for PV images with the error that the xcenter and ycenter do not have the same base unit (arcsec and Hz).  Instead, the LCEllipsoid for the reference image is created from its control points.

WCRegion is used to convert the region to an LCRegion in another image.  Since the WCEllipsoid cannot be created for PV images, this region type cannot be converted to a non-reference image and will fail.  If this is needed, the conversion would have to be done manually.

### DIFF
--- a/src/Region/Region.cc
+++ b/src/Region/Region.cc
@@ -796,23 +796,21 @@ casacore::LCRegion* Region::GetConvertedLCRegion(
     }
 
     try {
-        if (is_reference_image) {
+        // Convert reference WCRegion to LCRegion in image with output csys and shape
+        // Create reference WCRegion if not set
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):
this fix works fine. 👍 

spatial matching for pv images is out of scope so we don't have to worry about the conversion of "world" coordinates in this case.

markccchiang (approved):

confluence (approved):

Issue #813, opened 2021-05-11
Generate random FITS and HDF5 files for unit tests


confluence:
Many unit tests which need to be added require files with specific large dimensions (e.g. tests of tile or chunk behaviour). It would be impractical to store these files. Instead, we should integrate the unit tests with the [random FITS file generator](https://github.com/idia-astro/image-generator) and the [HDF5 converter](https://github.com/idia-astro/fits2idia/), create these files as they are needed, and delete them afterwards.

- [x] Enhance the FITS generator to allow specifying arbitrary headers (for header tests) and a random seed (for file reproducibility)
- [x] Add generator and converter repos as submodules (to be used only for the test build). This may mean that there are additional dependencies for building and running the tests.
- [ ] Rewrite existing tests which use files so that they generate files instead

veggiesaurus:
I suggest we split the last checkbox into a separate issue

PR #822, opened 2021-05-21
Generate files for tests and create paths to data consistently

confluence:
This implements #813. I haven't converted _all_ the old tests to use generated files, since some of them specifically test existing directory structures or test casa files (which we can't generate easily). The existing test files are small, so it's not a problem if we keep them (and I don't think it's really a problem if we continue to use them for tests which don't require big files). But if anyone feels strongly that we should eliminate all test data files that aren't strictly necessary, more of the tests can be rewritten.

This PR also adds helper functions for obtaining paths to existing or generated files, and for finding the test root. We should always use these functions in future tests instead of constructing paths manually, and in particular we should _not_ assume that the current working directory is the test root. With these changes it is currently possible to run the tests from outside the test root (for example from the main build directory, which is useful) and I would like us to keep it that way. For the handful of tests which test relative paths to data I explicitly change directory to the test root at the start of the test (and add setup and teardown code to reset to the original working directory, although this shouldn't matter).

@ajm, This PR adds two submodules: a Python script for generating FITS files and the C++ FITS to HDF5 converter. The converter is built as an external project when tests are enabled. Python 3, numpy and astropy are runtime dependencies when tests are run. I *think* that those are the only additional dependencies, and that the dependencies for building the converter are all dependencies we already have for the backend.


ajm-ska:
Looks like our CentOS7 and Ubuntu 16.04 systems had a problem building the FITS to HDF5 converter: 
`error: 'fits_read_string_key'` 

The CentOS7 and Ubuntu 16.04 systems both had cfitsio 3.37 by default e.g.
`-- Found CFITSIO: /usr/lib64/libcfitsio.so Version: 3.37`

I think the `fits_read_string_key` function was introduced in cfitsio version 3390?

So I installed the latest cfitsio (3.49) from source on those two systems.

When rebuilding carta-casacore, there was a cfitsio problem on CentOS7 for some unknown reason, so instead I tried installing an older cfitsio on it; 3.43. It worked. carta-casacore then built without problems.
Meanwhile, carta-casacore built fine with cfitsio 3.49 on Ubuntu 16.04. 

Now this new carta-backend branch can build without problems on Ubuntu 16.04.

Unfortunately now on CentOS7, the fits2idia CMakeLists.txt has a problem finding the CFITSIO... I'll try to work around it in the morning.

ajm-ska:
Speaking of `third-party/fits2idia/CMakeLists.txt`, it looks like line 23 should say HDF5:
```
find_package(HDF5 REQUIRED COMPONENTS CXX)
if (HDF5_FOUND)
    include_directories(${HDF5_INCLUDE_DIR})
    set(LINK_LIBS ${LINK_LIBS} ${HDF5_LIBRARIES})
else ()
    message(FATAL_ERROR "Could not find CFITSIO.")
endif ()
```

veggiesaurus:
@confluence is there any alternative to `fits_read_string_key` that we could use to get aroudn the centos7 issue? I'm not bothered by ubuntu 16.04 issues, but centos7 is still a CARTA supported platform.

ajm-ska:
Well I got around the cfitsio pkgconfig problem for now by prefixing the cmake command with 
`env PKG_CONFIG_PATH=/usr/lib/pkgconfig` 
(That is where the `cfitsio.pc` file was installed on the CentOS7 server).

veggiesaurus:
> Well I got around the cfitsio pkgconfig problem for now by prefixing the cmake command with
> `env PKG_CONFIG_PATH=/usr/lib/pkgconfig`
> (That is where the `cfitsio.pc` file was installed on the CentOS7 server).

@ajm-asiaa sure, but didn't that still require the installation of a newer version of cfitsio than available on CentOS by default?

confluence:
Don't we already require some newer versions of libraries to be installed on certain platforms?

ajm-ska:
> > Well I got around the cfitsio pkgconfig problem for now by prefixing the cmake command with
> > `env PKG_CONFIG_PATH=/usr/lib/pkgconfig`
> > (That is where the `cfitsio.pc` file was installed on the CentOS7 server).
> 
> @ajm-asiaa sure, but didn't that still require the installation of a newer version of cfitsio than available on CentOS by default?

Oh that's right. It is not an ideal solution.

confluence:
@ajm is there no RPM package available for a sufficiently new version of cfitsio, even in the additional repositories that we require? If not, it may be a good idea for us to provide one (the way we provide some additional dependency packages for Ubuntu Bionic). It may be possible to rebuild an existing package for a newer RPM distro release.

@veggiesaurus I don't think that supporting older platforms should necessarily mean supporting them with their out-of the box libraries. I will see how easy it would be to downgrade the version required by the converter, but I would prefer not to do this in general unless upgrading the library is going to be problematic.

ajm-ska:
> Don't we already require some newer versions of libraries to be installed on certain platforms?

We did. For CentOS7 I think it needs newer protobuf, grpc, and tbb, I think. Luckily those issues were eliminated by using the OpenStack repo.

ajm-ska:
I spoke too soon last night about `env PKG_CONFIG_PATH=/usr/lib/pkgconfig` being a solution. Although it worked when doing it manually through the terminal to just build the fit2idia submodule, it didn't work when it was added for Jenkins to perform the build for some reason. Neither did `export PKG_CONFIG_PATH=/usr/lib/pkgconfig`.

Instead I found another solution. The pkg-config version with CentOS7 is too old for PkgConfig to find the manually installed cfitsio. The default pkg-config version on CentOS7 is 0.27.1. So instead I installed pkg-config 0.29 from source and now PkgConfig can find cfitsio.

Now this seems to be just an issue with the way I installed the the newer cfitsio from source. Previously pkg-config 0.27.1 was detecting the original package manager installed cfitsio. So if we were to create a proper RPM of a newer cfitsio, then the original pkg-config 0.27.1 may be OK.

Anyway, now Jenkins can finally build this branch on CentOS7.

ajm-ska:
> @ajm is there no RPM package available for a sufficiently new version of cfitsio, even in the additional repositories that we require? If not, it may be a good idea for us to provide one (the way we provide some additional dependency packages for Ubuntu Bionic). It may be possible to rebuild an existing package for a newer RPM distro release.

I can't see any newer el7 RPM for cfitsio. But it should be easy to create one. Plus, although not entirely necessary, it could help us on the way of removing the need of any EPEL packages.

ajm-ska:
I finally got Python3 with numpy and astropy working on the three server now. 
The latest problem is the `ProgramSettingsTest` tests on MacOS. e.g.

```
/Users/acdc/jenkins-ci-mac/workspace/fluence_generate_files_for_tests/test/TestProgramSettings.cc:162: Failure
Expected equality of these values:
  settings.files[0]
    Which is: "Users/acdc/jenkins-ci-mac/workspace/fluence_generate_files_for_tests/build/test/data/images/fits/noise_10px_10px.fits"
  fits_image_path.substr(1)
    Which is: "Users/acdc/jenkins-ci-mac/workspace/fluence_generate_files_for_tests/build/test/./data/images/fits/noise_10px_10px.fits"
[  FAILED  ] ProgramSettingsTest.FileImageFromPositional (0 ms)
```

It looks like it might because there is a dot coming up between the 'test' and 'data' folders. That must not happen on Linux as the tests pass there.  Maybe MacOS handles paths differently?

confluence:
> The latest problem is the `ProgramSettingsTest` tests on MacOS.

@ajm which tests are failing, and do these tests definitely pass on MacOS in the dev branch?

ajm-ska:
@confluence I hope you don't mind but I updated the Dockerfiles and pushed them directly to this branch.
It is basically adding numpy and astropy to Python3. For the CentOS7 Dockerfile, I also installed the newer cfitsio from source. That should do for now. I also finally removed the Ubuntu 16.04 Dockerfile so that we won't need to worry about it anymore.

Also, I updated the `.circleci/config.yml`. Now CircleCI only runs the format check, plus a carta_backend build and Unit Tests in a CentOS7 Docker container.

confluence:
I don't think that we should stop running unit tests on MacOS; if we support MacOS we need to know if there is unexpected Mac-specific behaviour. I'm not sure if we need to run them both on CentOS and on Ubuntu, but they shouldn't take long to run, so I don't think there's a reason not to.

ajm-ska:
> > The latest problem is the `ProgramSettingsTest` tests on MacOS.
> 
> @ajm which tests are failing, and do these tests definitely pass on MacOS in the dev branch?

Failing tests are:
```
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] ProgramSettingsTest.FileImageFromPositional
[  FAILED  ] ProgramSettingsTest.RelativeFileImageFromPositional
[  FAILED  ] ProgramSettingsTest.TrimExtraFolders
[  FAILED  ] ProgramSettingsTest.CasaImageSetFromPositional
[  FAILED  ] ProgramSettingsTest.MultipleImagesFromPositional
[  FAILED  ] UtilTest.ParentNotSubdirectory
```
They appear to pass on `dev` branch. 

You can see them on Jenkins:
[`dev` branch test information](https://carta.asiaa.sinica.edu.tw/blue/organizations/jenkins/carta-backend/detail/dev/358/tests)
[`confluence/generate_files_for_tests` branch test information](https://carta.asiaa.sinica.edu.tw/blue/organizations/jenkins/carta-backend/detail/confluence%2Fgenerate_files_for_tests/21/tests)

I see the HDF5 tests are failing this time on ubuntu 16.04, but were working previously, so I can fix that (It isn't finding the libcfitsio.so.8).

ajm-ska:
> I don't think that we should stop running unit tests on MacOS; if we support MacOS we need to know if there is unexpected Mac-specific behaviour. I'm not sure if we need to run them both on CentOS and on Ubuntu, but they shouldn't take long to run, so I don't think there's a reason not to.

We haven't stopped running unit tests on MacOS?!
The local Jenkins still runs unit tests on MacOS, CentOS7, and Ubuntu 16.04.
All I did was modify the cloud based CircleCI to retire the unsupported CentOS6 and Ubuntu 16.04. I have replaced them with one CentOS7 build.

confluence:
Oh, I see. I misunderstood your comment about CircleCI.

confluence:
The issue with the `ProgramSettingsTest` tests is that their behaviour has changed -- they were previously using the current path as the test root (which is incorrect). Now they are attempting to use the executable path first, which is correct, but apparently an additional `.` is being inserted into this path on MacOS, which is (correctly) being removed by the program settings code, and this is causing the test to fail. I need to investigate further to see where the `.` is coming from.

ajm-ska:
A new strange problem; CircleCI seems to have difficulty cloning the fit2idia repository:

```
Cloning into 'third-party/fits2idia'...
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
Clone of 'https://github.com/idia-astro/fits2idia/' into submodule path 'third-party/fits2idia' failed

Exited with code exit status 1
```
I wonder why? It seems to be public.

confluence:
Could it be a temporary failure?

ajm-ska:
It appears it is not temporary. I tried restarting it 3 times. 
It also gave the same error every time when it was using the old form of the circleci/config.yml file.
You can see it on CircleCI [here](https://app.circleci.com/pipelines/github/CARTAvis/carta-backend?branch=confluence%2Fgenerate_files_for_tests).

ajm-ska:
@confluence I think I found the reason [here](https://discuss.circleci.com/t/cloning-of-a-new-public-github-submodule-fails/26592) 
It says CircleCI doesn't like trailing slashes. 
Could you remove the trailing slash in .gitmodules?
```
[submodule "test/fits2idia"]
        path = third-party/fits2idia
        url = https://github.com/idia-astro/fits2idia/
```

confluence:
I'm going to see if removing the trailing `/` from the submodule URL makes a difference.

confluence:
Snap. :) I have no idea how it got in there.

ajm-ska:
@confluence Good! Now MacOS is down to [one failed test](https://carta.asiaa.sinica.edu.tw/blue/organizations/jenkins/carta-backend/detail/confluence%2Fgenerate_files_for_tests/24/pipeline/141): 

```
[ RUN      ] UtilTest.ParentNotSubdirectory
/Users/acdc/jenkins-ci-mac/workspace/fluence_generate_files_for_tests/test/TestUtil.cc:62: Failure
Value of: IsSubdirectory(pwd.parent_path().string(), pwd.string())
  Actual: true
Expected: false
[  FAILED  ] UtilTest.ParentNotSubdirectory (0 ms)```

confluence:
This is a very weird failure; I'm temporarily changing the test so that I can see what the root directory is actually being set to.

confluence:
This particular test doesn't need to be using the root directory at all, so I can just switch back to using the current directory -- but I would like to understand what is happening.

confluence:
OK, I can see what's happening; I'm trying to fix it. `parent_path` treats paths with and without trailing slashes differently, and I'm assuming that the existing normalization step is stripping off a `.` on macOS, which is why there's a trailing slash on macOS only.

confluence:
OK, all the tests seem to be passing now.

ajm-ska:
@confluence Nice work! 
The only problem now are 4 tests failing in the CircleCI CentOS7 Docker container.

> BasicLoadingTest
> ExampleFriendTest
> CorrectShape2dImage
> CorrectShapeDegenerate3dImages

You can see the details [here](https://app.circleci.com/pipelines/github/CARTAvis/carta-backend/3030/workflows/f15eccc5-ed5f-46e4-b1a1-69a56f3fe02c/jobs/8599/tests#failed-test-0)
It is strange as I don't get the failures when I run the tests in the same Docker container locally. 
I wonder if the problem could be the fact there is no Xvfb present?
I'm thinking to just remove that CentOS7 build from CircleCI and only have CircelCI do the format check.

confluence:
That is weird; they're very basic HDF5 tests. They don't have anything to do with X. I wonder it it's some kind of race condition, and the check is being done before the object is constructed. But I would expect all of these statements to be blocking.

confluence:
I'm going to optimize the file generation and deletion to avoid recreating and deleting the same files repeatedly -- that may fix this as a side effect, if the problem is that one test's teardown method is deleting another test's file with the same name. Either way, it should speed up the execution of the unit tests -- not a significant speedup now, but it will be once we start generating larger files.

confluence:
That did not fix the tests -- the exact same ones are failing.

confluence:
It looks like the HDF5 converter always crashes on two specific FITS files in the circleci build -- but it doesn't crash on any other files, and it doesn't crash when converting the same files anywhere else. I don't know if this could be an issue with one of the installed dependencies. In any case, I think running the unit tests there is redundant (and very very slow), so maybe it is a good idea to remove this build.

confluence:
This might be caused by some uninitialized values in the converter, which I will try to fix.

confluence:
Looks like that was the problem!

ajm-ska:
> Looks like that was the problem!

@confluence Nice work fixing the problem! Although you are right that it is redundant running a CircleCI CentOS7 build and unit tests. The biggest annoyance is that it is so slow to run. So I'm OK with removing it. However, I guess it did serve a useful purpose here by revealing that issue with the converter.

@veggiesaurus What do you think? Should we remove the CircleCI CentOS7 part now and only have CircleCI run the format check? Or add a code coverage check (codecov.io?) to it?

veggiesaurus:
> @veggiesaurus What do you think? Should we remove the CircleCI CentOS7 part now and only have CircleCI run the format check? Or add a code coverage check (codecov.io?) to it?

Ideally, I'd like us to be running our CI stuff on our own hardware. Between the three institutes, we have a wealth of computing resources and I see no real reason to rely on cloud-based CI. I think we can do the code coverage check on our own hardware as well.

ajm-ska:
> > @veggiesaurus What do you think? Should we remove the CircleCI CentOS7 part now and only have CircleCI run the format check? Or add a code coverage check (codecov.io?) to it?
> 
> Ideally, I'd like us to be running our CI stuff on our own hardware. Between the three institutes, we have a wealth of computing resources and I see no real reason to rely on cloud-based CI. I think we can do the code coverage check on our own hardware as well.

Good point. OK. I will remove it now.

veggiesaurus (approved):

Issue #617, opened 2020-08-26
Compressed FITS file support


Kechil:

kswang1029:
test images are available via #682

pford:
Issue #360 (fz) has been included in Epic #788 Carta FITS Support, which has an open PR #803 under review for release 2.0.  Issue #384 (fits.gz) has proven to be trickier and will be removed from #788/#803 for separate development, possibly pushed back to release 3.0 due to the code freeze and beta testing deadlines.

PR #834, opened 2021-06-09
Support fits.gz images using zlib

### LINKED ISSUE #648: feature request: handle fits.gz files

pford:
Closes #648 and #384 in Epic #617 .

New class CompressedFits uses zlib to read headers, find size of decompressed file, and decompress to temporary directory if it will not fit in memory when accessed with cfitsio.

To read headers and find decompressed size, CompressedFits decompresses one FITS block (2880 bytes) at a time to read headers (gzread) and then skips data blocks (gzseek).  All blocks are decompressed (gzread) to be written to a temporary uncompressed file.  The temporary uncompressed file is removed when the FitsLoader exits.

I added debug and performance logging to CompressedFits.  On my system, a 20GB fits.gz image takes ~2 min for headers, ~2 min for size, and ~6 min for decompression to file since it will not fit in memory.  Any suggestions for improvement welcome.


kswang1029:
@pford I see the file info for the original fits and the compressed fits are different. Any idea?
Screen Shot 2021-06-22 at 2 48 40 PM
Screen Shot 2021-06-22 at 2 32 48 PM

kswang1029:
@pford found an issue when the fits.gz is decompressed in RAM: switching to a different channel is extremely slow (it works but very slow) and the animation playback hangs (probably same root cause). 

In addition, the per cube histogram calculation is very slow. If all the data are in RAM, I guess per cube histogram calculation should be quite fast.

Could you have a look?

pford:
Fixed bug in resetting index into FITS block for headers, and read data slice in one cfitsio read to improve performance.

@kswang1029 please re-test.

kswang1029:
@pford yes confirmed that performance issue is gone. 👍 If the image is in RAM, cube histogram or animation playback is faster as expected. 

I spotted a frontend issue: if we request a file info from a fits.gz image but close the file browser before we see the file info, then if we re-launch the file browser, we will see a spinning wheel in the file info tab forever. @veggiesaurus do you have any idea on this UX issue?

pford:
I added debug logging for incoming messages and the time between FILE_INFO_REQUEST (for large fits.gz) and FILE_LIST_REQUEST (for reopened file browser) was ~2 minutes.  Is the frontend waiting for a response before sending another request?  I think the backend would process the file list request if it were sent, while the previous file info request was in progress.

kswang1029:
> I added debug logging for incoming messages and the time between FILE_INFO_REQUEST (for large fits.gz) and FILE_LIST_REQUEST (for reopened file browser) was ~2 minutes. Is the frontend waiting for a response before sending another request? I think the backend would process the file list request if it were sent, while the previous file info request was in progress.

I think the backend does things right and is responsive to file info request instead of queuing up. It appears that it is the frontend UI in a non-ideal state. 🤔

veggiesaurus:
If we're happy with limiting our fits.gz support to just those files that have a single HDU, then the problem is _much_ simpler:

- read the primary HDU in a block-by-block fashion until you get the `END` keyword (this is what you're currently doing, right?). This should be near-instant.
- use the header information to determine the size of the data (e.g. `NAXIS1 * NAXIS2 * NAXIS3 * num_bytes_per_pixel`)

That should be all that is needed for getting file info, and whether or not we can load the file into memory.

pford:
Limiting the file info to the first hdu would definitely be simpler, but incomplete if image extensions are not included.  The decision whether to access in memory or decompress to file is done when the OpenFile message is sent, not FileInfoRequest.  Smaller fits.gz images are handled fairly quickly, it is just the large files (such as KSW's 20G compressed, 28G decompressed file) that have the file info delay.

As I understand it, Cfitsio decompresses the entire fits.gz file to memory, not just the requested HDU, so knowing the data size of the first HDU is insufficient.  This is not the case with 'fz' compression, which Cfitsio decompresses in blocks as needed.

https://heasarc.gsfc.nasa.gov/docs/software/fitsio/compression.html

veggiesaurus:
> Limiting the file info to the first hdu would definitely be simpler, but incomplete if image extensions are not included. 

Right, I guess we'll see what @kswang1029 (and possibly the rest of the science team?) say.

kswang1029:
> > Limiting the file info to the first hdu would definitely be simpler, but incomplete if image extensions are not included.
> 
> Right, I guess we'll see what @kswang1029 (and possibly the rest of the science team?) say.

I think we can keep the current implementation as it is now. It is understandable that accessing a large fits.gz is slower. Equivalently it is CARTA unzipping for users. If users unzip the file beforehand then load it in CARTA, it is also slow for large files. I suggest that we emphasize to users that fits.gz is not an ideal format for image visualization with CARTA. It is just a straightforward format for image archives. It is still recommended to unzip the images first. What CARTA supports now is just a handy shortcut for small gzipped fits images. Would this be sensible? @veggiesaurus @pford

pford:
I agree with adding a warning about fits.gz and providing complete file info, or else it will be confusing if they only see one HDU and they know there are additional extensions.

I updated the branch with dev and resolved conflicts.

kswang1029:
@pford could you add the restoring beam to the file info? It is currently missing for fits.gz.
Screen Shot 2021-07-06 at 12 27 35 PM
Screen Shot 2021-07-06 at 12 27 24 PM
Screen Shot 2021-07-06 at 12 26 58 PM
Screen Shot 2021-07-06 at 12 26 44 PM

After this I will approve the PR. 👍

veggiesaurus:
@kswang1029 do those images have a per-plane beam? that might complicate things considerably

kswang1029:
> @kswang1029 do those images have a per-plane beam? that might complicate things considerably

Hmm those do not have per-plane-beam. I should test the case with that

kswang1029:
I tested a fits.gz cube with per-plane-beam and the beam info is missing so that we do not see a beam is rendered at the bottom-left corner.

pford:
@kswang1029 I have added restoring beam info for a single beam (headers BMAJ, BMIN, BPA).

For per-plane beams, casacore::FITSImage looks for header `CASAMBM = T` (not a FITS standard keyword).  If this header is found, the HDUs are iterated through to find a binary table with EXTNAME containing "BEAMS".  This table is read to create an image beam set, used to calculate a median area beam for the file info and a beam table when the image is loaded.  Obviously, this will affect the performance of the file info and loading an image.  Should multi-beam info be included in both places?

kswang1029:
> @kswang1029 I have added restoring beam info for a single beam (headers BMAJ, BMIN, BPA).
> 
> For per-plane beams, casacore::FITSImage looks for header `CASAMBM = T` (not a FITS standard keyword). If this header is found, the HDUs are iterated through to find a binary table with EXTNAME containing "BEAMS". This table is read to create an image beam set, used to calculate a median area beam for the file info and a beam table when the image is loaded. Obviously, this will affect the performance of the file info and loading an image. Should multi-beam info be included in both places?

Yes please. Performance impact is understandable in this case.

pford:
I added beam info from the headers or a beam table to the file info (using CompressedFits) and OpenFileAck (using CartaFitsImage).  Ready to re-test.

kswang1029:
@pford thanks. The beam info part is fine. While testing the case with per plane beam, I noticed that there might be some conversion errors in the file info generation step (headers are identical). In the following screenshots, we see the reference world coordinates are not identical. In addition, "spectral frame" is duplicated. Could you have a look? The test images are attached [here](https://drive.google.com/file/d/1rkgJ9NrDupA83EM5Tzk9kpeur--9Smfe/view?usp=sharing)
 
Screen Shot 2021-07-14 at 10 05 28 AM

Screen Shot 2021-07-14 at 10 05 37 AM

pford:
I removed the repeated entries, but I do not see a difference in the reference coordinates between the fits and fits.gz files with an Ubuntu backend.  Both files use the same code to convert from header value and unit to a time/angle string using casacore Measures.

kswang1029:
> I removed the repeated entries, but I do not see a difference in the reference coordinates between the fits and fits.gz files with an Ubuntu backend. Both files use the same code to convert from header value and unit to a time/angle string using casacore Measures.

Hmm this probably a Mac only issue. I will do some more tests and file a new issue.

kswang1029 (approved):
Overall this is working well👍

A separate issue about the world coordinates precision on Mac will be filed

src/FileList/FitsHduList.cc
```diff
@@ -18,11 +18,8 @@ FitsHduList::FitsHduList(const std::string& filename) {
 }
 
 void FitsHduList::GetHduList(std::vector<std::string>& hdu_list, std::string& error) {
-    // Returns list of hdu num and ext name for primary array and image extensions
-    if (IsCompressedFits(_filename)) {
-        error = "Compressed FITS gz/bz format not supported yet.";
-        return;
-    }
+    // Returns list of hdu num and ext name for primary array and image extensions.
+    // Note: DO NOT USE for compressed FITS, open decompresses entire file.
```
confluence: Should we not continue to disable this function for compressed files, to safeguard against this being ignored in the future?

confluence (approved):
This looks good; I just have a question about futureproofing the use of `GetHduList`.

confluence:
Did we decide that the read-only setting would not affect writing the unzipped file to disk? I think that this is sensible; writing temporary data to the system temporary directory doesn't really count as "writing". But perhaps this should be documented in the manual, so that it's 100% clear.

src/FileList/FitsHduList.cc
pford: This now throws an exception rather than the previous error message since the error is for the developer not the user.

markccchiang (approved):

Issue #814, opened 2021-05-11
Add tests for spatial profiles


confluence:
This depends on #813 and should be done before #430 is merged so that it can be tested more effectively.

PR #841, opened 2021-06-15
Add basic tests for spatial profiles

confluence:
The bulk of this PR is the implementation of test helper classes for reading data directly from a FITS or HDF5 file (to provide expected data for tests). For this test I implemented utility functions for reading a point or a spatial profile, but arbitrary regions can also be read and more utility functions can be added (e.g. for spectral profiles).

I have only implemented two basic tests (for a small FITS and HDF5 file) -- I intend to add many more tests in the mipmaps branch, which introduces mip and range to spatial profile requests, as well as decimation or downsampling of the returned data (with diverging behaviour for FITS and HDF5). Are there any other tests that I should add to this branch now?

@ajm-asiaa this code uses Google Mock, which is part of Google Test, but is provided in a separate binary package (`libgmock-dev`) on Ubuntu (and possibly on other platforms). We need to install the appropriate packages in our containers to get this to compile. I'm also linking to the C++ API for HDF5, but that must already be installed if the converter (added earlier) can compile.

(This implements #814.)

### DIFF
--- a/test/CommonTestUtilities.cc
+++ b/test/CommonTestUtilities.cc
@@ -78,6 +78,18 @@ std::string FileFinder::XmlTablePath(const std::string& filename) {
     return (TestRoot() / "data" / "tables" / "xml" / filename).string();
 }
 
+float DataReader::ReadPointXY(hsize_t x, hsize_t y, hsize_t channel, hsize_t stokes) {
+    return ReadRegion({x, y, channel, stokes}, {x + 1, y + 1, channel + 1, stokes + 1})[0];
+}
+
+std::vector<float> DataReader::ReadProfileX(hsize_t y, hsize_t channel, hsize_t stokes) {
+    return ReadRegion({0, y, channel, stokes}, {_width, y + 1, channel + 1, stokes + 1});
+}
+
+std::vector<float> DataReader::ReadProfileY(hsize_t x, hsize_t channel, hsize_t stokes) {
+    return ReadRegion({x, 0, channel, stokes}, {x + 1, _height, channel + 1, stokes + 1});
+}
+
 FitsDataReader::FitsDataReader(const std::string& imgpath) {
     int status(0);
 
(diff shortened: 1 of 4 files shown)


confluence:
It looks like it's compiling, so I guess the package is already installed. 🤔

test/CommonTestUtilities.cc
```diff
@@ -77,6 +78,147 @@ std::string FileFinder::XmlTablePath(const std::string& filename) {
...
+
+Hdf5DataReader::Hdf5DataReader(const std::string& imgpath) {
+    _imgfile = H5::H5File(imgpath, H5F_ACC_RDONLY);
+    auto group = _imgfile.openGroup("0");
+    _dataset = group.openDataSet("DATA");
+
+    auto data_space = _dataset.getSpace();
+    _N = data_space.getSimpleExtentNdims();
+    _dims.resize(_N);
+    data_space.getSimpleExtentDims(_dims.data(), NULL);
```
veggiesaurus: Shouldn't those be nullptr for consistency?

test/CommonTestUtilities.cc
```diff
@@ -77,6 +78,147 @@ std::string FileFinder::XmlTablePath(const std::string& filename) {
...
+
+    _stokes = _N == 4 ? _dims[3] : 1;
+    _depth = _N >= 3 ? _dims[2] : 1;
+    _height = _dims[1];
+    _width = _dims[0];
+}
+
+Hdf5DataReader::~Hdf5DataReader() {}
```
veggiesaurus: You can replace this with =default in the header

veggiesaurus (changes requested):
Minor comments

test/CommonTestUtilities.cc
confluence: Shouldn't what be nullptr?

confluence: You mean that `_imgfile` and `_dataset` should be pointers?

veggiesaurus: Sorry, I meant all the NULL arguments should be nullptr instead

confluence: Ah, gotcha.

ajm-ska (approved):
@confluence You might have noticed in the beginning there was trouble building the carta_backend on the ubuntu 16.04 system (I know 16.04 is unsupported, but we are currently stuck with that version as it is the only one compatible with our old Lustre disk array system).

I tried installing a more recent googletest version. That fixed the build issue, but then running the tests gave a segmentation fault. I have since tried installing things again, and now everything compiles and runs OK on ubuntu 16.04.

There was also the Jenkins problem last night so I think Jenkins sending the status back to Github is messed up at the moment, but if you check on Jenkins manually, all tests have passed. 

So everything seems to be OK now.

veggiesaurus (approved):

PR #842, opened 2021-06-16
Refactor ImageMoments and Image2DConvolver

markccchiang:
The original thought is to rewrite the image moments calculator. But I think it is not a good idea to reinvent the wheel. Instead, we can just remove unused codes and some of the casa code dependencies to let the code looks more clean and tidy. The performance of the original code is good enough. So far I do not find a better way to improve it further, especially for HDF5 files with ZXY data (#625 method is still the best one). Changing the chunk data size from 20 MB to 32 MB maybe can improve the performance of FITS files by 5% (according to @acdo2002's tests). But the other image types' performances are roughly the same.


veggiesaurus:
> So far I do not find a better way to improve it further, especially for HDF5 files with ZXY data

HDF5 files have very poor performance with the moment generator. We really need to use the ZYX data to extract decent performance here. Calculating a moment map has a similar complexity in terms of I/O and CPU usage to calculating a region. You're just collapsing a 3D region to a single 2D image instead of a 1D profile. 

A quick test with a real-world image from MeerKAT (roughly 2500 x 2500 x 500) shows the following (after carefully clearing the disk cache), using a 300x300 square region:

FITS region: 2400 ms
HDF5 region: 350 ms (~7 x speedup)

Generating a moment map with the FITS image then takes 1200 ms, because a fair amount of the data on disk has now been cached due to the region calculation. If I clear the disk cache after the region is calculated, but _before_ the moment map is generated, it takes 3300 ms.

On the other hand, the HDF5 moment map is generated with XYZ, so the cached ZYX data doesn't help, and the calculation takes 2600 ms with or without a cache clear (the slight advantage over FITS here is due to the tiles).

Based on the above, I think there is still room for very significant improvements when using HDF5 files. I don't think refactoring things is a bad idea, but I do think we need to make a plan for HDF5 files (#602 assigned to @confluence for later in the 3.0 dev cycle) . We're missing a very significant speedup here.

The code in `Hdf5Loader.cc` would just need to be adapted by @confluence to generate moment maps instead of single 2D profiles.

veggiesaurus:
@markccchiang I thought the plan was to try to submit a PR to CASA, so that we don't retain ownership of this large chunk of code?

markccchiang:
> @markccchiang I thought the plan was to try to submit a PR to CASA, so that we don't retain ownership of this large chunk of code?

As my understanding that I have no right to submit a PR to CASA. It may need @pford's help again to do so.

pford:
I created a new issue in carta-casacore to update casa6 imageanalysis code: https://github.com/CARTAvis/carta-casacore/issues/15.  I will use this PR for the changes needed, and note the progress in the carta-casacore issue.

markccchiang:
@pford Since the code style is different between CARTA and CASA. Is it a proper way that I modify original CASA files `ImageMoments.h`, `ImageMoments.tcc`, `Image2DConvolver.h` and `Image2DConvolver.tcc`. And then send these files to you to submit a PR to CASA? Should I write a brief technical report and summarize the change of the code for the submission?

pford:
You should match the imageanalysis code style since we are changing the source files there.  You can send me the modified files, or put the changes in the carta-backend/src/Moment files in a carta-backend branch. Either is fine, whatever is easier for you.  A description of the code changes and why they are necessary for CARTA would be useful for the casa6 issue.  Is it just to enable cancellation of the generator?  Do the changes keep the same behavior for the moment generator in the casa viewer?

CASA is currently in the 6.3 development cycle, and any new development would have to wait until release 6.4.  The release/6.3 branch will happen sometime in August, then I can open the PR for these changes to go into the CASA master branch for 6.4.  Until then, we can use the casa6 branch commit, which can happen right away, for carta-casacore.  We will need a carta-backend branch with the Moment files removed to test the carta-casacore branch.

PR #827, opened 2021-06-08
757 Use casacore FitsInput for FITS headers

pford:
Closes #757 

The backend will always use casacore::FitsInput (cfitsio interface) for FITS files to show actual headers in file info.  Specifically for this issue, the CTYPE and projection string will include the "SIP" suffix for use by the frontend.

This change breaks FILEINFO_FITS.test.ts since some file info header entries have changed. For example, keyword "PC1_1" is now "PC01_01" for set_QA/M17_SWex.fits.

PR #843, opened 2021-06-17
460 Use HDF5 headers for file info

### LINKED ISSUE #460: casacore processed headers when exceptions happen
I tested with a set of images with some missing headers (set_spectral_conversion). In general it works fine. A peculiar thing I noticed is that depending on cases, casacore may "smartly" fill up missing headers if those can be computed from other relevant headers (e.g., an image with missing OBSGEO but having TELESCOP: casacore will return headers with OBSGEO recovered). Other thing is with a cube in VRAD but with RESTFRQ missing, the casacore returns in FREQ.

test [images](https://drive.google.com/file/d/1GYjF1hOeoL4CMRVY-U1eFDMlXkGWyK6_/view?usp=sharing)

@pford could you help?

pford:
Closes #460 with FITS fix in PR #827.

Uses Hdf5Attributes to get header strings (via CartaHdf5Image) and set header_entries in FileInfoExtended.  This fixes shortcomings in casacore ImageHeaderToFITS including (1) removing headers with non-FITS keywords (such as HDF5 schema and converter headers), (2) adding FITS keywords (BLANK), and (3) sometimes computing headers which do not exist in image headers as described in #460.

### DIFF
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -14,7 +14,6 @@
 #include <casacore/casa/Quanta/MVAngle.h>
 #include <casacore/coordinates/Coordinates/DirectionCoordinate.h>
 #include <casacore/coordinates/Coordinates/SpectralCoordinate.h>
-#include <casacore/fits/FITS/fits.h>
 #include <casacore/fits/FITS/hdu.h>
 #include <casacore/images/Images/FITSImage.h>
 #include <casacore/images/Images/ImageFITSConverter.h>
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -14,6 +14,7 @@
 #include <casacore/casa/Quanta/MVAngle.h>
 #include <casacore/coordinates/Coordinates/DirectionCoordinate.h>
 #include <casacore/coordinates/Coordinates/SpectralCoordinate.h>
+#include <casacore/fits/FITS/fits.h>
 #include <casacore/fits/FITS/hdu.h>
 #include <casacore/images/Images/FITSImage.h>
 #include <casacore/images/Images/ImageFITSConverter.h>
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):
as far as I can tell with some hdf5 images I have, this PR works fine. 👍

confluence:
When I run this branch (with the latest frontend) the backend crashes after logging `[critical] stod` when I attempt to display any HDF5 file in the file browser.

kswang1029:
> When I run this branch (with the latest frontend) the backend crashes after logging `[critical] stod` when I attempt to display any HDF5 file in the file browser.

Could you send me a small hdf5 image to test with?

confluence:
Hmm, it looks like it's not *any* file -- I'm checking in more detail which files trigger the issue and which do not.

confluence:
I have identified two images that are affected. They are both very large, but I'm attaching the headers here (since this is a header parsing issue).

I think that I know what the issue is with `XMMLSS_3_contcube` -- it has a `RESTFRQ` value of `4.940656458412E-324`, which is too small to be representable by a double, and this is causing `stod` to fail. I had to [add  a workaround to the HDF5 converter](https://github.com/idia-astro/fits2idia/commit/a3d7343d46881cfefab9007d6caeaf05125da803) to deal with this by parsing the value as a long double and converting to double (I don't recall whether we previously also had to handle this case in any of the CARTA header parsing code).

I can't locate a similar small value in the `ultravista` header, so that seems to be a different string to double parsing issue.

* [ultravista-dr3-k-band.fits.header.txt](https://github.com/CARTAvis/carta-backend/files/6693401/ultravista-dr3-k-band.fits.header.txt)
* [XMMLSS_3_contcube.fits.header.txt](https://github.com/CARTAvis/carta-backend/files/6693402/XMMLSS_3_contcube.fits.header.txt)

pford:
Pushed a fix for the headers:  if the value starts with a single quote, do not convert to int/double.  If conversion to int/double throws out_of_range exception, try to convert to long int/double.

confluence:
I can confirm that this fixes the issue with the two files I tested. Every HDF5 file I tried now works.

confluence (approved):

kswang1029:
Re-tested with 6dad03e and it still works fine with my few test hdf5 images. Ready to merge. 👍

PR #852, opened 2021-06-22
Angus/848 garbled messages

### LINKED ISSUE #848: fix issues with contour animation when contour compression is disabled
As discovered by @kswang1029, playing back an animation with contours at a high frame rate can result in garbled ICD messages when using the MacOS electron app. @ajm-asiaa worked out that this only occurs with the electron app (and not the homebrew install) because contour compression is disabled. 

This may be because we're instructing uWebSockets to compress the uncompressed contour messages using `deflate`. Perhaps the buffer size is too small :thinking:

veggiesaurus:
Closes #848 by upgrading to uWebSockets v19.2.0

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -46,12 +46,6 @@ carta::ProgramSettings settings;
 // Sessions map
 std::unordered_map<uint32_t, Session*> sessions;
 
-// Apply ws->getUserData and return one of these
-struct PerSocketData {
-    uint32_t session_id;
-    string address;
-};
-
 void DeleteSession(int session_id) {
     Session* session = sessions[session_id];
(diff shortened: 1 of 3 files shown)


confluence (approved):

markccchiang (approved):

ajm-ska (approved):
I did the standard test (Set Contour Compression Level to 0, open HD163296_CO_2_1.fits, generate 5 min-max-scaling contours, set animator frame rate to 15, play the animation, and keep an eye on the console log) on MacOS, CentOS7, and Ubuntu. 
There were no garbled console log messages, so this has certainly fixed the problem.

PR #855, opened 2021-06-25
Angus/853 directory info

### LINKED ISSUE #853: Populate file list response with modification time and number of items for each directory
Request from a group of users at U. Toronto to also display the modification time for directories in CARTA. I think we should also display the number of items (easy enough to determine using `fs::directory_iterator`).

Example from the Ubuntu file browser below:

image

veggiesaurus:
Closes #853


confluence (approved):
Looks good to me. Please also make a PR for the protobuf change.

PR #859, opened 2021-06-29
Confluence/729 consolidate style scripts

### LINKED ISSUE #729: Consolidate all formatting and style checks in a single script
We currently have separate scripts for checking formatting style, reformatting, checking and fixing headers, and checking more high-level coding style (which we don't use much at the moment). I propose folding all these scripts into what is now the header script, which is written in Python. This will simplify both using and maintaining this functionality.

We can also take this opportunity to add more custom features, like enforcing newlines at end-of-file; something which `clang-format` currently doesn't do and probably won't do.

confluence:
This fixes #729. Style checking with clang-tidy is still not in a usable state, but it is now partially implemented in the same script. I have added a check and a fix for missing newlines at the ends of files.

The new usage for doing all the checks (except for the style) is: `./scripts/style.h all check`, and the new usage for fixing all the problems automatically is `./scripts/style.h all fix`. This replaces the separate header and format checks, and should be fixed in the automated tests.

In "check" mode the script should exit with a return code of 1 if any issues are found, and 0 otherwise.

I would specifically like to check if everyone can see the format output diff in the correct order -- the "wrong" format should be displayed as "subtracted" and the "correct" format should be displayed as "added". We have previously swapped the diff order around several times, and I'm not sure if this is because BSD diff behaviour is different. Does this look right on a Mac?

This PR also fixes all the missing trailing newlines.


ajm-ska:
@confluence We can easily fix the old CircleCI format check by changing the command in line 9 of `.circleci/config.yml` and removing the second run step that checked the headers (lines 12-15). However, it looks like I need to update the Docker container with Python 3.7 first. At the moment it is using Python 3.6.9, and it looks a function you are using was introduced in Python 3.7:
   
     TypeError: __init__() got an unexpected keyword argument 'capture_output'

I will update the Docker container soon.

I also tested it on MacOS (python 3.9.5). `./scripts/style.py all check` works if the header is missing and `./scripts/style.py all fix` puts it back, but I can't seem to test it for formatting. For example, if I modify line 74 of `src/AnimationObject.h` to put the bracket on the next line e.g.

        if (delta_frame.channel() < 0 || delta_frame.stokes() < 0) 
        {

Then `./scripts/style.py all check` reports nothing wrong!?

confluence:
@ajm that's weird. What version of `clang-format` do you have installed?

ajm-ska:
@confluence Looks like there was no `clang-format` installed. I just installed it now using Homebrew. 
It installs `clang-format version 12.0.0`. 
However, `./scripts/style.py all check` still gives no output.

Shall I make the modification to `.circleci/config.yml` on this branch?

confluence:
Yes, please edit `.circleci/config.yml`.

Is `clang-format` on your path? What happens if you run it manually on the modified file? Does the output contain a fix?

confluence:
@ajm it's weird that you didn't get any error output when clang-format was not installed. What do you see if you pass in a non-existent custom executable? E.g. `./scripts/style.py all check --clang-format zzzzzzzzzzz`

ajm-ska:
`which clang-format` shows that it is at `/usr/local/bin/clang-format`.
`echo $PATH` shows `/usr/local/bin` so it is in the path.

Doing `clang-format src/AnimationObject.h` shows the file in the terminal.

Doing `./scripts/style.py all fix --clang-format zzzzzzzzzzz` gives no output.

I'm trying to figure it out.

confluence:
> Doing `clang-format src/AnimationObject.h` shows the file in the terminal.

Yes, but does the file have the expected fix applied?

You should definitely be seeing error output from the command with the incorrect executable, so something appears to be suppressing errors.

Is your shell Bash? And is there a bash executable at `/bin/bash` on the Mac?

ajm-ska:
Yes. When `clang-format src/AnimationObject.h` lists the file contents in the terminal, it does show it with the expected fix.

The shell is `/bin/zsh`, but `/bin/bash` is present too. I switched to bash and tried it, but it is the same; no output after running.

I'm delayed by another problem at the moment. I can't push the `.circleci/config.yml` modification so trying to read up how to set up and use the personal access token.

> remote: Password authentication is temporarily disabled as part of a brownout. Please use a personal access token instead.
> remote: Please see https://github.blog/2020-07-30-token-authentication-requirements-for-api-and-git-operations/ for more information.
> fatal: unable to access 'https://github.com/CARTAvis/carta-backend.git/': The requested URL returned error: 403

confluence:
You can check out the ssh url and use ssh keys.

veggiesaurus (approved):

ajm-ska:
@veggiesaurus Does it work on MacOS for you?

veggiesaurus:
> @veggiesaurus Does it work on MacOS for you?

No, it doesn't work for me on MacOS. However, the old `reformat` and `checkformat` scripts didn't work either. @confluence I think your `find` call might be acting up on MacOS for some reason.

Edit: here's the reason: https://stackoverflow.com/questions/5905493/why-isnt-this-regex-working-find-regex-m-h

veggiesaurus (changes requested):
`find` command needs adjusting for MacOS support

confluence:
That may explain the lack of error output. I can eliminate the `find` calls entirely and iterate over the files in Python. It probably won't be any slower. I will look at this tomorrow.

ajm-ska:
@confluence Nice! That works now on MacOS, and continues to work on Linux.
The - and + have turned into < and >, but that is still logical. One would assume the original lines are on the left and the new lines on the right.

```
Bad format found in src/AnimationObject.h
60,61c60
<         if (!matched_frames.empty()) 
<         {
---
>         if (!matched_frames.empty()) {
```

I just have one question; How do you invoke the "Missing newline at end of" checks?
If there is no newline at the end of a file, it gives no error.
If I add some newlines (returns) at the end, the format check picks them up:

```
Bad format found in src/Main.cc
735,736d734
< 
< 

```

confluence:
Missing newline character at end of file, not missing blank lines. `cat` the file and see if your prompt is appended to the last line.

Try making a file like this: `echo -n "foo" > src/foo.cc`

ajm-ska:
Thanks. I understand now.

ajm-ska (approved):
Works fine with MacOS and Linux.

PR #860, opened 2021-06-30
774 Add a mutex lock on setting image channels

### LINKED ISSUE #774: backend crash when attempting to load a per-plane-beam cube as new while the current one is animating
to reproduce:
1) load set_s255/S255_IR.spw25.cube.I.pbcor.fits and trigger channel animation
2) load a new image (not appending) set_perPlaneBeam/SDC335.579-0.292.spw0.line.fits
3) the backend should crash.

changing the order of step 1 and 2 seems fine and the backend survives.

markccchiang:
It is to fix #774. The backend crashes if we delete a Frame when its animation process is not finished yet. This could be solved by simply adding a mutex lock on setting image channels.

### DIFF
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -584,6 +584,7 @@ void Session::OnAddRequiredTiles(const CARTA::AddRequiredTiles& message, bool sk
 
 void Session::OnSetImageChannels(const CARTA::SetImageChannels& message) {
     auto file_id(message.file_id());
+    std::unique_lock<std::mutex> lock(_frame_mutex);
     if (_frames.count(file_id)) {
         auto frame = _frames.at(file_id);
         std::string err_message;


veggiesaurus (approved):

veggiesaurus:
We're already limiting the number of `SetImageChannels` messages we handle at once to 1, so I guess adding a mutex won't make any difference to performance.

confluence (approved):

kswang1029 (approved):
issue solved. 👍

veggiesaurus:
It seems to be failing an ICD test 🤔

kswang1029:
I tried to close an image while it is streaming region spectral profile and it works fine. @acdo2002 could you check why the ICD test-close_file is failing?

acdo2002:
I am investigating now, so far it looks like the internet problem. Because from the ICD message shows the spectral profile streaming does not complete within a given time. It only completes the progress of 0.24, but it should complete the progress of 0.5.

And also I cannot connect to asiaa vpn now...

acdo2002:
I build mark's backend on MacOS & Ubuntu, and run the mark's branch on acdc CentOS server.
And running the ICD tests of all CLOSE_FILE*(using ws), they are all passed.

Then I may suggest it is because the internet(streaming is too slow) problem.

PR #864, opened 2021-07-02
Use H5Aiterate2 to iterate over HDF5 attributes; add legacy fallback for HDF5 boolean value detection

### LINKED ISSUE #861: Improve speed of HDF5 header parsing
The FITS file `hlsp_candels_hst_acs_gn-tot-60mas_f606w_v1.0_drz.fits` has roughly 13000 header entries. The file info takes ~2 seconds to load. After converting to HDF5, it takes about 30 seconds to load the file info in CARTA. 

I don't think this is an inherent limitation of the HDF5 file format though, because viewing the attributes in HDF Compass takes about 2 seconds as well. We should be able to get the HDF5 load time down to ~4 seconds as a maximum in CARTA.

confluence:
Should fix #861. I haven't yet tested the performance.


src/ImageData/Hdf5Attributes.cc
```diff
@@ -16,35 +16,40 @@
...
+
+        H5Tclose(atype);
+        H5Sclose(aspace);
+        H5Aclose(attr);
+
+        return 0;
+    };
+
+    // Iterate over the attributes using the order in which they were written
+    H5Aiterate2(group_hid, H5_INDEX_CRT_ORDER, H5_ITER_INC, NULL, attr_info, &headers);
```
veggiesaurus: `NULL` -> `nullptr`

veggiesaurus (changes requested):

src/ImageData/Hdf5Attributes.cc
veggiesaurus: (and above)

veggiesaurus (approved):

confluence:
I see a speed improvement when I load the large test file from SSD. It's still not instantaneous, but definitely much faster than `dev`.

veggiesaurus:
I can confirm the significant speed improvement over `dev`. For `hlsp_candels_hst_acs_gn-tot-60mas_f606w_v1.0_drz.fits` (13 000 header entries):

FITS (`dev`): 0.65 s
HDF5 (`dev`): 35.1 s
HDF5 (this branch): 2.0 s

veggiesaurus:
@confluence I've just noticed that `Hdf5Attributes::ReadAttributes` is getting called twice when responding to a single file info. Each time takes around 800 ms, so we could probably improve this quite a bit more. Any idea why its getting called twice?

edit: 

### First stack:
```
Hdf5Attributes::ReadAttributes(long) Hdf5Attributes.cc:19
carta::CartaHdf5Image::Hdf5ToFITSHeaderStrings() CartaHdf5Image.cc:244
carta::CartaHdf5Image::SetUpImage() CartaHdf5Image.cc:162
carta::CartaHdf5Image::CartaHdf5Image(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, casacore::MaskSpecifier) CartaHdf5Image.cc:34
carta::Hdf5Loader::OpenFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) Hdf5Loader.cc:21
FileExtInfoLoader::FillFileInfoFromImage(CARTA::FileInfoExtended&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) FileExtInfoLoader.cc:51
FileExtInfoLoader::FillFileExtInfo(CARTA::FileInfoExtended&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) FileExtInfoLoader.cc:41
Session::FillExtendedFileInfo(std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, CARTA::FileInfoExtended, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, CARTA::FileInfoExtended> > >&, CARTA::FileInfo&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) Session.cc:215
Session::OnFileInfoRequest(CARTA::FileInfoRequest const&, unsigned int) Session.cc:321
...
```

### Second stack:

```
Hdf5Attributes::ReadAttributes(long) Hdf5Attributes.cc:19
carta::CartaHdf5Image::Hdf5ToFITSHeaderStrings() CartaHdf5Image.cc:244
FileExtInfoLoader::FillFileInfoFromImage(CARTA::FileInfoExtended&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) FileExtInfoLoader.cc:75
FileExtInfoLoader::FillFileExtInfo(CARTA::FileInfoExtended&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) FileExtInfoLoader.cc:41
Session::FillExtendedFileInfo(std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, CARTA::FileInfoExtended, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, CARTA::FileInfoExtended> > >&, CARTA::FileInfo&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) Session.cc:215
Session::OnFileInfoRequest(CARTA::FileInfoRequest const&, unsigned int) Session.cc:321
...
```

confluence:
The first time is when the HDF5 image is opened; the second time is in the info loader, because we don't cache the values. Should be easy to fix; I see no reason not to cache this.

confluence:
Something weird is happening with the CI tests -- the unit tests are failing on Ubuntu because booleans are being treated as integers. But I tested this on Ubuntu. This looks suspiciously like an architecture-dependent type difference.

We've had problems with the detection of booleans before, but IIRC it was a converter version issue.

I have checked the code and the files produced by the converter. The booleans are definitely still stored as unsigned 8-bit integers, and this is definitely still what we're checking for. It's the same code. The function that we use to get the type of an attribute is also the same.

We also know that the same converter version is running, because the previous assertion in the test checks the converter version attribute in the header.

confluence:
I think that this ended up being an [HDF5 library version issue](https://github.com/CARTAvis/carta-backend/pull/750). If we're running these unit tests on Bionic, then we are probably using an old version of the library, which is causing these attributes to be converted incorrectly. Which I now realise is a problem that we may need to fix in the converter, since we are supporting it on Bionic and providing a Bionic package.

confluence:
This doesn't mean that we can't merge this PR -- it isn't introducing a new bug; it's just exposing one with a new unit test. But the tests should all pass, so I will make a converter issue and prioritise it.

confluence:
Upon reflection, I think it may be more sensible for us to check for both the old and the new native boolean type. This will allow us to process both old and new files correctly, and there should be no confusion with "real" integer attributes because we use signed 64-bit integers for those. I will update this PR.

confluence:
I am investigating the failing test on MacOS. It looks like the statistics datasets aren't being written correctly by the converter on MacOS, but that should be independent of this PR.

I re-ran the tests and they passed -- looks like an intermittent error.

veggiesaurus:
> The first time is when the HDF5 image is opened; the second time is in the info loader, because we don't cache the values. Should be easy to fix; I see no reason not to cache this.

This branch is down to 1.2 seconds with the caching approach now :rocket: just 30x faster than before :+1:

confluence:
I think that it would be good to patch the release version with this -- we've just started using the packaged version of the converter on ILIFU, and ILIFU is on Bionic, which means that all our users are about to start making files with the "wrong" booleans.

veggiesaurus:
> I think that it would be good to patch the release version with this -- we've just started using the packaged version of the converter on ILIFU, and ILIFU is on Bionic, which means that all our users are about to start making files with the "wrong" booleans.

good idea

confluence:
OK, I've checked, and Bionic is on HDF5 1.10.0. Are we using Xenial in the CI tests?

Going by the git version, I assume so. I think we should keep the change, so that we can display legacy files correctly, but a backport is not necessary (since we only officially support the converter on Focal and Bionic, and the ILIFU converter is going to be the Bionic package).

veggiesaurus:
I've added @kswang1029 as a reviewer so that he can just do a sanity check and ensure that the set of HDF5 test images still load ok. All the ones that loaded in `dev` still load fine for me in this branch

kswang1029 (approved):
This works fine as far as I can tell with my hdf5 images.

markccchiang (approved):

PR #867, opened 2021-07-08
866 Set FileLoader _width, _height to fix TileCache seg fault

### LINKED ISSUE #866: Multiple problems with specific HDF5 files
The [latest commit](https://github.com/CARTAvis/carta-backend/commit/c9d08197e69f4ef384840cd12785b6651198da19) seems to have introduced several issues which affect specific HDF5 files. Some files (e.g. `M17_SWex.hdf5`) cause the backend to crash if they are loaded, while some files cause the backend to freeze / hang after they are loaded (a symptom of this was [this frontend issue](https://github.com/CARTAvis/carta-frontend/issues/1552); spatial profiles don't update; etc.).

These issues are not present in the previous commit.

pford:
Closes #866

Restore setting _width and _height in FileLoader, used in TileCache.

### DIFF
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -127,9 +127,9 @@ bool FileLoader::FindCoordinateAxes(IPos& shape, int& spectral_axis, int& z_axis
 
     // Determine which axes will be rendered
     std::vector<int> render_axes = GetRenderAxes();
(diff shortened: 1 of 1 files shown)


confluence (approved):
I have re-tested all the files which were previously triggering errors, and it looks like this fixes the issue. Thanks!

Issue #796, opened 2021-04-13
support spatial profile extraction from a line/polyline region 


kswang1029:
the relevant frontend issue is https://github.com/CARTAvis/carta-frontend/issues/1617

This is to extract mean values over a width along the line/polyline.

- [x] v3b1: support point region request
- [ ] v3b3: support line region request

Detailed doc will be provided.  
https://docs.google.com/document/d/1mE-JeZjczT_GEPfllm-aNWWUZ903wTJy4JurxBq537c/edit#

kswang1029:
@markccchiang for v3b1 will just need to support point region. line region will be scheduled for v3b2.

kswang1029:
We will need to reschedule this to beta3.

PR #872, opened 2021-07-16
796 Support spatial profile for a point region

markccchiang:
It is for the first task in #796. The backend supports the spatial profile for a point region. This branch works with the frontend branch [jason/issue_1399](https://github.com/CARTAvis/carta-frontend/tree/jason/issue_1399) with respect to the PR [#1576](https://github.com/CARTAvis/carta-frontend/pull/1576).


veggiesaurus (dismissed):
Looks good to me. Just need to make sure it still works ok with hdf5 files

veggiesaurus (approved):
Looks good to me. Just need to make sure it still works ok with hdf5 files

kswang1029 (approved):
this works fine. also tested with hdf5 images.

src/Session.cc
```diff
@@ -1387,10 +1390,23 @@ bool Session::SendSpatialProfileData(int file_id, int region_id) {
                 data_sent = true;
             }
         }
+    } else {
+        string error = fmt::format("Spatial profiles not valid for non-cursor or non-point region ", region_id);
```
confluence: Minor text correction: `non-cursor or non-point` -> `non-cursor and non-point`

Or maybe just `non-point`, since a cursor region is a kind of point region?

confluence (approved):
I have a minor comment about a text message; otherwise this looks good.

src/Session.cc
veggiesaurus: right, `non-point` should be sufficient.

confluence (approved):

pford (approved):

markccchiang:
According to the discussion from the [frontend PR 1576](https://github.com/CARTAvis/carta-frontend/pull/1576), I modified the backend so that can get spatial profile data with different stokes for a cursor/point region without an additional `SET_IMAGE_CHANNELS`.

veggiesaurus:
> According to the discussion from the [frontend PR 1576](https://github.com/CARTAvis/carta-frontend/pull/1576), I modified the backend so that can get spatial profile data with different stokes for a cursor/point region without an additional `SET_IMAGE_CHANNELS`.

@markccchiang doesn't this mean that we can end up with the whole channel for each Stokes in memory? What happens when you change channels? Does it need to load the new channel for each Stokes?

(I'm not saying this is the wrong way to approach the issue, I can't really offer a better suggestion :thinking: )

src/Frame.cc
```diff
@@ -998,192 +1008,227 @@ bool Frame::FillSpatialProfileData(int region_id, CARTA::SpatialProfileData& spa
...
-        int mip(config.mip());
+    // Get point region spatial profile data with respect to the stokes (key)
+    for (auto& point_regions_spatial_config : point_regions_spatial_configs) {
+        int stokes = point_regions_spatial_config.first;
 
-        if (!end) {
-            end = config.coordinate() == "x" ? _width : _height;
+        bool is_current_stokes(true);
+        if (stokes != CurrentStokes()) {
+            if (_image_caches[stokes].empty() && !FillImageCache(stokes)) {
```
confluence: If the image cache is going to be per-Stokes, then the ` _image_cache_valid ` flag must also be per-Stokes, and you should check that consistently instead of checking either whether the per-Stokes cache is empty or whether a single flag is set (which is no longer meaningful).

src/Frame.cc
```diff
@@ -91,7 +91,7 @@ Frame::Frame(uint32_t session_id, carta::FileLoader* loader, const std::string&
     _num_stokes = (_stokes_axis >= 0 ? _image_shape(_stokes_axis) : 1);
 
     // load full image cache for loaders that don't use the tile cache and mipmaps
-    if (!(_loader->UseTileCache() && _loader->HasMip(2)) && !FillImageCache()) {
+    if (!(_loader->UseTileCache() && _loader->HasMip(2)) && !FillImageCache(CurrentStokes())) {
```
confluence: I suggest making Stokes an optional parameter which defaults to `CURRENT_STOKES` and is set to `CurrentStokes()` _inside_ the function. We could then avoid changing this everywhere.

It would also be cleaner to write accessor functions for the image cache, mutexes, etc., which have the same behaviour. Then instead of using `_image_caches[CurrentStokes()]` everywhere we could just write `ImageCache()` for the default or `ImageCache(some_other_stokes)`.

src/Frame.cc
```diff
@@ -998,192 +1008,227 @@ bool Frame::FillSpatialProfileData(int region_id, CARTA::SpatialProfileData& spa
...
+    std::unordered_map<int, std::vector<CARTA::SetSpatialRequirements_SpatialConfig>> point_regions_spatial_configs;
+
+    for (auto& config : _point_regions_spatial_configs[region_id]) {
+        // Get stokes
+        std::string coordinate(config.coordinate());
+        int stokes;
+        if (!GetStokesTypeIndex(coordinate, stokes)) {
+            continue;
+        }
+        if (stokes < 0) {
```
confluence: I would prefer to check for equality to `CURRENT_STOKES` explicitly. This should probably also be changed in `GetStokesTypeIndex`.

confluence (changes requested):

pford (changes requested):
My understanding is that the image cache holds the data for the current channel and current stokes which is rendered in the frontend.  The cache should not be reloaded just for a spatial profile.  Frame::GetImageSlicer can be used to select other channels/ranges and stokes and then Frame::GetSlicerData is used to load the profile.  Or am I missing something?

veggiesaurus:
> My understanding is that the image cache holds the data for the current channel and current stokes which is rendered in the frontend. The cache should not be reloaded just for a spatial profile. Frame::GetImageSlicer can be used to select other channels/ranges and stokes and then Frame::GetSlicerData is used to load the profile. Or am I missing something?

That's more in line with what I was expecting.

pford:
GetImageSlicer does not allow you to select x and y (gets the entire xy plane for channel and stokes), so either it will need to be modified with x and y parameters (default ALL), or create a new GetImageSlicer(x, y, z, stokes) where any of these parameters can be the entire range ALL or perhaps an AxisRange.  We currently only have a constant for ALL_Z (-2) so we will need to generalize it. CURRENT_Z is -1 and CURRENT_STOKES is -1.  But I think a Slicer is the way to go here.

markccchiang:
> GetImageSlicer does not allow you to select x and y (gets the entire xy plane for channel and stokes), so either it will need to be modified with x and y parameters (default ALL), or create a new GetImageSlicer(x, y, z, stokes) where any of these parameters can be the entire range ALL or perhaps an AxisRange. We currently only have a constant for ALL_Z (-2) so we will need to generalize it. CURRENT_Z is -1 and CURRENT_STOKES is -1. But I think a Slicer is the way to go here.

@pford Thank you for the suggestions! Have changed the code to get spatial profile data for different stokes in this way.

src/Frame.cc
```diff
@@ -998,192 +1041,232 @@ bool Frame::FillSpatialProfileData(int region_id, CARTA::SpatialProfileData& spa
...
-                    auto tile_start = y_offset + max(start - tile_x, 0);
-                    auto tile_end = y_offset + min(end - tile_x, tile_width);
-                    auto profile_start = profile.begin() + max(tile_x - start, 0);
-                    std::copy(tile_start, tile_end, profile_start);
-                }
+            profile.clear();
+            bool have_profile(false);
 
-                have_profile = true;
+            if (_loader->UseTileCache() && (mip < 2 || !_loader->HasMip(2)) &&
```
pford: This feels confusing, with many `if` statements at the same level, testing combinations of mip and _loader->HasMip and is_current_stokes; most of these existed prior to this PR.  What do you think about refactoring this? (double check my logic is correct!)  I also think GetDownsampledRasterData can be used for any stokes since it has a `stokes` parameter (@confluence) ?
```
bool downsample(mip >= 2);

if (downsample && _loader->HasMip(2)) {
    // GetDownSampledRasterData with stokes in config
} else {
    if (downsample) {
        // Set decimated_start/end and update start/end
    }

    if (is_current_stokes) {
        if (_loader->UseTileCache()) {
            // Get profile from _tile_cache
        } else {
            // Get profile from _image_cache
        }
    } else {
        // GetSlicerData
    }

    if (have_profile && downsample) {
        // decimate profile
    }
}
```

src/Frame.cc
```diff
@@ -998,192 +1041,237 @@ bool Frame::FillSpatialProfileData(int region_id, CARTA::SpatialProfileData& spa
...
-        // Round the endpoints if we're going to decimate
-        if (mip >= 2 && !_loader->HasMip(2)) {
-            // These values will be used to resize the decimated data
-            decimated_start = std::ceil((float)start / (mip * 2)) * 2;
-            decimated_end = std::ceil((float)end / (mip * 2)) * 2;
-            // These values will be used to fetch the data to decimate
-            start = decimated_start * mip;
-            end = decimated_end * mip;
-            end = config.coordinate() == "x" ? std::min(end, _width) : std::min(end, _height);
+        bool is_current_stokes(true);
```
pford: bool is_current_stokes(stokes == CurrentStokes());

src/Frame.cc
```diff
@@ -998,192 +1041,237 @@ bool Frame::FillSpatialProfileData(int region_id, CARTA::SpatialProfileData& spa
...
+                    bounds.set_x_min(x_floor);
+                    bounds.set_x_max(x_floor + mip);
+                    bounds.set_y_min(start);
+                    bounds.set_y_max(end);
+                }
 
-                    for (int j = tile_start; j < tile_end; j++) {
-                        profile[profile_start + j - tile_start] = (*tile)[(j * tile_width) + (x - tile_x)];
-                    }
+                if (_loader->GetDownsampledRasterData(profile, CurrentZ(), stokes, bounds, mip, _image_mutex)) {
```
pford: have_profile = _loader->GetDownsampledRasterData(...);

src/Frame.cc
```diff
@@ -998,192 +1041,237 @@ bool Frame::FillSpatialProfileData(int region_id, CARTA::SpatialProfileData& spa
...
-                    auto idx = x_start + j;
-                    profile.push_back(_image_cache[idx]);
-                }
-                cache_lock.release();
-            } else if (config.coordinate() == "y") {
-                tbb::queuing_rw_mutex::scoped_lock cache_lock(_cache_mutex, write_lock);
-                for (unsigned int j = start; j < end; ++j) {
-                    auto idx = (j * _width) + x;
-                    profile.push_back(_image_cache[idx]);
+                    if (GetSlicerData(section, profile)) {
```
pford: have_profile = GetSlicerData(...);

pford (changes requested):
Remove a few more unnecessary `if` blocks, then it's good.

pford (approved):

kswang1029:
@markccchiang @duidae The point region spatial profiles can be derived and rendered properly for each of the loaded images. After matching, point region spatial profile disappears. The error message from the backend is "Spatial requirements not valid for non-cursor or non-point region ". Any idea?

https://user-images.githubusercontent.com/20819712/128133868-d9bc4edb-33c4-4f19-94a0-b5ec3b82e523.mov

confluence (changes requested):
Could you please extend the spatial profile unit tests to check that profiles for other Stokes are fetched correctly? Ideally this should be added to each of the existing test cases (for all the different file formats and sizes), but there should at least be one for FITS files and one for large HDF5 files.

markccchiang:
> @markccchiang @duidae The point region spatial profiles can be derived and rendered properly for each of the loaded images. After matching, point region spatial profile disappears. The error message from the backend is "Spatial requirements not valid for non-cursor or non-point region ". Any idea?

@duidae Could you please check does the frontend sends the right `file id` when requiring the spatial profile data? In @kswang1029's test as shown above. When two images are matched on `xy` plane, there is no problem getting point regions spatial profile data for the first image. But it has a problem getting point regions spatial profile data for the second image. It seems the frontend does not change the `file id` when switching to the second image.

confluence (approved):

duidae:
> > @markccchiang @duidae The point region spatial profiles can be derived and rendered properly for each of the loaded images. After matching, point region spatial profile disappears. The error message from the backend is "Spatial requirements not valid for non-cursor or non-point region ". Any idea?
> 
> @duidae Could you please check does the frontend sends the right `file id` when requiring the spatial profile data? In @kswang1029's test as shown above. When two images are matched on `xy` plane, there is no problem getting point regions spatial profile data for the first image. But it has a problem getting point regions spatial profile data for the second image. It seems the frontend does not change the `file id` when switching to the second image.

@markccchiang As we can see when switching to the second file, setSpatialRequirement gives the new fileId but no spatial profile is returned, the problem should be here.

https://user-images.githubusercontent.com/25971696/128654491-0756b83a-bcdb-4c77-a2fb-a43b20630d66.mp4

kswang1029 (approved):

Issue #812, opened 2021-05-10
Improve --browser command-line argument behaviour


veggiesaurus:
Need to emulate behaviour found in jupyter-notebook (browser opening up as a separate process, etc).

veggiesaurus:
assigning to @jolopezl. @confluence can help with the investigations.

confluence:
I think examining what Jupyter / `webbrowser.open` / `subprocess.Popen` is doing in more detail will help us figure out what to do. We need to 1) launch the browser in a new session, 2) disown it so that closing the terminal doesn't kill it and 3) decide what to do with the output. My preference is to suppress it, for parity with the default browser command.

jolopezl:
@confluence, @veggiesaurus: I tried several things for improving the browser command, including `boost::process` and system calls such as `fork()`, `setsid()`, and `execv()`, but none of them was actually the correct approach. Something along the lines of:
```
    pid_t pid = fork();
    if (pid < 0) {
        // fork fails
    } else if (pid == 0) {
        pid_t sid = setsid();
        exec(cmd.c_str(), NULL); // std::system(cmd.c_str());
    } else {
        // parent
    }
```
works relatively ok, but it changes the backend process, and it doesn't really exit when killing it as usual but stays in the background so that if you launch it again using the same port it will fail.

I suggest that we don't add more changes to the browser command now, and better understand what's the correct way of handling this.

PS: just noticed that I have a carta_backend zombie process now eating all CPU.

PR #874, opened 2021-07-20
WebBrowser: Create web browser class to improve --browser setting (alternative)

jolopezl:
This is an alternative implementation for #812 with respect to #818. It exploits the use of a bash call to launch a new process redirecting the output and disowning the process.


jolopezl:
@confluence, @veggiesaurus, this is a follow-up of #818. I am still a bit puzzled about making this implementation work, so I am calling for help. I reduced all the previous code to a simple `bash`-call for launching a new process, then using `disown` to let the process run after closing the terminal. If I do in a terminal:
```
bash -c "google-chrome URL & disown"
```
the web browser starts, and after closing the terminal, the browser continues working. However, when I run this bash'ism within the backend, it doesn't work after fishing the backend (the browser closes). So I find that the only solution for this is using `fork()`, basically, in the same way as in the previous pull-request.

Do you have any suggestions to get the correct behavior without falling back to #818?

jolopezl:
Closing in favor of #818.

PR #875, opened 2021-07-23
579 fix cached image

### LINKED ISSUE #579: Re-open image probably uses cache
**System:** Linux, Firefox 79.0-1 via carta.idia.ac.za

**Issue description:** Re-opening CASA image files after re-creation shows the first image not the updated one.

**To reproduce:**
1. Opening a CASA image file in CARTA.
2. Delete the file on the disk.
3. Re-create the CARA image with different image parameters (only tested with same file size).
4. In CARTA: File --> open image will not show the updated file, but the deleted one.

**Workaround:** In CARTA: Before opening the updated image do File --> Close image

pford:
Closes #579 .

This solves the issue of the image data being rewritten but the original data is displayed when the image is already opened by CARTA.  The casacore::Table data manager does not detect that the data changed when the headers and columns are unchanged.

Uses shared_ptr for the image in the FileLoader so that when its use count is 1 (image pointer created by the loader), its underlying lattice is closed if the image file has been modified unless it is decompressed FITS.  In normal use, the image lattice will not be closed.

When functions requiring image access are complete, `FileLoader::CloseImageIfUpdated()` is called to temporarily close the image lattice if the modify time has changed.  This check is also done across all sessions from Main when a new file is opened so that outdated cached image data is dropped and new data is reloaded when needed.

### DIFF
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -66,7 +66,10 @@ FileLoader* FileLoader::GetLoader(std::shared_ptr<casacore::ImageInterface<float
 }
 
 FileLoader::FileLoader(const std::string& filename, bool is_gz)
-    : _filename(filename), _is_gz(is_gz), _num_dims(0), _has_pixel_mask(false) {}
+    : _filename(filename), _is_gz(is_gz), _modify_time(0), _num_dims(0), _has_pixel_mask(false) {
+    // Set initial modify time
+    ImageUpdated();
+}
 
 bool FileLoader::CanOpenFile(std::string& /*error*/) {
     return true;
@@ -80,13 +83,24 @@ typename FileLoader::ImageRef FileLoader::GetImage() {
     return _image;
 }
 
-void FileLoader::CloseImage() {
-    // Destroy image if only the loader owns it unless it has been decompressed
-    if (!_is_gz) {
-        if (_image.unique()) {
-            _image.reset();
-        }
+void FileLoader::CloseImageIfUpdated() {
+    // Close image if updated when only the loader owns it unless decompressed
+    if (!_is_gz && _image.unique() && ImageUpdated()) {
+        _image->tempClose();
     }
+}
+
+bool FileLoader::ImageUpdated() {
+    bool changed(false);
+    casacore::File ccfile(_filename);
+    auto updated_time = ccfile.modifyTime();
+
+    if (updated_time != _modify_time) {
+        changed = true;
+        _modify_time = updated_time;
+    }
+
+    return changed;
 }
 
 bool FileLoader::HasData(FileInfo::Data dl) const {
@@ -829,8 +843,7 @@ double FileLoader::CalculateBeamArea() {
 
     auto& info = image->imageInfo();
 
-    image.reset();
-    CloseImage();
+    CloseImageIfUpdated();
 
     if (!info.hasSingleBeam() || !_coord_sys.hasDirectionCoordinate()) {
         return NAN;
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -296,7 +296,7 @@ bool FileLoader::GetSlice(casacore::Array<float>& data, const casacore::Slicer& 
     try {
         auto image = GetImage();
 
-        if (!_image) {
+        if (!image) {
             return false;
         }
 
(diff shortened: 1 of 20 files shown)


veggiesaurus:
@pford what sort of overhead do we incur for opening the file each time we need to read? Presumably this issue is only for CASA images, right?

> The casacore::Table data manager does not detect that the data changed when the headers and columns are unchanged.

If we made a pr to casacore to check the modified date of the data?

confluence:
How does this impact large HDF5 images which use the tile cache to load full-resolution data?

acdo2002:
I found two problems with this backend branch (with the latest frontend dev branch):

1. if the first image is *.hdf5, there is no image tile rendered on the frontend canvas (Adrianna also found this PR impacts hdf5)
2. for all file types (i.e. *.fits, *.image, and *.hdf5), the moment generator will always induce the backend crash (segmentation fault (core dumped))

pford:
Modified `FileLoader::CloseImageIfUpdated` to call image `tempClose` for underlying lattice when modify time has changed.  The lattice is reopened automatically when needed.  Also called from Main across all sessions/frames/loaders with given filename when OpenFile message is received so that image is not cached in another session or frame.

Updating original pull request description.

acdo2002:
I have confirmed that previous reported problems have been solved after this commit! 
hdf5 works fine and moment generator comes back!

kswang1029:
I tested bf67a15 with the save image function. The "image" data of the old one do have been updated after a new one is loaded in CARTA. 👍 However, I noticed an issue when image matching is enabled at the frontend. The "header" of the old one persists because the frontend does not receive a header update from the backend. Therefore, the image is matched to a "wrong" place as the frontend still uses the old header to do so. I am not sure how we can fix this though... Overall this backend fix works as expected.  

The following video demonstrates the issue. The second image was the old one and the "image data" got updated after the third image is loaded. The header of the second image is still the original one so the updated image data are not matched to the right location.

https://user-images.githubusercontent.com/20819712/127271223-200a6f63-56bd-4ddc-a6f2-c6d0ba160269.mov

kswang1029:
Regarding the header issue I reported above, maybe we can consider to have a backend-triggered “close_file” to the frontend? for example image A and B are loaded in carta and I regenerate image B and “append” it. Then the backend can trigger a close_file to the frontend to close the old image B and append the new image B. This would guarantee image data and header are always self-consistent.

veggiesaurus:
> Regarding the header issue I reported above, maybe we can consider to have a backend-triggered “close_file” to the frontend? for example image A and B are loaded in carta and I regenerate image B and “append” it. Then the backend can trigger a close_file to the frontend to close the old image B and append the new image B. This would guarantee image data and header are always self-consistent.

I guess we could do something like that, but this is all to make sure results are consistent when a file is updated while open _and_ the header's wcs is also adjusted? I don't think CARTA should be expected to behave itself under those circumstances. As long as opening the file again works properly, that should be sufficient.

kswang1029:
> > Regarding the header issue I reported above, maybe we can consider to have a backend-triggered “close_file” to the frontend? for example image A and B are loaded in carta and I regenerate image B and “append” it. Then the backend can trigger a close_file to the frontend to close the old image B and append the new image B. This would guarantee image data and header are always self-consistent.
> 
> I guess we could do something like that, but this is all to make sure results are consistent when a file is updated while open _and_ the header's wcs is also adjusted? I don't think CARTA should be expected to behave itself under those circumstances. As long as opening the file again works properly, that should be sufficient.

Right, this PR does ensure the newly loaded image is CORRECT both in image data and header. What’s been left in the frontend still works if the header remains unchanged when the image is recreated. The issue I reported can be considered as an edge case.

src/ImageData/FileLoader.cc
```diff
@@ -65,28 +65,80 @@ FileLoader* FileLoader::GetLoader(std::shared_ptr<casacore::ImageInterface<float
...
+    // Close image if updated when only the loader owns it unless decompressed
+    if (!_is_gz && _image.unique() && ImageUpdated()) {
+        _image->tempClose();
+    }
+}
+
+bool FileLoader::ImageUpdated() {
+    bool changed(false);
+    casacore::File ccfile(_filename);
+    auto updated_time = ccfile.modifyTime();
```
veggiesaurus, line 96: at some stage I'd like us to move all our filesystem stuff to `std::filesystem` instead of `casacore::File`, because I think we should be sticking with STL variants of things where they exist. But that's a separate issue for another time :smile:

pford: The return value of std::filesystem::last_write_time is std::chrono::time_point.  It is non-trivial (if not impossible) to convert this to std::time_t in a portable way until we use C++20.

kswang1029 (approved):

src/ImageData/FileLoader.cc
```diff
@@ -315,43 +356,48 @@ bool FileLoader::GetSlice(casacore::Array<float>& data, const casacore::Slicer&
 
 bool FileLoader::GetSubImage(const casacore::Slicer& slicer, casacore::SubImage<float>& sub_image) {
     // Get SubImage from Slicer
-    ImageRef image = GetImage();
-    if (!image) {
+    auto image = GetImage();
+    if (!_image) {
```
veggiesaurus, line 360: shouldn't this still be `image`?

veggiesaurus: I think this should still be `image`, rather than `_image`.

src/ImageData/FitsLoader.h
```diff
@@ -31,17 +31,12 @@ class FitsLoader : public FileLoader {
...
 
-    bool HasData(FileInfo::Data ds) const override;
-    ImageRef GetImage() override;
-
 private:
-    bool _is_gz;
     std::string _unzip_file;
-    casacore::uInt _hdu;
-    std::unique_ptr<casacore::ImageInterface<float>> _image;
+    casacore::uInt _hdunum;
```
veggiesaurus: should we call this `_hdu_num` instead, for consistency?

veggiesaurus (changes requested):
minor variable-name changes required

veggiesaurus (approved):

markccchiang (approved):

Issue #840, opened 2021-06-15
Add token-based authentication to the gRPC service


confluence:
Add authentication to the gRPC service as described in the [scripting security design document](https://docs.google.com/document/d/1UVsyDd4FB8iFCdy6nsKdMoB28gpIhS55pJE0evWn1Oc/) and in the [more general wrapper issue](https://github.com/idia-astro/carta-python/issues/13).

confluence:
This was implemented in PR #881.

PR #881, opened 2021-07-29
Add security token to gRPC interface

confluence:
This implements #840. However, we do not currently use an encrypted gRPC channel between the wrapper and the backend, so the token is sent unencrypted. We would have to implement SSL at both ends in order to encrypt the connection. Is the current solution sufficient for UDM? When we add SDM support we could handle this in the controller, and have the option of using a secure connection in the wrapper.

This has a [companion wrapper PR](https://github.com/idia-astro/carta-python/pull/18).

### DIFF
--- a/src/GrpcServer/CartaGrpcService.cc
+++ b/src/GrpcServer/CartaGrpcService.cc
@@ -21,9 +21,7 @@
 
 uint32_t CartaGrpcService::_scripting_request_id = 0;
 
-CartaGrpcService::CartaGrpcService() : _auth_token(""), _debug_no_auth(true) {}
-
-CartaGrpcService::CartaGrpcService(std::string auth_token) : _auth_token(auth_token), _debug_no_auth(false) {}
+CartaGrpcService::CartaGrpcService(std::string auth_token) : _auth_token(auth_token) {}
 
 void CartaGrpcService::AddSession(Session* session) {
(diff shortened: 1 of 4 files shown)


src/GrpcServer/CartaGrpcService.cc
```diff
@@ -21,7 +21,9 @@
 
 uint32_t CartaGrpcService::_scripting_request_id = 0;
 
-CartaGrpcService::CartaGrpcService(std::string auth_token) : _auth_token(auth_token) {}
+CartaGrpcService::CartaGrpcService() : _auth_token(""), _debug_no_auth(true) {}
+
+CartaGrpcService::CartaGrpcService(std::string auth_token) : _auth_token(auth_token), _debug_no_auth(false) {}
```
veggiesaurus: a neater way to do this might be to just pass in an empty auth token when `debug_no_auth` is `true`. That would be consistent with `SimpleFrontendServer` as well

veggiesaurus (changes requested):

src/GrpcServer/CartaGrpcService.cc
confluence: I thought that an explicit solution would be better, but I like consistency -- I will change it.

veggiesaurus (approved):

PR #883, opened 2021-07-30
ajm/jenkins macos unit test adjustment

ajm-ska:
Let's try this for now.
The HDF5 related unit tests are now excluded from MacOS.
For other improvements, I have limited the number of cores that 'make' can use for MacOS because when it is unlimited, the system keeps running out of application memory for some reason and the Terminal etc. becomes paused.
I have also temporarily removed 3 ICD tests that don't seem to work 100% of the time.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -144,7 +144,7 @@ pipeline {
                         catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE')
                         {
                         dir ('build/test') {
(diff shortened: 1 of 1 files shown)


veggiesaurus:
@ajm-asiaa will you make an issue to put the misbehaving ICD tests back / resolve why they don't work?

ajm-ska:
@veggiesaurus Yes. @acdo2002 and myself discussed it earlier. We will have it figured out in time to re-implement them in the brand new Jenkins Docker system.

Issue #854, opened 2021-06-25
Add code coverage metrics to CI


veggiesaurus:
We have a bunch of unit tests for the backend, and @markccchiang is adding new ones for existing features as part of the 3.0-beta.1 release (#797). 

We should add some basic code coverage statistics to the CI. `gcovr` can be used to generate a report from `gcov`, or we can use some sort of online platform for this.

veggiesaurus:
We're currently sitting at just under 17% code coverage, by the way :nauseated_face:

ajm-ska:
@veggiesaurus This was initially not automatically working for all branches when I was using a Docker Pipeline in Jenkins. But I have since managed to update the RHEL7 server that is running the Jenkins Mix 'N Match. Now Jenkins can natively build the carta-backend with -Dtest=on and run the coverage script.

The page looks a bit primitive, but it seems to be working: https://carta.asiaa.sinica.edu.tw/coverage/

veggiesaurus:
ok, it's getting there. Can you make the commit hashes a link to the actual commit on github? And the branch names a link to the branch on github.

I noticed some odd behaviour: 

![image](https://user-images.githubusercontent.com/592504/132317003-1e5d5f82-e4ff-4dd1-8bc7-53f62d1420b6.png)

Same commit hash, very different results

ajm-ska:
@veggiesaurus OK. I will try to add appropriate hyperlinks. 

Well spotted about the the 14.5% and 18.5%. I see the latest two `78c4ff2` entries exhibit the same behaviour. They are same commit, but the job was just re-run.
![Screen Shot 2021-09-07 at 6 01 51 PM](https://user-images.githubusercontent.com/20802551/132326582-c42610da-c8c2-4b9b-8a2c-bb49c9953c48.png)
If you click through to the details, they are both 14.4%, so I guess the problem is my attempt to use grep and awk to grab the percentage value from the html:

> percentage="$(grep -m 1 '<td class="coverage-low"' coverage/index.html | cut -d'>' -f2 | awk '{ print substr( $0, 1, length($0)-4 ) }')"

I will try to improve it.

ajm-ska:
@veggiesaurus The requested changes have been made and it seems to be working: https://carta.asiaa.sinica.edu.tw/coverage/

Now, clicking on the Commit ID takes you to the commit on GitHub, clicking on Branch Name takes you to the branch on GitHub, and clicking on the percentage value takes you to the full code coverage report.  

[...]

PR #885, opened 2021-07-30
script for running gcovr to generate coverage report in html

veggiesaurus:
Just a minor PR to add script for generating a coverage report using `gcovr` _after_ the unit tests have been run with coverage enabled.

Likely to be useful for #854


scripts/ci_mac_casacore.sh
```diff
@@ -1,39 +0,0 @@
-#!/bin/bash
```
veggiesaurus, line 1: @ajm-asiaa has confirmed we don't need this anymore

confluence (approved):

PR #890, opened 2021-08-04
only export SERVER_IP env variable if its not empty

veggiesaurus:
To fix an issue with running `carta` on machines without network access, as raised in the helpdesk:

> When I'm disconnected from the internet, "hostname -I" returns
nothing.
>
>If I'm connected to the internet (and Carta works), I get e.g.
> hostname -I
192.168.1.107
>
>(this is my laptop at home w/ wireless router)
>
>The OS is Ubuntu 18.04.5 LTS.
I tried: "carta --host=localhost", and that works whether my internet
connection is on or not. Shouldn't the "localhost" option just be the
default in that case?
>
>michael

With the current `carta` script, we get the following auto-generated URL when no IP is detected:
`CARTA is accessible at http://:3002/?token=0d927379-3513-41e1-a2da-7854b723bd4b`

With the modified script:
`CARTA is accessible at http://localhost:3002/?token=3587a442-b9ff-4cd3-babe-f9db2315dc35`

### DIFF
--- a/scripts/carta
+++ b/scripts/carta
@@ -1,13 +1,14 @@
 #!/bin/bash
 
-if [ "$(uname)" == "Darwin" ]
-then
-    SERVER_IP=$(ipconfig getifaddr en0)
-elif [ "$(uname)" == "Linux" ]
-then
-    SERVER_IP=$(hostname -I | cut -d' ' -f1)
+if [ "$(uname)" == "Darwin" ]; then
+    FIRST_IP=$(ipconfig getifaddr en0)
+elif [ "$(uname)" == "Linux" ]; then
+    FIRST_IP=$(hostname -I | cut -d' ' -f1)
 fi
 
-export SERVER_IP
+# Only export env variable if it's not empty
+if [ ! -z $FIRST_IP ]; then
+    export SERVER_IP=$FIRST_IP
+fi
 
 carta_backend "$@"


ajm-ska (approved):
It appears this was never an issue on MacOS. If I run the current 'dev' branch on MacOS without a network connection active, it still adds 'localhost' to the URL. 
However, I confirm that with this fix, it continues to work on MacOS.

I am actually unable to test on Linux at present as I don't have a local Linux machine on hand.

I did try with Docker on MacOS, but can't reproduce it there either. If I turn off the Mac's network connection and start the carta_backend in a Docker container, it still produces 'localhost'. I guess that is because Docker hosts its own private network, at least on MacOS.

ajm-ska:
@veggiesaurus I know it shows Jenkins is still running here. There was a problem with Jenkins earlier where it got stuck. Some current branches still do not use the updated Jenkinsfile where we have deactivated the HDF5 unit tests on MacOS. There appears to be some kind of memory leak, and the Mac runs out of RAM for some reason. I think it is because of this HDF5 problem, but I haven't had time to investigate yet. 

I manually re-ran this commit and if you click on "Details" you can see it did actually pass all the tests although it didn't report back to Github.

veggiesaurus:
> It appears this was never an issue on MacOS. If I run the current 'dev' branch on MacOS without a network connection active, it still adds 'localhost' to the URL. 
> However, I confirm that with this fix, it continues to work on MacOS.
> 
> I am actually unable to test on Linux at present as I don't have a local Linux machine on hand.
> 
> I did try with Docker on MacOS, but can't reproduce it there either. If I turn off the Mac's network connection and start the carta_backend in a Docker container, it still produces 'localhost'. I guess that is because Docker hosts its own private network, at least on MacOS.

You have to disable that from outside of docker once you've got the backend installed. At least that's how I tested it.

confluence (approved):

PR #898, opened 2021-08-10
891 Fix control points record for polygon/line/polyline

### LINKED ISSUE #891: backend crash when exporting region file including polyline
@YuHsuan-Hwang found this while implementing the frontend support of polyline (https://github.com/CARTAvis/carta-frontend/pull/1592). After some investigations, it appears that exporting a polyline with specific numbers of points would cause the backend crash. They are 4, 8, 12 and so on.

pford:
Closes #891 

Probable merge error.  Polygon region requires adding first point as last point to close the region when creating the record for region export.  This point was being added incorrectly for polygon and line-type regions, but vector was size npoints instead of npoints+1.  Fixed so that vector is resized and point added for polygon only.

@veggiesaurus and @kswang1029 after this is tested, should this be added to v3b1?

### DIFF
--- a/src/Region/Region.cc
+++ b/src/Region/Region.cc
@@ -940,10 +940,6 @@ casacore::TableRecord Region::GetControlPointsRecord(const casacore::IPosition& 
                 y(i) = _region_state.control_points[i].y();
             }
 
-            // LCPolygon::toRecord includes last point as first point to close region
-            x(npoints) = _region_state.control_points[0].x();
-            y(npoints) = _region_state.control_points[0].y();
-
             if (_region_state.type == CARTA::RegionType::POLYGON) {
                 // LCPolygon::toRecord includes first point as last point to close region
                 x.resize(npoints + 1, true);


veggiesaurus (approved):

kswang1029 (approved):
yes the fix works. 

@veggiesaurus weirdly, the color code defined in the region file is `2ee6d6` (cyan) but one of the regions is rendered in magenta. It appears that the "complex polygon" coloring logic is applied to polyline. LOL. 

Screen Shot 2021-08-11 at 3 35 52 PM

veggiesaurus:
> yes the fix works.
> 
> @veggiesaurus weirdly, the color code defined in the region file is `2ee6d6` (cyan) but one of the regions is rendered in magenta.

Is it possible that the polyline intersects itself?

kswang1029:
> > yes the fix works.
> > @veggiesaurus weirdly, the color code defined in the region file is `2ee6d6` (cyan) but one of the regions is rendered in magenta.
> 
> Is it possible that the polyline intersects itself?

It appears that the "complex polygon" coloring logic is applied to polyline.

YuHsuan-Hwang (approved):

PR #902, opened 2021-08-11
Don't calculate the cube histogram if it's available in the HDF5 file.

### LINKED ISSUE #899: cube histogram of hdf5 is not derived from the pre-calculated one
When requesting a cube histogram from hdf5 image, the backend should just read the pre-calculated one instead of performing real calculations from the image cube.

confluence:
Fixes #899.

This reorders the conditions so that the cube histogram isn't always calculated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


acdo2002 (approved):
I check with performance test. The per cube histogram for hdf5 now is back to 3ms. 
The last performance tests measured was on `a8831fc` with 196594ms.

kswang1029 (approved):
Confirmed with mix n match. HDF5 cube histogram works as expected. Other image formats work as expected too.

markccchiang (approved):

markccchiang:
Thank you for fixing it!

PR #903, opened 2021-08-11
Allow sending the cursor value if cursor spatial config is empty

### LINKED ISSUE #900: cursor info widget does not show updated values for all matched images
To reproduce:
1) load two matchable images and match them
2) launch the cursor info widget
3) move cursor around and observe the reported pixel values. We should see pixel values from the two images are updating but there is only one updating.

markccchiang:
Fixes #900. Allow sending the cursor value if cursor spatial config is empty. Also fixes the cursor value for different stokes if it is required in the spatial config.


confluence (approved):

kswang1029 (approved):
yes this fixes the cursor info widget.

YuHsuan-Hwang (approved):

acdo2002:
The ICD tests: MATCH*  now works with this branch. (they are failed with the current dev branch)

PR #905, opened 2021-08-12
Remove a duplicate histogram calculation

markccchiang:
It is my fault that introduced this bug in [PR 790](https://github.com/CARTAvis/carta-backend/pull/790/files). The image histogram configs cache should be cleared when resetting the histogram requirements for a Frame.


confluence (approved):

confluence:
What is the impact of this -- do we just send an extra histogram? Does this need to be fixed in the current beta release, or can it wait until the next one?

Please also add this to the changelog.

markccchiang:
> What is the impact of this -- do we just send an extra histogram? Does this need to be fixed in the current beta release, or can it wait until the next one?
> 
> Please also add this to the changelog.

It just sends an extra histogram. Since this extra histogram calculation has been cached, it spends at most a few milli-secondes to do so. I think it can wait until the next one.

confluence:
Long-term suggestion: we wouldn't have this problem if these containers were sets rather than vectors -- but then we'd have to make sure the objects we put in them are hashable (possibly by specifying an appropriate hash function).

confluence (changes requested):
This new changelog entry needs to go in a new `[Unreleased]` section -- we have already tagged the last release.

confluence (approved):

PR #907, opened 2021-08-13
Fixes region histograms for moment images

### LINKED ISSUE #906: region histogram broken with the in-memory moment images
After generating a moment image in memory, if we request a region histogram with the histogram widget, we will see no histogram showing up.

markccchiang:
Fix #906 by using a different way to get region data for an image in memory.


kswang1029 (approved):
the fix works 👍

confluence (approved):
There will probably be a merge conflict in the changelog -- we should leave enough time between merges of these bugfix PRs so that it can be fixed in whichever one is merged second.

Issue #895, opened 2021-08-09
Support other polarization types


YuHsuan-Hwang:
Support all the stokes value of fits files:
![Screen Shot 2021-08-04 at 2 53 46 PM](https://user-images.githubusercontent.com/43841102/128664733-7644f102-d156-4cbc-928a-6a0ef13a8686.png)
This includes:

- [x] Extend protobuf message `StokesType`
- [x] Extend stokes type of Z profile, XY profile, histogram data, and statistics data
- [x] Extend types of hypercube

requiring frontend: https://github.com/CARTAvis/carta-frontend/issues/1606

veggiesaurus:
@kswang1029 can we close this now?

kswang1029:
yes

PR #910, opened 2021-08-24
Yuhsuan/895 polarization types

YuHsuan-Hwang:
This PR is for the first two items of #895:
* extend protobuf message `StokesType`
* extend stokes type of Z profile, XY profile, histogram data, and statistics data

The last item will be implemented in a separate PR.

Requiring protobuf branch: `yuhsuan/polarization_types`
Requiring frontend branch: `yuhsuan/1606_polarization_types`


src/Util.h
```diff
@@ -65,9 +65,22 @@ casacore::String GetResolvedFilename(const std::string& root_dir, const std::str
...
+    {7, CARTA::StokesType::RL}, {8, CARTA::StokesType::LR}, {9, CARTA::StokesType::XX}, {10, CARTA::StokesType::YY},
+    {11, CARTA::StokesType::XY}, {12, CARTA::StokesType::YX}};
 int GetStokesValue(const CARTA::StokesType& stokes_type);
 CARTA::StokesType GetStokesType(int stokes_value);
 
+static std::unordered_map<string, CARTA::StokesType> StokesStringTypes{{"I", CARTA::StokesType::I}, {"Q", CARTA::StokesType::Q},
+    {"U", CARTA::StokesType::U}, {"V", CARTA::StokesType::V}, {"RR", CARTA::StokesType::RR}, {"LL", CARTA::StokesType::LL},
+    {"RL", CARTA::StokesType::RL}, {"LR", CARTA::StokesType::LR}, {"XX", CARTA::StokesType::XX}, {"YY", CARTA::StokesType::YY},
+    {"XY", CARTA::StokesType::XY}, {"YX", CARTA::StokesType::YX}};
+
```
confluence: It would be nice to eliminate some of this repetition by defining one map mapping the constants to an index + string pair and then constructing the three maps actually used for lookups at runtime. But I'm not sure if there's a clean way to do this while keeping these as static variables, so this may be the clearest solution for now.

confluence (approved):

markccchiang (approved):

src/Util.h
markccchiang: Maybe something likes [Bimap](https://theboostcpplibraries.com/boost.bimap) from the Boost. But we should not use the Boost library actually.

acdo2002 (approved):
So far the performance tests are all passed without any problem

ajm-ska (approved):
I'm not sure what image or images that I should use to directly test this. However, in general the frontend and backend still appears to work fine.

YuHsuan-Hwang:
Additional changes: renamed protobuf messages based on branch `yuhsuan/polarization_types`
Some ICD tests failed and need further checks

ajm-ska:
> Some ICD tests failed and need further checks

@YuHsuan-Hwang A reboot of the Ubuntu server appears to have fixed that issue. Now all ICD tests have passed.

acdo2002 (approved):

kswang1029 (approved):

PR #915, opened 2021-09-06
Angus/multiple file args

### LINKED ISSUE #84: Feature request: open multiple files via the command line
It would be useful to be able to open mulitple files via the command line, ideally with wildcard support. The current behaviour is:

```
$ carta 1*/*MTP*pcalmask-MFS-image.fits
Error: Requested file or directory does not exist, or image is a non standard type. Please check.
```

and when many files are spread across multiple paths loading them all individually via the GUI is tedious. 

Thanks.

veggiesaurus:
closes https://github.com/CARTAvis/carta/issues/84.

The backend already had support for handling multiple file arguemtns. However, it would always only send the first one, because (at the time of implementation) the frontend didn't support the `files=` URL query parameter.

This PR fixes that. If multiple file arguments are passed (e.g. `carta *.fits` or `carta m51/xray.fits m42.fits other/m81.fits`), the backend will check whether all files are in the same folder. If they are, it will generate a URL in the form`&folder=<common_folder>&files=<filename1>,<filename2>,...`. 

If there is no common folder, the URL will be in the form `&files=<filepath1>,<filepath2>,...`. This is more verbose, but more flexible.


jolopezl (commented):
I wonder if it is possible to encapsulate these lines inside a new method in ProgramSettings to keep `Main.cc` _cleaner_. Since `files` is in the program settings struct, `curl.h` can be extracted from the main, too, making the main program smaller. 

**Maybe a better alternative:** another possible place could be the `SimpleFrontendServer` that could return the `frontend_url` via `GetFrontendURL(program_settings.files)` with all the `query_url` processing encapsulated there. This looks more reasonable than making the change in program settings since the calculating is the frontend url.

confluence:
I'm in the process of refactoring `Main.cc`. I hope to create a draft PR soon -- we'll be able to discuss more general refactoring issues there.

confluence:
@angus I haven't run this yet, but have you tested this with a mixture of file and directory parameters? What happens if the user passes in two directories, or a directory and a file?

jolopezl:
> I'm in the process of refactoring `Main.cc`. I hope to create a draft PR soon -- we'll be able to discuss more general refactoring issues there.

That's great; we can do that. I just think that with this pull request extracting https://github.com/CARTAvis/carta-backend/blob/78c4ff230ffad7011818e4ca9737f9da091b7982/src/Main.cc#L711-L756 into a method into `SimpleFrontendServer` is a free lunch.

veggiesaurus:
@jolopezl I agree. The lack of unit tests in my code are also making me nervous :laughing: So I think I'll do that minor refactoring and add some unit tests. 

Moving PR back to draft until then

veggiesaurus:
moved the URL generation to a static function in `SimpleFrontendServer` and added unit tests.

PR ready for review

confluence:
@veggiesaurus what happens if some or all of the parameters are directories?

veggiesaurus:
> @veggiesaurus what happens if some or all of the parameters are directories?

that behaviour is unchanged. If any of the positional arguments are folders (other than casacore image folders), then the folder gets set as the starting folder, and no file arguments are processed

confluence (approved):

jolopezl (approved):
Thanks for addressing the refactoring suggestion. Adding some testing is another plus from going in that direction.

PR #917, opened 2021-09-09
Fix the file info inconsistency for FITS

### LINKED ISSUE #916: file info inconsistency
The generated file info from FITS and CASA images has inconsistent entries, likely due to truncation error:

Screen Shot 2021-09-08 at 10 03 35 AM
Screen Shot 2021-09-08 at 10 03 22 AM

markccchiang:
Closes #916. Fix the file info inconsistency for FITS.

### DIFF
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -813,6 +813,7 @@ void FileExtInfoLoader::AddComputedEntriesFromHeaders(CARTA::FileInfoExtended& e
     casacore::String ctype1, ctype2, cunit1("deg"), cunit2("deg"), frame, radesys, specsys, bunit;
     double min_float(std::numeric_limits<float>::min());
     double crval1(min_float), crval2(min_float), crpix1(min_float), crpix2(min_float), cdelt1(min_float), cdelt2(min_float);
+    int velref(-1);
 
     // Quit looking for key when have needed values
     bool need_ctype(true), need_crpix(true), need_crval(true), need_cdelt(true), need_frame(true), need_radesys(true);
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
Yes, this fixes the issue as far as I can tell from my test images. 👍

veggiesaurus:
I presume we now need to fix the file info ICD test?

veggiesaurus (approved):

confluence (approved):

kswang1029:
> I presume we now need to fix the file info ICD test?

@acdo2002 could you check?

acdo2002:
Yes, I am going to prepare a ICD update branch. 
Only update the snapshot for fits, also the hdf5 header too. 
but all these three type images will show the same Image reference coords with this backend branch.

kswang1029:
> Yes, I am going to prepare a ICD update branch.
> Only update the snapshot for fits, also the hdf5 header too.
> but all these three type images will show the same Image reference coords with this backend branch.

Such inconsistency _should not_ ever exist in the ICD test actually. We should not apply an incorrect answer just to make the test pass. When did the ICD test get updated with this inconsistency? In my old robot test, reference world coordinates are identical and that’s why I spotted this problem.

kswang1029:
> > Yes, I am going to prepare a ICD update branch.
> > Only update the snapshot for fits, also the hdf5 header too.
> > but all these three type images will show the same Image reference coords with this backend branch.
> 
> Such inconsistency _should not_ ever exist in the ICD test actually. We should not apply an incorrect answer just to make the test pass. When did the ICD test get updated with this inconsistency? In my old robot test, reference world coordinates are identical and that’s why I spotted this problem.

Do we ever test the computed entries? ie the context in the file info tab?

acdo2002:
the last time I update is in Pam's branch.
Sorry I didn't notice the snapshot changes the Image reference coords.
I will be more careful about using snapshot.

Issue #909, opened 2021-08-24
crash when opening a region file on CentOS8


ajm-ska:
Our ICD tests have revealed a crash that only occurs on CentOS8.

When a carta_backend running on CentOS8 tries to open our region test file `M17_SWex_regionSet1_world.crtf` it results in an immediate crash of the backend:

```
 2021-08-24 01:57:14	INFO	RegionTextParser::_determineVersion	RegionTextParser::_determineVersion: Found spec version 0
/usr/include/c++/8/bits/stl_vector.h:932: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = int; _Alloc = std::allocator<int>; std::vector<_Tp, _Alloc>::reference = int&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Aborted (core dumped)
```

It is strange because the same region file can be opened without problems on CentOS7, Ubuntu 18.04, and Ubuntu 20.04.

At first I thought it was the GCC version used to build the backend. CentOS8 uses gcc 8.4.1.
(CentOS7 uses gcc 8.3.1, Ubuntu 18.04 uses gcc 8.4.0, and Ubuntu 20.04 uses gcc 9.3.0)
I tried installing devtoolset-9 and devtoolset-10 on CentOS8. They use gcc 9.2.1 and 10.2.1, respectively, but the problem remains.

Do you have any idea what could be wrong? I guess something small needs to be adjusted in `CrtfImportExport.cc` to make it more robust?

kswang1029:
> #symbol [[275.13653085deg, -16.17909524deg], .] coord=ICRS, corr=[I], linewidth=1, linestyle=-, symsize=1, symthick=1, color=green, font=Helvetica, fontsize=10, fontstyle=bold, usetex=false
> #centerbox [[275.13683271deg, -16.18844433deg], [30.0324arcsec, 30.0324arcsec]] coord=ICRS, corr=[I], linewidth=1, linestyle=-, symsize=1, symthick=1, color=green, font=Helvetica, fontsize=10, fontstyle=bold, usetex=false
> centerbox [[275.13683271deg, -16.18844433deg], [30.0324arcsec, 30.0324arcsec]]

the first one “symbol” (aka point region) works fine.
[...]

PR #888, opened 2021-08-03
797 ICD tests in the backend unit tests

markccchiang:
This PR is the trial for #797 to move some ICD tests back to the backend unit tests. I will add more testing cases according to the [document](https://docs.google.com/spreadsheets/d/1qwdTTKdNA1oNnNAUNgyPZH5HOVyt2Uw0vPT2_UsEaXk/edit#gid=0)).

PR #919, opened 2021-09-13
matrix style Jenkinsfile

### LINKED ISSUE #865: Get CI running on supported platforms
We currently have our CI running on Ubuntu 16.04, CentOS (7?) and MacOS (10.15?)

We need to run our CI on the following officially supported platforms:
- Ubuntu 18.04 and 20.04
- RHEL 7/8 or CentOS 7/8 (We could also use AlmaLinux or RockyLinux for CentOS 8 support)
- MacOS 10.15 and 11

We can use Docker for most of these (I think we can even use Docker for RHEL if we have an account).

ajm-ska:
@veggiesaurus I think it is about time to switch to the new Jenkinsfile. This will free up our existing Ubuntu server so I'll finally be able to upgrade it and use it for frontend ICD testing in parallel. 

Here is some information:

- All the four supported Linux systems are now run through Docker containers on a single AMD server. Test images are read from an internal SSD.

- Due to CentOS8 reaching EOL soon and the fact that things might look complicated if I continue to use CentOS7 with either AlmaLinux or Rocky Linux, I decided to switch to actual RHEL. So now we are using real RHEL7 and RHEL8 Docker containers created from the RedHat Universal Base Images (UBI).

- I know we support two macOS versions, but we only have one Intel MacMini with Big Sur available at present. The backend build, unit tests, and ICD tests run natively on it. We plan to purchase a 2nd Intel MacMini shortly, perhaps for when macOS Monterey comes out as it will be easier to upgrade Monterey rather than downgrade Big Sur to Catalina. At that time we will cover the latest two macOS versions; 11 and 12.

- There is an annoying limit to the size of a declarative pipeline in a Jenkinsfile. Even after moving as many lines as possible into definition blocks, the file is still too large to be parsed when covering the 5 systems. Therefore, I have had to use 'matrix' style of configuration. Unfortunately, it is not visualised very well by Blue Ocean. I have tried to make it as clear as possible. The first 3 stages (Build, Unit Tests, and prepare ICD tests) are kept in the old way. The ICD tests themselves appear in a single stage. Throughout that stage, it is not clear which row represents which OS. They don't line up with those of the first column. But you can find out by inspecting the stage or remembering that they are in alphabetical order (bionic, focal, macos11, rhel7, rhel8). I tried to put the first 3 stages in the same order, but I have their names more explicit e.g. Ubuntu 20.04. I needed to use shorter single word names for the matrix variable.

- As before, if an ICD test fails, it waits 30 seconds and runs the test again, using the same backend process. Also, when a test fails, it runs `cat /root/.carta/log/carta.log`. Hopefully both may be helpful in some cases, but hasn't been useful so far.

- There is a very strange issue where a few ICD tests fail on the Ubuntu 20.04 Docker container when run through Jenkins. They do not fail when I run them manually in the same container. Plus they work fine in the three other Docker containers when Jenkins runs them. I decided to temporarily remove them but I have made an issue in the ICD test repository [#277](https://github.com/CARTAvis/carta-backend-ICD-test/issues/277)

- We have discovered a carta-backend bug occurring only on RHEL8/CentOS8 regarding region files. It causes the backend to crash. We have filed a backend issue [#909](https://github.com/CARTAvis/carta-backend/issues/909).  I thought I would leave the failing test in for now. Hopefully it could be fixed soon.

- I decided to add macOS Unit Tests back. Using the older hdf5@1.10 actually appears to prevent the previous HDF5 problems.

- The list of ICD tests are still hard coded into the Jenkinsfile, so it is not so easy to change tests or add more as it would then need to be merged into every existing branch.

- In the case of @markccchiang's branch [PR#888  ](https://github.com/CARTAvis/carta-backend/pull/888), and future additions, I guess we could modify the Jenkinsfile in the same pull request to remove the duplicated tests from the Jenkinsfile. e.g. Session and Animator.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ pipeline {
         preserveStashes() 
     }
     stages {
-        stage ('Build') {
+        stage('Build') {
             parallel {
                 stage('Ubuntu 18.04') {
                     agent {
@@ -167,7 +167,7 @@ pipeline {
                         }
                     }
                 }
-                stage('Ubuntu 20.04') {          
+                stage('Ubuntu 20.04') {
                     agent {
                         label "focal-docker-agent"
                     }
@@ -301,7 +301,7 @@ pipeline {
                         steps {
                             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
                                 unstash "${PLATFORM}-backend"
-                                sh "./carta_backend /images --top_level_folder /images --port 3002 --omp_threads 8 --debug_no_auth true --no_http true &"
+                                sh "./carta_backend /images --top_level_folder /images --port 3112 --omp_threads 8 --debug_no_auth true --no_http true --verbosity=5 &"
                                 unstash "${PLATFORM}-ICD"
                                 session()
                             }
@@ -314,7 +314,7 @@ pipeline {
                         steps {
                             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
                                 unstash "${PLATFORM}-backend"
-                                sh "./carta_backend /images --top_level_folder /images --port 3002 --omp_threads 8 --debug_no_auth true --no_http true &"
+                                sh "./carta_backend /images --top_level_folder /images --port 3112 --omp_threads 8 --debug_no_auth true --no_http true --verbosity=5 &"
                                 unstash "${PLATFORM}-ICD"
                                 file_browser()
                             }
@@ -327,7 +327,7 @@ pipeline {
                         steps {
                             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
                                 unstash "${PLATFORM}-backend"
-                                sh "./carta_backend /images --top_level_folder /images --port 3002 --omp_threads 8 --debug_no_auth true --no_http true &"
+                                sh "./carta_backend /images --top_level_folder /images --port 3112 --omp_threads 8 --debug_no_auth true --no_http true --verbosity=5 &"
                                 unstash "${PLATFORM}-ICD"
                                 animator()
                             }
@@ -340,7 +340,7 @@ pipeline {
                         steps {
                             catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
                                 unstash "${PLATFORM}-backend"
-                                sh "./carta_backend /images --top_level_folder /images --port 3002 --omp_threads 8 --debug_no_auth true --no_http true &"
+                                sh "./carta_backend /images --top_level_folder /images --port 3112 --omp_threads 8 --debug_no_auth true --no_http true --verbosity=5 &"
                                 unstash "${PLATFORM}-ICD"
                                 contour()
                             }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -9,1206 +9,945 @@ void setBuildStatus(String message, String state) {
 }
 pipeline {
     agent none
+    options {
+        preserveStashes() 
+    }
     stages {
-        stage('Build') {
+        stage ('Build') {
```
veggiesaurus: formatting inconsistency.

Jenkinsfile
```diff
@@ -9,1206 +9,945 @@ void setBuildStatus(String message, String state) {
...
                         }
+                    }
+                    post {
+                        always {
+                            junit 'test/ubuntu_bionic_test_detail.xml'
                         }
                     }
                 }
-                stage('Ubuntu') {
+                stage('Ubuntu 20.04') {
```
veggiesaurus: whitespace

Jenkinsfile
```diff
@@ -9,1206 +9,945 @@ void setBuildStatus(String message, String state) {
...
-                            dir ('carta-backend-ICD-test') {
-                                match_tests()
-                            }
-                        }
-                        }
-                    }
-                }
+    }
+}
+def prepare_focal_ICD() {
```
veggiesaurus, line 467: newline?

veggiesaurus (changes requested):
I'm out of my depth WRT jenkins itself, but I spotted a couple of minor formatting issues

kswang1029 (commented):
we are using `--omp_threads 8` for all OSs. This means in parallel, we will use 32 (8x4) cores in one single AMD server. Is this correct? For macOS, we declare to use 8 cores too?

ajm-ska:
@veggiesaurus Thanks. I tried to address the formatting issues.

On MacOS, I noticed a problem where it tries to run `cat /root/.carta/log/carta.log` but fails because there is no log file there. Now I realise MacOS would write to `/Users/<username>/.carta/log`. As I'm using the matrix style configuration, it is not easy to have different parameters for different systems. However, I went through a similar situation for the `/images` directory. On MacOS, we can't make folders in the root directory. But I found a handy way to create symlinks to the root directory on [Stackoverflow](https://stackoverflow.com/questions/58396821/what-is-the-proper-way-to-create-a-root-sym-link-in-catalina) using a `/etc/synthetic.conf` file. I have added /root symlink to it now, so our file looks like this (In that file make sure to use a tab instead of spaces) and the problem is fixed.

```
images  Users/acdc/almalustre-images
root    Users/acdc
```

I have also decided to remove the MacOS unit tests again because from these few attempts, they still tend to fail more often than not.

@kswang1029 Yes. For each Linux Docker container, I let them build the backend using 16 cores, and run the backends with 8 cores. So yes, during the ICD tests, it should be using 32 cores in total. That is fine as it is a 2x24 core CPU server, plus there should be no other users (For MacOS, the resources are more limited, so I only let it build with 8 cores and run with 8 cores).

If you consider RAM usage, that is not an issue either. The server has 128GB. I have monitored it using `docker stats`. During the backend build process, no single Docker container has used more than 12% of the RAM at any one time. During the Unit Tests and ICD tests, I haven't observed it going any higher than 1.6% memory usage per container.

ajm-ska:
Another disadvantage of the matrix configuration in the Jenkinsfile is that now all the ICD tests appear grouped into one stage on Github. For example:

Screen Shot 2021-09-13 at 9 17 15 PM

So you need to go and click "Details" to see which ones failed. 

Previously each stage of ICD test appeared on separate rows:

Screen Shot 2021-09-13 at 9 19 51 PM

But I can't figure out any other way of doing it.

veggiesaurus:
@ajm-asiaa isn't that what env variables are for? (I mean the macos versus Linux paths)

ajm-ska:
@veggiesaurus Good point. I guess that could have been an alternative. But both are solutions to the same problem. You can see that the `/etc/synthetic.conf` method is working perfectly fine.

veggiesaurus (approved):

PR #920, opened 2021-09-16
Refactor messages handlers on the Main and OnMessageTask

markccchiang:
This is the part of code changes on #888. I refactored messages handlers from the Main and OnMessageTask. The incoming protobuf messages are decoded only on the Main. And a new `GeneralMessageTask` from the OnMessageTask handles messages based on the protobuf message type. The `MultiMessageTask` is removed. It allows the backend ICD testers to create a testing message without encoding/decoding steps.

### DIFF
--- a/src/Main.cc
+++ b/src/Main.cc
@@ -172,6 +172,7 @@ void OnMessage(uWS::WebSocket<false, true, PerSocketData>* ws, std::string_view 
             LogReceivedEventType(event_type);
 
             auto event_type_name = CARTA::EventType_Name(CARTA::EventType(event_type));
+            bool message_parsed(false);
             OnMessageTask* tsk = nullptr;
 
             switch (event_type) {
@@ -179,8 +180,7 @@ void OnMessage(uWS::WebSocket<false, true, PerSocketData>* ws, std::string_view 
                     CARTA::RegisterViewer message;
                     if (message.ParseFromArray(event_buf, event_length)) {
                         session->OnRegisterViewer(message, head.icd_version, head.request_id);
-                    } else {
-                        spdlog::warn("Bad {} message!", event_type_name);
+                        message_parsed = true;
                     }
                     break;
                 }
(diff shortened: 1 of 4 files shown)


src/Main.cc
```diff
@@ -167,18 +167,20 @@ void OnMessage(uWS::WebSocket<false, true, PerSocketData>* ws, std::string_view
...
+            OnMessageTask* tsk = nullptr;
+
+            switch (event_type) {
                 case CARTA::EventType::REGISTER_VIEWER: {
                     CARTA::RegisterViewer message;
                     if (message.ParseFromArray(event_buf, event_length)) {
                         session->OnRegisterViewer(message, head.icd_version, head.request_id);
                     } else {
-                        spdlog::warn("Bad REGISTER_VIEWER message!");
+                        spdlog::warn("Bad {} message!", event_type_name);
```
veggiesaurus: Since we're refactoring things anyway, I'd prefer to get rid of all this repetition. What about having a `bool message_parsed = false` that gets set to `true` in each `if (message.ParseFromArray(...)) {` branch?

And then at the end we can just have a single `spdlog::warn`. That way, if we want to change how it is formatted, we only need to change one line

src/OnMessageTask.h
```diff
@@ -27,40 +27,61 @@ class OnMessageTask : public tbb::task {
...
         _session = nullptr;
     }
 };
 
-class MultiMessageTask : public OnMessageTask {
-    carta::EventHeader _header;
-    int _event_length;
-    char* _event_buffer;
-    tbb::task* execute() override;
+template <typename T>
```
veggiesaurus: please move this and the rest of the function implementations into a `.tcc` file for consistency with the rest of the codebase

veggiesaurus (changes requested):

veggiesaurus (approved):

confluence (approved):

PR #921, opened 2021-09-16
Create a protobuf messages wrapper

markccchiang:
This is the part of code changes on #888. I created `src/Util/Message.h/cc` for protobuf messages helpers for creating or filling messages. There are some creating requested messages functions on the `Message` class which will be used in ICD tests.

### DIFF
--- a/src/Util/Message.cc
+++ b/src/Util/Message.cc
@@ -278,22 +278,6 @@ CARTA::EventType Message::EventType(std::vector<char>& message) {
     return static_cast<CARTA::EventType>(head.type);
 }
 
-void Message::FillHistogram(CARTA::Histogram* histogram, const carta::BasicStats<float>& stats, const carta::Histogram& hist) {
-    FillHistogram(histogram, hist.GetNbins(), hist.GetBinWidth(), hist.GetBinCenter(), hist.GetHistogramBins(), stats.mean, stats.stdDev);
-}
-
-void Message::FillHistogram(CARTA::Histogram* histogram, int num_bins, double bin_width, double first_bin_center,
-    const std::vector<int>& bins, double mean, double std_dev) {
-    if (histogram) {
-        histogram->set_num_bins(num_bins);
-        histogram->set_bin_width(bin_width);
-        histogram->set_first_bin_center(first_bin_center);
-        *histogram->mutable_bins() = {bins.begin(), bins.end()};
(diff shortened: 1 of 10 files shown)


veggiesaurus (approved):

src/Util/Message.cc
```diff
@@ -0,0 +1,340 @@
...
+            double nan_value = nan("");
+            new_profile->set_raw_values_fp64(&nan_value, sizeof(double));
+        } else {
+            new_profile->set_raw_values_fp64(spectral_data[stats_type].data(), spectral_data[stats_type].size() * sizeof(double));
+        }
+    }
+    return profile_message;
+}
+
+void Message::FillStatisticsValue(
```
confluence: This should be called `FillStatisticsValues`, or just `FillStatistics`.

src/Util/Message.cc
```diff
@@ -0,0 +1,340 @@
...
+CARTA::EventType Message::EventType(std::vector<char>& message) {
+    carta::EventHeader head = *reinterpret_cast<const carta::EventHeader*>(message.data());
+    return static_cast<CARTA::EventType>(head.type);
+}
+
+void Message::FillHistogram(CARTA::Histogram* histogram, const carta::BasicStats<float>& stats, const carta::Histogram& hist) {
+    FillHistogram(histogram, hist.GetNbins(), hist.GetBinWidth(), hist.GetBinCenter(), hist.GetHistogramBins(), stats.mean, stats.stdDev);
+}
+
+void Message::FillHistogram(CARTA::Histogram* histogram, int num_bins, double bin_width, double first_bin_center,
```
confluence: If we're going to keep these function signatures (which I think is fine for now), these should *not* be methods on the `Message` class, since they are not like the functions which return messages. They should be functions outside of the class (as they were before), but still defined in the same files (`Message.h` and `Message.cc`).

src/Util/Message.cc
confluence: My comment about the `FillHistogram` functions also applies to this. The `Fill...` functions should be outside the `Message` class (but still in the `Message` files).

confluence (changes requested):
Apart from the naming issues, I think this is fine. We should extend these wrappers to _all_ messages that we're using, but I don't think that needs to be done in this PR.

confluence (approved):

Issue #744, opened 2021-02-24
Move message handling code to a separate class


veggiesaurus:
`Main.cc` is currently far too large, and has a variety of responsibilities. The majority of message-handling code can be moved to a new class `SessionManager`

veggiesaurus:
Increasing to 2 because we need to extract some of the code from `Session` that manages the exit timeouts (previously done with static variables)

pford:
We have FileListHandler and RegionHandler, should they be Manager instead for consistency?  Or is there a difference between "managing" a Session vs "handling" a file list request or region request?

veggiesaurus:
> We have FileListHandler and RegionHandler, should they be Manager instead for consistency? Or is there a difference between "managing" a Session vs "handling" a file list request or region request?

That's a good question. I do sort of feel like the region requests and file list requests are handled rather than managed, whereas the sesssion manager should manage all incoming requests, creating and deleting sessions when necessary. I'm open to counterarguments though :smile:

confluence:
This was closed in #923.

PR #923, opened 2021-09-21
Refactor constants, utils and main

### LINKED ISSUE #761: refactor Constants.h

confluence:
The main focus of this PR is to fix #761 and #744. It splits `Constants.h` and `Util.*` into multiple smaller files (currently in a `Util` subdirectory), and moves the messaging code from `Main.cc` into a new `SessionManager` class. Some gRPC-specific code has also been moved out of main, and some other common code has been rearranged.

Is this a sensible split? Should `Util` be called something else, like `Common`?

I have a few other refactoring suggestions, which could be done in this PR or in separate PRs:

* Move `MimeTypes` to `Util`
* Fix all the unnecessary relative imports (imports relative to the top-level directory work fine)
* Remove `using namespace std` everywhere -- we don't use it consistently anyway, and I think that it makes the code more confusing to read
* Fix the inconsistent use of namespaces
* Move `ProgramSettings` into the root directory -- it makes no sense for it to be in `SessionManager`, because the settings are not per-session
* Move `SimpleHttpServer` into the `SessionManager` directory?
* Move `Logger` into `Util`?


pford:
I'm not sure I understand how refactoring 3 files into 15 Util/ files is a simplification of the code, especially when the lines are blurred e.g. file code in Casacore.h or File.h?  But I agree with putting constants where they are used when they are not shared.

confluence:
> I'm not sure I understand how refactoring 3 files into 15 Util/ files is a simplification of the code, especially when the lines are blurred e.g. file code in Casacore.h or File.h? But I agree with putting constants where they are used when they are not shared.

The main motivation is to avoid putting all shared imports in a single header, so that any change to that header requires recompiling everything. Division into smaller and more specific files also makes the individual dependencies clearer conceptually.

I'm not completely sold on the Casacore subdivision; I could move the two formatting functions into a header specific to moments and divide the rest between File and Image. Does that sound like a better split?

pford:
If it helps with the goal of reducing compilation, keep the Casacore header separate.

pford (approved):

src/GrpcServer/CartaGrpcService.cc
```diff
@@ -88,3 +89,41 @@ grpc::Status CartaGrpcService::CallAction(
 
     return status;
 }
+
+GrpcManager::GrpcManager() : _selected_port(-1) {}
```
veggiesaurus, line 93: is this the "dummy" constructor for tests?

veggiesaurus (approved):
looks fine. We can merge as soon as the conflicts are resolved

src/GrpcServer/CartaGrpcService.cc
confluence: It's the dummy constructor for the default case of running the backend with gRPC disabled.

PR #927, opened 2021-09-23
Fix loading error for the compressed FITS file

### LINKED ISSUE #924: load fz and fits.gz via the command line
Reported by a user with v3b1 release:

It seems not being able to load fits.gz image directly via the command line (eg `carta xxx.fits.gz`). Tested with the ubuntu package 

Screen Shot 2021-09-23 at 1 43 53 PM

(The same issue can be reproduced by loading the image in URL such as
`http://localhost:3002/?file=/Users/SpongeBob/carta_image_pool/set_compressed_fits/spire500_ext.fits.gz`. If we use double-click to load the image, we see the same issue too)

and the macOS electron app (`Error: Possibly an unsupported file type`).

We should also need to ensure fz can be loaded via the command line.

markccchiang:
This PR tries to fix #924. When loading a compressed FITS file via URL query parameter or double click, the frontend sends an empty HDU string. And the backend tries to get an HDU list through `FitsHduList,` thus an exception throw: `Use CompressedFits for HDU header map.` I just added a condition check that the FitsHduList object can not process the compressed FITS file. @kswang1029 could you please check doest it solve the problem?

### DIFF
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -213,10 +213,9 @@ bool Session::FillExtendedFileInfo(CARTA::FileInfoExtended& extended_info, CARTA
                 hdu = file_info.hdu_list(0);
             }
 
-            if (hdu.empty() && (file_info.type() == CARTA::FileType::FITS)) {
+            if (hdu.empty() && (file_info.type() == CARTA::FileType::FITS) && !IsCompressedFits(fullname)) {
                 // File info adds empty string for FITS
                 std::vector<std::string> hdu_list;
-                std::string message;
                 FitsHduList fits_hdu_list(fullname);
                 fits_hdu_list.GetHduList(hdu_list, message);
 
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -31,6 +31,7 @@
 #include "FileList/FileExtInfoLoader.h"
 #include "FileList/FileInfoLoader.h"
 #include "FileList/FitsHduList.h"
+#include "ImageData/CompressedFits.h"
 #include "Logger/Logger.h"
 #include "OnMessageTask.h"
 #include "SpectralLine/SpectralLineCrawler.h"
@@ -220,12 +221,10 @@ bool Session::FillExtendedFileInfo(CARTA::FileInfoExtended& extended_info, CARTA
             if (hdu.empty() && (file_info.type() == CARTA::FileType::FITS)) {
                 // File info adds empty string for FITS
                 if (IsCompressedFits(fullname)) {
-                    std::map<std::string, CARTA::FileInfoExtended> hdu_info_map;
-                    file_info_ok = ext_info_loader.FillFitsFileInfoMap(hdu_info_map, fullname, message);
-                    if (hdu_info_map.empty()) {
+                    CompressedFits cfits(fullname);
+                    if (!cfits.GetFirstImageHdu(hdu)) {
                         return file_info_ok;
                     }
-                    hdu = hdu_info_map.begin()->first; // get the key (hdu name) of the first element of a map
                 } else {
                     std::vector<std::string> hdu_list;
                     FitsHduList fits_hdu_list(fullname);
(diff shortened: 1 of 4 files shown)


kswang1029:
Yes, this partly fixed the issue.

with the attached image, we still cannot load the image as the first HDU is not an image. 
[spire500_ext.fits.gz](https://github.com/CARTAvis/carta-backend/files/7221786/spire500_ext.fits.gz)

The backed log says `[error] Image must be at least 2D.`

I think we still need to get the HDU list first then load the "first" "image" HDU (not always the first HDU as it may not be an image at all)

markccchiang:
> Yes, this partly fixed the issue.
> 
> with the attached image, we still cannot load the image as the first HDU is not an image.
> [spire500_ext.fits.gz](https://github.com/CARTAvis/carta-backend/files/7221786/spire500_ext.fits.gz)
> 
> The backed log says `[error] Image must be at least 2D.`
> 
> I think we still need to get the HDU list first then load the "first" "image" HDU (not always the first HDU as it may not be an image at all)

Thanks for the comment! Please check again does the new commit solve the problem?

kswang1029:
> > Yes, this partly fixed the issue.
> > with the attached image, we still cannot load the image as the first HDU is not an image.
> > [spire500_ext.fits.gz](https://github.com/CARTAvis/carta-backend/files/7221786/spire500_ext.fits.gz)
> > The backed log says `[error] Image must be at least 2D.`
> > I think we still need to get the HDU list first then load the "first" "image" HDU (not always the first HDU as it may not be an image at all)
> 
> Thanks for the comment! Please check again does the new commit solve the problem?

yes now with command line or URL query parameter, fits.gz and fz images can be loaded successfully. 👍

src/Session.cc
```diff
@@ -215,21 +219,27 @@ bool Session::FillExtendedFileInfo(CARTA::FileInfoExtended& extended_info, CARTA
 
             if (hdu.empty() && (file_info.type() == CARTA::FileType::FITS)) {
                 // File info adds empty string for FITS
-                std::vector<std::string> hdu_list;
-                std::string message;
-                FitsHduList fits_hdu_list(fullname);
-                fits_hdu_list.GetHduList(hdu_list, message);
+                if (IsCompressedFits(fullname)) {
+                    std::map<std::string, CARTA::FileInfoExtended> hdu_info_map;
+                    file_info_ok = ext_info_loader.FillFitsFileInfoMap(hdu_info_map, fullname, message);
```
pford: This works but is a lot of wasted effort to fill FileInfoExtended header_entries and computed_entries for all HDUs just to throw away everything but the first HDU string.  We could modify `CompressedFits::GetFitsHeaderInfo` (e.g. add arguments `int& first_hdu, bool first_hdu_only=false`, set to true for this), or create a similar much smaller function `int CompressedFits::GetFirstImageHdu()`, which stops after IsImageHdu returns true and returns the hdu.

I could do this if you'd rather not work with zlib, but it's really just adding gzclose and returning the hdu number when you find the first image HDU.

pford (changes requested):

src/Session.cc
pford: Also check whether there is an image HDU detected, perhaps there are FITS files with none?

markccchiang: Thank you for the suggestions! I created a new smaller function to get the first image HDU.

src/Session.cc
```diff
@@ -215,21 +220,25 @@ bool Session::FillExtendedFileInfo(CARTA::FileInfoExtended& extended_info, CARTA
...
             if (hdu.empty() && (file_info.type() == CARTA::FileType::FITS)) {
                 // File info adds empty string for FITS
-                std::vector<std::string> hdu_list;
-                std::string message;
-                FitsHduList fits_hdu_list(fullname);
-                fits_hdu_list.GetHduList(hdu_list, message);
+                if (IsCompressedFits(fullname)) {
+                    CompressedFits cfits(fullname);
+                    if (!cfits.GetFirstImageHdu(hdu)) {
+                        return file_info_ok;
```
pford, line 227: Add message that no image HDU found.  The rest looks good!

pford (changes requested):

pford (approved):

kswang1029 (approved):

PR #928, opened 2021-09-24
Fix the crash problem when loading a non-image via the command line

### LINKED ISSUE #926: backend crash when loading a non-image via the command line
Tested with Ubuntu package. If we do something like `carta Jy_per_beam_TO_Kelvin.png`, the frontend loads up in the browser but we will immediately see a pop up due to backend crash.

markccchiang:
This PR is to fix #926. The backend returns an error message without crash if it can not open an image file.


src/FileList/FileExtInfoLoader.cc
```diff
@@ -107,6 +107,10 @@ bool FileExtInfoLoader::FillFileExtInfo(
     bool info_ok(false);
     if (_loader && _loader->CanOpenFile(message)) {
```
pford: _loader is a nullptr for an unsupported file or image type.  Wouldn't it be better to catch this in Session, before it is passed to the FileExtInfoLoader constructor and FillFileExtInfo() is called?

The message should probably use the word "file" instead of "image" since it could be, as in the subject of this PR, a non-image file.  It could also be an unsupported image type, a file permissions issue, or other problems. @kswang1029  how can this message cover these different scenarios and not be confusing?  Or am I overthinking this?

pford (changes requested):

src/FileList/FileExtInfoLoader.cc
kswang1029: Good point. I think we simply say “the file cannot be loaded”. If we, however, can know _why_ it cannot be loaded _without extra work_, we can also show the reasons (unsupported format, permission, broken file etc) to users.

kswang1029: if there is an equivalent handy “file” command that we can use, it may be good to use it and return the output as part of our error message.

pford: Agree, can check if file does not exist else if file is not readable else if !_loader, with a message for each.  We have to do extra checking since this is user input.

markccchiang: I think the backend already checks if the file exists or is readable @kswang1029 could you please confirm it?

kswang1029: If it is permission issue, we will see a popup error message on the frontend. 
Screen Shot 2021-09-27 at 12 55 38 PM

If it is a file that does not exist, the frontend loads without any error/warning messages. The file browser shows up just like usual.

confluence (approved):
This looks fine to me.

src/FileList/FileExtInfoLoader.cc
pford: @veggiesaurus @kswang1029 is this resolved?  The permissions error message is coming from casacore::RegularFileIO and could be improved.  And according to the previous comment, if the file does not exist it is ignored.  Is there a message in the log/console?  If I were a user I would be confused.

kswang1029: I think if we load a file that does not exist, we should show an error message like "Input file does not exist.".

markccchiang: The `carta_backend` can print an error message "File does not exist." if we load a file that does not exist. But I am not sure does the `carta` wrapper can catch this message? Or the `carta_backend` should throw an exception with the message "File does not exist." to let the `carta` catch this error message?

kswang1029: I think we should let `carta_backend` throw the error message. One possible scenario is a user uses the GUI to load an image but after receiving the file list, the selected image is deleted somehow. So in this case, when the `carta_backend` loads the image, it can warn the user.

markccchiang: @kswang1029 With current code changes, can the `carta` show the error message `Unsupported format.` if we load a non-image file? If not, this error message should also come up with an exception thrown.

kswang1029: Hmm I am not sure if I can test the wrapper properly. I do see unsupported image error on the frontend.

markccchiang: In my test, the frontend can show `File does not exist` message if we load a file that does not exist. Perhaps we can verify it after the new `carta` is ready.

kswang1029: Ok

kswang1029 (approved):

pford (approved):

veggiesaurus:
@markccchiang can you just solve the merge conflicts and then we can merge this in?

PR #929, opened 2021-09-27
Respect minimum CMake version of 3.10

### LINKED ISSUE #904: ensure CMakeLists.txt is fully compatible with the specified cmake_minimum_required
Our `CMakeLists.txt` has `cmake_minimum_required` as `3.10`, but we are using some functions from later versions. We need to make sure that we conform to 3.10.

Main offender seems to be `add_compile_definitions` , which was introduced in 3.12.

confluence:
This should fix #904; as far as I can tell these were the only functions which were too new.


ajm-ska (approved):
I tested it in an Ubuntu 18.04 Docker container. Everything seems to work fine with Ubuntu 18.04's default cmake version 3.10.2.

veggiesaurus (approved):

PR #934, opened 2021-10-04
Pv generator

### LINKED ISSUE #794: in-house implementation of a pv image generator
The reason why we cannot re-use the code of casa-impv for the pv image generation is performance. casa-impv attempts to generate a _rotated_ cube first so that the pv cut is either vertical or horizontal before actual image slicing. This is a slow process and seems unnecessary and does not scale well when image cube becomes larger. 

A more scalable approach (and cancellable) is to sample the pv cut as a series of rectangle regions with 1 x cut_width (default 3pix) shape and request region spectral profiles to form a 2D image. Based on the experiments with astropy, we see at least an order of magnitude performance boost. This approach can be also used to support generating a pv image along a polyline (hence an arbitrary curve). 

Screen Shot 2021-04-13 at 12 07 59 PM

Once uncertainty of this proposed approach is how we form a proper header for the offset axis, especially when projection effect is not negligible. 

Science requirements: https://docs.google.com/document/d/1Z1fCDj9IujGvV6ER_bzjjL2j_bF5ykc34GRhPdOiXI4/edit

pford:
Closes #794.

The implementation is done in RegionHandler, which approximates the line as a series of box regions with the requested width.  The PV image data is the spectral profile of each box region set in an array.  This code can be reused for the mean spatial profile along a line by getting the mean of each region for the current channel instead of per-channel.  Polyline is not currently supported.

For images with fixed pixel separation, a series of temporary rectangle regions is created along the line.  If the angular separation of the fixed pixel points varies, a fixed angular separation (variable pixel interval) is used to create a series of temporary polygon regions to describe a box with four corners.  This is because the pixel center for a rectangle may not be the angular center.  The mean of three adjacent boxes was used for the angular separation to avoid a region with no pixel centers, resulting in a nan spectral profile.

Each rectangle or polygon region is set with a temporary region id, then its spectral profile is calculated and added as a row in the 2D array which will be used for the PV image data.  Then the temporary region is destroyed.

Once the data array is set, it is passed to the PvGenerator which creates an image with the appropriate CoordinateSystem and sets the data.  The x-axis is an Offset axis with offset 0 at the line center, positive offsets toward endpoint 0, and negative offsets toward endpoint 1.  This image is passed back to Session which opens it, similar to the MomentGenerator.

I created a directory ImageGenerators for the MomentGenerator and PvGenerator with a shared header ImageGenerator.h.  The CollapseResult struct has been generalized to a GeneratedImage struct in this header file for the MomentGenerator and PvGenerator.

Region implementation document:
https://docs.google.com/document/d/1VZqXM-A8Djmsybz4kNJsXudoghnjtqFvadB8SxlqEUQ/edit#


kswang1029:
@pford for a cube with a pixel size of 0.05 arcsec, the offset pixel of the pv image generated with carta is 0.0494949". Could you please check possible places for the error here?

I also noticed that given the same line region definition as from-to (eg [100, 100] to [200, 100]) in image coordinate, the offset axis has two less pixels (eg 99) comparing to the pv image generated with impv (eg 101). Could you check too?

kswang1029:
@pford When a pv cut is partially outside the image or totally outside the image, we need a proper handle of these cases. For the partially outside case, we should still be able to generate a pv image with those outside part as NaN. For the totally outside case, the backend should provide a proper error message to the frontend (and maybe the frontend can check if the line region and the image have intersection before enabling the generate button (@TienHao)
.

kswang1029:
@pford are you able to generate a pv image if the input is a cube with IQUV as Stokes? I see it runs with few progress update messages and stops. The backend log shows 

> TBB Warning: Exact exception propagation is requested by application but the linked library is built without support for it

Not sure if this is due to my build env or else. (adding @ajm-asiaa in the loop)

pford:
@kswang1029 Several fixes have been added:

- Fix calculation of increment
- Update progress for null regions and when complete
- When input image has a beam set, use first beam to set pv image restoring beam (as done in imageanalysis)

When the PV generator fails (response success=false with message), this error is not displayed by the frontend, which is confusing when nothing happens.  I sent a message to @TienHao .

I tested the pv generator for two IQUV images:

- The first image failed because it had no spectral coordinate, so the backend sent an error message in the response, "Cannot generate PV image with no valid spectral axis".  Casacore failed to create the spectral coordinate from the headers since its axis CTYPE was "Frequency" instead of "FREQ".  Should I add a workaround for this?  The check was added for e.g. 2D images. And normally, the PV image spectral coordinate is simply a copy of the input image spectral coordinate.
- The second image succeeded after the beam set fix.  Before the fix, this issue caused the backend to hang.
- I did not see the TBB warning.

pford:
@kswang1029 more detail on pv cuts outside the image: the line box regions outside the image are null.  When the box regions vector ended in null regions, the progress was never updated to completion, so the PV image was not created. The error "PV generator failed to complete" was sent but not displayed.  This has been fixed so that, as you indicated, the null regions produce NaN data in the generated PV image.

When ALL regions are null, the error message "PV calculation failed for input region" is sent.  This message may be too vague (and poorly worded), but I did not know if there may be another case where all regions are null besides being outside the image (?) so I did not want to be too specific.

kswang1029:
@pford regarding the TBB warning (and backend hangs) issue, could you try the attached image and a cut like the screenshot? In my test env, the TBB warning shows up and the backend hangs.

Screen Shot 2021-10-13 at 3 09 03 PM

https://drive.google.com/file/d/1oVt7awaUOk1tebmuaIaBPwY7crM7zY7Y/view?usp=sharing

update: we need to consider to following cases
SDC335 579-0 292 spw0 line fits-image-2021-10-13-15-44-48

pford:
@kswang1029 I did not get the TBB message but the pv generator was hanging when trying to assign the spectral profile to the PV image data array row.  It turns out the profile is empty when all input values are NAN.  Fixed.

kswang1029:
> @kswang1029 I did not get the TBB message but the pv generator was hanging when trying to assign the spectral profile to the PV image data array row. It turns out the profile is empty when all input values are NAN. Fixed.

YES! Now the hanging issue is gone. I will do more tests and report back.

kswang1029:
@pford @TienHao 
I found two issues:
1) If I request a cancellation once, I cannot request a new generation anymore.
2) If I have two cubes matched and if we use the GUI to request a pv generation from the 2nd image, the backend will hang. (and I see the TBB warning - this is probably just my build env problem).

kswang1029:
@pford @TienHao I found two issues:

 1. If I request a cancellation once, I cannot request a new generation anymore.
 2. If I have two cubes matched and if we use the GUI to request a pv generation from the _2nd_ image, the backend will hang. (and I see the TBB warning - this is probably just my build env problem).

kswang1029:
@pford good news to report:
In order to test the case when the input cube is a wide field image, I made a test cube with a set of sources sitting at the celestial equator with _equal_ angular separation. Duo to SIN projection, those sources have variable separations on the raster image as you can see on the left panel. Then I made a pv cut alone the equator and request the pv generation. If the line box generation with equal angular separation works, we should see sources appear with equal separation on the OFFSET axis, and YES, I think they appear at right locations as we can see on the right panel. This is GREAT!.

Screen Shot 2021-10-15 at 1 41 07 PM

PS. ~~we see some line boxes results in all NaN near the image edge, probable due to sampling issue and projection effect. This should be okay I think as I am testing an edge use case.~~
hmm for those NaN stripes, do you still use 1px as the width of the line boxes? If so I think we need to refine it to make the width be constant angular width. An approximation would be using the middle point of adjacent line box centers to build up the width. And about height, we will need to have an adaptive one. That is, calculate the angular size of 3px from header, then use the value at different places to get the height in pixel (eg 3.5px). This way, each box should approximately have equal sky coverage (aka solid angle). 

PS2: @pford @TienHao do you think we can have an option for the unit of the OFFSET axis? We should consider arcsec, arcmin, and degree. (if this will take significant extra time, we can leave it for now)

kswang1029:
@pford how do we determine when we trigger "fixed angular increment" sampling? Is it something like
Screen Shot 2021-10-15 at 2 19 08 PM

pford:
I fixed the matched-image pv generator issue, and improved cancellation.  It is possible that the dialog was hanging after cancel due to no pv response; please re-test.

The unit can be changed for the OFFSET axis, this is simply one casacore::Quantity conversion to the new unit.

I will write a short doc describing the fixed pixel vs angular separation.

pford:
Line box regions implementation is described in this document (in CARTA shared drive backend development folder):
https://docs.google.com/document/d/1VZqXM-A8Djmsybz4kNJsXudoghnjtqFvadB8SxlqEUQ/edit?usp=sharing

In both cases, the rectangle region's width is set to the width in the PvRequest message.  But this should be variable like the height for the angular separation case? (width is perpendicular to line, height is along line)

pford:
Converted this PR to draft while testing and tuning.  Still could use a code review of the basic strategy.

@kswang1029: I am testing a widefield image that uses the fixed angular increment which produces null box regions (casacore error = "LCPolygon - entire polygon is outside the lattice") between two valid regions, hence the NaN offsets.  Investigating this error.

pford:
Removing code reviewers while refactoring and until user testing is further along.

pford:
Pushed refactoring change - spectral profiles are calculated from each temporary box region rather than making a vector of box regions and getting profiles later.  Also added debug output; each rectangle region along the box is printed as "polygon[[...]]" in CRTF pixel format for pasting into a file to import and check, as well as max of NumPixels spectral profile for each region (always seems to be nan, but spectral profile is not nan).  This was not nan when spectral profiles were calculated in the PvGenerator, usually 0-3 pixels for width=3.

Preliminary testing looks okay.  I will continue testing and investigate the NumPixels issue.  Still need to implement fixed angular width.

pford:
NumPixels = nan was because this stats type was not enabled for spectral profiles.  It is fixed now.

pford:
Fixed angular regions were changed to use polygon; will update implementation doc.  Ready for testing.  I will also do more testing on Monday.

kswang1029:
> Fixed angular regions were changed to use polygon; will update implementation doc. Ready for testing. I will also do more testing on Monday.

I will start testing and report asap.

pford:
Converted from draft PR at @kswang1029 's request for code review.  Most of the implementation is in RegionHandler, not sure who should be assigned to this PR. 

Remaining issue: when a line is matched between two widefield images and the PV image is requested for the matched image, the PV image is incorrect.  The converted line regions are the 1000-point polygon approximations, but this works when a single region is set and the z profile is requested for the matched region.  Still investigating this bug.

pford:
Remaining issue has been fixed (subject to further testing)

kswang1029 (approved):
Thank you @pford to bring the implementation of pv generator this far! 👍 I have tested some use cases including cancellation with small and large cubes, pv generation with narrow/wide field images, pv generation from matched narrow/wide field images etc. No obvious issues are seen for now. Hope we can learn more from users with the v3beta2 release.

kswang1029:
@pford I think we need a changelog for this PR.

pford:
I updated the backend and protobuf changelogs.  @kswang1029 let me know if I need to remove any debug output for the line regions (polygon points, num pixels).

kswang1029:
> I updated the backend and protobuf changelogs. @kswang1029 let me know if I need to remove any debug output for the line regions (polygon points, num pixels).

Let’s keep it in the debug mode for now just in case. 🙂

markccchiang (approved):

veggiesaurus:
@pford can you make a separate PR for the protobufs please?

src/ImageGenerators/PvGenerator.h
```diff
@@ -0,0 +1,46 @@
...
+
+#include "../ImageData/FileLoader.h"
+#include "ImageGenerator.h"
+
+namespace carta {
+
+class PvGenerator {
+public:
+    PvGenerator(int file_id, const std::string& filename);
+    ~PvGenerator(){};
```
veggiesaurus: perhaps replace with `= default`?

veggiesaurus:
@pford I don't have many comments about the code itself, it looks good. But could we add some basic unit tests for the PV generator? I'm happy to have them done in a separate PR if we want to get beta-2 out the door soon, but I do think this could benefit from some unit tests to catch potential regressions as we add functionality going forward.

src/ImageGenerators/PvGenerator.h
pford: Or just remove?  I think this is a refactoring relic.

I also think I don't need to have a map of pv generators in RegionHandler, since the cancellation is done in RegionHandler.  The pv generator is now a local variable after the profiles (pv data) are complete.

pford:
I added TestPvGenerator, which uses test images in the backend repo (test/data/images/) since the generated images do not seem to have a valid coordinate system, which is required.  I check the PV image coordinate system, data shape, and that the data is not all NAN, and that an image with no spectral coordinate fails.

veggiesaurus:
> I added TestPvGenerator, which uses test images in the backend repo (test/data/images/) since the generated images do not seem to have a valid coordinate system, which is required. I check the PV image coordinate system, data shape, and that the data is not all NAN, and that an image with no spectral coordinate fails.

Ok. I think that's a good start. We should probably (a) adjust the generator to include a valid coord system and (b) add more robust tests in separate issues.

veggiesaurus (approved):

pford:
@veggiesaurus I thought I could easily add a PV generator test with a horizontal line and width 1, which would be equivalent to an image data slice with an x-range at the value of y.  I found that the PV data was the same no matter which way the line start/end was set, which turned out to be a bug in the offset direction for a horizontal line with fixed pixel spacing.  I should add a vertical line test as well.

I fixed the bug, added the test, and updated protobuf with the new commit and major version.  Will do more testing tomorrow, but @kswang1029 you might want to have another look at the pv image offset direction with fixed pixel increment if you have time.  The fixed angular offset direction should be correct since the separation is calculated toward the starting or ending endpoint.

pford:
Testing may have to wait until frontend ICD version is updated...

kswang1029:
> @veggiesaurus I thought I could easily add a PV generator test with a horizontal line and width 1, which would be equivalent to an image data slice with an x-range at the value of y. I found that the PV data was the same no matter which way the line start/end was set, which turned out to be a bug in the offset direction for a horizontal line with fixed pixel spacing. I should add a vertical line test as well.
> 
> I fixed the bug, added the test, and updated protobuf with the new commit and major version. Will do more testing tomorrow, but @kswang1029 you might want to have another look at the pv image offset direction with fixed pixel increment if you have time. The fixed angular offset direction should be correct since the separation is calculated toward the starting or ending endpoint.

will do shortly and report back

confluence (approved):

pford:
Created pv_generator branch in ICD tests to set IcdVersion = 25 and opened PR.

kswang1029:
this commit 5ef8d36 helps a lot to remove the sampling artifacts. 
The middle image is from CARTA pv generator and the right image is from CASA-impv. It is hard to tell differences visually. Numerical values differ a bit but this is understandable as CASA-impv does image regrid and CARTA does box averaging.

Screen Shot 2021-12-08 at 12 01 42 PM

@veggiesaurus ready to merge in. @pford thank you again. This is one of the big milestones in the CARTA dev history.

PR #936, opened 2021-10-13
Angus/plaftorm info

veggiesaurus:
Fills `plaftorm_strings` map with OS information when sending `RegisterViewerAck` messages.

Protobuf PR [here](https://github.com/CARTAvis/carta-protobuf/pull/57), but changes are non-breaking.

The frontend telemetry PR is https://github.com/CARTAvis/carta-frontend/pull/1670, but this can be merged in before that PR is sorted out.

### DIFF
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -312,9 +312,9 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
     auto& platform_string_map = *ack_message.mutable_platform_strings();
     platform_string_map["release_info"] = GetReleaseInformation();
 #if __APPLE__
-    platform_string_map["plaform"] = "MacOS";
+    platform_string_map["plaform"] = "macOS";
 #else
-    platform_string_map["plaform"] = "Linux";
+    platform_string_map["platform"] = "Linux";
 #endif
 
     uint32_t feature_flags;
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -312,7 +312,7 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
     auto& platform_string_map = *ack_message.mutable_platform_strings();
     platform_string_map["release_info"] = GetReleaseInformation();
 #if __APPLE__
-    platform_string_map["plaform"] = "macOS";
+    platform_string_map["platform"] = "macOS";
 #else
     platform_string_map["platform"] = "Linux";
 #endif
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -40,7 +40,9 @@
 #ifdef _ARM_ARCH_
 #include <sse2neon/sse2neon.h>
 #else
+#include <Util/App.h>
 #include <xmmintrin.h>
+
 #endif
 
 int Session::_num_sessions = 0;
@@ -306,6 +308,14 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
     ack_message.set_success(success);
     ack_message.set_message(status);
     ack_message.set_session_type(type);
+
+    auto& platform_string_map = *ack_message.mutable_platform_strings();
+    platform_string_map["release_info"] = GetReleaseInformation();
+#if __APPLE__
+    platform_string_map["plaform"] = "MacOS";
+#else
+    platform_string_map["plaform"] = "Linux";
+#endif
 
     uint32_t feature_flags;
     if (_read_only_mode) {
--- a/src/Util/App.cc
+++ b/src/Util/App.cc
@@ -8,10 +8,19 @@
 
 #include <unistd.h>
 #include <climits>
+#include <fstream>
+#include <vector>
 
 #ifdef __APPLE__
 #include <mach-o/dyld.h>
+#include <stdio.h>
 #endif
+
+#include "FileSystem.h"
+#include "Logger/Logger.h"
+
+#define MAX_PLATFORM_INFO_LENGTH 1024
+#define MAX_PLATFORM_LINE_LENGTH 256
 
 bool FindExecutablePath(std::string& path) {
     char path_buffer[PATH_MAX + 1];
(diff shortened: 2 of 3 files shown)


src/Session.cc
```diff
@@ -307,6 +309,14 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
     ack_message.set_message(status);
     ack_message.set_session_type(type);
 
+    auto& platform_string_map = *ack_message.mutable_platform_strings();
+    platform_string_map["release_info"] = GetReleaseInformation();
+#if __APPLE__
+    platform_string_map["plaform"] = "MacOS";
+#else
+    platform_string_map["plaform"] = "Linux";
+#endif
```
jolopezl: Here, `"plaform"` should be `"platform"`, and `"MacOS"` should be `"macOS"`, in case this needs an exact match with the system information.

src/Util/App.cc
```diff
@@ -33,3 +42,40 @@ bool FindExecutablePath(std::string& path) {
...
+    while (fgets(line, sizeof(line), file_handle) != nullptr) {
+        int l = snprintf(info_buffer + buffer_length, sizeof(info_buffer) - buffer_length, "%s", line);
+        buffer_length += l;
+        if (buffer_length > MAX_PLATFORM_INFO_LENGTH) {
+            spdlog::warn("Problem reading platform information");
+            return std::string("Platform information not available");
+        }
+    }
+    pclose(file_handle);
+    return info_buffer;
```
jolopezl, line 62: I think that this part is susceptible to changes in the OS. Maybe, it will be less trouble in the future to read the fields directly from the command line, without parsing using the following:
```
~$ sw_vers -productName
macOS
~$ sw_vers -productVersion
11.6
~$ sw_vers -buildVersion
20G165
```
This alternative would involve just two calls to `popen()`.

There are other choices as well that I am taking from this reference: https://www.cyberciti.biz/faq/mac-osx-find-tell-operating-system-version-from-bash-prompt/. I would also like to parse the content of `/System/Library/CoreServices/SystemVersion.plist`:
```
~$ cat /System/Library/CoreServices/SystemVersion.plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>ProductBuildVersion</key>
	<string>20G165</string>
	<key>ProductCopyright</key>
	<string>1983-2021 Apple Inc.</string>
	<key>ProductName</key>
	<string>macOS</string>
	<key>ProductUserVisibleVersion</key>
	<string>11.6</string>
	<key>ProductVersion</key>
	<string>11.6</string>
	<key>iOSSupportVersion</key>
	<string>14.7</string>
</dict>
</plist>
```
We already have `pugixml` in the codebase, and this could be another choice to keep things less prone to errors due to software upgrades. However, maybe the only advantage of this approach is avoiding a system call, but I would expect Apple not to change the keys of this XML file!

jolopezl (commented):
Looks good, I just have a couple of comments on the way you are getting the macOS information.

src/Session.cc
veggiesaurus: :see_no_evil: nice catch. I swear I've made that typo a million times!

src/Util/App.cc
veggiesaurus: I considered doing this, but I'd rather handle the platform interpretation on the telemetry server side of things. That way, if Apple changes the output of `sw_vers` in some way, we can adapt our server-side code, instead of needing to push a fix to every CARTA instance in the world.

The approach I've taken for MacOS is somewhat consistent with the Linux approach. `sw_vers` output looks like

```
ProductName:    Mac OS X
ProductVersion: 10.12.5
BuildVersion:   16F73
```

and `/etc/os-release` looks like

```
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.3 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

both will require some regex work on the server-side, but that's much easier to iterate on than altering CARTA itself.

src/Session.cc
kswang1029: as it is called Mac OS X or OS X or macOS, please refer to https://zh.m.wikipedia.org/wiki/MacOS 
Apple renamed it few times in the past

src/Session.cc
```diff
@@ -307,6 +309,14 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
     ack_message.set_message(status);
     ack_message.set_session_type(type);
 
+    auto& platform_string_map = *ack_message.mutable_platform_strings();
+    platform_string_map["release_info"] = GetReleaseInformation();
+#if __APPLE__
+    platform_string_map["plaform"] = "macOS";
```
confluence: You changed `macOS`, but it's still `plaform` :D

src/Util/App.cc
confluence: I'm happy with the approach of parsing this in the server code.

confluence (changes requested):

confluence (approved):

src/Util/App.cc
jolopezl: I think my comment was misleading. I didn't mean to do things through the command line, but instead of parsing the entire output of `sw_vers`use the command-line options that come with it to get macOS, and the version number from two `popen` calls. Anyway, the way it is now is ok too, and maybe simpler.

jolopezl (approved):

PR #937, opened 2021-10-14
Use cfitsio for FITS headers

### LINKED ISSUE #912: Incorrect projection in v3b1 when CROTA keys present
An IDIA user has some mosaics created from CASA and WSClean images, which both include the deprecated CROTA FITS header keys. I've asked for a copy of these images to provide for @kswang1029.

carta.idia (v2.0.1) projects it fine, including with diagonal grid lines for the mosaic of the WSClean images. But carta-beta.idia (v3.0.0-beta.1b) incorrectly projects it without the diagonal grid lines. Spatially matching them doesn’t change it either.

pford:
Closes #912 

Workaround for casacore::ImageFITSConverter::ImageHeaderToFITS, which loses the index _n_ on headers CXXXXn (for example CTYPE1, CRPIX2, CUNIT3, CROTA2).  They are returned in order so we have been incrementing _n_ for each header name each time it appears.  However, there is no way to guess _n_ for CROTAn which is not in order.

Instead, we will use native libraries where possible to create the header entries: (1) HDF5 for HDF5 images (retrieved from CartaHdf5Image) and (2) cfitsio for FITS images (retrieved from CartaFitsImage, and generated in FileExtInfoLoader for casacore::FITSImage),  but continue to convert images to headers using casacore as before for CASA and MIRIAD images.

Tested using the image in issue #912 by verifying the CROTA2 header keyword and the WCS grid on the loaded image.

### DIFF
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -455,15 +455,14 @@ void FileExtInfoLoader::FitsHeaderInfoToHeaderEntries(casacore::ImageFITSHeaderI
             }
         }
 
-        auto header_entry = extended_info.add_header_entries();
-        header_entry->set_name(name);
-        header_entry->set_comment(comment);
-
         switch (fkw->type()) {
             case casacore::FITS::LOGICAL: {
                 bool value(fkw->asBool());
                 std::string bool_string(value ? "T" : "F");
 
+                auto header_entry = extended_info.add_header_entries();
+                header_entry->set_name(name);
+                header_entry->set_comment(comment);
                 *header_entry->mutable_value() = bool_string;
                 header_entry->set_entry_type(CARTA::EntryType::INT);
                 header_entry->set_numeric_value(value);
@@ -473,6 +472,9 @@ void FileExtInfoLoader::FitsHeaderInfoToHeaderEntries(casacore::ImageFITSHeaderI
                 int value(fkw->asInt());
                 std::string string_value = fmt::format("{:d}", value);
 
+                auto header_entry = extended_info.add_header_entries();
+                header_entry->set_name(name);
+                header_entry->set_comment(comment);
                 *header_entry->mutable_value() = string_value;
                 header_entry->set_entry_type(CARTA::EntryType::INT);
                 header_entry->set_numeric_value(value);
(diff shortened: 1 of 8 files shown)


pford:
Both tests fail due to additional header entries.  @kswang1029 do you want the COMMENT and HISTORY headers shown in the file browser and File Header widget?

kswang1029:
> Both tests fail due to additional header entries. @kswang1029 do you want the COMMENT and HISTORY headers shown in the file browser and File Header widget?

Yes please. This would be ideal and consistent with ds9

pford:
I enabled history for ImageHeaderToFITS used for CASA and MIRIAD images; already included for HDF5 and FITS in this branch.

kswang1029:
@pford @veggiesaurus Got an issue loading an image (see link). The image loads but the WCS grid cannot be rendered properly. It appears that with this backend change, the frontend needs to be updated too to accommodate some exceptional (non-standard?) wcs headers. 

https://drive.google.com/file/d/1mxo4YO6H7zLq9T1eZnHBnGpfe1ITKoEl/view?usp=sharing

Screen Shot 2021-10-29 at 11 13 38 AM

pford:
This is a backend error.  It is reading the primary header.

pford:
@kswang1029 I fixed the header issue, please re-test.

kswang1029 (approved):
Have tried FITS images in our test image pool and it works fine as far as I can tell. 👍

src/FileList/FileExtInfoLoader.cc
```diff
@@ -425,120 +455,118 @@ void FileExtInfoLoader::FitsHeaderInfoToHeaderEntries(casacore::ImageFITSHeaderI
...
+                header_entry->set_comment(comment);
+                header_entry->set_entry_type(CARTA::EntryType::STRING);
+                break;
             }
+            case casacore::FITS::NOVALUE: {
+                auto header_entry = extended_info.add_header_entries();
+                header_entry->set_name(name);
+                header_entry->set_comment(comment);
+                header_entry->set_entry_type(CARTA::EntryType::STRING);
+                break;
```
confluence: 1. It looks like the calculated `string_value` not actually being used in any of these cases.
2. Would it be possible to refactor this case statement to move the common elements of the `header_entry` creation outside?

confluence (changes requested):

confluence (approved):

veggiesaurus:
this looks good. @kswang1029 I guess we just need to update the ICD tests now, since the `HISTORY` headers are parsed correctly as of this PR.

PR #941, opened 2021-10-18
Made the fits2idia converter a program dependency

### LINKED ISSUE #940: cfitsio 3.47 dependency for RHEL7/CentOS7
I will put this here for now even though technically it should go to the fits2idia repository?

For RHEL7/CentOS7 we can not build the carta-backend with the unit-tests (-Dtest=on). This is because RHEL7 comes with cfitsio 3.37 by default (through the [EPEL repository](https://centos.pkgs.org/7/epel-x86_64/cfitsio-3.370-10.el7.x86_64.rpm.html)). The older cfitsio 3.37 works fine if simply building the carta-backend as a release version, but there is a problem if we try to build it with the unit-tests.

A workaround is to install a newer cfitsio, such as version 3.47 (the same version used by Ubuntu 20.04 and RHEL8). That could be easily supplied on our cartavis rpm repository. However, it is not ideal. As @veggiesaurus points out, installing a newer cfitsio on a users machine could conflict with other astronomy software that use that library, plus we should ensure that we run the tests using the same version of cfitsio as the packaged version. This is important because cfitsio is quite a crucial library for CARTA and we don't want it to potentially produce incorrect results.

The error at build time is:
```
/root/carta-backend/third-party/fits2idia/Util.cc: In function 'void readFitsStringAttribute(fitsfile*, const string&, std::string&)':
/root/carta-backend/third-party/fits2idia/Util.cc:142:5: error: 'fits_read_string_key' was not declared in this scope; did you mean 'fits_read_key'?
  142 |     fits_read_string_key(filePtr, name.c_str(), 1, 255, strValueTmp, &strLen, NULL, &status);
      |     ^~~~~~~~~~~~~~~~~~~~
      |     fits_read_key
make[5]: *** [CMakeFiles/fits2idia.dir/Util.cc.o] Error 1
make[4]: *** [CMakeFiles/fits2idia.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [test/fits2idia-prefix/src/fits2idia-stamp/fits2idia-build] Error 2
make[1]: *** [test/CMakeFiles/fits2idia.dir/all] Error 2
make: *** [all] Error 2
```

It is a bit awkward to reproduce through the current [Dockerfile-CentOS7](https://github.com/CARTAvis/carta-backend/blob/dev/Dockerfiles/Dockerfile-CentOS7) as it requires some other updates (Now needs gtest 1.10, plus there appears to be a gcc problem). For now the fastest way may be:

1. Remove the lines 20-27 that install cfitsio 3.47 from source.
2. Add `cfitsio-devel` to the yum install command at line 16.
3. Instead of building the carta-backend, try building fits2idia converter directly e.g.
```
cd /root/carta-backend/third-party/fits2idia
mkdir build
cd build
cmake ..
make
```

confluence:
This partially fixes #940. In addition to this we need to install `fits2idia` separately as a dependency for the tests. Packages are already available for Ubuntu Bionic and Focal; the simplest thing to do on RHEL is probably to provide packages as well. We can then patch the RHEL7/CentOS7 package to use a newer (statically linked?) CFITSIO without affecting compilation of the backend.


ajm-ska:
I have updated all the Dockerfiles so that they are capable of building and running the unit-tests.

Dockerfile-CentOS7:
- Created a special el7 fits2idia RPM using a static libcfitsio.a 3.47 so that the carta-backend may continue to use the default cfitsio 3.37 system version.
- Created an el7 googletest 1.10 RPM so that we don't need to build it from source.
- Removed building carta-casacore from source and downloading the measures data with svn. Instead it now uses the cartavis-repo RPMS. This brings it more inline with the Ubuntu Dockerfiles (The only package remaining that is built from source is TBB, but that will be able to be removed soon).

Dockerfile-CentOS8:

- Created a normal el8 fits2idia RPM and el8 googletest 1.10 RPM and using carta-casacore and measures-data RPMs. Now no packages are built from source.
- Install newer numpy and astropy from pip3.
- Limit it to use CentOS 8.4.2105 incase it starts using CentOS stream. 

Dockerfile-ubuntu-18.04:

- Installs fits2idia from cartavis-team PPA.
- Installs newer googletest 1.10 from Xbmc PPA.
- Installs newer numpy and astropy from pip3.

Dockerfile-ubuntu-20.04:

- Installs fits2idia from cartavis-team Launchpad.

ajm-ska:
I wonder if we should mention the carta-casacore build instructions in the README.md now, just in case anyone still wishes to build it from source (plus the svn commands to get the latest measures-data)?

ajm-ska:
We also need to inform developers that they need to install the fits2idia dependency if running the unit tests.
On MacOS they can build and install it from source. It will be very simple.

Dockerfiles/Dockerfile-CentOS7
```diff
@@ -10,63 +10,22 @@ RUN \
...
-
-# Get libsofa.a for casacore
+# CARTA will have a problem finding the .h files in /usr/include/cfitsio/
 RUN \
   source /etc/bashrc && \
-  wget http://www.iausofa.org/2018_0130_F/sofa_f-20180130.tar.gz && \
-  tar xzf sofa_f-20180130.tar.gz && rm sofa_f-20180130.tar.gz && \
-  cd sofa/20180130/f77/src && make && cp libsofa.a /usr/lib/libsofa.a && \
-  cd /root && rm -rf sofa
+  cp /usr/include/cfitsio/* /usr/include/
```
veggiesaurus: Shouldn't this be solved by our use of find_package in the cmake file?

ajm-ska: That doesn't work for some reason.
The `find_package` seems to only find the .so files. 
The .so file is found properly at `/usr/lib64/libcfitsio.so`, but the include files are not found. 
For some reason on RHEL, the fitsio include files are located in a seemingly non-standard location of `/usr/include/cfitsio/` whereas on Ubuntu they are found in `/usr/include`.

I guess another solution would be to modify all relevant the carta-backend includes to also search for the standard RHEL /usr/include/cfitsio/ location?

Or alternatively we can just do `cmake .. -DCMAKE_CXX_FLAGS="-I/usr/include/cfitsio"`

ajm-ska: Or maybe add it into an INCLUDE_DIRECTORIES in CMakeLists.txt?

ajm-ska: @veggiesaurus Should we add `/usr/include/cfitsio` to the CMakeLists.txt. Maybe on this line?

> INCLUDE_DIRECTORIES(/opt/carta-casacore/include /opt/carta-casacore/include/casacore /opt/carta-casacore/include/casacode /usr/include/cfitsio)

ajm-ska:
I checked how casacore deals with the cfitsio include files. They have a part like this:
```
if (CFITSIO_FOUND)
    include_directories (${CFITSIO_INCLUDE_DIRS})
endif (CFITSIO_FOUND)
```

I see our CentOS7 and CentOS8 finds cfitsio with "pkg-config". I assume it will always find it with "pkg-config" so I put the ` include_directories (${CFITSIO_INCLUDE_DIRS})` line in the part of the "if" loop where pkg-config finds cfitsio.

Then I removed the manual copying of the cfitsio include file location from the CentOS7 and CentOS8 Dockerfiles and tested them. It all seems to work.

How about this solution?

ajm-ska (approved):
Jenkins won't be able to build the last commit because the server ran out of hard disk space again. When that happens, it gets stuck with the orange dot until a new commit is made. I'm sorry about that. It is a full time job to continuously monitor all the servers for issues like this.

But I think everything is OK with this commit and branch.

veggiesaurus (approved):

PR #942, opened 2021-10-18
Yuhsuan/895 hypercube types

YuHsuan-Hwang:
This PR is for the second part of #895:
1. Extended the types of hypercube:
apply a general check of the FITS stokes axis;
convert `CARTA::PolarizationType` to `casacore::Stokes`;
apply error handling for `stokesCoordinate()`.
2. Fixed issues when reading stokes info from the header:
allow `CTYPE` to be "Stokes" or "stokes" for CASA and FITS images;
get stokes indices using `CRPIX`, `CRVAL`, and `CDELT` to correctly read headers with `CRPIX != 1`;
convert FITS index to `CARTA::PolarizationType`.

requires frontend branch: `yuhsuan/1606_hypercube_types`


confluence (changes requested):
I would prefer to avoid putting anything in `Util` unnecessarily -- it should only be required if a function is used in multiple files which do not and should not depend on each other.

`GetCasaStokesType` is only used by `StokesFilesConnector`, so I think that it should be defined in that header (and if we do decide to put it in `Util`, for consistency it should go in `Util/Casacore`).

`ConvertFitsStokesValue` could be defined in the `FileLoader` header, since `StokesFilesConnector` imports it.

Apart from that, I think this all looks good.

YuHsuan-Hwang:
@confluence I moved the functions to `StokesFilesConnector.h` and `FileLoader.h`. Please check, thanks!

confluence (approved):

markccchiang (approved):

kswang1029 (approved):
this works well 👍

PR #946, opened 2021-10-20
additional error handling when checking for file existence

### LINKED ISSUE #945: Backend crashes on RedHat platforms when certain folders aren't accesible
The backend throws a `std::filesystem_error` exception and crashes when some folders aren't accessible on certain linux distributions (currently only reproducible on RHEL7) .

For example, if `/etc/carta` or `~/.carta` isn't accessible, the backend can crash with an error in the log:

```[critical] filesystem error: status: Permission denied```.

I suspect it has to do with not catching exceptions thrown from `std::filesystem::exists` calls.

### PR DESCRIPTION
**Pull request by veggiesaurus**
Closes #945 by adding additional error handling when calling `fs::exists`

### DIFF
--- a/src/Frame.cc
+++ b/src/Frame.cc
@@ -1841,7 +1841,7 @@ bool Frame::ExportCASAImage(casacore::ImageInterface<casacore::Float>& image, fs
 
     // Remove the old image file if it has a same file name
     std::error_code error_code;
-    if (fs::exists(output_filename, error_code) && !error_code) {
+    if (fs::exists(output_filename, error_code)) {
         fs::remove_all(output_filename);
     }
 
--- a/src/Frame.cc
+++ b/src/Frame.cc
@@ -1840,7 +1840,8 @@ bool Frame::ExportCASAImage(casacore::ImageInterface<casacore::Float>& image, fs
     bool success(false);
 
     // Remove the old image file if it has a same file name
-    if (fs::exists(output_filename)) {
+    std::error_code error_code;
+    if (fs::exists(output_filename, error_code) && !error_code) {
         fs::remove_all(output_filename);
     }
 
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -39,13 +39,15 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
     const fs::path system_settings_path = "/etc/carta/backend.json";
 
     json settings;
-    if (fs::exists(system_settings_path) && !no_system_config) {
+    std::error_code error_code;
+
+    if (!no_system_config && fs::exists(system_settings_path, error_code) && !error_code) {
         settings = JSONSettingsFromFile(system_settings_path.string());
         system_settings_json_exists = true;
         debug_msgs.push_back(fmt::format("Reading system settings from {}.", system_settings_path.string()));
     }
 
-    if (fs::exists(user_settings_path) && !no_user_config) {
+    if (!no_user_config && fs::exists(user_settings_path, error_code) && !error_code) {
         auto user_settings = JSONSettingsFromFile(user_settings_path.string());
         user_settings_json_exists = true;
         debug_msgs.push_back(fmt::format("Reading user settings from {}.", user_settings_path.string()));
@@ -61,6 +63,12 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
 json ProgramSettings::JSONSettingsFromFile(const std::string& json_file_path) {
     std::ifstream ifs(json_file_path, std::ifstream::in);
     json j;
+
+    if (!ifs.good()) {
+        warning_msgs.push_back(fmt::format("Error reading config file {}.", json_file_path));
+        return j;
+    }
+
     try {
         j = json::parse(ifs, nullptr, true, true);
     } catch (json::exception& err) {
@@ -269,8 +277,8 @@ end.
     no_browser = result["no_browser"].as<bool>();
     read_only_mode = result["read_only_mode"].as<bool>();
 
-    no_user_config = result.count("no_user_config") ? true : false;
-    no_system_config = result.count("no_system_config") ? true : false;
+    no_user_config = result.count("no_user_config") != 0;
+    no_system_config = result.count("no_system_config") != 0;
 
     applyOptionalArgument(top_level_folder, "root", result);
     // Override deprecated "root" argument
@@ -295,7 +303,8 @@ end.
 
     for (const auto& arg : positional_arguments) {
         fs::path p(arg);
-        if (fs::exists(p)) {
+        std::error_code error_code;
+        if (fs::exists(p, error_code) && !error_code) {
             if (fs::is_directory(p)) {
                 auto image_type = casacore::ImageOpener::imageType(p.string());
                 if (image_type == casacore::ImageOpener::AIPSPP || image_type == casacore::ImageOpener::MIRIAD ||
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -305,7 +305,7 @@ end.
         fs::path p(arg);
         std::error_code error_code;
         if (fs::exists(p, error_code)) {
-            if (fs::is_directory(p)) {
+            if (fs::is_directory(p, error_code)) {
                 auto image_type = casacore::ImageOpener::imageType(p.string());
                 if (image_type == casacore::ImageOpener::AIPSPP || image_type == casacore::ImageOpener::MIRIAD ||
                     image_type == casacore::ImageOpener::IMAGECONCAT || image_type == casacore::ImageOpener::IMAGEEXPR ||
@@ -317,7 +317,7 @@ end.
                     file_paths.clear();
                     break;
                 }
-            } else if (!fs::is_regular_file(p)) {
+            } else if (!fs::is_regular_file(p, error_code)) {
                 // Ignore invalid files
                 file_paths.clear();
                 break;
--- a/src/SessionManager/ProgramSettings.cc
+++ b/src/SessionManager/ProgramSettings.cc
@@ -41,13 +41,13 @@ ProgramSettings::ProgramSettings(int argc, char** argv) {
     json settings;
     std::error_code error_code;
 
-    if (!no_system_config && fs::exists(system_settings_path, error_code) && !error_code) {
+    if (!no_system_config && fs::exists(system_settings_path, error_code)) {
         settings = JSONSettingsFromFile(system_settings_path.string());
         system_settings_json_exists = true;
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: src/Util/App.cc [hunk 1]
```diff
@@ -64,14 +64,17 @@ std::string GetReleaseInformation() {
     // Unix solution just attempts to read from /etc/os-release. This works with Ubuntu, RedHat, CentOS, Arch, Debian and Fedora,
     // and should work on any system that has systemd installed
     fs::path path = "/etc/os-release";
+    std::error_code error_code;
 
-    if (fs::exists(path) && fs::is_regular_file(path)) {
+    if (fs::exists(path, error_code) && !error_code, fs::is_regular_file(path)) {
```
confluence: Should this be `&&` and not `,`?

src/Table/Table.cc
```diff
@@ -23,8 +23,8 @@ using namespace std;
 
 Table::Table(const string& filename, bool header_only) : _valid(false), _filename(filename), _num_rows(0), _available_rows(0) {
     fs::path file_path(filename);
-
-    if (!fs::exists(file_path)) {
+    std::error_code error_code;
+    if (!fs::exists(file_path, error_code) || error_code) {
         _parse_error_message = "File does not exist!";
```
confluence: Should we print a more specific message if there is an error code?

src/SessionManager/ProgramSettings.cc
```diff
@@ -269,8 +277,8 @@ end.
     no_browser = result["no_browser"].as<bool>();
     read_only_mode = result["read_only_mode"].as<bool>();
 
-    no_user_config = result.count("no_user_config") ? true : false;
-    no_system_config = result.count("no_system_config") ? true : false;
+    no_user_config = result.count("no_user_config") != 0;
+    no_system_config = result.count("no_system_config") != 0;
```
confluence, line 281: Can you assign an integer directly to boolean? If yes, why not just do that instead? Or cast to boolean?

src/Util/App.cc
veggiesaurus: how did it even compile with that typo? :see_no_evil:

kswang1029: Eyeballing compiler😆

veggiesaurus: huh, I'll be damned: https://en.wikipedia.org/wiki/Comma_operator

src/Table/Table.cc
veggiesaurus: I've just added `or cannot be read` here. Digging into the filesystem error codes would be beyond the scope, I think

src/SessionManager/ProgramSettings.cc
veggiesaurus: we use `!= 0` quite a bit in the codebase, so I figured I'd keep it in. The really unneccessary part was the `? true : false` bit, since it's basically saying `if (true) return true; else return false;`

confluence (approved):

src/Frame.cc
```diff
@@ -1840,7 +1840,8 @@ bool Frame::ExportCASAImage(casacore::ImageInterface<casacore::Float>& image, fs
     bool success(false);
 
     // Remove the old image file if it has a same file name
-    if (fs::exists(output_filename)) {
+    std::error_code error_code;
+    if (fs::exists(output_filename, error_code) && !error_code) {
```
jolopezl: Why do you check twice the result of `fs::exists`? The return value is `true` if the path exists and `false` if it does not. You can use `error_code` just in case something goes wrong. I suggest you change this line and use `error_code` afterward only to investigate what went wrong (a system-dependent error code). Do you really need it?

src/SessionManager/ProgramSettings.cc
jolopezl: I wouldn't say I like this change. IMO, The former version is more explicit (not that I originally wrote it...), while the second version is less friendly. Also, on the other hand, you can do:
```
    no_user_config = result.count("no_user_config");
    no_system_config = result.count("no_system_config");
```
With this, the code looks less complicated. It also avoids any comparison, which is a plus. 

In the end, all versions will look the same when compiled (thank you compiler), but reading is quite important for the future.

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -139,11 +146,10 @@ void SimpleFrontendServer::AddNoCacheHeaders(Res* res) {
...
-        return {{"version", 1}};
-    }
-
     try {
+        if (!fs::exists(preferences_path)) {
+            return {{"version", 1}};
+        }
         ifstream file(preferences_path.string());
         string json_string((std::istreambuf_iterator<char>(file)), std::istreambuf_iterator<char>());
         return json::parse(json_string);
```
jolopezl, line 155: Not entirely related to this pull request, but what's the purpose of the try-catch construct here if the exception isn't handled? Maybe something is missing like:
```
catch (exception e) {
        spdlog::warn(e.what());
        return {};
    }
```

src/Frame.cc
jolopezl: This comment propagates to the rest of the changes, for both cases:
```
fs::exists(p, ec) && !ec
```
and
```
!fs::exists(p, ec) || ec
```

jolopezl (changes requested):
I have a few comments. A couple of minor comments, and another more general across the changes regarding the use of `fs::exists`. Is the purpose of using `fs::exists(p, ec)` just using the overload of `fs::exists` that doesn't throw an exception?

src/Frame.cc
jolopezl: I suggest adding across the changes something like:
```
if (fs::exists(p, ec) {
    // ...
} else {
    spdlog::debug(fmt::format("failed with error code {}", ec));
}
```
or just ignore the error_code result.

veggiesaurus: I think we can just ignore the error code, you're right. It's really just there so that we can use the `no_except` version of the call

src/SimpleFrontendServer/SimpleFrontendServer.cc
veggiesaurus: good point, this was inconsistent. Fixed

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -76,14 +76,19 @@ void SimpleFrontendServer::HandleStaticRequest(Res* res, Req* req) {
     bool gzip_compressed = false;
     auto gzip_path = path;
     gzip_path += ".gz";
-    if (accepts_gzip && fs::exists(gzip_path) && fs::is_regular_file(gzip_path)) {
+    std::error_code error_code;
+    if (accepts_gzip && fs::exists(gzip_path, error_code) && fs::is_regular_file(gzip_path)) {
```
jolopezl: Here `fs::is_regular_file(gzip_path)` should also be `fs::is_regular_file(gzip_path, error_code)` so that it doesn't throw. It has almost the same signatures as `fs::status` (https://en.cppreference.com/w/cpp/filesystem/is_regular_file).

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -76,14 +76,19 @@ void SimpleFrontendServer::HandleStaticRequest(Res* res, Req* req) {
...
         gzip_compressed = true;
         path = gzip_path;
     }
 
-    if (fs::exists(path) && fs::is_regular_file(path)) {
+    if (fs::exists(path, error_code) && fs::is_regular_file(path)) {
```
jolopezl: Idem for `fs::is_regular_file`, and in more places in the rest of the changes.

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -113,13 +118,15 @@ void SimpleFrontendServer::HandleStaticRequest(Res* res, Req* req) {
 }
 
 bool SimpleFrontendServer::IsValidFrontendFolder(fs::path folder) {
+    std::error_code error_code;
+
     // Check that the folder exists
-    if (!fs::exists(folder) || !fs::is_directory(folder)) {
+    if (!fs::exists(folder, error_code) || !fs::is_directory(folder)) {
```
jolopezl: Idem for `fs:: is_directory` (https://en.cppreference.com/w/cpp/filesystem/is_directory),  and in more places in the rest of the changes.

src/SimpleFrontendServer/SimpleFrontendServer.cc
jolopezl: Great!, but now I got a bit confused with this try-catch construct. 
Right now, `fs::exists` and `json::parse` can throw. Shouldn't we try to catch both exceptions? So, I think the previous position of `fs::exists` was fine, and you can use the `noexcept` version of it, but the try-catch construct should be changed a little bit to catch the json parse error: ` catch(json::parse_error& e) spdlog::warn(e.what());`

src/SimpleFrontendServer/SimpleFrontendServer.cc
```diff
@@ -359,7 +366,9 @@ void SimpleFrontendServer::HandleClearObject(const std::string& object_type, Res
 nlohmann::json SimpleFrontendServer::GetExistingObjects(const std::string& object_type) {
     auto object_folder = _config_folder / (object_type + "s");
     json objects = json::object();
-    if (fs::exists(object_folder)) {
+    std::error_code error_code;
+
+    if (fs::exists(object_folder, error_code)) {
         for (auto& p : fs::directory_iterator(object_folder)) {
             try {
                 string filename = p.path().filename().string();
```
jolopezl, line 374: Following up on my last thought. Two lines below here, the try-catch construct will need to catch two exceptions. Maybe it's better to change `fs::is_regular_file(p)` to `fs::is_regular_file(p, error_code)` and just try to catch the possible exception from `json::parse`. What do you think?

jolopezl (changes requested):

confluence:
I agree that we should be consistent about either using the functions that throw or the functions that do not.

src/SimpleFrontendServer/SimpleFrontendServer.cc
veggiesaurus: good point. Have done the same for `is_directory`

src/SimpleFrontendServer/SimpleFrontendServer.cc
veggiesaurus: have fixed this too :+1:

veggiesaurus:
> I agree that we should be consistent about either using the functions that throw or the functions that do not.

I have changed them all, except for one: `SimpleFrontendServer::ClearObjectFromString` has calls in a try-catch and we actually _want_ to get hold of the exception information in this case, so I've left it is.

confluence (approved):

jolopezl (approved):
Thanks for addressing the comments!. Looks great.

Issue #433, opened 2020-02-05
Support PI (polarized intensity) and PA (polarized angle) request from the frontend


kswang1029:
Currently the backend can supply profiles and images of any Stokes parameter. We need to extend this to PI (polarized intensity) and PA (polarized angle) so the backend will use Q and U profiles to derived PI and PA profiles. Fractional quantities should be supported too.

This includes:
- [x] profiles
- [x] contours
- [x] raster image

ref: https://github.com/CARTAvis/carta-frontend/issues/713

markccchiang:
Is it similar to the way in image moments generation? First, the frontend sends requirements to the backend, generating polarized intensity, fractional polarized intensity, or polarized angle raster images. Then, the backend opens these generated raster images and sends open files acks to the frontend. Using these raster images, we can do profiles or contours analysis.

PR #948, opened 2021-10-26
Support new polarization types

markccchiang:
#433. This is the trial to support the new five polarization types: total polarized intensity, fractional total polarized intensity, linear polarized intensity, fractional linear polarized intensity, and polarized angle. The raster images of these polarization types are calculated when required from the frontend. The spatial/spectral profiles, region statistics, and histograms for these images are also supported. 

A new struct `StokesSource` is used to get the original image interface or a calculated polarization image interface from the `FileLoader`. It binds with a specific slicer or a lattice region. Polarization calculations are triggered when the frontend requests specific stokes indices (13, 14, 15, 16, 17) for raster images or stokes configs ("Ptotal", "PFtotal", "Plinear", "PFlinear", "Pangle") + ("x", "y", "z") for spatial/spectral profiles. Unit tests of these functionalities are provided. But it also needs to test with the modified frontend.


kswang1029:
@markccchiang @TienHao I briefly tested `mark/433_compute_polarization_types` + `thchang/support_computed_polarizations_714` and found the following issues:

1. which we select a computed image (all the P*) the histogram in the render config widget is empty so that the raster image is not rendered properly (clip min/max as 0/1). This happens when the input cube has the channel axis (ie 4D cube as spatial-spatial-channel-stokes). 
2. the polarization slider label gets crowded with those extra computed images. Is it possible to rotate the label when the space is not enough?
Screen Shot 2022-03-29 at 14 11 53

Alternatively, we can try to use acronyms and use tool tip for full name or show text label to the right of the slider (similar to the channel slider) as 
- TPI: total polarized intensity
- LPI: linear polarized intensity
- FTPI: fractional total polarized intensity
- FLPI: fractional linear polarized intensity
- PA: polarization angle
We might need to reduce Stokes I/Q/U/V to just I/Q/U/V but show Stokes I/Q/U/V either in tooltip or text label next to the polarization slider. 

3. when it is Pangle selected, the image pixel unit and profile unit should be "degree" (not Jy/beam, not radian).

kswang1029:
@markccchiang could you check the unit of the computed polarized angle? Based on my test, it is about from -1.56 to +1.56 or -pi/2 to +pi/2. We need the angle to be in degree not in radian.

kswang1029:
@markccchiang @TienHao when we animate the computed quantities, in the forward mode we see unexpected behavior after the last channel is reached. It should loop back to the first channel _using the same computed quantities_. It _may_ hang at the last channel with the following debug message from the backend:

> [2022-04-05 11:56:38.899] [CARTA] [debug] Session 972574378: Channel 35 or Stokes 162693452 is invalid in image

  Please refer to the following movie for the issue. 


https://user-images.githubusercontent.com/20819712/161676186-87ec8660-6e00-4787-b5f0-dc73278e8837.mov

kswang1029:
@markccchiang @TienHao When we select PFtotal and PFlinear, the displayed unit should be "%". At the moment, the computed value lacks a multiplier of 100 (eg, 0.23, it should be 23%).

kswang1029:
@markccchiang Using a test Stokes 4D cube, I am seeing a weird issue that I cannot explain. There are consistent parts but not always.

If we request a cursor or point region Pangle profile using the Z profiler, we see a consistent profile as computed in the Stokes widget (the PA panel).
Screen Shot 2022-04-06 at 10 51 51

However, if we make a region and compare the two profiles again, we see inconsistent results. I have tried to use the region Q and U profile to compute Pangle profile manually using the formula 0.5*atan2(U/Q) and can get the profile as in the green box. Weirdly, if I use casa-immath with mode='pola' I can get the exact profile as in the magenta box. 
Screen Shot 2022-04-06 at 10 52 27

I think the profile in the green box is the true solution as I am using a test Stokes 4D cube with known distribution of Pangle along the spectral axis. Not sure what's happening with casa/casacore routines. 🤔

kswang1029:
@markccchiang I found an issue that I cannot find a robust way to reproduce so please just consider this is a note only until I can find a good way to reproduce the problem. 

Once I saw the histogram in the render configure widget has an unexpected distribution with just two spikes near the two ends. This happens when I was seeing Pangle of certain channels. The image appears abnormal with just two values.

kswang1029:
@TienHao @markccchiang It seems that if there are other matched image (**UPDATE: we don't need a matched image actually. Region histogram of a computed image cannot be displayed.**), the histogram widget cannot display a region histogram when the computed polarization image is selected. cf the attached screenshot. The image on the left is a Stokes 4D cube and the image on the right is a 4D cube with the 4th dimension of length 1. 
Screen Shot 2022-04-06 at 11 17 31

kswang1029:
> @markccchiang Using a test Stokes 4D cube, I am seeing a weird issue that I cannot explain. There are consistent parts but not always.
> 
> If we request a cursor or point region Pangle profile using the Z profiler, we see a consistent profile as computed in the Stokes widget (the PA panel). Screen Shot 2022-04-06 at 10 51 51
> 
> However, if we make a region and compare the two profiles again, we see inconsistent results. I have tried to use the region Q and U profile to compute Pangle profile manually using the formula 0.5*atan2(U/Q) and can get the profile as in the green box. Weirdly, if I use casa-immath with mode='pola' I can get the exact profile as in the magenta box. Screen Shot 2022-04-06 at 10 52 27
> 
> I think the profile in the green box is the true solution as I am using a test Stokes 4D cube with known distribution of Pangle along the spectral axis. Not sure what's happening with casa/casacore routines. 🤔

@markccchiang I have sorted this out. The inconsistent results came from user error (me!). The two profiles are computed differently (the order between mean and division). So this is NOT an issue.

kswang1029:
Regarding region spectral profiler when a computed polarization quantity is selected, we _cannot_ treat the computed polarization image cube as a regular cube to extract region spectral profile. Instead, we will need to extract region spectral profiles from Stokes IQUV _first_ then use those profiles to compute the selected polarization quantity.

kswang1029:
@markccchiang This is probably the last small issue I can spot. If we generate moment or pv images from a computed component (such as PFlinear, Pangle), the BUNIT header entry of the computed moment or pv image is not correct. For PFtotal and PF linear, it should be %. For Pangle it should be degree. Ptotal and Plinear have the same unit as the input cube.

kswang1029 (approved):
This is a great enhancement of the usability about polarization data. Thanks @markccchiang @TienHao  👍

src/Frame/Frame.cc
```diff
@@ -342,10 +374,11 @@ bool Frame::FillImageCache() {
     }
 
     auto t_start_set_image_cache = std::chrono::high_resolution_clock::now();
-    casacore::Slicer section = GetImageSlicer(AxisRange(_z_index), _stokes_index);
+    casacore::Slicer section = GetImageSlicer(AxisRange(_z_index), _stokes_index).second;
     _image_cache_size = section.length().product();
     _image_cache = std::make_unique<float[]>(_image_cache_size);
-    if (!GetSlicerData(section, _image_cache.get())) {
+    std::pair<StokesSrc, casacore::Slicer> stokes_slicer = GetImageSlicer(AxisRange(_z_index), _stokes_index);
```
pford: Why is GetImageSlicer being called twice?

src/Frame/Frame.cc
```diff
@@ -1116,10 +1149,12 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
         if (is_current_stokes) {
             cursor_value = cursor_value_with_current_stokes;
         } else {
-            casacore::Slicer section = GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes);
+            casacore::Slicer section = GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes).second;
             const auto N = section.length().product();
-            std::unique_ptr<float[]> data(new float[N]); // zero initialization
-            if (GetSlicerData(section, data.get())) {
+            std::pair<StokesSrc, casacore::Slicer> stokes_slicer =
+                GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes);
```
pford: Same here, GetImageSlicer twice with same arguments

src/Frame/Frame.h
```diff
@@ -307,6 +310,7 @@ class Frame {
 
     // Moment generator
     std::unique_ptr<MomentGenerator> _moment_generator;
+    StokesSrc _stokes_src;
```
pford: Why is this a class member, it is only used internally in Frame::CalculateMoments

src/ImageData/FileLoader.cc
```diff
@@ -864,6 +886,35 @@ bool FileLoader::GetStokesTypeIndex(const CARTA::PolarizationType& stokes_type,
     return false;
 }
 
+carta::CasaLoader::ImageRef FileLoader::GetStokesImage(const StokesSrc& stokes_src) {
```
pford: Why is this CasaLoader::ImageRef not FileLoader::ImageRef?  I was expecting a special typedef in CasaLoader but there is none.

pford (changes requested):
Minor changes requested

Also please update carta-backend changelog and protobuf changelog

pford (approved):

src/Util/Message.h
```diff
@@ -75,6 +76,7 @@ class Message {
     static CARTA::ImageProperties ImageProperties(std::string directory, std::string file, std::string hdu, int32_t file_id,
         CARTA::RenderMode render_mode, int32_t channel, int32_t stokes);
     static CARTA::ResumeSession ResumeSession(std::vector<CARTA::ImageProperties> images);
+    static CARTA::SetSpectralRequirements_SpectralConfig SpectralConfig(const std::string& coordinate);
```
jolopezl, line 79: Is this a bit out of the naming conventions? You might have a good argument to keep it like this.

jolopezl, line 79: Is this a bit out of the naming conventions? You might have a good argument to keep it like this.

test/CommonTestUtilities.cc
```diff
@@ -254,3 +397,14 @@ void CmpValues(float data1, float data2, float abs_err) {
...
+void CmpHistograms(const carta::Histogram& cube_histogram1, const carta::Histogram& cube_histogram2) {
+    auto hist1 = cube_histogram1.GetHistogramBins();
+    auto hist2 = cube_histogram2.GetHistogramBins();
+    EXPECT_EQ(hist1.size(), hist2.size());
+    if (hist1.size() == hist2.size()) {
+        for (int i = 0; i < hist1.size(); ++i) {
+            EXPECT_EQ(hist1[i], hist2[i]);
+        }
+    }
+}
```
jolopezl: I am ashamed that I didn't put an operator overload in Histogram for this. That said, I think that the comparison needs to be stricter (if needed). See: https://github.com/CARTAvis/carta-backend/blob/edc2561dbc5878d219e4dc77d5c6128edd6809f3/test/TestHistogram.cc#L31-L51.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -0,0 +1,266 @@
...
+casacore::LatticeExprNode PolarizationCalculator::MakeTotalPolarizedIntensityNode() {
+    casacore::LatticeExprNode lin_node = casacore::LatticeExprNode(
+        casacore::pow(*_stokes_images[V], 2) + casacore::pow(*_stokes_images[U], 2) + casacore::pow(*_stokes_images[Q], 2));
+    return casacore::sqrt(lin_node);
+}
+
+casacore::LatticeExprNode PolarizationCalculator::MakePolarizedIntensityNode() {
+    casacore::LatticeExprNode lin_node =
+        casacore::LatticeExprNode(casacore::pow(*_stokes_images[U], 2) + casacore::pow(*_stokes_images[Q], 2));
+    return casacore::sqrt(lin_node);
```
jolopezl, line 128: I'd prefer to use `std::pow`, `std::sqrt`. Is there any reason not to use STL?

src/Util/Image.h
```diff
@@ -108,20 +119,63 @@ struct PointXy {
...
 CARTA::PolarizationType GetStokesType(int stokes_value);
+bool ComputedStokes(int stokes);
+bool ComputedStokes(const std::string& stokes_type);
+
+// The struct StokesSrc is used to tell the file loader to get the original image interface, or get the computed stokes image interface.
+// The x, y, and z ranges from the StokesSrc indicate the range of image data to be calculated (for the new stokes type image).
+// We usually don't want to calculate the whole image data, because it spends a lot of time.
+// StokesSrc will bind casacore::Slicer or casacore::ImageRegion, because the coordinate of a computed stokes image is different from
+// the original image coordinate.
+struct StokesSrc {
```
jolopezl: I think the full name improves readability: `StokesSource`. Shorter name are more variable-like:
```
src = StokesSource();
```

jolopezl (changes requested):
Large pull request. I have a few minor suggestions.

src/ImageData/PolarizationCalculator.cc
markccchiang: It seems `std::pow` and `std::sqrt` can not process the variable type `casacore::LatticeExprNode`. So I did not change this part of the code.

src/Util/Message.h
```diff
@@ -75,6 +76,7 @@ class Message {
     static CARTA::ImageProperties ImageProperties(std::string directory, std::string file, std::string hdu, int32_t file_id,
         CARTA::RenderMode render_mode, int32_t channel, int32_t stokes);
     static CARTA::ResumeSession ResumeSession(std::vector<CARTA::ImageProperties> images);
+    static CARTA::SetSpectralRequirements_SpectralConfig SetSpectralRequirements_SpectralConfig(const std::string& coordinate);
```
jolopezl: My previous comment was misleading and I was looking at the wrong part of the code. I was thinking that:
```
CARTA::SetSpectralRequirements_SpectralConfig
```
should be:
```
CARTA::SetSpectralRequirementsSpectralConfig
```
To follow the naming conventions. I think you could revert this change if you want. A function name like `SetSpectralConfig` is totally fine.

jolopezl (approved):
Looks good. Now I am just amending a comment that was misleading, I think.

src/Frame/Frame.cc
```diff
@@ -132,18 +132,24 @@ std::string Frame::GetFileName() {
     return filename;
 }
 
-std::shared_ptr<casacore::CoordinateSystem> Frame::CoordinateSystem() {
+std::shared_ptr<casacore::CoordinateSystem> Frame::CoordinateSystem(const StokesSource& stokes_src) {
```
confluence: I have a slight preference for renaming this variable from `stokes_src` to `stokes_source`.

src/Util/Image.h
```diff
@@ -108,20 +119,63 @@ struct PointXy {
...
     {"LR", CARTA::PolarizationType::LR}, {"XX", CARTA::PolarizationType::XX}, {"YY", CARTA::PolarizationType::YY},
-    {"XY", CARTA::PolarizationType::XY}, {"YX", CARTA::PolarizationType::YX}};
+    {"XY", CARTA::PolarizationType::XY}, {"YX", CARTA::PolarizationType::YX}, {"Ptotal", CARTA::PolarizationType::Ptotal},
+    {"Plinear", CARTA::PolarizationType::Plinear}, {"PFtotal", CARTA::PolarizationType::PFtotal},
+    {"PFlinear", CARTA::PolarizationType::PFlinear}, {"Pangle", CARTA::PolarizationType::Pangle}};
 
 int GetStokesValue(const CARTA::PolarizationType& stokes_type);
 CARTA::PolarizationType GetStokesType(int stokes_value);
+bool ComputedStokes(int stokes);
+bool ComputedStokes(const std::string& stokes_type);
```
confluence: These should be renamed to `IsComputedStokes`.

src/Util/Image.h
```diff
@@ -108,20 +119,63 @@ struct PointXy {
...
+    AxisRange y_range;
+
+    StokesSource() : stokes(-1), z_range(AxisRange(ALL_Z)), x_range(AxisRange(ALL_X)), y_range(AxisRange(ALL_Y)) {}
+
+    StokesSource(int stokes_, AxisRange z_range_) : stokes(stokes_), z_range(z_range_), x_range(ALL_X), y_range(ALL_Y) {}
+
+    StokesSource(int stokes_, AxisRange z_range_, AxisRange x_range_, AxisRange y_range_)
+        : stokes(stokes_), z_range(z_range_), x_range(x_range_), y_range(y_range_) {}
+
+    bool OriginalImage() const {
```
confluence: This should be renamed to `IsOriginalImage`.

src/Util/Image.cc
```diff
@@ -21,3 +21,13 @@ CARTA::PolarizationType GetStokesType(int stokes_value) {
...
+
+bool ComputedStokes(int stokes) {
+    return ((stokes == COMPUTE_STOKES_PTOTAL) || (stokes == COMPUTE_STOKES_PLINEAR) || (stokes == COMPUTE_STOKES_PFTOTAL) ||
+            (stokes == COMPUTE_STOKES_PFLINEAR) || (stokes == COMPUTE_STOKES_PANGLE));
+}
+
+bool ComputedStokes(const std::string& stokes_type) {
+    return ((stokes_type == "Ptotal") || (stokes_type == "Plinear") || (stokes_type == "PFtotal") || (stokes_type == "PFlinear") ||
+            (stokes_type == "Pangle"));
+}
```
confluence: You can avoid redundancy here by using the mappings defined in the header to convert the string type name to the `CARTA:PolarizationType` and that to the integer value, and then using the integer function above.

src/Util/Image.cc
```diff
@@ -21,3 +21,13 @@ CARTA::PolarizationType GetStokesType(int stokes_value) {
     }
     return stokes_type;
 }
+
+bool ComputedStokes(int stokes) {
+    return ((stokes == COMPUTE_STOKES_PTOTAL) || (stokes == COMPUTE_STOKES_PLINEAR) || (stokes == COMPUTE_STOKES_PFTOTAL) ||
+            (stokes == COMPUTE_STOKES_PFLINEAR) || (stokes == COMPUTE_STOKES_PANGLE));
+}
```
confluence: Because these are sequential integer values, I think it would be cleaner and more efficient to check if the value falls within the integer range.

src/Frame/Frame.cc
```diff
@@ -1572,41 +1608,44 @@ bool Frame::HasSpectralConfig(const SpectralConfig& config) {
...
-bool Frame::GetImageRegion(int file_id, const AxisRange& z_range, int stokes, casacore::ImageRegion& image_region) {
+bool Frame::GetImageRegion(
+    int file_id, const AxisRange& z_range, int stokes, std::pair<StokesSource, casacore::ImageRegion>& stokes_region) {
     if (!CheckZ(z_range.from) || !CheckZ(z_range.to) || !CheckStokes(stokes)) {
         return false;
     }
     try {
-        casacore::Slicer slicer = GetImageSlicer(z_range, stokes);
-        casacore::LCSlicer lcslicer(slicer);
+        std::pair<StokesSource, casacore::Slicer> stokes_slicer = GetImageSlicer(z_range, stokes);
```
confluence: Given how frequently this pair is used, perhaps it would be clearer to create a `StokesSlicer` struct with these elements as members instead.

confluence: Perhaps the same should be done for `std::pair<StokesSource, casacore::LattRegionHolder>`? If so, would we also need a separate struct for `std::pair<StokesSource, casacore::ImageRegion>`, or could we cast `LattRegionHolder` to `ImageRegion` where required?

confluence: This is not strictly necessary, but I prefer structs with named members to bare pairs / tuples, especially when we have several similar pairs in use in the same area of the code.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -0,0 +1,266 @@
...
+
+std::shared_ptr<casacore::ImageInterface<float>> PolarizationCalculator::ComputeTotalPolarizedIntensity() {
+    if (!_image_valid) {
+        return nullptr;
+    }
+
+    auto q = GetStokesImage(Q);
+    auto u = GetStokesImage(U);
+    auto v = GetStokesImage(V);
+    if (!q || !u || !v) {
```
confluence: These temporary variables should be eliminated and the function should be called directly from the conditional. This is bypassing the short-circuited `||` -- all of the Stokes images are looked up every time, even if e.g. `q` is null and the conditional exits immediately.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -0,0 +1,266 @@
...
+std::shared_ptr<casacore::ImageInterface<float>> PolarizationCalculator::ComputeTotalFractionalPolarizedIntensity() {
+    if (!_image_valid) {
+        return nullptr;
+    }
+
+    auto i = GetStokesImage(I);
+    auto q = GetStokesImage(Q);
+    auto u = GetStokesImage(U);
+    auto v = GetStokesImage(V);
+    if (!i || !q || !u || !v) {
```
confluence: As above.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -0,0 +1,266 @@
...
+}
+
+std::shared_ptr<casacore::ImageInterface<float>> PolarizationCalculator::ComputePolarizedIntensity() {
+    if (!_image_valid) {
+        return nullptr;
+    }
+
+    auto q = GetStokesImage(Q);
+    auto u = GetStokesImage(U);
+    if (!q || !u) {
```
confluence: As above.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -0,0 +1,266 @@
...
+
+std::shared_ptr<casacore::ImageInterface<float>> PolarizationCalculator::ComputeFractionalPolarizedIntensity() {
+    if (!_image_valid) {
+        return nullptr;
+    }
+
+    auto i = GetStokesImage(I);
+    auto q = GetStokesImage(Q);
+    auto u = GetStokesImage(U);
+    if (!i || !q || !u) {
```
confluence: As above.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -0,0 +1,266 @@
...
+}
+
+std::shared_ptr<casacore::ImageInterface<float>> PolarizationCalculator::ComputePolarizedAngle() {
+    if (!_image_valid) {
+        return nullptr;
+    }
+
+    auto q = GetStokesImage(Q);
+    auto u = GetStokesImage(U);
+    if (!q || !u) {
```
confluence: As above.

src/Region/RegionHandler.cc
```diff
@@ -1256,16 +1345,55 @@ bool RegionHandler::GetRegionSpectralData(int region_id, int file_id, std::strin
...
+                !get_partial_profiles(profiles_u, "Uz")) {
+                return false;
+            }
+            GetStokesPflinear(profiles_i, profiles_q, profiles_u, partial_profiles);
+        } else if (stokes_index == COMPUTE_STOKES_PANGLE) {
+            if (!get_partial_profiles(profiles_q, "Qz") || !get_partial_profiles(profiles_u, "Uz")) {
+                return false;
+            }
+            GetStokesPangle(profiles_q, profiles_u, partial_profiles);
+        } else {
```
confluence: This is identical to the block above except for the lambda used. It could be factored out into a lambda or helper function.

src/Region/RegionHandler.cc
```diff
@@ -2154,4 +2282,168 @@ casacore::Vector<float> RegionHandler::GetTemporaryRegionProfile(
...
+
+    for (auto stats_q : profiles_q) {
+        for (auto stats_u : profiles_u) {
+            if (stats_q.first == stats_u.first) {
+                std::vector<double>& results = profiles_ptotal[stats_q.first];
+                results.resize(stats_q.second.size());
+                std::transform(stats_q.second.begin(), stats_q.second.end(), stats_u.second.begin(), results.begin(), calc_step1);
+            }
+        }
+    }
```
confluence: This block is frequently repeated with different objects (and different lambdas which have the same signatures), and could be factored out into a helper function or lambda.

src/Region/RegionHandler.cc
```diff
@@ -2154,4 +2282,168 @@ casacore::Vector<float> RegionHandler::GetTemporaryRegionProfile(
...
+        }
+        return std::numeric_limits<double>::quiet_NaN();
+    };
+
+    auto calc_step3 = [&](double step2, double i) {
+        if (!std::isnan(step2) && !isnan(i)) {
+            return 100.0 * (step2 / i);
+        }
+        return std::numeric_limits<double>::quiet_NaN();
+    };
```
confluence: Some of these lambdas are duplicated across the different functions. A single set of unique functions could be perhaps be defined outside these functions to avoid duplication. This is also not absolutely necessary, but it may improve legibility of these calculations.

I wonder if it would make sense to create a single lambda to do the `isnan` check and call another lambda -- that would reduce the inner calculation lambdas to one-liners, which could then be defined much less verbosely (even if the definitions are repeated inside these functions).

confluence (changes requested):

confluence:
Please treat the lambda refactoring suggestions as optional.

src/Region/RegionHandler.cc
```diff
@@ -2154,4 +2282,168 @@ casacore::Vector<float> RegionHandler::GetTemporaryRegionProfile(
     return profile;
 }
 
+void RegionHandler::GetStokesPtotal(
+    const ProfilesMap& profiles_q, const ProfilesMap& profiles_u, const ProfilesMap& profiles_v, ProfilesMap& profiles_ptotal) {
+    auto calc_step1 = [&](double q, double u) {
+        if (!std::isnan(q) && !isnan(u)) {
```
confluence: One more thing I just noticed: why is the second `isnan` not a `std::isnan` in all these functions?

test/TestPolarizationCalculator.cc
```diff
@@ -0,0 +1,825 @@
+/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
```
confluence: As in the case of #1016, I would like to split out this test code into a separate PR.

markccchiang:
Todo:
- [x]  Refactor `RegionHandler::GetRegionSpectralData` function
- [x] Move testing file `TestPolarizationCalculator.cc` to a separate PR
- [x] Solve merge conflicts

src/Util/Image.cc
```diff
@@ -22,12 +22,6 @@ CARTA::PolarizationType GetStokesType(int stokes_value) {
     return stokes_type;
 }
 
-bool ComputedStokes(int stokes) {
-    return ((stokes == COMPUTE_STOKES_PTOTAL) || (stokes == COMPUTE_STOKES_PLINEAR) || (stokes == COMPUTE_STOKES_PFTOTAL) ||
-            (stokes == COMPUTE_STOKES_PFLINEAR) || (stokes == COMPUTE_STOKES_PANGLE));
-}
-
-bool ComputedStokes(const std::string& stokes_type) {
```
confluence: I was not suggesting that you remove this function, but that you change the implementation to look up the integer index and then call the function which takes an integer index.

src/Frame/Frame.cc
```diff
@@ -2179,7 +2176,7 @@ bool Frame::GetStokesTypeIndex(const string& coordinate, int& stokes_index) {
             CARTA::PolarizationType stokes_type = StokesStringTypes[stokes_string];
             if (_loader->GetStokesTypeIndex(stokes_type, stokes_index)) {
                 stokes_ok = true;
-            } else if (ComputedStokes(stokes_string)) {
+            } else if (IsComputedStokes(StokesValues[StokesStringTypes[stokes_string]])) {
```
confluence: See other comment about the `IsComputedStokes` function; my suggestion is that you keep the string function to use here and make this the internal implementation.

src/ImageData/PolarizationCalculator.cc
```diff
@@ -154,10 +154,7 @@ std::shared_ptr<casacore::ImageInterface<float>> PolarizationCalculator::Compute
         return nullptr;
     }
 
-    auto q = GetStokesImage(Q);
-    auto u = GetStokesImage(U);
-    auto v = GetStokesImage(V);
-    if (!q || !u || !v) {
+    if (!GetStokesImage(Q) || !GetStokesImage(U) || !GetStokesImage(V)) {
```
confluence: Since these functions have access to the private class members, why not use `_stokes_images[Q]`, etc., here? This would be consistent with the other functions in the class.

src/Region/RegionHandler.cc
```diff
@@ -2285,165 +2285,87 @@ casacore::Vector<float> RegionHandler::GetTemporaryRegionProfile(
 void RegionHandler::GetStokesPtotal(
     const ProfilesMap& profiles_q, const ProfilesMap& profiles_u, const ProfilesMap& profiles_v, ProfilesMap& profiles_ptotal) {
     auto calc_step1 = [&](double q, double u) {
-        if (!std::isnan(q) && !isnan(u)) {
-            return (pow(q, 2) + pow(u, 2));
-        }
-        return std::numeric_limits<double>::quiet_NaN();
+        return (IsValid(q, u) ? (std::pow(q, 2) + std::pow(u, 2)) : std::numeric_limits<double>::quiet_NaN());
```
confluence: I was suggesting eliminating the repeated check entirely. I think that you could do that by creating a function that takes a calculation function as a parameter and returns a function with the same signature which does the check and then either returns the result of the original function or NaN. But that may be too convoluted, and this already looks much more legible, so I'm happy to leave it as it is.

confluence: Actually, I think that you can do this quite cleanly by defining a new lambda inside your `CombineStokes` function. Instead of using `func` directly, you could define a `func_if_valid` which does the check and either calls `func` on the operands or returns NaN. Then you could use `func_if_valid` in the transform, and you could strip out the checks from all the original lambdas so that they contain only the unique calculation code.

kswang1029:
@confluence I requested myself as a reviewer. Once code review and refactoring are done I will have one final quick test.

confluence (approved):
I think this is also good to go once @kswang1029 does a final check.

kswang1029 (approved):
Rock n' Roll~ 👍 @markccchiang this is a very nice new feature enabling new use cases!

kswang1029:
@confluence @YuHsuan-Hwang this backend PR and the companion frontend PR (and protobuf PR) are ready to merge. Thank you @markccchiang @TienHao !

PR #949, opened 2021-10-26
Fix TestProgramSettings.cc in CentOS7

jolopezl:
@ajm-asiaa, this may fix the problem observed in CentOS7 with devtoolset-8 (GCC 8.3.1).


ajm-ska (approved):
@jolopezl Thank you for fixing the problem! I tried it in a CentOS7 Docker container and it now builds successfully. In addition, Mix 'N Match automatically succeeded and generated the code coverage, so it is working there too. Also the CI worked fine for Ubuntu and MacOS, so everything seems to be perfect now.

veggiesaurus (approved):

PR #951, opened 2021-10-27
unicode XML string columns

### LINKED ISSUE #950: votable catalog column with unicodeChar is not recognized
as reported by a user.

[carta_votable_test.xml.zip](https://github.com/CARTAvis/carta-backend/files/7422901/carta_votable_test.xml.zip)

all formats in https://www.ivoa.net/documents/VOTable/20191021/REC-VOTable-1.4-20191021.pdf Section 2.1

veggiesaurus:
closes #950 by treating `unicodeChar` columns as `char` columns. Seems to work ok

### DIFF
--- a/src/Table/Columns.cc
+++ b/src/Table/Columns.cc
@@ -32,7 +32,7 @@ std::unique_ptr<Column> Column::FromField(const pugi::xml_node& field) {
 
     unique_ptr<Column> column;
 
-    if (type_string == "char") {
+    if (type_string == "char" || type_string == "unicodeChar") {
         column = make_unique<DataColumn<string>>(name);
     } else if (!array_size_string.empty()) {
         // Can't support array-based column types other than char


veggiesaurus:
not sure why the `close_file` ICD test is failing :thinking: @ajm-asiaa how to I view the backend log for that stage?

ajm-ska:
@veggiesaurus It is only the annoying glitch where some tests randomly fail for mysterious reasons. I tried running the test manually and it passes fine, so it should work. I'm re-running the tests now in a 2nd run, but looks like my manually testing accidentally caused some trouble and other tests failed! Once it finishes I'll restart again and the 3rd run should be OK.

veggiesaurus:
> @veggiesaurus It is only the annoying glitch where some tests randomly fail for mysterious reasons. I tried running the test manually and it passes fine, so it should work. I'm re-running the tests now in a 2nd run, but looks like my manually testing accidentally caused some trouble and other tests failed! Once it finishes I'll restart again and the 3rd run should be OK.

It looks to me like a backend crash though. Can we store the backend log in a unique location for each stage?

ajm-ska:
@veggiesaurus To answer your question, unfortunately you can't see the backend log at the moment, but hopefully you will with the new system that is coming soon. In the new Jenkinsfile I have put in a section where if a test fails, it will show the backend log. e.g.

```
            retry(3) {
                if (ret) {
                    sleep(time:30,unit:"SECONDS")
                    sh "cat /root/.carta/log/carta.log"
```

ajm-ska:
@veggiesaurus 3rd time lucky! It passed all the tests.

veggiesaurus:
> @veggiesaurus 3rd time lucky! It passed all the tests.

that in itself is worrying... because we can't really tell whether its something in the branch or something in the CI that is going wrong

ajm-ska:
> that in itself is worrying... because we can't really tell whether its something in the branch or something in the CI that is going wrong

That's right. It has been an ongoing issue since the beginning. We are trying to ensure that each CI test can reliably pass.

veggiesaurus:
> > that in itself is worrying... because we can't really tell whether its something in the branch or something in the CI that is going wrong
> 
> That's right. It has been an ongoing issue since the beginning. We are trying to ensure that each CI test can reliably pass.

Well, it's also ok if they _reliably_ fail. It's not your job to make sure they all pass, that's the developer's responsibility :smile:

kswang1029 (approved):

PR #954, opened 2021-10-29
check that task queue is not empty before attempting to execute task

veggiesaurus:
fixes hard crash on startup with CentOS 7.


ajm-ska (approved):
This fixes the segmentation fault issue on CentOS7 when using devtoolset.
It appears to still work fine on other OSs too.

Issue #830, opened 2021-06-09
Don't establish readlocks on CASA images


keflavich:
I'd like to prevent CARTA from establishing a "read lock" on image cubes.

I've had several (dozen) experiences where I open an image to inspect it, find something to fix, re-run the imaging, then find my supercomputer has been spinning its wheels for days waiting for a read/write lock.  It would be better not to establish those locks.

If needed, this may be doable with https://github.com/radio-astro-tools/casa-formats-io instead of casacore, though the implementation in casa-formats-io is not 100% complete (I think it doesn't read _all_ of the header information as of today)

veggiesaurus:
I've done a quick test using `lsof` after opening a FITS file, HDF5 file and CASA file in CARTA.

FITS files open with read-only access and no lock. HDF5 files open with read-only access _and_ a read lock. CASA images open `table.f0_TSM0` and `mask0/table.f0_TSM0` with read-only access and no lock, as well as `table.lock` and `mask0/table.lock` in read-write access and a read lock on part of the file.

I'm not really too clued up on the CASA file format, so @pford should be able to clarify whether those locks are really necessary.

In terms of removing the read locks entirely, I'm not sure how we should approach it. It seems you can build with `HDF5_USE_FILE_LOCKING=FALSE` to disable the read locks, but I haven't investigated that.

edit: opening `carta_backend` with then env variable `HDF5_USE_FILE_LOCKING=FALSE` works, it results in removing the read lock

veggiesaurus:
Also, I guess one thing we should expect is that things may go horribly wrong with CARTA if you attempt to continue reading from a file after its been rewritten!

keflavich:
I think I've already done this in some cases; I used the filesystem to move a file that had a lock on it (`mv` ignores readlocks, at least on this filesystem?) and... CARTA just used what was cached, I think.  It just redisplayed the same data.  It was weird, but no major crashes happened.  Again, not sure I can reproduce _exactly_ what I did...

veggiesaurus:
@pford is this fixed now?

kswang1029:
should be fixed now.

PR #956, opened 2021-11-06
Stop reusing a single pointer to cache image loaders

### LINKED ISSUE #896: backend crash when loading multiple HDF5 images in one shot
with my local mac build, loading four hdf5 images in one shot would "usually" crash the backend. Sometimes I see

> [2021-08-09 14:21:53.639] [error] Error loading image data: HDF5 error: reading slab from data set array DATA
[2021-08-09 14:21:53.640] [error] Error opening HDF5 image: HDF5 error: Unknown data type of scalar attribute PC02_03
[2021-08-09 14:21:53.641] [error] Error loading image data: HDF5 error: reading slab from data set array DATA
[2021-08-09 14:21:53.787] [error] Error loading image data: HDF5 error: reading slab from data set array DATA
Segmentation fault: 11

Screen Shot 2021-08-09 at 2 31 02 PM

Tested the the demo server and it can be reproduced.

confluence:
Now that we can open multiple images simultaneously, it no longer makes sense to reuse the same per-session pointer to cache the image loader (so that the loader used to read file info can be reused when the image is opened), since different images opening in parallel could overwrite this variable and cause undefined behaviour.

I have added a simple LRU cache which stores loaders up to a maximum capacity. When an image is opened, its loader is explicitly removed from the cache. Cached loaders are invalidated if their files have changed on disk.

While the original intention of this change was to remove a thread-unsafe variable, a nice side effect is that if a file has a lot of headers which take a long time to parse, the file info will appear immediately as long as the loader remains in the cache.

Outstanding questions:

1. I started work on this in an attempt to fix #896, but that bug appears to have been fixed as a side-effect of the recent TBB / threading changes (or at least @kswang1029 reported being unable to reproduce it). I would like to confirm that this branch doesn't reintroduce this bug.

2. What is a sensible value for the capacity of the cache? Should it be a fixed number of loaders, or should we estimate the actual memory usage of different loader types and base the capacity on total memory usage? The HDF5 loaders store cached statistics.

3. Does it actually make sense to remove the loader from the cache when an image is opened? The user is unlikely to append the same image again (unless I'm missing a use case), but if they click on the file in the browser while browsing, there may be an annoying and unexpected delay in displaying the file info just for that file (which may be avoided if we leave the loader in the cache).


confluence:
I can see that this is breaking the file browser ICD test -- I'll see if I can figure out why.

Edit: I see that this is unrelated -- we're just parsing `HISTORY` correctly now. :)

acdo2002:
From the ICD test console.log, because there are too many header.entries so the file_info and moment ICD tests are failed. 
https://carta.asiaa.sinica.edu.tw/blue/organizations/jenkins/carta-backend/detail/confluence%2Fmultiple_cached_image_loaders/1/pipeline/336

confluence:
Followup question: should we be bypassing this entirely for internally generated images? They never have their info loaded in the file browser, so there's nothing to reuse.

kswang1029:
> 3\. Does it actually make sense to remove the loader from the cache when an image is opened? The user is unlikely to append the same image again (unless I'm missing a use case), but if they click on the file in the browser while browsing, there may be an annoying and unexpected delay in displaying the file info just for that file (which may be avoided if we leave the loader in the cache).

@confluence regarding loading the same image file multiple times, yes, this could happen when there are multiple line features in the cube and users would like to compare different lines. This is now a more realistic use case with this https://github.com/CARTAvis/carta-frontend/pull/1647 merged, meaning users can set different rest frequencies and compare different line features matched in the same velocity reference.

confluence:
In that case I think we should leave the loader in the cache and rely only on the normal capacity limit to remove loaders.

I've realised that the current implementation is actually broken for generated images, because the loaders have a blank filename, and a valid filename is needed for the modification time check.

confluence:
The loader is no longer removed from the cache when an image is opened. I opted to bypass the cache when loading internally generated images -- the benefit of caching loaders in this case is unclear, and may require more extensive changes (e.g. right now we recreate all the individual loaders in the Stokes connector anyway). But I can extend this PR to cover that case if it seems like a good idea.

pford:
It was not really intended for the Session to hold a _loader (shared_ptr for a FileLoader) permanently.  It should be reset when the image is loaded (FileLoader* handed to Frame) or reset to the next selected image in the file list.  The unintended consequence I just realized here is that if the image is not loaded, the last loader for file info remains in scope for the duration of the Session, holding the image file opened by FileExtInfoLoader.  And if the image is loaded, the image file remains open until it is unloaded.  Either way, this may cause readlock problems when using the image outside carta (issue #830).  Within carta, I don't know if loading an image in different Frames in the same or different Sessions causes a crash; this needs investigation and possibly moving the Frame image mutex to a shared-FileLoader image mutex.

So the issues I see are:
1. Besides their own cached data (e.g. stats), loaders have open image files (fix separately for #830?)
2. Putting these loaders into a cache just compounds issue 1
3. Used and unused loaders may fill the cache, causing a used loader to "age out" when the capacity is reached.  There is no check of use count before deletion.  For example, after image1 is loaded the user clicks around the file list (each file info is a cached loader) and perhaps appends then closes some images (does not remove loaders from cache).  The cache reaches capacity so the loader for image1 is deleted, user interacts with image1 and gets errors or a crash.

I don't see the need to store the loaders for images clicked in the file browser which may never be used again.  Instead, the Session could create a local loader in OnFileInfoRequest for the FileExtInfoLoader (deleted on exit from function) and in OnOpenFile for the FileExtInfoLoader and the Frame.  If you think a shared FileLoader for a loaded image would be useful, then the OnOpenFile loader could be added to LoaderCache and the shared_ptr passed to each Frame holding that image.  The cached loader could be used for file info in OnFileInfoRequest if it exists, otherwise create a local instance.  A cached loader should be removed when its last Frame is closed.

confluence:
I remember that #830 used to be a problem for HDF5 files, but as far as I know we changed the mode that we use to open HDF5 files and they are no longer affected. Is a similar solution available for the other image formats? #830 would be triggered by loaders in open frames even without any caching, although I see how the cache compounds the issue. That's probably a good enough reason not to implement the cache now unless we can fix #830 for all file types.

Re: 3, surely the loader used by an open frame will not be affected by a loader's removal from the cache? Both the cache and the frame store shared pointers to loaders; I only remove the shared pointers from the cache containers, and don't delete the underlying objects. The shared pointer itself is what tracks references before deleting the loader object.

My understanding was that the point of storing a persistent loader pointer at all was to reuse an existing loader created to display file info in the file browser if that file is opened immediately afterwards (since creating the loader may take time, for various reasons -- like parsing long headers or precalculating derived HDF5 stats). My original solution to avoid using the same pointer for multiple images was to use a map (I realised that my choice of key in my implementation didn't make any sense, but in principle I could use the normalized filenames as keys as I'm doing here in the cache).

The problem I ran into was how to trigger removal of a loader from this map if it was used to display file info but never used to open a file. There was never a mechanism for this in the original code (which leads to the side effect you noticed), but because a single pointer variable is overwritten, at least there we don't accumulate infinite loader objects. I introduced a cache with a fixed capacity to get around this issue.

The simplest short-term solution is to abandon reusing the loader and create a new local variable for every request. This would mean a slowdown in some circumstances. I think what we should do depends on how easy it is to fix #830.

pford:
You make very good points, and you are correct about removing the shared_ptr from the cache (which is the whole point of shared_ptr, duh!).  I suppose this is another case of memory usage vs. performance, always a tradeoff.

The issue in #830 had to do with writing data to a CASA image which was opened in CARTA, so it's possible that accessing the image headers for file info is not a problem; I'll have to do more testing.

pford (approved):

kswang1029 (approved):
from user testing, I don't see odd behavior nor backend crash with my crash recipe. I see the performance boost via header caching. 👍

markccchiang (approved):

veggiesaurus (approved):

acdo2002:
I have tested the performance regression tests, this backend branch looks fine! 
No significant change in the performance regression tests.

PR #962, opened 2021-11-11
Angus/1 display files by ext

### LINKED ISSUE #1: Long load times in directories with many files
CARTA has a long load time when I call CARTA (v1.2) in a directory with a large number of files (>100), but I'm calling it with a specific file from the command line, e.g.,

```
carta file0001.fits
```

The long load time appears to come from running the scan across all the files in directory before loading.  Is it possible to skip the scan when called from the command line in this fashion?

### LINKED ISSUE #87: Long wait time in file browser

veggiesaurus:
Closes https://github.com/CARTAvis/carta/issues/1 and closes https://github.com/CARTAvis/carta/issues/87. Companion PR of https://github.com/CARTAvis/carta-frontend/pull/1699.

Minor refactoring to extract the logic of guessing file types to `Util/File.cc` (happy to move it from here if there's a better place)


src/Util/File.cc
```diff
@@ -67,3 +67,88 @@ fs::path SearchPath(std::string filename) {
...
+            case GZ_MAGIC_NUMBER:
+                fs::path gz_path(path_string);
+                return (gz_path.stem().extension().string() == ".fits") ? CARTA::FITS : CARTA::UNKNOWN;
+        }
+    } else {
+        // Guess file type by extension
+        fs::path path(path_string);
+        auto filename = path.filename().string();
+
+        if (EndsWith(filename, ".fits") || EndsWith(filename, ".FITS") || EndsWith(filename, ".fz") || EndsWith(filename, ".fits.gz")) {
```
confluence: I would make this a case-insensitive regular expression (or three of them). What if it's `.FITS.GZ` or `.FITS.gz`, etc.?

src/Util/File.cc
```diff
@@ -67,3 +67,88 @@ fs::path SearchPath(std::string filename) {
...
+        // Guess file type by magic number
+        auto magic_number = GetMagicNumber(path_string);
+        switch (magic_number) {
+            case FITS_MAGIC_NUMBER:
+                return CARTA::FITS;
+            case HDF5_MAGIC_NUMBER:
+                return CARTA::HDF5;
+            case GZ_MAGIC_NUMBER:
+                fs::path gz_path(path_string);
+                return (gz_path.stem().extension().string() == ".fits") ? CARTA::FITS : CARTA::UNKNOWN;
```
confluence: What if the suffix is a different case here?

src/Util/File.cc
```diff
@@ -67,3 +67,88 @@ fs::path SearchPath(std::string filename) {
...
+    } else {
+        // Guess file type by extension
+        fs::path path(path_string);
+        auto filename = path.filename().string();
+
+        if (EndsWith(filename, ".crtf")) {
+            return CARTA::CRTF;
+        } else if (EndsWith(filename, ".reg")) {
+            return CARTA::DS9_REG;
+        }
```
confluence: More case sensitivity.

src/Util/File.cc
```diff
@@ -67,3 +67,88 @@ fs::path SearchPath(std::string filename) {
...
+    } else {
+        // Guess file type by extension
+        fs::path path(path_string);
+        auto filename = path.filename().string();
+
+        if (EndsWith(filename, ".fits") || EndsWith(filename, ".FITS") || EndsWith(filename, ".fz") || EndsWith(filename, ".fits.gz")) {
+            return CARTA::CatalogFileType::FITSTable;
+        } else if (EndsWith(filename, ".xml") || EndsWith(filename, ".vot") || EndsWith(filename, ".votable")) {
+            return CARTA::CatalogFileType::VOTable;
+        }
```
confluence: As above. :)

confluence (changes requested):

src/Util/String.cc
```diff
@@ -22,3 +22,11 @@ void SplitString(std::string& input, char delim, std::vector<std::string>& parts
...
     }
 }
+
+bool EndsWith(const std::string& haystack, const std::string& needle) {
+    if (needle.size() > haystack.size()) {
+        return false;
+    }
+
+    return std::equal(needle.rbegin(), needle.rend(), haystack.rbegin());
+}
```
confluence: I'd replace this with a case-insensitive regex check, and rename it to something more specific like `HasSuffix`, `DetectSuffix`, `SuffixFuzzy`, `SuffixIgnoreCase`, `EndsWithIgnoreCase`, etc.. Not sure how clear we need to be, but `EndsWith` would probably be misleading in a function that's in `Util`.

src/Util/File.cc
```diff
@@ -86,9 +86,9 @@ CARTA::FileType GuessImageType(const std::string& path_string, bool check_conten
         fs::path path(path_string);
         auto filename = path.filename().string();
 
-        if (EndsWith(filename, ".fits") || EndsWith(filename, ".FITS") || EndsWith(filename, ".fz") || EndsWith(filename, ".fits.gz")) {
+        if (HasSuffix(filename, ".fits") || HasSuffix(filename, ".FITS") || HasSuffix(filename, ".fz") || HasSuffix(filename, ".fits.gz")) {
```
confluence: 1. It looks like you're using the case sensitive check everywhere.
2. If this becomes case-insensitive, one of the conditions will be redundant.

confluence (changes requested):

src/Util/File.cc
```diff
@@ -67,3 +67,89 @@ fs::path SearchPath(std::string filename) {
...
+                fs::path gz_path(path_string);
+                std::string extension = gz_path.stem().extension().string();
+                return HasSuffix(extension, ".fits") ? CARTA::FITS : CARTA::UNKNOWN;
+        }
+    } else {
+        // Guess file type by extension
+        fs::path path(path_string);
+        auto filename = path.filename().string();
+
+        if (HasSuffix(filename, ".fits") || HasSuffix(filename, ".fz") || HasSuffix(filename, ".fits.gz")) {
```
confluence, line 90: These are all case-sensitive checks, because the flag defaults to false.

(I would consider making it default to true, although that may be too misleading.)

confluence (changes requested):

src/Util/File.cc
veggiesaurus: the flag is `case_sensitive`, so setting it to false means the checks are all case-insensitive.

confluence: Ohhh, I see now. :)

confluence (approved):

src/FileList/FileListHandler.cc
```diff
@@ -137,11 +138,11 @@ void FileListHandler::GetFileList(CARTA::FileListResponse& file_list, std::strin
                     bool is_region_file(false);
 
                     if (region_list && cc_file.isRegular(true)) {
-                        CARTA::FileType file_type(GetRegionType(full_path)); // CRTF, DS9, or UNKNOWN
+                        auto file_type = GuessRegionType(full_path, filter_mode == CARTA::Content);
```
pford, line 141: I would think from the user comment in carta issue 1 ("I would greatly prefer if we just saw the filenames and sizes (simple ls -lh output) rather than the metadata") and the CARTA::AllFiles mode that you would skip all type guessing, else isn't this actually by extension?  In AllFiles mode, the frontend should leave the file browser Type column blank or hide it if possible rather than list all files as Unknown.

src/FileList/FileListHandler.cc
```diff
@@ -187,24 +188,9 @@ void FileListHandler::GetFileList(CARTA::FileListResponse& file_list, std::strin
...
-                                    file_type = CARTA::FileType::HDF5;
-                                    add_file = true;
-                                } else if (region_list) {
-                                    // List all regular files in region list
-                                    add_file = true;
-                                }
-                            }
+                        } else if (!region_list && cc_file.isRegular(true)) {
+                            file_type = GuessImageType(full_path, filter_mode == CARTA::Content);
+                            add_file = filter_mode == CARTA::AllFiles || file_type != CARTA::UNKNOWN;
```
pford, line 193: Same here, I think AllFiles means skip the guessing.  They want the file list ASAP.

src/Util/File.cc
```diff
@@ -67,3 +67,89 @@ fs::path SearchPath(std::string filename) {
...
+    if (check_content) {
+        // Guess file type by magic number
+        auto magic_number = GetMagicNumber(path_string);
+        switch (magic_number) {
+            case FITS_MAGIC_NUMBER:
+                return CARTA::FITS;
+            case HDF5_MAGIC_NUMBER:
+                return CARTA::HDF5;
+            case GZ_MAGIC_NUMBER:
+                fs::path gz_path(path_string);
```
pford, line 81: Duplicates code in IsCompressedFits (in this file), which also checks the magic number but only checks for lower case ".fits".

src/FileList/FileListHandler.cc
veggiesaurus: @kswang1029 what do you think? I'm assuming that guessing the file type by extension is trivially fast (just some string comparisons), so I can't imagine skipping this process for "All files" would make any difference. If we keep it in, users can at least see which files CARTA _thinks_ it can support.

veggiesaurus: I've just checked, and the performance difference is negligible, so I can't see any practical reason why we shouldn't guess the file type by extension for "All files".

See tests for 25 000 FITS files below. Disk caches cleared between each test.

```shell
# listing 25 000 files from an SSD, filtering by content
[2021-11-15 11:08:17.873] [debug] [performance] File list for directory /home/angus/cubes/fits/lots_dummy filled in 3038.569 ms at 8227.557 entries/s

# listing 25 000 files from an SSD, filtering by extension
[2021-11-15 11:08:37.931] [debug] [performance] File list for directory /home/angus/cubes/fits/lots_dummy filled in 355.259 ms at 70371.194 entries/s

# listing 25 000 files from an SSD, showing all files and not checking extension
[2021-11-15 11:08:48.008] [debug] [performance] File list for directory /home/angus/cubes/fits/lots_dummy filled in 347.240 ms at 71996.314 entries/s


# listing 25 000 files from an HDD, filtering by content
[2021-11-15 11:13:34.370] [debug] [performance] File list for directory /media/angus/ImageData/dummy filled in 123947.127 ms at 201.699 entries/s

# listing 25 000 files from an HDD, filtering by extension
[2021-11-15 11:13:52.568] [debug] [performance] File list for directory /media/angus/ImageData/dummy filled in 364.601 ms at 68568.106 entries/s

# listing 25 000 files from an HDD, showing all files and not checking extension
[2021-11-15 11:14:04.578] [debug] [performance] File list for directory /media/angus/ImageData/dummy filled in 342.513 ms at 72989.930 entries/s
```

src/Util/File.cc
veggiesaurus: ok, I've made the two functions consistent now.

src/FileList/FileListHandler.cc
kswang1029: Good to guess file type by file extension if the performance impact is negligible. Question here: by extension does this mean only fits and hdf5 images can be guessed? casa and miriad will be listed as folder?

veggiesaurus: > Question here: by extension does this mean only fits and hdf5 images can be guessed? casa and miriad will be listed as folder?

No, they are handled the same way as before. This PR only affects how regular files are handled. I guess if you had thousands of CASA images, this might be a problem? We may need to implement a faster image check for those image types. In particular, something that doesn't involve actually _opening_ the files within the CASA image folder.

Based on [this code](https://github.com/casacore/casacore/blob/e793b3d5339d828a60339d16476bf688a19df3ec/images/Images/ImageOpener.cc#L73) in cassacore it looks like it attempts to open the table file before determining that it's a CASA image (when calling [`Table::isReadable`](https://github.com/casacore/casacore/blob/e793b3d5339d828a60339d16476bf688a19df3ec/tables/Tables/Table.cc#L1009)).

pford: I was trying to understand the difference between "All files" and "By extension", which seemed to do the same thing.  It appears that Unknown files are listed for the first but not the second.

pford (approved):

Issue #963, opened 2021-11-17
Bump preferences schema version


confluence:
The backend is currently using `preference_schema_1.json` when writing the preferences file. It should be using `preferences_schema_2.json`.

PR #964, opened 2021-11-17
Switch to new version of preferences schema

confluence:
This updates the version of the preferences schema that we use when writing preferences files, fixing #963.


veggiesaurus (changes requested):
we also need to change the version number we write in `SimpleFrontendServer::WritePreferencesFile`

veggiesaurus (approved):

PR #972, opened 2021-11-22
Remove the loader from the cache after opening an image.

### LINKED ISSUE #971: Cursor-related crash in spectral multi-profile plot of image matched with itself
The segfault reported by @kswang1029 in #955 is actually present in `dev` (but not in the last beta, so it should be possible to bisect to find the commit which introduced it).

Steps to reproduce:

1. Load an image
2. Append the same image
3. Match them
4. Open the Z profile widget
5. Check the "Image" checkbox to plot multiple profiles
6. Move the cursor around

The crash is in `memmove` in `casacore::TSMCube::accessSection` and follows two consecutive `SET_CURSOR` requests (but not always).

confluence:
Fixes #971. This restores the previous image opening behaviour. Reusing the same loader object when opening the same image multiple times causes problems in casacore (probably race conditions if the same casacore objects are accessed simultaneously).

### DIFF
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -483,6 +483,9 @@ bool Session::OnOpenFile(const CARTA::OpenFile& message, uint32_t request_id, bo
         // query loader for mipmap dataset
         bool has_mipmaps(loader->HasMip(2));
 
+        // remove loader from the cache (if we open another copy of this file, we will need a new loader object)
+        _loaders.Remove(fullname);
+
         if (frame->IsValid()) {
             // Check if the old _frames[file_id] object exists. If so, delete it.
             if (_frames.count(file_id) > 0) {


kswang1029:
@confluence it seems that I cannot reproduce #971 (using my steps and the image cube) anymore. 👍

confluence:
If we want to eliminate the small delay the next time the user views the headers of the same image, we could attempt to get the loader from the cache immediately after removing it (which would trigger the caching of a new copy), but I don't know if this really makes sense (is the user more likely to view that file again than to view a completely different file?) so I'm inclined to leave things as they are.

kswang1029:
> If we want to eliminate the small delay the next time the user views the headers of the same image, we could attempt to get the loader from the cache immediately after removing it (which would trigger the caching of a new copy), but I don't know if this really makes sense (is the user more likely to view that file again than to view a completely different file?) so I'm inclined to leave things as they are.

Mostly user will append different images. Having a delay when appending the same image is okay I think.

veggiesaurus (approved):

Issue #797, opened 2021-04-13
Migrate possible ICD test back to backend unit test


kswang1029:
This is a step of making the automated test robust and informative. Detailed list of which ICD tests to be migrated is TBD (work-in-progress list [here](https://docs.google.com/spreadsheets/d/1qwdTTKdNA1oNnNAUNgyPZH5HOVyt2Uw0vPT2_UsEaXk/edit#gid=0))

todo list based on the C column of the above spreadsheet (each should be a separate PR):

- [x] ACCESS_CARTA
- [x] ANIMATION
- [x] REGION_REGISTER
- [x] GET_FILELIST_DEFAULT_PATH, GET_FILELIST_UNKNOWN_PATH, FILETYPE_PARSER
- [x] FILE_INFO*
- [ ] REGION_LIST, REGION_INFO
- [ ] CASA_REGION*
- [ ] DS9_REGION*
- [ ] CATALOG_LIST
- [ ] CATALOG_FITS_VOT, CATALOG_GENERAL
- [ ] CHECK_RASTER_TILE_DATA
- [x] CONTOUR_IMAGE_DATA, CONTOUR_IMAGE_DATA_NAN
- [ ] CURSOR_SPATIAL_PROFILE, CURSOR_SPATIAL_PROFILE_NAN
- [ ] POINT_REGION_SPATIAL_PROFILE, POINT_REGION_SPATIAL_PROFILE_NAN
- [ ] CURSOR_SPECTRAL_PROFILE, CURSOR_SPECTRAL_PROFILE_NAN
- [ ] POINT_REGION_SPECTRAL_PROFILE, POINT_REGION_SPECTRAL_PROFILE_NAN
- [ ] REGION_SPECTRAL*
- [ ] REGION_STATISTICS*
- [ ] MATCH_*
- [ ] MOMENT_GENERATOR_*
- [ ] OPENFILE_CASA_BEAMTABLE, OPENFILE_FITS_BEAMTABLE, OPEN_IMAGE_APPEND, OPEN_IMAGE_PV, OPEN_IMAGE_CASA_VARIENTS
- [ ] PER_CUBE_HISTOGRAM, PER_CUBE_HISTOGRAM_CANCELLATION, PER_CUBE_HISTOGRAM_HDF5

veggiesaurus:
My feeling on the migration is that we should rely on unit tests for:
- Checking the validity of calculations
- Comparing results between different file formats
- Ensuring we handle edge cases correctly

[...]

PR #973, opened 2021-11-23
Unit tests for the file list and file info

markccchiang:
This is part of #797. Migrating the ICD tests of file list and file info to the backend unit tests. Thanks for @acdo2002 preparing sample image files.


confluence (changes requested):
As we previously discussed, these should be lower-level unit tests which bypass the messaging interface and directly call functions on internal classes (in this case the classes in `src/FileList/`).

veggiesaurus (changes requested):
echoing what @confluence said: We don't just want an ICD test translated to the backend. Unit tests of the actual functionality are what we want.

I'd also like to have them in separate files i.e. `TestFileInfo.cc` and `TestFileList.cc`.

Do we really need the very long history entries in the test data?

markccchiang:
> 
> Do we really need the very long history entries in the test data?

I think we don't need very long history entries in the test data. @acdo2002 could you please remove `HISTORY` entries in sample images `M17_SWex_unit.image` and `M17_SWex_unit.fits`?

acdo2002:
@markccchiang I have put the update unit tests image in [here](https://drive.google.com/drive/folders/1Wi1WxXk6gg-aQQXhXTTEN8m8vM4bWMcX?usp=sharing), please  download "unit-test.zip".

I have remove the HISTORY related header in fits.
However casa format is converted from fits via CASA, there will generate few HISTORY header. But the new version is still smaller than previous one.
I also update the hdf5 & miriad format.

Now the exact image size:
(1)"M17_SWex.image" in files
type == 1
size = 112308
HDU_List = [""]

(2)"M17_SWex.fits" in files
type == 0
size = 8640
HDU_List = [""]

(3)"M17_SWex.miriad" in files
type == 3
size = 3040
HDU_List = [""]

(4)"M17_SWex.hdf5" in files
type == 2
size = 25040
HDU_List = ["0"]

markccchiang:
As for the file list test, it tests the `FileListHandler` class with sample files and folders. And I added a condition check that prevents users access the folder beyond the top-level folder via the `../` command. For the file info test, it tests classes `FileInfoLoader`, `FileExtInfoLoader` and the `FillExtendedFileInfo` function from `Session` with different file formats. The time spends on different classes with respect to file formats can also be seen from tests. The Casa file spends almost twice the time on getting file extended info than the other file formats.

confluence (approved):

PR #975, opened 2021-11-25
backport of #946

veggiesaurus:
Backport of #946 in order to fix issues with the CARTA controller when running from a directory


veggiesaurus:
@ajm-asiaa we don't need this PR anymore, do we? Now that we're not trying to read the `true` directory all the time?

ajm-ska:
@veggiesaurus OK. Let's close it.

Issue #978, opened 2021-11-29
Unit Test problem with IcdTest.AnimatorPlayback 


ajm-ska:
It has been discovered that the `IcdTest.AnimatorPlayback` Unit Test is does not run reliably on Ubuntu 20.04.
Sometimes it passes, sometimes it gives a segmentation fault, and sometimes it just hangs indefinitely. 
The later case makes CI difficult as it holds everything up.

To reproduce the issue, please try running the test multiple times:
`./carta_backend_tests --gtest_filter=IcdTest.AnimatorPlayback --gtest_repeat=10`

veggiesaurus:
@markccchiang can you look at this?

Actually, not sure who should investigate this. I suggested @markccchiang as he wrote the test, but I can't really see anything wrong with the test itself.

markccchiang:
> @markccchiang can you look at this?
> 

I will investigate this problem.

veggiesaurus:
Stack trace for the segfault I get on ubuntu:

```
Session::WaitingFlowEvent() Session.h:199
AnimationTask::execute() OnMessageTask.cc:41
operator() Threading.cc:53
std::__invoke_impl<void, carta::ThreadManager::StartEventHandlingThreads(int)::<lambda()> >(std::__invoke_other, <> &&) invoke.h:60
std::__invoke<carta::ThreadManager::StartEventHandlingThreads(int)::<lambda()> >(<> &&) invoke.h:95
std::thread::_Invoker<std::tuple<carta::ThreadManager::StartEventHandlingThreads(int)::<lambda()> > >::_M_invoke<0>(std::_Index_tuple<>) const thread:244
std::thread::_Invoker<std::tuple<carta::ThreadManager::StartEventHandlingThreads(int)::<lambda()> > >::operator()() const thread:251
std::thread::_State_impl<std::thread::_Invoker<std::tuple<carta::ThreadManager::StartEventHandlingThreads(int)::<lambda()> > > >::_M_run() const thread:195
<unknown> 0x00007f0544093de4
start_thread 0x00007f05446f3609
__clone 0x00007f0543ed2293
```

The other relevant threads are:

```
std::mutex::lock() std_mutex.h:103
carta::concurrent_queue<std::pair<std::vector<char, std::allocator<char> >, bool> >::try_pop(std::pair<std::vector<char, std::allocator<char> >, bool>&) Concurrency.h:32
[...]

PR #979, opened 2021-11-30
Changing the type of reference count from Session to atomic int

markccchiang:
It tries to solve #978. The reference count from Session is increased when `OnMessageTask` is created to run in a different thread. It is decreased when the task is done and `OnMessageTask` is destructed. So the type of reference count is better to be atomic int to avoid unexpected behavior in animation.

### DIFF
--- a/src/Session.h
+++ b/src/Session.h
@@ -327,7 +327,7 @@ protected:
 
     SessionContext _animation_context;
 
-    int _ref_count;
+    std::atomic<int> _ref_count;
     int _animation_id;
     bool _connected;
     static int _num_sessions;
--- a/test/BackendModel.cc
+++ b/test/BackendModel.cc
@@ -283,6 +283,6 @@ void BackendModel::ClearMessagesQueue() {
 
 void BackendModel::WaitForJobFinished() {
     while (_session->GetRefCount() > 1) {
-        std::this_thread::sleep_for(std::chrono::microseconds(1));
+        std::this_thread::sleep_for(std::chrono::milliseconds(1));
     }
 }
--- a/test/TestIcd.cc
+++ b/test/TestIcd.cc
@@ -436,6 +436,8 @@ public:
         }
 
         EXPECT_EQ(_message_count, 0); // make sure there is no data stream when animation stopped
+
+        _dummy_backend->WaitForJobFinished();
     }
 
     void RegionRegister() {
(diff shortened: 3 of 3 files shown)


ajm-ska:
I'm sorry to report that the issue appears to still remain.

veggiesaurus:
I do think we should make this change anyway, even if it doesn't fix the issue. @robsimmonds do you agree?

markccchiang:
It seems two animation tasks can not run at the same time. In this unit test, it runs an animation task play forward and then another task playback. I added a busy waiting step between these two tasks. In a rare case, the backend can not generate all the data messages we expected. So the while loop can not be broken and it hangs here. I added a timeout mechanism to break this infinite loop. Now it looks running smoothly on Ming-Yi's Ubuntu computer. @ajm-asiaa could you please test it again?

ajm-ska (approved):
@markccchiang Nice work! That appears to have fixed the issue all right. I ran the test on Ubuntu with `--gtest_repeat=100` and they all ran successfully without hanging.

acdo2002:
Yes, I also test it with --gtest_repeat=100 on my Ubuntu laptop, they are all passed!

veggiesaurus:
> It seems two animation tasks can not run at the same time. In this unit test, it runs an animation task play forward and then another task playback. I added a busy waiting step between these two tasks. In a rare case, the backend can not generate all the data messages we expected. So the while loop can not be broken and it hangs here. I added a timeout mechanism to break this infinite loop. Now it looks running smoothly on Ming-Yi's Ubuntu computer. @ajm-asiaa could you please test it again?

@markccchiang just to be sure, does the test still fail if it times out?

veggiesaurus:
these changes look fine to me, but I know @robsimmonds is working on fixing the underlying issue with handling animation messages in an unexpected order, so I suggest we merge this in, so as not to lose our CI functionality, and then revisit the issue when @robsimmonds makes his PR.

We can thenclose the associated issue and make a new, more relevant one (specific to the animation tasks).

veggiesaurus (approved):

confluence (approved):

markccchiang:
> @markccchiang just to be sure, does the test still fail if it times out?

Yes, it reports an error if times out. Since the number of response messages does not fit our expectations.

PR #983, opened 2021-12-09
Rwjs/anifixd21

robsimmonds:
Fix for the queuing metrix code.
Added an additional task to avoid possible race condition when starting an animation if there is an existing animation running.


veggiesaurus:
@robsimmonds can you merge `dev` into your branch so that the ICD tests pass

src/Concurrency.h
```diff
@@ -83,21 +82,26 @@ class queuing_rw_mutex {
         if ((_reader_count > 0) || (_writer_count > 1)) {
             queue_writer(lock);
         }
+        short rc2 = _reader_count;
```
veggiesaurus: What's the point of these two lines? (and the ones below). `rc` and `wc` are never used.

src/Session.h
```diff
@@ -236,6 +236,14 @@ class Session {
 
     // Close cached image if it has been updated
     void CloseCachedImage(const std::string& directory, const std::string& file);
+    bool AnimationActive() {
+        return _animation_active;
+    }
+    void SetAnimationActive(bool val) {
+        fprintf(stderr, " SetAniActive %d\n", val);
```
veggiesaurus: please use `spdlog::debug` for this

src/SessionManager/SessionManager.cc
```diff
@@ -230,6 +229,7 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
                 }
                 case CARTA::EventType::ANIMATION_FLOW_CONTROL: {
                     CARTA::AnimationFlowControl message;
+                    fprintf(stderr, " F ");
```
veggiesaurus: please remove this

src/Concurrency.h
```diff
@@ -83,21 +82,26 @@ class queuing_rw_mutex {
...
+        short wc2 = _writer_count;
```
veggiesaurus: What's the point of these two lines? (and the ones below). `rc` and `wc` are never used.

veggiesaurus (changes requested):

src/Session.h
confluence: Please also use the full function name.

(Unless this was meant to be temporary and should just be removed.)

src/Session.cc
```diff
@@ -2055,7 +2056,7 @@ bool Session::ExecuteAnimationFrame() {
 }
 
 void Session::StopAnimation(int file_id, const CARTA::AnimationFrame& frame) {
-    if (!_animation_object) {
+    if (_animation_object == nullptr) {
```
veggiesaurus: isn't this the same thing?

confluence (commented):
I have the same comments and questions as @veggiesaurus ; otherwise this looks OK.

src/Concurrency.h
robsimmonds: for debugging. gone now.

src/Concurrency.h
```diff
@@ -90,14 +89,15 @@ class queuing_rw_mutex {
...
         if ((_reader_count == 0) && (_writer_count > 0)) {
-            --_writer_count;
             dequeue_one_writer();
         }
     }
     void writer_leave() {
         std::unique_lock<std::mutex> lock(_mtx);
-
         --_writer_count;
+        short rc = _reader_count;
```
veggiesaurus: these ones are still here?

veggiesaurus (changes requested):

src/Concurrency.h
robsimmonds: ok. i will remove and push. The tests have all completed successfully at this point.

confluence (approved):

veggiesaurus (approved):

PR #986, opened 2021-12-11
Added join on thread exit.

robsimmonds:
Join threads before exit to make sure that all threads associated with process exit.


veggiesaurus:
this doesn't solve the issue. The actual issue is because `ExitEventHandlingThreads` isn't always called when `exit(0)` is called.

I think this is the solution: https://github.com/CARTAvis/carta-backend/pull/988

confluence (approved):
It wasn't clear to me that this was intended to fix that issue (I could still reproduce it when I checked). I'm happy for this to be merged anyway if it makes the code more robust. I have also approved the other PR.

PR #988, opened 2021-12-12
exit threads when calling exit(0) due to timeout

veggiesaurus:
Fixes problem with backend hanging rather than exiting after all sessions have disconnected and the backend is run with the `--exit_timeout` or `--initial_timeout` arguments.


confluence (approved):
I can confirm that this fixes the issue.

PR #993, opened 2022-01-05
Update the animation tester

markccchiang:
This PR updates the start animation message handler in the testing code. Also adds a lock on the `AnimationObject`. The current Carta backend can only execute an animation task in one session at a time. This lock prevents two animation tasks overlap at the same time, which would lead to the backend crash.


robsimmonds:
I am not sure why this new mutex was needed. This could should not be executed concurrently. Is there are particular problem that this mutex fixed?

veggiesaurus:
> I am not sure why this new mutex was needed. This could should not be executed concurrently. Is there are particular problem that this mutex fixed?

Doesn't the mutex ensure the code isn't executed concurrently?

markccchiang:
> I am not sure why this new mutex was needed. This could should not be executed concurrently. Is there are particular problem that this mutex fixed?

This crash problem only happens in the ICD animation test (#978). Adding this mutex can solve it according to @acdo2002's test. But since the current frontend already prevents running animations concurrently, this lock is not very necessary I agree.

veggiesaurus:
> This crash problem only happens in the ICD animation test (#978). Adding this mutex can solve it according to @acdo2002's test. But since the current frontend already prevents running animations concurrently, this lock is not very necessary I agree.

I think we should be defensive here. We shouldn't assume the frontend always behaves correctly. Ideally, _nothing_ the frontend does should be able to crash the backend. So I support the idea of fixing this issue.

robsimmonds:
> > I am not sure why this new mutex was needed. This could should not be executed concurrently. Is there are particular problem that this mutex fixed?
> 
> This crash problem only happens in the ICD animation test (#978). Adding this mutex can solve it according to @acdo2002's test. But since the current frontend already prevents running animations concurrently, this lock is not very necessary I agree.

Do you know if the ICD test still crashes without the mutex? I thought what I did before Christmas should have solved that issue. My main concern is if adding this mutex is hiding a bigger problem, if what I did before Christmas didn't solve the issue. If the crash is still happening, please let me know how to recreate the crash so I can figure out if there is still another issue.

markccchiang:
> Do you know if the ICD test still crashes without the mutex? I thought what I did before Christmas should have solved that issue. My main concern is if adding this mutex is hiding a bigger problem, if what I did before Christmas didn't solve the issue. If the crash is still happening, please let me know how to recreate the crash so I can figure out if there is still another issue.

Unfortunately, the crash still happens after the fixing. To reproduce it, remove the [code line 384](https://github.com/CARTAvis/carta-backend/blob/dev/test/TestIcd.cc#L384) and [450](https://github.com/CARTAvis/carta-backend/blob/dev/test/TestIcd.cc#L450) from the file `test/TestIcd.cc`, which means don't wait for the previous animation task to finish. Then execute the animation test with command `./carta_backend_tests --gtest_filter=IcdTest.AnimatorPlayback --gtest_repeat=100`. It would crash in a random test iteration.

markccchiang:
@acdo2002 I just merged @robsimmonds's fixing codes for the animation crash problem, and remove the animation object mutex. Could you please test it on the Ubuntu computer and see if the crash problem is solved. Thank you!

acdo2002:
I tested the update `mark/test_animator_playback (merged Rob's fixing code)` with my Ubuntu laptop and Ubuntu Desktop.

Unfortunately, if I run the following command:
` ./carta_backend_tests --gtest_filter=IcdTest.AnimatorPlayback --gtest_repeat=100`
There is a core dump message shows up:

> Repeating all tests (iteration 55) . . .
> 
> Note: Google Test filter = IcdTest.AnimatorPlayback
> [==========] Running 1 test from 1 test suite.
> [----------] Global test environment set-up.
> [----------] 1 test from IcdTest
> [ RUN      ] IcdTest.AnimatorPlayback
> terminate called after throwing an instance of 'casacore::AipsError'
>   what():  File::getstat error on /home/mingyi/carta-backend-unit-animator-playback/build/test/data/generated/3d847a36f637acc2.fits: No such file or directory
> Aborted (core dumped)

Such message even shows up if I set `--omp_threads=4`

veggiesaurus:
> terminate called after throwing an instance of 'casacore::AipsError'
> what(): File::getstat error on /home/mingyi/carta-backend-unit-animator-playback/build/test/data/generated/3d847a36f637acc2.fits: No such file or directory
> Aborted (core dumped)

this looks like a different error though, like it's just missing a file?

acdo2002:
If it's just missing a file, why the core dumped happened after 55 runs (the previous ones are passed)...?

veggiesaurus:
> If it's just missing a file, why the core dumped happened after 55 runs (the previous ones are passed)...?

oops, sorry! you're right. I didn't read that part of the log :laughing:

confluence:
The current crash is happening because the file generated by the test does not exist while casa is trying to read from it. My best guess is that this is happening because the test isn't correctly waiting for the animation code to stop running before exiting, which means that it may still be running when the global test teardown deletes all the generated files.

I'm re-running the test with the two `WaitForJobFinished` calls restored, to see if I can still reproduce the crash.

Either way, this crash is not revealing a problem in the code; it's an artifact of the test setup (I think we can expect many things to break if we delete a file on disk while CARTA is using it!).

confluence:
I repeated the test 1000 times after adding the `WaitForJobFinished` calls back in and did not trigger the crash (I was previously able to trigger it). If @acdo2002 can confirm this, I suggest that we merge the test with those lines added back. It's possible that adding the second call is sufficient.

confluence:
I re-ran the test 1000 times with only the second call added back, and could not reproduce the crash.

acdo2002:
With the current codes, I run 
`./carta_backend_tests --gtest_filter=IcdTest.AnimatorPlayback --gtest_repeat=1000`
They are all passed.

confluence (approved):

PR #995, opened 2022-01-07
Fix the problem of linear coordinate conversion for matched images

### LINKED ISSUE #982: no region stat for matched images generated with CASA ia.fft
Reported by a user

To reproduce the issue:
1) load the attached images and match them
2) make a region on the first image
3) set the first image as active and pan around the region to see region statistics. This is working well.
4) set the 2nd image (matched) as active and pan around the region to see region statistics. This triggers the issue.

The backend log says:
[2021-12-07 15:07:19.375] [debug] Region conversion failed or was distorted, using polygon approximation for rectangle
[2021-12-07 15:07:19.375] [error] Error converting region to output image: (/Users/kswang/carta_build/carta-casacore/casa6/casatools/casacore/coordinates/Coordinates/CoordinateSystem.cc : 1143) Failed AlwaysAssert world.nelements() == nWorldAxes()
[2021-12-07 15:07:19.375] [error] Error approximating rectangle as polygon in matched image.

With other images generated with CASA tclean, there is no such issue.

Screen Shot 2021-12-07 at 3 07 32 PM

[tmp.zip](https://github.com/CARTAvis/carta-backend/files/7666230/tmp.zip)

markccchiang:
Fix #982. According to the error message from casacore, the vector size of the input world point values should be equal to the number of world coordinate axes. If we want to convert the world coordinate to pixel coordinate using a linear coordinate system.

### DIFF
--- a/src/Region/Region.cc
+++ b/src/Region/Region.cc
@@ -1302,11 +1302,17 @@ bool Region::ConvertWorldToPixel(std::vector<casacore::Quantity>& world_point, c
         // Input and output linear frames
         casacore::Vector<casacore::String> output_units = output_csys.worldAxisUnits();
         casacore::Vector<casacore::Double> world_point_value(output_csys.nWorldAxes(), 0);
-        world_point_value(indices[0]) = world_point[0].get(output_units(indices[0])).getValue();
-        world_point_value(indices[1]) = world_point[1].get(output_units(indices[1])).getValue();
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
Tested with matched uvamp and uvphase images, region statistics works fine. Also tested with other matched regular images and no odd things are observed.

veggiesaurus (approved):

confluence (approved):

pford (approved):

PR #997, opened 2022-01-10
Stop using namespace std #932

### LINKED ISSUE #932: Stop using namespace std
We currently import the `std` namespace in several `.cc` files. I suggest eliminating this everywhere, because:

1. It means importing a lot of generic symbol names which are difficult to distinguish at a glance within the code.
2. It may lead to unpredictable behaviour, since certain mathematical functions may silently resolve to either a C function or an STL function depending on what headers have been imported.
3. Even when `using namespace std;` is declared in a file, we often use the fully qualified names anyway, since this is a common convention. So our code is inconsistent.
4. `std::` isn't a long string.

Are there good reasons to *keep* importing the namespace?

jolopezl:
This pull request closes #932. With these changes, we stop importing all symbols from `std::`. A couple of additional changes include a catch statement (as a consequence of removing `std::`) and removing a TBB definition (which isn't used anymore, just as a minor catch while editing the code).

### DIFF
--- a/src/DataStream/Compression.cc
+++ b/src/DataStream/Compression.cc
@@ -17,10 +17,8 @@
 #include <x86intrin.h>
 #endif
 
-using namespace std;
-
-int Compress(vector<float>& array, size_t offset, vector<char>& compression_buffer, size_t& compressed_size, uint32_t nx, uint32_t ny,
-    uint32_t precision) {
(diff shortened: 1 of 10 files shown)


confluence (approved):
This looks OK to me.

veggiesaurus (approved):

Issue #959, opened 2021-11-10
Backend refactoring


confluence:
This epic collects all outstanding issues related to backend code refactoring and code style.

pford:
I just ran across the backend Timer class which was modified slightly in a PR but does not seem to be used in the code.  Should this be added to the list (use-it-or-lose-it)?  It would make the performance logging code throughout the backend look nicer but adds overhead for map access.

And speaking of maps, should we prefer map::find over map::count to check for a given key and make this consistent overall?

confluence:
Agreed on both counts -- we should make a list of classes and functions which are no longer used. And I think `find` is more idiomatic for checking for a key.

pford:
I stand corrected, the carta::Timer will be used by the new C++ ICD tests when performance is enabled.

PR #998, opened 2022-01-11
Rearrange source code structure

### LINKED ISSUE #933: Use carta namespaces consistently
We currently don't consistently put all of our classes in the `carta` namespace. We should decide what structure the namespace should have (should there be nested namespaces corresponding to subdirectories?) and apply this system consistently.

jolopezl:
This should close issue #933, which is part of the [backend refactoring](https://github.com/CARTAvis/carta-backend/issues/959). Several changes are included moving files from one subdirectory to another (see discussion in #933), and adding almost everything into the `carta` namespace.


CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ INCLUDE_DIRECTORIES(${HDF5_INCLUDE_DIR})
 ADD_SUBDIRECTORY(carta-protobuf)
 ADD_SUBDIRECTORY(carta-scripting-grpc)
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})
-INCLUDE_DIRECTORIES(/opt/carta-casacore/include /opt/carta-casacore/include/casacore /opt/carta-casacore/include/casacode)
+INCLUDE_DIRECTORIES(SYSTEM /opt/carta-casacore/include /opt/carta-casacore/include/casacore /opt/carta-casacore/include/casacode)
```
veggiesaurus: what does the `SYSTEM` label do?

jolopezl: Sorry, I didn't mean to push that change. I just use it while developing to better look at what interests me. Here, `SYSTEM` turns off the warnings from external libraries. It's very useful since get a lot of warning clutter from casacore.

jolopezl: I just dropped this change. I like adding SYSTEM while working on something particular, but I think it's better to always keep all those warnings visible.

confluence (approved):

markccchiang (approved):

markccchiang:
It looks OK to me.

pford (changes requested):
Thanks for reviewing the backend code and its organization.  This looks much better.

My recommendation is to change FileSettings -> CursorSettings.  This was named generically when it also used to manage ImageBounds, for when the user zoomed in and out quickly and we only needed the last setting, similar to setting the cursor.  I would also move this to the Session directory since it is only used by Session and OnMessageTask in that directory.

That leaves ProgramSettings in the Settings directory, which is mostly used by Main (global settings, as Adrianna pointed out in #933) then handed to SessionManager, which only uses 5 of the settings.  Could we create a SessionManager/SessionSettings struct for the needed values only, to avoid confusion as to why per-app settings are in SessionManager?

WebBrowser in Util/ seems odd to me with the other files there, which are header files and, well, utilities for higher-level code.  ProgramSettings and WebBrowser are only used by Main (after removing ProgramSettings from SessionManager as suggested).  Perhaps group these together in a renamed Settings directory (since WebBrowser is not a setting)?

confluence:
Perhaps we should have a `Main` subdirectory for `Main.cc`, `ProgramSettings` and `WebBrowser`?

I agree that `FileSettings` should be renamed and moved -- I missed that!

kswang1029:
I tested this PR with our e2e tests currently available and they run well. I also did some manual tests and did not see obvious regressions.

jolopezl:
@pford, thanks for the code review.

> Thanks for reviewing the backend code and its organization. This looks much better.
> 
> My recommendation is to change FileSettings -> CursorSettings. This was named generically when it also used to manage ImageBounds, for when the user zoomed in and out quickly and we only needed the last setting, similar to setting the cursor. I would also move this to the Session directory since it is only used by Session and OnMessageTask in that directory.

done.

> 
> That leaves ProgramSettings in the Settings directory, which is mostly used by Main (global settings, as Adrianna pointed out in #933) then handed to SessionManager, which only uses 5 of the settings. Could we create a SessionManager/SessionSettings struct for the needed values only, to avoid confusion as to why per-app settings are in SessionManager?

I agree. Can we handle this in a separate issue? Further changes in that struct shouldn't produce too many conflicts and will be easy to treat. There are other places as well that need better separation.

> 
> WebBrowser in Util/ seems odd to me with the other files there, which are header files and, well, utilities for higher-level code. ProgramSettings and WebBrowser are only used by Main (after removing ProgramSettings from SessionManager as suggested). Perhaps group these together in a renamed Settings directory (since WebBrowser is not a setting)?

I agree. Then, I will regroup Main, WebBrowser, and ProgramSettings (to be refactored separately) in a new folder.

jolopezl:
@confluence, thanks for looking at the changes once again.

> Perhaps we should have a `Main` subdirectory for `Main.cc`, `ProgramSettings` and `WebBrowser`?
> 
> I agree that `FileSettings` should be renamed and moved -- I missed that!

I will move Main, WebBrowser, and ProgramSettings to another folder.

confluence (approved):

veggiesaurus (approved):

veggiesaurus:
@jolopezl just have some merge conflicts to resolve and then we'll be good to go once @confluence signs off on it :)

confluence:
It looks like the merge has introduced a namespace-related build error.

confluence:
Now it needs a formatting pass. :)

pford (approved):

PR #1003, opened 2022-01-17
Avoid the file id repetition in generated moments or PV images

markccchiang:
This bug was found by @YuHsuan-Hwang. The file id of the generated moment image is repeated with the PV image if the number of moment type is 0 [(PV generator file id def)](https://github.com/CARTAvis/carta-backend/blob/dev/src/ImageGenerators/PvGenerator.cc#L29). So I increase the file id for moments images by 1 to solve this problem.


veggiesaurus (approved):

kswang1029 (approved):
@YuHsuan-Hwang thanks for spotting this issue. @markccchiang thanks for fixing this proactively. Tested with multiple cubes loaded and generated PV each and moments -1 and 0 to verify the fix. All working properly. 👍

PR #1008, opened 2022-01-24
Add casa_data_autoupdate bash script

### LINKED ISSUE #961: auto update casa measure data
based on https://casa.nrao.edu/casadocs/casa-6.1.0/external-data/casa-data-repository it is suggested to update casa measure data _regularly_. The above URL contains the information on how to get it updated.

jolopezl:
This is one part of https://github.com/CARTAvis/carta-backend/issues/961.

To do:

- [x] Complete this description. (now see below: https://github.com/CARTAvis/carta-backend/pull/1008#issuecomment-1092547670)
- [x] Improve the time interval calculation.
- [x] Hook this script to the main CARTA execution script (`scripts/carta`).
- [x] ~Define when and where to export CASAPATH to a new place.~ (not required)
- [ ] If we read key-pair values from JSON, add these to the schema (will add these changes once this gets approved.)


veggiesaurus:
we need to consider a few things here:
- In multi-user systems (e.g. the controller), we shouldn't update casacore data in multiple locations (e.g. each and every user directory).
- When using the controller, the preferences.json file is never used (preferences are stored in a database)
- Ubuntu users have the option of simply using the casacore-data package from KERN in order to keep their data up to date.
- Only a subset of users actually need up-to-date measures data

Personally, I think we could just keep this as a script for users to run at their convenience, but not sure how @kswang1029 feels about that. Another option is to add updated RPMs and homebrew packages that link to the latest casacore data

jolopezl:
> we need to consider a few things here:
> 
> * In multi-user systems (e.g. the controller), we shouldn't update casacore data in multiple locations (e.g. each and every user directory).
> * When using the controller, the preferences.json file is never used (preferences are stored in a database)

In the script auto-update starts disabled by default and only changes if `preferences.json` is found in `$HOME/.carta/config`. Then, it finishes and nothing happens.

> * Ubuntu users have the option of simply using the casacore-data package from KERN in order to keep their data up to date.

Ok, but that requires sudo access (they might have... or maybe not). We thought that it was ok to keep the auto-update this way, it's only 50 to 51 Mb of data.

> * Only a subset of users actually need up-to-date measures data

Those can enable the feature, as proposed using the front-end preferences panel. 

> 
> Personally, I think we could just keep this as a script for users to run at their convenience, but not sure how @kswang1029 feels about that. Another option is to add updated RPMs and homebrew packages that link to the latest casacore data

Ok, but can we move that part of the discussion to the [issue](https://github.com/CARTAvis/carta-backend/issues/961)?

veggiesaurus:
> Ok, but can we move that part of the discussion to the [issue](https://github.com/CARTAvis/carta-backend/issues/961)?

good idea

ajm-ska:
Or maybe I should post this here:

- The current plan is that the `carta` script executes `casa_data_autoupdate` every time (Note the Debian and RPM packages will need to copy `casa_data_autoupdate` to /usr/bin/`).

- `casa_data_autoupdate` only takes action if the user has manually added the following lines to their `~/.carta/config/preferences.json` file.  Instructions about this would be added to the User Manual. It is still undecided whether there would be a carta-frontend option to control this.
```
    "auto_update_casa_data": true,
    "auto_update_casa_data_interval": 30
```

- It relies on the change to carta-casacore that checks for the presence of a `~/.carta/casarc` file. When present, this file overwrites the default `/usr/share/casacore/data/geodetic` location and points towards a user-accessible location:
```
measures.observatory.directory: ~/.carta/data/geodetic
```
-  From testing, CARTA appears to work perfectly fine with only `geodetic/Observatories` present. I'm not sure if the rest of the geodetic folder and all of the ephemerides folder are actually needed.

-  The only contents of casa-data that appear to regularly update are `geodetic/TAI_UTC`. Does CARTA use that?

- As the old ` https://svn.cv.nrao.edu/svn/casa-data` repository will apparently become private soon, we need to get the updated measures data elsewhere. Here we can get it from `ftp://ftp.astron.nl/outgoing/Measures/WSRT_Measures.ztar` using wget. That might be good because I imagine it is more likely for `wget` to be installed on a computer rather than `svn`.  However, it appears even though `WSRT_Measures.ztar` is updated daily, only the contents of `geodetic/TAI_UTC` is actually updated.
It is unfortunate that we need to duplicate most of the data by downloading the entire 19MB file. 
From the NRAO SVN repo, we could just download the contents of `geodetic/Observatories` and `geodetic/TAI_UTC` which are only a few kB.

- The "datediff" function needs to be improved.

- One shortcoming of this entire scheme is that it will not work for beta releases as they use `~/.carta-beta` and we have built carta-casacore  with`~/.carta/casarc` hard-coded. We could easily change carta-casacore when building the macOS Electron and Linux AppImage versions because they bundle separate casacore lib files, but it would be more difficult to have a separate carta-casacore version for the Debian and RPM versions, and probably not worth it. The CARTA beta versions are not intended for long-term use anyway, so it probably isn't a problem.

confluence (commented):
1. If the public svn repo is going to go away, I think we shouldn't depend on it at all, and instead look for either wget (which I believe is installed on most Linux distros by default) or curl (which I believe is installed on Macs by default), and otherwise instruct the user to install one or the other (if they have the preference set).
2. I would like to make sure that it's *possible* for someone to use this solution with a beta package (even if it means some hairy symlinking from inside `.carta-beta` to locations in `.carta`). This is unlikely to matter on Ubuntu because users can just use the KERN data package, but if there is no equivalent RPM solution available it may come up on RPM systems.

confluence:
Has anyone tested if it's possible to use this with a beta package (by setting up the appropriate symlinks)?

jolopezl:
Making symbolic links seem to work for me, but it requires several steps and instructions which we can avoid. As discussed in [carta-casacore](https://github.com/CARTAvis/carta-casacore/pull/21), it makes sense to move the data to `./carta-casacore/data`, and also this script, but that's open to discussion.

confluence:
@jolopezl @ajm I am reopening this so that we can add the change to the launcher script here (as discussed in [the carta-casacore PR](https://github.com/CARTAvis/carta-casacore/pull/22)).

It may be a good idea to add a check to call the update script only if the command can be found on the path, but maybe it's fine if we print an error if it can't be found.

ajm-ska:
@confluence @jolopezl The `casa_data_autoupdate` command was already in the `carta` script here, but we need to remove the old `scripts/casa_data_autoupdate` as that is now in carta-casacore.

Here I took your suggestion and it will only run the command if it exists:
```
if [ -x "$(command -v casa_data_autoupdate)" ]; then
    casa_data_autoupdate
fi
```
However, maybe it would be better to just print the error if the script is not found as that might help the user to debug problems if they are wishing to use the `casa_data_autoupdate` script. Although I don't mind either way.

jolopezl:
> @confluence @jolopezl The `casa_data_autoupdate` command was already in the `carta` script here, but we need to remove the old `scripts/casa_data_autoupdate` as that is now in carta-casacore.
> 
> Here I took your suggestion and it will only run the command if it exists:
> 
> ```
> if [ -x "$(command -v casa_data_autoupdate)" ]; then
>     casa_data_autoupdate
> fi
> ```
> 
> However, maybe it would be better to just print the error if the script is not found as that might help the user to debug problems if they are wishing to use the `casa_data_autoupdate` script. Although I don't mind either way.

I would keep it silent, but that's just a personal preference. If you want to print a message I would print something like `"CASA data auto-update is not available in this system"`.

confluence:
I would prefer to leave it completely silent; unfortunately the current solution means that we always run the script even if updates are not enabled, and printing a message may imply to the user that something is broken, or trying to do a download when it it's actually, etc..

confluence (approved):

jolopezl:
> I would prefer to leave it completely silent; unfortunately the current solution means that we always run the script even if updates are not enabled, and printing a message may imply to the user that something is broken, or trying to do a download when it it's actually, etc..

I agree, I vote for silent execution. If you want, we could check the JSON here for the update field existence, too (we can change this behavior in another iteration). I suggest we merge this as is. Still, I think that:
```
if [ -x "$(command -v casa_data_autoupdate)" ]; then
    casa_data_autoupdate
fi
```
is harmless if `casa_data_autoupdate` isn't in the PATH.

confluence:
> If you want, we could check the JSON here for the update field existence

I don't think there's any advantage to doing this; I would prefer to keep this launcher script as simple as possible until we move to a longer-term solution of doing it in the backend (which would be aware of its own config directory location).

ajm-ska (approved):

PR #1010, opened 2022-01-25
Replaced mutable with volatile where required.

robsimmonds:
Small fix that hopefully solves the crash problem when an animation stops and restarts immediately in the test code.


robsimmonds:
Looks like this is now being used in the animation test branch so does not need reviewing here.

robsimmonds:
> 

Actually maybe just get this into dev since this is similar than the other branch it has been merged into.

confluence (approved):

PR #1015, opened 2022-02-08
665 LEL image creation

### LINKED ISSUE #665: Create CASA LEL images dynamically
LEL images are created dynamically. The frontend will have an interface (similar to the CASA viewer) where one can enter the mathematical expression. The frontend will send the expression to the backend, which will create and open the LEL image.

https://casadocs.readthedocs.io/en/stable/notebooks/image_analysis.html?highlight=LEL#Lattice-Expression-Language

Requires [CARTAvis/carta-frontend#1264](https://github.com/CARTAvis/carta-frontend/issues/1264)

pford:
Closes #665 .

Added boolean lel_expr field to OPEN_FILE; when true, filename is an LEL expression.  FileLoader returns the ExprLoader which parses the expression and directory into a LatticeExprNode to create a new casacore::ImageExpr.  ImageExpr is used for doGetSlice (for applying LEL math) and for saving the image in CASA format as an LEL image.

Development complete for LEL expression support, pending frontend development and question about HDF5 support which fails in casacore::ImageExprParse.


veggiesaurus:
@pford I'm going to add a placeholder branch for testing this on the frontend. But can you give me an example of what you expect in terms of the following (@kswang1029 can chime in too):
- a LEL expression with a single image
- a LEL expression combining two images in the same directory
- a LEL expression combining images in _different_ directories (I'm quite happy if we just don't support this, personally)

kswang1029:
> @pford I'm going to add a placeholder branch for testing this on the frontend. But can you give me an example of what you expect in terms of the following (@kswang1029 can chime in too):
> 
> * a LEL expression with a single image

`'HD163296_CO_2_1.image.mom1.fits'*2
`
> * a LEL expression combining two images in the same directory

`'HD163296_CO_2_1.image.mom1.fits'+'HD163296_CO_2_1.image.mom1.fits'
`
> * a LEL expression combining images in _different_ directories (I'm quite happy if we just don't support this, personally)

`'HD163296_CO_2_1.image.mom1.fits'+'../HD163296_CO_2_1.image.mom1.fits'`

We should see identical results (pixel value doubled) with the above three LEL expressions.

pford:
> > * a LEL expression combining two images in the same directory
> 
> `'HD163296_CO_2_1.image.mom1.fits'+'HD163296_CO_2_1.image.mom1.fits' `
> 
I did some testing with the backend unit tests, and you do not need the quote marks unless the name contains something in the LEL syntax (`HD163296_CO_2_1.image.mom1.fits` is okay without quotes).  For example, `set_QA/image.mom1.fits` would try to divide image `set_QA` by image `image.mom1.fits` and the first "image" would fail, so you need `'set_QA/image.mom1.fits'`. With the backend test directory `test/data/images/fits` you can use the expression `noise_10px_10px.fits + '../casa/noise_10px_10px.im'` with quotes around the casa path only.

kswang1029:
@pford @veggiesaurus I have done some tests with the combined branches. Mostly this works nicely 👍 
Tested the following cases so far:
- absolute and relative paths 👍 
- simple expression such as +, /, sqrt. 👍 I will try others as highlighted from the user manual later on. 
- enhanced input with mouse click 👍 
- save the computed image as a new image file 👍 
- error handling of images with inconsistent shape 👍 

Few issues that could be addressed further (sorry I have to mix up both frontend and backend issues as I think they are related mostly):
1) (backend + frontend) it appears that we can have only one computed cube loaded at once. That is if we use the LEL expression to load one image and try to "append" another with LEL expression, the first one will be closed. The button in the file browser is "Load expression" instead of "Append expression" as the browser is launched via "Append image". 

2) (frontend) I suggest every time we launch the file browser via "Load image" or "Append image", we switch back to "List filtering" mode as I think it is the more commonly used mode.

veggiesaurus:
I suspect these are both front-end issues. I'll try sort them out today 🙂

kswang1029:
The two issues are fixed. 👍

kswang1029 (approved):
Have performed some basic tests and they work fine. I probably cannot test all cases that LEL can do.

kswang1029:
With https://github.com/CARTAvis/carta-casacore/pull/17, hdf5 images can be the input for LEL image generation (even mixed with other image formats). Verified image statistics and it seems working correctly.

Just to confirm, when an HDF5 image is loaded via LEL (eg, "a.hdf5" * 2), we ignore low resolution mip maps and use the full resolution tiles to perform LEL image generation, right? When requesting a full image statistics of a given channel, we ignore the pre-calculated channel statistics, right? That is, the LEL generated image behaves just like other image formats.

veggiesaurus:
> That is, the LEL generated image behaves just like other image formats.

That's my understanding, yes. I wonder if it should be reported as "FITS" type to the frontend, actually? So that we don't treat it like an HDF5 file with all the pre-computed stuff.

veggiesaurus:
@kswang1029 can you load a LEL expression combining different file types?

kswang1029:
> @kswang1029 can you load a LEL expression combining different file types?

yes. I tested a case like "a.fits" + "b.hdf5" and it works correctly.

pford:
> I wonder if it should be reported as "FITS" type to the frontend, actually? So that we don't treat it like an HDF5 file with all the pre-computed stuff.

casacore::ImageExpr is reported as CARTA::FileType::CASA.

I verified that HDF5-based LEL images call CartaHdf5Image::doGetSlice() to get the image data for rendering and calculations, so we could look into improving this method for better performance.  Currently, it always uses DATA for the HDF5Lattice set up at construction.

Issue #1002, opened 2022-01-15
Vector field data calculations


kswang1029:
This is paired with the frontend feature https://github.com/CARTAvis/carta-frontend/issues/1155

Please refer to the google doc https://docs.google.com/document/d/1S8nNvonlowNwX_uPoPspxugUEnw5wRxbAS6y2HScTzs/edit# for details.

veggiesaurus:
adjusting the story points to 5 because HDF5 tile cache needs to be taken into account

PR #1016, opened 2022-02-08
Support vector field calculations

markccchiang:
#1002, the backend calculates polarized intensities and polarized angles that form vectors. The downsampled data in the HDF5 file are used if it has `mip` value that fits the required smoothing factor. However, it will not be used to do another downsampling. For example, use `mip=2` data to calculate the required value `mip=10`. It seems the errors would increase significantly compared to downsampling from the full resolution data.


veggiesaurus:
> t seems the errors would increase significantly compared to downsampling from the full resolution data.

I thought @kswang1029 tested this and found it was ok?

markccchiang:
> 
> I thought @kswang1029 tested this and found it was ok?

Maybe it depends on the pixel values distribution in an image. Will check it in detail.

kswang1029:
> > I thought @kswang1029 tested this and found it was ok?
> 
> Maybe it depends on the pixel values distribution in an image. Will check it in detail.

Assuming we have an image of 1000 pixel x 1000 pixel and we create the following rebinned image:
1) with a bin factor of 10, resulting an image of 100 pixel x 100 pixel
2) with a bin factor of 2, resulting an image of 500 pixel x 500 pixel, then with another bin factor of 5, resulting 100 pixel x pixel
3) with a bin factor of 5, resulting an image of 200 pixel x 200 pixel, then with another bin factor of 2, resulting 100 pixel x pixel

As the rebin is done with "mean", the three 100 pixel x 100 pixel images from 1, 2, and 3 should be identical (within numerical error). Or did I miss something? 🤔

Note that the downsampling step has to happen on Stokes Q and Stokes U images (and Stokes I if fractional). Then we use the downsampled Stokes images to compute polarized intensity and angle. We should not apply downsampling directly to the polarized intensity and angle images as it is mathematically incorrect.

markccchiang:
> As the rebin is done with "mean", the three 100 pixel x 100 pixel images from 1, 2, and 3 should be identical (within numerical error). Or did I miss something? 🤔
> 
> Note that the downsampling step has to happen on Stokes Q and Stokes U images (and Stokes I if fractional). Then we use the downsampled Stokes images to compute polarized intensity and angle. We should not apply downsampling directly to the polarized intensity and angle images as it is mathematically incorrect.

I have added a tester to check the error of downsampled data from smaller downsampled data in an HDF5 image. The larger error (from 1e-6 to 0.3) happens only when an image in `mip X mip` block contains several NaN pixels. @kswang1029 if this error is acceptable, then I will apply this way for HDF5 images (?)

kswang1029:
> > As the rebin is done with "mean", the three 100 pixel x 100 pixel images from 1, 2, and 3 should be identical (within numerical error). Or did I miss something? 🤔
> > Note that the downsampling step has to happen on Stokes Q and Stokes U images (and Stokes I if fractional). Then we use the downsampled Stokes images to compute polarized intensity and angle. We should not apply downsampling directly to the polarized intensity and angle images as it is mathematically incorrect.
> 
> I have added a tester to check the error of downsampled data from smaller downsampled data in an HDF5 image. The larger error (from 1e-6 to 0.3) happens only when an image in `mip X mip` block contains several NaN pixels. @kswang1029 if this error is acceptable, then I will apply this way for HDF5 images (?)

Hmm… I will investigate this and report back.

veggiesaurus:
> > As the rebin is done with "mean", the three 100 pixel x 100 pixel images from 1, 2, and 3 should be identical (within numerical error). Or did I miss something? 🤔
> > Note that the downsampling step has to happen on Stokes Q and Stokes U images (and Stokes I if fractional). Then we use the downsampled Stokes images to compute polarized intensity and angle. We should not apply downsampling directly to the polarized intensity and angle images as it is mathematically incorrect.
> 
> I have added a tester to check the error of downsampled data from smaller downsampled data in an HDF5 image. The larger error (from 1e-6 to 0.3) happens only when an image in `mip X mip` block contains several NaN pixels. @kswang1029 if this error is acceptable, then I will apply this way for HDF5 images (?)

Right, I think this is what @kswang1029 and I expected. As soon as you have NaN pixels, the downsampling is no longer separable. I think we concluded that it would only affect areas _near_ NaNs and was therefore not a problem.

The smaller errors are just numeric errors and can be ignored, since neither is more correct than the other.

ajm-ska:
@markccchiang I know it might be hard to spot due to the annoying unit test issue with MacOS, but it appears the `VectorFieldTest.TestZFPCompression` unit test fails on Ubuntu 18.04.

> [  FAILED  ] 1 test, listed below:
> [  FAILED  ] VectorFieldTest.TestZFPCompression

The error seems to be:

```
   File "/root/carta-backend/build/test/bin/make_image.py", line 177, in <module>
    make_image(args)
  File "/root/carta-backend/build/test/bin/make_image.py", line 54, in make_image
    rng = np.random.default_rng(args.seed)
AttributeError: module 'numpy.random' has no attribute 'default_rng'
2022-03-04 03:47:02	INFO	FITSCoordinateUtil::fromFITSHeader	passing empty or nonexistant spectral Coordinate axis
/root/carta-backend/test/TestVectorField.cc:1164: Failure
Expected: (count_pi) > (0), actual: 0 vs 0
/root/carta-backend/test/TestVectorField.cc:1175: Failure
Expected: (count_pa) > (0), actual: 0 vs 0
```

Do you think the numpy version that I'm using on Ubuntu 18.04 is too old?

```
root@b93969b619f7:~/carta-backend/build/test# python3 --version
Python 3.6.9
root@b93969b619f7:~/carta-backend/build/test# python3
Python 3.6.9 (default, Jan 26 2021, 15:33:00) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> numpy.version.version
'1.13.3'
>>> 
```

markccchiang:
> 
> Do you think the numpy version that I'm using on Ubuntu 18.04 is too old?

@ajm-asiaa I think you are right. The numpy version is too old. Which can refer to this [stack overflow discussions](https://stackoverflow.com/questions/62077194/can%C2%B4t-import-qiskit-attribute-error-in-numpy-numpy-random-has-no-attribute).

ajm-ska:
@markccchiang OK. I have just updated the NumPy version in the Ubuntu 18.04 CI Docker container. 
Now it uses NumPy version '1.19.5'. 
I guess we do not have any minimum supported NumPy version @confluence?, particularly as NumPy is only used by the unit tests. Anyway, it shouldn't be an issue for much longer as we will be dropping support for Ubuntu 18.04 soon.

I just noticed a new issue on the macOS11 CI unit tests. During the last run the `macOS 11 / TestHdf5DownSampledData – VectorFieldTest` test failed. It appears it calculated the wrong values for some reason:

```
CommonTestUtilities.cc:263: Failure
The difference between data1 and data2 is 0.0051380693912506104, which exceeds abs_err, where
data1 evaluates to 0.18984830379486084,
data2 evaluates to 0.18471023440361023, and
abs_err evaluates to 9.9999997473787516e-06.
```

But strangely, after re-running the tests on the same commit, the test passed. Maybe it is related the ongoing HDF5 unit tests issue on macOS.

kswang1029:
@markccchiang @veggiesaurus these might be the backend issue too.

[backend / frontend] possibly there are vector components rendered on top of empty pixels (ie outside the image) or on top of NaN pixels. 

[frontend / backend] I see the default threshold is 0 and is set in the SET_VECTOR_OVERLAY_PARAMETERS message and the backend applies it to the vector field data calculation. The default threshold should be NaN so that the backend does not apply a threshold in the calculations.

veggiesaurus:
> @markccchiang @veggiesaurus these might be the backend issue too.
> 
> [backend / frontend] possibly there are vector components rendered on top of empty pixels (ie outside the image) or on top of NaN pixels. 
> 
> [frontend / backend] I see the default threshold is 0 and is set in the SET_VECTOR_OVERLAY_PARAMETERS message and the backend applies it to the vector field data calculation. The default threshold should be NaN so that the backend does not apply a threshold in the calculations. 
> 

I think the first is because the average position of a group may lie on top of a NaN, and if you block average a 13x13 pixel image by 4x4, then the center of the last block will actually lie outside the image

kswang1029:
> > @markccchiang @veggiesaurus these might be the backend issue too.
> > [backend / frontend] possibly there are vector components rendered on top of empty pixels (ie outside the image) or on top of NaN pixels.
> > [frontend / backend] I see the default threshold is 0 and is set in the SET_VECTOR_OVERLAY_PARAMETERS message and the backend applies it to the vector field data calculation. The default threshold should be NaN so that the backend does not apply a threshold in the calculations.
> 
> I think the first is because the average position of a group may lie on top of a NaN, and if you block average a 13x13 pixel image by 4x4, then the center of the last block will actually lie outside the image

Yes the root cause is understandable. Please consider this as low priority or none issue. Usually we apply a threshold so this “issue” is less important.

markccchiang:
@veggiesaurus is it possible that the frontend receives PA values in the unit of `degrees` rather than rads? Since according to @kswang1029 the default unit of the image pixel is `degrees` if we take it as the polarized angle.

veggiesaurus:
I'll make the change to assume degrees

src/Util/Image.h
```diff
@@ -124,4 +129,48 @@ static std::unordered_map<std::string, CARTA::PolarizationType> StokesStringType
...
+        threshold = std::numeric_limits<double>::quiet_NaN();
+        debiasing = false;
+        q_error = 0;
+        u_error = 0;
+        stokes_intensity = -1;
+        stokes_angle = -1;
+        compression_type = CARTA::CompressionType::NONE;
+        compression_quality = 0;
+    }
+};
```
jolopezl: Could you consider creating `VectorFieldSettings.h` and `VectorFieldSettings.cc`?. This struct is just used in Frame, so it should not just be dropped as a utility struct. Instead, I suggest you create these two files in Frame. This will help with the readability and maintainability of the code.

src/Util/Image.cc
```diff
@@ -21,3 +21,29 @@ CARTA::PolarizationType GetStokesType(int stokes_value) {
...
+    tiles.resize(num_tile_rows * num_tile_columns);
+
+    for (int i = 0; i < num_tile_columns; ++i) {
+        for (int j = 0; j < num_tile_rows; ++j) {
+            tiles[j * num_tile_columns + i].x = i;
+            tiles[j * num_tile_columns + i].y = j;
+            tiles[j * num_tile_columns + i].layer = tile_layer;
+        }
+    }
+}
```
jolopezl: This is only used in Frame in one single place. Do you really need a different function somewhere else? If this is not a common utility function, I suggest you move it to Frame (or maybe Tile.h?). You are actually passing two parameters which are members of Frame.

Also, this is used only in `Frame::VectorFieldImage`, which is already a very long method. There is no harm with just placing it there unless you want to test this function in isolation, for example.

src/Util/Message.cc
```diff
@@ -406,3 +426,28 @@ void FillStatistics(CARTA::RegionStatsData& stats_data, const std::vector<CARTA:
...
+            std::vector<char> compression_buffer;
+            size_t compressed_size;
+            int precision = lround(compression_quality);
+            carta::Compress(array, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
+            tile->set_image_data(compression_buffer.data(), compressed_size);
+        } else {
+            tile->set_image_data(array.data(), sizeof(float) * array.size());
+        }
+    }
+}
```
jolopezl: Idem, this is only used in `Frame`. No need to bring it to the utility functions set. I suggest moving this and the other functions to Frame or maybe creating a FrameHelperFunctions header and source file.

src/Frame/Frame.cc
```diff
@@ -2181,4 +2189,292 @@ void Frame::CloseCachedImage(const std::string& file) {
     }
 }
 
+bool Frame::SetVectorOverlayParameters(const CARTA::SetVectorOverlayParameters& message) {
+    VectorFieldSettings new_settings = {(int)message.smoothing_factor(), message.fractional(), message.threshold(), message.debiasing(),
+        message.q_error(), message.u_error(), message.stokes_intensity(), message.stokes_angle(), message.compression_type(),
+        message.compression_quality()};
```
jolopezl: It would look neat if you could do it from a VectorFieldSettings constructor:
```
VectorFieldSettings settings(message);
```

src/Frame/Frame.cc
```diff
@@ -2181,4 +2189,292 @@ void Frame::CloseCachedImage(const std::string& file) {
...
+
+    if (_vector_field_settings != new_settings) {
+        _vector_field_settings = new_settings;
+        return true;
+    }
+    return false;
+}
+
+bool Frame::VectorFieldImage(VectorFieldCallback& partial_vector_field_callback) {
```
jolopezl: Not compulsory at all, but It almost looks like this could be moved to an isolated struct/class (VectorField?), is that possible? I think that would make its purpose more readable than inside Frame.

jolopezl (changes requested):
This is quite a significant change. It looks good. Still, I suggest a few changes, primarily based on the last codebase restructuring and making the code more readable and maintainable (too many things happening in Frame). A few functions could be better located, and `VectorFieldSettings` can have its own location/file (it could also be `VectorField`).

markccchiang:
@jolopezl Thank you for the suggestions! Please check do new code changes fit your expectations. As for the last comment, I am trying a good way to refactor the `Frame::VectorFieldImage` function and move it to an isolated class. Will push another commit if it is done.

src/Util/Image.cc
```diff
@@ -21,3 +21,13 @@ CARTA::PolarizationType GetStokesType(int stokes_value) {
...
+
+CARTA::ImageBounds GetImageBounds(const carta::Tile& tile, int image_width, int image_height, int mip) {
+    int tile_size_original = TILE_SIZE * mip;
+    CARTA::ImageBounds bounds;
+    bounds.set_x_min(std::min(std::max(0, tile.x * tile_size_original), image_width));
+    bounds.set_x_max(std::min(image_width, (tile.x + 1) * tile_size_original));
+    bounds.set_y_min(std::min(std::max(0, tile.y * tile_size_original), image_height));
+    bounds.set_y_max(std::min(image_height, (tile.y + 1) * tile_size_original));
+    return bounds;
+}
```
jolopezl: I just checked and this is only used in Frame too. Maybe move it there too as you did with GetTiles?

src/Util/Image.h
```diff
@@ -124,4 +129,9 @@ static std::unordered_map<std::string, CARTA::PolarizationType> StokesStringType
 int GetStokesValue(const CARTA::PolarizationType& stokes_type);
 CARTA::PolarizationType GetStokesType(int stokes_value);
 
+// Polarization vector field callback (intensity_tile, angle_tile, progress)
+using VectorFieldCallback = const std::function<void(CARTA::VectorOverlayTileData&)>;
```
jolopezl: Should go in VectorFieldSettings or just somewhere in Frame.h.

src/Frame/VectorFieldSettings.h
```diff
@@ -0,0 +1,70 @@
...
+
+#ifndef CARTA_BACKEND__FRAME_VECTORFIELDSETTINGS_H_
+#define CARTA_BACKEND__FRAME_VECTORFIELDSETTINGS_H_
+
+#include <cmath>
+
+#include <carta-protobuf/enums.pb.h>
+#include <carta-protobuf/vector_overlay.pb.h>
+
+struct VectorFieldSettings {
```
jolopezl, line 15: Great! Could you please create VectorFieldSettings.cc just with
```
/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
   Copyright 2018, 2019, 2020, 2021 Academia Sinica Institute of Astronomy and Astrophysics (ASIAA),
   Associated Universities, Inc. (AUI) and the Inter-University Institute for Data Intensive Astronomy (IDIA)
   SPDX-License-Identifier: GPL-3.0-or-later
*/

#include "VectorFieldSettings"
```
then add it to CMakeLists.txt. That way the compiler will process it as a standalone unit and show problems more clearly and not inside Frame (See for instance https://github.com/CARTAvis/carta-backend/blob/dev/src/DataStream/Tile.cc). This is something we should probably do with other pieces of the code, for example, CountorSettings in Frame.h should be extracted, but that's for another issue.

jolopezl (changes requested):
Thanks for taking care of those suggestions. It looks great. I have a few more minor comments though.

src/Frame/Frame.cc
```diff
@@ -2181,4 +2189,345 @@ void Frame::CloseCachedImage(const std::string& file) {
...
+    int32_t tile_layer = carta::Tile::MipToLayer(mip, image_width, image_height, TILE_SIZE, TILE_SIZE);
+    tiles.resize(num_tile_rows * num_tile_columns);
+
+    for (int i = 0; i < num_tile_columns; ++i) {
+        for (int j = 0; j < num_tile_rows; ++j) {
+            tiles[j * num_tile_columns + i].x = i;
+            tiles[j * num_tile_columns + i].y = j;
+            tiles[j * num_tile_columns + i].layer = tile_layer;
+        }
+    }
```
jolopezl: One more comment. I am not 100% sure, but I think inverting the loop order will provide better CPU caching for traversing the vector. Now, the double for loop is accessing elements that aren't contiguous en each pass.

I suggest:
```
    for (int j = 0; j < num_tile_rows; ++j) {
        for (int i = 0; i < num_tile_columns; ++i) {
            tiles[j * num_tile_columns + i].x = i;
            tiles[j * num_tile_columns + i].y = j;
            tiles[j * num_tile_columns + i].layer = tile_layer;
        }
    }
```

jolopezl (changes requested):
Just one more suggestion!

src/Frame/Frame.h
```diff
@@ -203,6 +208,17 @@ class Frame {
...
 
+    // Polarization vector field
+    bool SetVectorOverlayParameters(const CARTA::SetVectorOverlayParameters& message);
+    inline VectorFieldSettings& GetVectorFieldParameters() {
+        return _vector_field_settings;
+    };
+    inline void ClearVectorFieldParameters() {
+        _vector_field_settings.ClearSettings();
+    };
+    bool GetDownSampledRasterData(std::vector<float>& data, int& down_sampled_width, int& down_sampled_height, int z, int stokes,
```
jolopezl: Downsampling in one word: `GetDownSampledRasterData -> GetDownsampledRasterData `

src/Frame/VectorFieldCalculator.cc
```diff
@@ -0,0 +1,296 @@
...
+*/
+
+#include "VectorFieldCalculator.h"
+
+#include "DataStream/Compression.h"
+
+namespace carta {
+
+bool VectorFieldCalculator::DoCalculations(VectorFieldCallback& callback) {
+    if (!frame) {
```
jolopezl: Is this enough? I assume that before the vector field calculator was running in a valid frame, should you check `frame->IsValid()`?.

jolopezl (approved):
Thanks for addressing all those requests. It looks great. I still have two tiny comments that you could handle or not.

src/Frame/VectorFieldCalculator.cc
```diff
@@ -0,0 +1,296 @@
...
+            response.set_progress(progress);
+            callback(response);
+        }
+        return true;
+    }
+
+    // Consider the case if an image has stokes axis
+
+    // Get Stokes I, Q, and U indices
+    int stokes_i, stokes_q, stokes_u;
```
confluence: There's some repetition in this part of the function which could be eliminated if the indices, the boolean flags determining whether they are required, and the data vectors, were stored in (unordered) maps with the same string names (`"Ix"`, etc.) as keys. Then you could loop over the keys to get the required indices, then later to get the data.

src/Frame/VectorFieldCalculator.cc
```diff
@@ -0,0 +1,296 @@
...
+            }
+            return std::numeric_limits<float>::quiet_NaN();
+        };
+
+        auto calc_pa = [&](float q, float u) {
+            if (!std::isnan(q) && !isnan(u)) {
+                return (float)(180.0 / casacore::C::pi) * atan2(u, q) / 2;
+            }
+            return std::numeric_limits<float>::quiet_NaN();
+        };
```
confluence: Here I have the same suggestion as in #948 -- perhaps a generic lambda which does the `isnan` check and calls another lambda. But it may be better to leave this for another PR.

src/Frame/VectorFieldCalculator.cc
```diff
@@ -0,0 +1,296 @@
...
+                    downsampled_data, downsampled_width, downsampled_height, channel, CURRENT_STOKES, bounds, mip)) {
+                return false;
+            }
+            // Apply a threshold cut
+            apply_threshold_on_data(downsampled_data);
+        }
+
+        // Lambda functions for calculating PI, fractional PI, and PA
+        auto calc_pi = [&](float q, float u) {
+            if (!std::isnan(q) && !isnan(u)) {
```
confluence: Why is the second `isnan` not a `std::isnan` in all these functions?

src/Session/Session.cc
```diff
@@ -1426,6 +1429,14 @@ void Session::OnStopPvCalc(const CARTA::StopPvCalc& stop_pv_calc) {
     }
 }
 
+void Session::OnSetVectorOverlayParameters(const CARTA::SetVectorOverlayParameters& message) {
+    if (_frames.count(message.file_id())) {
+        if (_frames.at(message.file_id())->SetVectorOverlayParameters(message)) {
```
confluence: This could be flattened into a single `&&` conditional.

src/Session/Session.cc
```diff
@@ -1861,6 +1872,41 @@ void Session::RegionDataStreams(int file_id, int region_id) {
...
+            partial_response.set_file_id(file_id);
+            SendFileEvent(file_id, CARTA::EventType::VECTOR_OVERLAY_TILE_DATA, 0, partial_response);
+        };
+
+        // Do PI/PA calculations
+        VectorFieldCalculator vector_field_calculator(frame);
+        if (vector_field_calculator.DoCalculations(callback)) {
+            return true;
+        }
+        SendLogEvent("Error processing vector filed image", {"vector field"}, CARTA::ErrorSeverity::WARNING);
```
confluence: Typo: filed -> field

src/Frame/VectorFieldCalculator.cc
```diff
@@ -0,0 +1,296 @@
...
+        };
+
+        auto calc_pa = [&](float q, float u) {
+            if (!std::isnan(q) && !isnan(u)) {
+                return (float)(180.0 / casacore::C::pi) * atan2(u, q) / 2;
+            }
+            return std::numeric_limits<float>::quiet_NaN();
+        };
+
+        auto set_nan_to_result = [&](float i, float result) {
```
confluence: This name is unclear. Maybe something like `apply_threshold`?

test/TestVectorField.cc
```diff
@@ -0,0 +1,1630 @@
...
+        BlockSmooth(
+            stokes_data_u.data(), downsampled_u.data(), image_width, image_height, downsampled_width, downsampled_height, 0, 0, mip);
+
+        // Reset Q and U errors as 0 if debiasing is not used
+        if (!debiasing) {
+            q_error = u_error = 0;
+        }
+
+        // Calculate PI, FPI, and PA
+        auto calc_pi = [&](float q, float u) {
```
confluence: A lot of this code is duplicated in the test and in the source. If the same calculations have to be used in both places, they should be factored out into common functions which both the source and the test can call.

confluence (changes requested):

test/TestVectorField.cc
```diff
@@ -0,0 +1,1630 @@
...
+
+static const std::string IMAGE_SHAPE = "1110 1110 25 4";
+static const std::string IMAGE_OPTS = "-s 0";
+static const std::string IMAGE_OPTS_NAN = "-s 0 -n row column -d 10";
+
+class TestFrame : public Frame {
+public:
+    TestFrame(uint32_t session_id, std::shared_ptr<carta::FileLoader> loader, const std::string& hdu, int default_z = DEFAULT_Z)
+        : Frame(session_id, loader, hdu, default_z) {}
+    FRIEND_TEST(VectorFieldTest, ExampleFriendTest);
```
confluence: This seems to be copied from another file. There is no `ExampleFriendTest` in this test.

test/TestVectorField.cc
```diff
@@ -0,0 +1,1630 @@
+/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
```
confluence: I am concerned that this test is adding too much complexity and custom code, and testing these features at too high a level (rather than testing individual low-level functions). In some cases I understand that it's difficult to call these functions cleanly from a test, and I think that we need to dedicate some time to creating a framework for writing low-level tests more effectively.

I would like to split out this test code into a separate PR -- I'm generally happy with the rest of the code, but I think that we should reevaluate the testing approach, and I don't want that to hold up merging this into the coming release.

kswang1029:
@confluence I requested myself as a reviewer. Once code review and refactoring are done I will have one final quick test.

confluence (commented):
I think that this is OK to merge after final approval by @kswang1029 .

kswang1029:
@markccchiang could you solve the merge conflict please?

kswang1029 (approved):

confluence (approved):

PR #1018, opened 2022-02-11
ajm/docker jenkinsfile4

ajm-ska:
Here is the updated Jenkinsfile.


ajm-ska:
It runs Docker containers for RHEL7, RHEL8, Ubuntu 18.04/20.04/21.04, and runs natively on an MacOS 11 Intel Mac Mini, and a MacOS 12 M1 Mac Mini.

Unit-tests are still problematic on MacOS and frequently fail, so I made a temporary modification that would attempt to re-run the unit tests on MacOS up to 3 times if they fail. I didn't make such a change to Linux unit-tests as I guess it is a good idea to catch when unit-tests fail because ideally they should be robust.

When all ICD tests pass, the total run-time now appears to be ~32 minutes. I imagine this will decrease when we convert more and more ICD tests to unit-tests. 
Another thing is that currently all 5 Linux Docker containers are running on the same server. Later I plan to divide them between at least 2 servers. I did a test before and when there are fewer simultaneous Docker containers running on the same server, they can run a bit faster. I expect it could make it about 25 minutes.

Once Ubuntu 22.04 is out, I can replace the Ubuntu 21.10.

The only downside I see, if that when the Jenkins sends the results back to Github now, the "ICD tests" stage appear to be  a single stage. So you would need to click through to the Jenkins interface to investigate which particular ICD tests may have failed. But I guess you would need to do that anyway in order to inspect the log output. I couldn't figure out any other way when using a Declarative Pipeline as the Jenkinsfile has a limited length.

veggiesaurus (approved):

PR #1020, opened 2022-02-11
796 spatial profile from line regions

### LINKED ISSUE #796: support spatial profile extraction from a line/polyline region

### LINKED ISSUE #1038: An offset in the offset axis of a PV image generated with fixed angular separation
With a test wide field image cube, if the pv cut is sampled with fixed pixel separation, the origin of the offset axis is correct as
Gaussian_array_wide fits-Gaussian_array_wide_pv fits-image-2022-03-09-15-31-56

However, if we simply increase the pv cut of the above example so that it is sampled with fixed angular separation, there seems to have an offset in the offset axis of the pv image as
Gaussian_array_wide fits-Gaussian_array_wide_pv fits-image-2022-03-09-15-31-27

We can see the origin of the offset axis does not overlay with the peak emission of the central gaussian source.

@pford do you have any idea on this? Could it be due to the precision issue when converting pixel coordinate to world coordinate in such noticeably distorted sky field?

pford:
Closes #796 .
Closes #1038 

Implemented using existing code for PV images which approximates the line as a series of box regions to get mean of regions with current channel and stokes.  Added `width` field to `SpatialConfig` submessage (ignored for point regions).  Polyline regions are not supported for PV but are supported for spatial profiles.

For a line region, box regions are generated from the line end to the start, starting at the pixel center and working out.  Box regions for a polyline are also generated from end to start, starting at the end of each line segment instead of the center.  If the residual is less than 0.5*increment, the first box region in the next line segment is skipped.

The mean of each box region for the current channel and requested stokes is saved as the spatial profile.

Reusing spatial requirements and spatial data messages, but ignoring some fields:

- Spatial requirements for line regions: use file_id, region_id, SpatialConfig coordinate (stokes only, ignore x/y) and width; ignore start, end, mip.
- Spatial data for line regions: set file_id, region_id, channel, and stokes; ignore x, y, and value; SpatialProfile set raw_values and coordinate, ignore start, end, and mip.


pford:
Completed polyline support for fixed pixel and fixed angular images.

kswang1029:
> Completed polyline support for fixed pixel and fixed angular images.

@pford could you briefly describe how polyline is handled, especially near the "kinks"?

pford:
@kswang1029 I added more detail in the PR description.  My current implementation gets the mean of the "kink" points twice (once for each line in which it is an endpoint); but if required only once, what should the angle of its box region be?  The requirements doc for handling polylines did not specify what should be done where the lines join.

kswang1029:
> @kswang1029 I added more detail in the PR description. My current implementation gets the mean of the "kink" points twice (once for each line in which it is an endpoint); but if required only once, what should the angle of its box region be? The requirements doc for handling polylines did not specify what should be done where the lines join.

@pford do you think the following makes sense? Basically we apply "trim" or not to the second line segment depending on the residual (< 0.5x unit separation or > 0.5x unit separation). If the residual is > 0.5x unit separation (case#1) we don't trim the first point of the second line segment. If the residual is < 0.5x unit separation (case#2) we trim the first point of the second line segment. 

carta_polyline_sampling_issue 001
carta_polyline_sampling_issue 002

acdo2002:
@pford I test this branch with given fixed width (width: 1) in `SET_SPATIAL_REQUIREMENTS`, please check the frontend branch `mylin/1617-support_spatial_profile_line_polyline`.

However if I open the frontend (default there is x, y spatial profile widget), then request a line region, the backend tend to crash. I put the tested image ("M17_SWex.fits") and the backend record ("pam-spatial-profile-line.txt") in [here](https://drive.google.com/drive/folders/1Wi1WxXk6gg-aQQXhXTTEN8m8vM4bWMcX?usp=sharing)

I am sure the returned SPATIAL_PROFILE_DATA at some situation is working.
If I
(1)open the frontend, 
(2)open the image, 
(3)then closed default x, y spatial profile widget, 
(4)then request a line 
(5) open the x, y spatial profile widget (send SET_SPATIAL_REQUIREMENTS)
There is a returned SPATIAL_PROFILE_DATA with a rawValuesFp32.

pford:
@acdo2002 I tracked down the backend crash.  The estimated number of profiles for the line needed to be updated so that the progress was complete and the profiles succeeded.

acdo2002:
@pford 
I git pull the update and rebuild the carta-backend.
So far the single line region works fine! However there are troubles when two line regions...

I still find two crashes:
(1) I think if there are at least two line region, if the first line is still running and I request the second line, then the carta-backend may crash.
1. open the image "M17_SWex.fits"
2. create two line regions
3. move the first line and quickly move the second line, then backend crash
4. the carta-backend shows `Segmentation fault (core dumped)` the backend log "two-lines-region-crash.txt" can be found [here](https://drive.google.com/drive/folders/1Wi1WxXk6gg-aQQXhXTTEN8m8vM4bWMcX?usp=sharing)

(2) 
1. open the image "M17_SWex.fits"
2. create two line regions
3. go to the right-hand-side widget X-profile widget and select region 1, and select Y-profile widget region 2.
4 the carta-backend shows 
```
[2022-03-07 14:56:45.263] [CARTA] [debug] Line 1 box region 124 max num pixels=nan

[2022-03-07 14:56:45.263] [CARTA] [debug] Angular box region corners: polygon[[311.40134pix, 348.42233pix], [308.46445pix, 349.0097pix], [308.66055pix, 349.9903pix], [311.59744pix, 349.40292pix]]
[2022-03-07 14:56:45.265] [debug] [performance] Fill region stats in 1.563 ms
[2022-03-07 14:56:45.265] [CARTA] [debug] Line 1 box region 125 max num pixels=3

terminate called after throwing an instance of 'casacore::ArrayShapeError'
  what():  ArrayBase::validateConformance shape [25] differs from [1]
Aborted (core dumped)
```
the backend log "line-select-region-crash.txt" can be found in [here](https://drive.google.com/file/d/1dzkP4FNMA6WNqtrM8RIuVeAG7APjric1/view?usp=sharing)

pford:
I added a mutex for obtaining separation between pixel points using the MVDirection from the DirectionCoordinate, since this is not thread-safe (static variables used).  I also added the test for whether to trim lines in a polyline so that a region at a vertex is not duplicated in the profile.

I noticed the profile was not being displayed without setting start and end in the SpatialProfile, so now start=0 and end=profile length.  However, the widget does not seem to resize based on this range.  The x-axis label is X Coordinate or Y Coordinate, but perhaps it should be Offset, similar to a PV image?

I also noticed that changing the line width in the profile settings does not send a new message with the updated value, and changing the region sends the new region requirements but does not "turn off" the previous region profile by sending spatial requirements for the previous region id with no SpatialConfig settings.

Spatial profiles for widefield images (not fixed pixel spacing) is still under development.

acdo2002:
@pford 

The frontend branch is `mylin/1617-support_spatial_profile_line_polyline`

> I noticed the profile was not being displayed without setting start and end in the SpatialProfile, so now start=0 and end=profile length. However, the widget does not seem to resize based on this range. The x-axis label is X Coordinate or Y Coordinate, but perhaps it should be Offset, similar to a PV image?

I think the step would start from the ICD `SET_REGION` (regionInfo contains the start and end point of x & y) and the backend return `SET_REGION_ACK` (at this moment, the line/polyline has regionId), then if the spatial profile widget exists (x & y Profile), the frontend would sent the ICD `SET_SPATIAL_REQUIREMENTS`, it includes the regionId and spatialProfiles(include coordinate x or/and y; mip; and width). 
Here is the movie for setting a region line and check the spatial profile for both x and y:

https://user-images.githubusercontent.com/34430648/160074835-d424d932-d53b-4ced-9449-30d87a72a131.mp4

> I also noticed that changing the line width in the profile settings does not send a new message with the updated value, and changing the region sends the new region requirements but does not "turn off" the previous region profile by sending spatial requirements for the previous region id with no SpatialConfig settings.

Could you tell me how you change the line width? Because I change the line width in the profile setting (under computation), if the spatial profile widget exists (x &/or y Profile), the frontend will send the ICD `SET_SPATIAL_REQUIREMENTS`. And the backend returns `SPATIAL_PROFILE_DATA` with regionId and profiles.
From my UI: I found when I change the region (e.g. regionId=1), the frontend will send `SET_REGION` and the backend send `SET_REGION_ACK` and if the spatial profile widget exists (x &/or y Profile), the backend will return `SPATIAL_PROFILE_DATA` with regionId and profiles.
Here is the movie for setting a polyline and change the line width

https://user-images.githubusercontent.com/34430648/160076604-970b816a-fde5-4780-bdcd-fd799851c358.mp4

acdo2002:
I tested with the latest backend branch and found few issues (all related to wide field all sky image, the testing image "Gaussian.fits" can be found in [here](https://drive.google.com/drive/folders/1Wi1WxXk6gg-aQQXhXTTEN8m8vM4bWMcX?usp=sharing)):
1) If I set a line region covers all the image peak in the Gaussian.fits, the X Profile shows incorrect profile. 
In principle, there are 10 peaks in the image, the X Profile only shows 8 peaks...(perhaps does not include all line region properly for a very long/wide line region)
Screen Shot 2022-04-14 at 4 36 24 PM

2) If I set a line region start point is beyond the image (i.e. region start coordinate in image/pixel is -47 and world coordinate is <bad>), then the X Profile and Y Profile do not updated (i.e. no <== SPATIAL_PROFILE_DATA returned). Such issue does not occur (works as expected) if the image is small.
Screen Shot 2022-04-14 at 5 04 58 PM

kswang1029:
@pford quick question: when a spatial profile derived from a line region, do we also add "Offset" array (plus a unit) in the SPATIAL_PROFILE_DATA message? And if it is polyline region, it should be "Distance" (plus a unit). The frontend would need two arrays (one for x and the other for y) to plot the profile properly.

acdo2002:
It looks the Issue (1) has been solved.
I set the polyline and cross the peaks in "Gaussian.fits" 3 times, the spatial profiler shows 30 peaks accurately.
Screen Shot 2022-04-25 at 2 38 17 PM

kswang1029:
@pford found a critical bug when a line region is too short. Polyline seems okay. 

https://user-images.githubusercontent.com/20819712/165230108-01b4780f-9be1-451f-9332-aa2451a766c8.mov

kswang1029:
@pford @acdo2002 could you check the order of the "distance" axis of a polyline (not sure it is frontend or backend)? From the test, if we define a two-segment polyline like the video, we would expect the bump be toward the origin instead of toward the end of the polyline. Now it seems the order is reversed. 

https://user-images.githubusercontent.com/20819712/165232840-633280d5-2ef0-4c4f-bc77-dce9a32dfad9.mov

veggiesaurus:
> @pford @acdo2002 could you check the order of the "distance" axis of a polyline (not sure it is frontend or backend)? 

Also check if it's reversed per-segment or entirely.

kswang1029:
@pford @acdo2002 here is a performance related issue which could be addressed post-beta3 if it is less trivial.

When a line region is selected (so it is "active") without any modification of the line region properties, the backend will recompute the profile. If the profile has been computed once, the frontend does show cached profile. When the image size is large, this performance issue is more prominent. 

https://user-images.githubusercontent.com/20819712/165236207-bb204eb1-9719-4b1c-813f-6e9139d8cd3d.mov

kswang1029:
> > @pford @acdo2002 could you check the order of the "distance" axis of a polyline (not sure it is frontend or backend)?
> 
> Also check if it's reversed per-segment or entirely.

I confirm it is entirely, not per-segment.
Screen Shot 2022-04-26 at 14 34 50

pford:
I fixed the reversed polyline profile and the short lines with 0- or 1-region profiles.  Also updated protobuf and backend branches with dev.

kswang1029:
> I fixed the reversed polyline profile and the short lines with 0- or 1-region profiles. Also updated protobuf and backend branches with dev.

@pford thanks. Confirmed the polyline profile order is correct now. 👍 
The crash issue remains. The backend will crash if we just create a very short line region.

kswang1029:
@pford accidentally I triggered a backend crash when setting a line region and a polyline region as active repeatedly on a _matched_ image. The crash log from macOS is attached.

Thread 3 Crashed:
0   libsystem_kernel.dylib        	    0x7ff8019d100e __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff801a071ff pthread_kill + 263
2   libsystem_c.dylib             	    0x7ff801952d24 abort + 123
3   libc++abi.dylib               	    0x7ff8019c3082 abort_message + 241
4   libc++abi.dylib               	    0x7ff8019b41a5 demangling_terminate_handler() + 242
5   libobjc.A.dylib               	    0x7ff8018b0e09 _objc_terminate() + 104
6   libc++abi.dylib               	    0x7ff8019c24a7 std::__terminate(void (*)()) + 8
7   libc++abi.dylib               	    0x7ff8019c4d05 __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 27
8   libc++abi.dylib               	    0x7ff8019c4ccc __cxa_throw + 116
9   carta_backend                 	       0x103d79728 std::__1::__throw_out_of_range(char const*) + 56 (stdexcept:267)
10  carta_backend                 	       0x103d7a039 std::__1::unordered_map<int, std::__1::shared_ptr<carta::Region>, std::__1::hash<int>, std::__1::equal_to<int>, std::__1::allocator<std::__1::pair<int const, std::__1::shared_ptr<carta::Region> > > >::at(int const&) + 53 (unordered_map:1792) [inlined]
11  carta_backend                 	       0x103d7a039 carta::RegionHandler::RegionSet(int) + 425 (RegionHandler.cc:116)
12  carta_backend                 	       0x103d907cd carta::RegionHandler::RegionFileIdsValid(int, int) + 39 (RegionHandler.cc:747) [inlined]
13  carta_backend                 	       0x103d907cd carta::RegionHandler::CancelLineProfiles(int, int, carta::RegionState&) + 61 (RegionHandler.cc:1773)
14  carta_backend                 	       0x103d91bd9 carta::RegionHandler::GetFixedPixelRegionProfiles(int, int, int, bool, int, carta::RegionState&, std::__1::shared_ptr<casacore::CoordinateSystem>, std::__1::function<void (float)>&, casacore::Matrix<float, std::__1::allocator<float> >&, double&, bool&) + 4521 (RegionHandler.cc:1880)
15  carta_backend                 	       0x103d8512e carta::RegionHandler::GetLineProfiles(int, int, int, bool, int, std::__1::function<void (float)>&, double&, casacore::Matrix<float, std::__1::allocator<float> >&, bool&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) + 766 (RegionHandler.cc:1747)
16  carta_backend                 	       0x103d8f410 carta::RegionHandler::GetLineSpatialData(int, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int, int, std::__1::function<void (std::__1::vector<float, std::__1::allocator<float> >, double)> const&) + 208 (RegionHandler.cc:1660)
17  carta_backend                 	       0x103d8f1a4 carta::RegionHandler::FillLineSpatialProfileData(int, int, std::__1::function<void (CARTA::SpatialProfileData)>) + 2100 (RegionHandler.cc:1634)
18  carta_backend                 	       0x103db4d96 carta::Session::SendSpatialProfileData(int, int) + 838 (Session.cc:1626)
19  carta_backend                 	       0x103db8b5e carta::Session::OnSetSpatialRequirements(CARTA::SetSpatialRequirements const&) + 1006 (Session.cc:961)
20  carta_backend                 	       0x103dd2bf4 carta::GeneralMessageTask<CARTA::SetSpatialRequirements>::execute() + 20 (OnMessageTask.tcc:24)
21  carta_backend                 	       0x103e073b1 carta::ThreadManager::StartEventHandlingThreads(int)::$_0::operator()() const + 114 (ThreadingManager.cc:54) [inlined]
22  carta_backend                 	       0x103e073b1 decltype(static_cast<carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(fp)()) std::__1::__invoke<carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(carta::ThreadManager::StartEventHandlingThreads(int)::$_0&&) + 114 (type_traits:3918) [inlined]
23  carta_backend                 	       0x103e073b1 void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>&, std::__1::__tuple_indices<>) + 114 (thread:287) [inlined]
24  carta_backend                 	       0x103e073b1 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0> >(void*) + 177 (thread:298)
25  libsystem_pthread.dylib       	    0x7ff801a074e1 _pthread_start + 125
26  libsystem_pthread.dylib       	    0x7ff801a02f6b thread_start + 15



UPDATE1: It is hard to reproduce but I got another one and see the log from the backend which may provide a clue.

libc++abi: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found


UPDATE2: it seems if we make a long line or a long polyline on a large image so that computation takes time to finish and while it is computing if we keep updating the line/polyline position, at some point the backend will crash with the log

libc++abi: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found

kswang1029:
@pford Found a werid spatial matching issue: if we have two wide field images matched and create a line region like the following attached image:
Screen Shot 2022-04-27 at 12 21 55

when the active image is the spatial reference image (left), the spatial profile along the line region can be displayed in the spatial profiler (five peaks). However, if we set the active image to the matched image (right), the spatial profile cannot be computed and the backend says

[2022-04-27 12:25:13.407] [CARTA] [error] Line region 2 spatial profile  failed: Cannot create profile for line center outside image.

Screen Shot 2022-04-27 at 12 22 12

pford:
I believe I have fixed most of the issues, including the out-of-range error in the region map, and matched images (however, I do not have Gaussian_SE2.fits to test with):

matched

Lines which go outside the image are no problem, but lines which go outside the **_coordinate system_** (the lower left point below) attempt to move the endpoint to the edge of the image and estimate the length outside of the image for NAN profiles:

outside_coord_sys

However, I think this needs more testing and debugging.  I have also not done any profile caching to improve performance, and could also improve checking that the requirement was not removed before completing the profile.

kswang1029:
@pford thanks for addressing issues. I will have a look shortly. Regarding performance improvement (caching, etc), we can try to do so for beta4 (aka v3 final).

kswang1029:
@pford I think this is now ready to be a PR. Please identify reviewers for code review and I will keep testing this. Mostly it is looking great so far! I will make a backend issue for the performance optimization for beta4.

acdo2002:
Here I confirmed the Issue (2) has been solved, thank you very much Pam!!!

pford:
I think this needs a little more work, for example in this profile there is no gap in the profile where the polyline is outside the image:
outside_coord_sys

kswang1029:
> I think this needs a little more work, for example in this profile there is no gap in the profile where the polyline is outside the image:
> outside_coord_sys

Hmm thanks for reporting this issue. I can reproduce the problem here. There are non-continuous NaN channels in that no-wcs and no data parts but there should be just one big chunk of NaN. 🤔

If you like, we can mark this as a known issue and try to find a solution in beta4. I wonder if we can simply ignore the request if _any_ parts of the polyline or line region have no WCS info? I believe this can be considered as an edge use case.

kswang1029:
@pford regarding the issue where part of the line/polyline does not have valid world coordinates, I think we should just ignore the profile request and throw an error message to frontend to be displayed. If the image is wide field so that each small region box has changing sizes, it is impossible to estimate the dynamic box size along the line/polyline at the places where world coordinates are undefined I suppose.

pford:
Additional fixes for line spatial profile:

* log error if fixed-angular profile line endpoints outside wcs
* check region type for "active region" requirements (e.g. spatial must be point/line, spectral must not be line, etc.)
* cancel line profile if requirements reset or region changes

@kswang1029 there are no success or message fields in the SpatialProfileData message, so if the profile fails, it just logs a message (backend log).  The spectral profile had the same behavior when moving a rectangle partially outside the coordinate system.

kswang1029 (approved):
@pford this looks great. Thanks for adding the cancellation feature which is a great plus to UX. 👍

markccchiang (approved):

src/Region/RegionHandler.cc
```diff
@@ -480,26 +571,36 @@ bool RegionHandler::SetStatsRequirements(
...
         _histogram_req.clear();
         _stats_req.clear();
-        std::unique_lock<std::mutex> ulock(_spectral_mutex);
+        std::unique_lock<std::mutex> ulock1(_spectral_mutex);
         _spectral_req.clear();
-        ulock.unlock();
+        ulock1.unlock();
+        std::unique_lock<std::mutex> ulock2(_spatial_mutex);
+        _spatial_req.clear();
+        ulock2.unlock();
```
confluence: This is very minor, but I would prefer to give these more descriptive names, like `ulock_spectral` and `ulock_spatial`.

src/Region/RegionHandler.cc
```diff
@@ -1557,214 +1776,336 @@ std::vector<int> RegionHandler::GetProjectedFileIds(int region_id) {
...
+    float progress(0.0);
 
-    if (per_z && _stop_pv[file_id]) {
-        cancelled = true;
-        return false;
-    }
+    if (num_lines == 1) {
+        // Get box centers along line; start at line center and add offsets
+        PointXy endpoint0(control_points[0].x(), control_points[0].y());
+        PointXy endpoint1(control_points[1].x(), control_points[1].y());
```
confluence: Perhaps these could also have more descriptive start / end names? Otherwise this could be a vector rather than two numbered variables. Would that have any effect on performance?

confluence (approved):
I have some variable naming suggestions, but they're not essential.

This needs a companion protobuf PR, right?

src/Region/RegionHandler.cc
pford: Renamed to line_start and line_end. I think it is more readable to keep them separate, and the endpoints are frequently used separately in the code (e.g. positive offsets vs negative offsets). I also changed to `vector<double>` instead of `PointXy` to make FixedPixelRegionProfiles consistent with FixedAngularRegionProfiles, and because the functions within the struct were not needed.

src/Region/RegionHandler.cc
pford: renamed to spectral_lock and spatial_lock

src/Region/RegionHandler.cc
```diff
@@ -318,24 +340,88 @@ bool RegionHandler::SetHistogramRequirements(
...
-        ConfigId config_id(file_id, region_id);
-        _histogram_req[config_id].configs = input_configs;
+    // Save frame pointer
+    _frames[file_id] = frame;
+
+    // Make HistogramConfig vector of requirements
+    std::vector<HistogramConfig> input_configs;
+    for (auto& config : configs) {
+        HistogramConfig hist_config(config.coordinate(), config.channel(), config.num_bins());
+        input_configs.push_back(hist_config);
```
jolopezl, line 360: I don't know how big this loop is, but I would suggest using `emplace_back(...)`.

jolopezl (approved):
Looks great. I have just one tiny comment.

src/Region/RegionHandler.cc
pford: For a region histogram, the only current setting is the stokes coordinate, so the max loop is nstokes (but more likely a single stokes that the user changes in the same widget)

kswang1029:
@pford This is picked up by the e2e tests. I found that when the image is a wide field one, if we create a line or a polyline region the backend will crash. Could you have a look please? 

With some comparisons of recent commits, it seems the crash started from

[Use vector<double> instead of PointXy for line profile points](https://github.com/CARTAvis/carta-backend/pull/1020/commits/144006d472cfb0a748e18cfcc18545eac5bc886d) [144006d](https://github.com/CARTAvis/carta-backend/pull/1020/commits/144006d472cfb0a748e18cfcc18545eac5bc886d)

the macOS console log:

Thread 2 Crashed:
0   carta_backend                 	       0x10c0ddaa4 carta::RegionHandler::FindPointAtTargetSeparation(std::__1::shared_ptr<casacore::CoordinateSystem>, std::__1::vector<double, std::__1::allocator<double> > const&, std::__1::vector<double, std::__1::allocator<double> > const&, double, double) + 772
1   carta_backend                 	       0x10c0d996e carta::RegionHandler::GetFixedAngularRegionProfiles(int, int, int, bool, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, carta::RegionState&, std::__1::shared_ptr<casacore::CoordinateSystem>, std::__1::function<void (float)>&, casacore::Matrix<float, std::__1::allocator<float> >&, double&, bool&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) + 2446 (RegionHandler.cc:2386)
2   carta_backend                 	       0x10c0ca502 carta::RegionHandler::GetLineProfiles(int, int, int, bool, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::function<void (float)>&, double&, casacore::Matrix<float, std::__1::allocator<float> >&, bool&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&) + 930 (RegionHandler.cc:1879)
3   carta_backend                 	       0x10c0d59ff carta::RegionHandler::GetLineSpatialData(int, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int, int, std::__1::function<void (std::__1::vector<float, std::__1::allocator<float> >, double)> const&) + 223 (RegionHandler.cc:1759)
4   carta_backend                 	       0x10c0d575c carta::RegionHandler::FillLineSpatialProfileData(int, int, std::__1::function<void (CARTA::SpatialProfileData)>) + 1868 (RegionHandler.cc:1733)
5   carta_backend                 	       0x10c0fb6e1 carta::Session::SendSpatialProfileData(int, int) + 785 (Session.cc:1621)
6   carta_backend                 	       0x10c0ff4de carta::Session::OnSetSpatialRequirements(CARTA::SetSpatialRequirements const&) + 1006 (Session.cc:956)
7   carta_backend                 	       0x10c119a74 carta::GeneralMessageTask<CARTA::SetSpatialRequirements>::execute() + 20 (OnMessageTask.tcc:24)
8   carta_backend                 	       0x10c14e811 carta::ThreadManager::StartEventHandlingThreads(int)::$_0::operator()() const + 114 (ThreadingManager.cc:54) [inlined]
9   carta_backend                 	       0x10c14e811 decltype(static_cast<carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(fp)()) std::__1::__invoke<carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(carta::ThreadManager::StartEventHandlingThreads(int)::$_0&&) + 114 (type_traits:3918) [inlined]
10  carta_backend                 	       0x10c14e811 void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>&, std::__1::__tuple_indices<>) + 114 (thread:287) [inlined]
11  carta_backend                 	       0x10c14e811 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0> >(void*) + 177 (thread:298)
12  libsystem_pthread.dylib       	    0x7ff8133f74e1 _pthread_start + 125
13  libsystem_pthread.dylib       	    0x7ff8133f2f6b thread_start + 15

kswang1029:
@pford with [7573e19](https://github.com/CARTAvis/carta-backend/pull/1020/commits/7573e1963cdc8976f9b380b72d44ea427f584985) the crash issue is gone. Thanks. 

However, I spotted another issue from an e2e test regarding matching wide field images and requesting pv generation (this can only be observed after the crash issue is fixed). The test tries to request pv generation from _each_ image with a shared line region. So there should be two pv image generated in the end.

https://user-images.githubusercontent.com/20819712/167536065-5c000109-d439-4ede-a514-cebbceddd225.mov

I don't see any useful log from the backend console nor the browser console.

kswang1029 (approved):
@pford this is good now! 👍 @confluence please have a final look, then we are ready for v3b3 release. 

I noticed a minor line sampling issue on a wide field image but we can worry about it later. It seems near the line ends there is a missing sample each side?

Screen Shot 2022-05-11 at 00 20 20

confluence (approved):

PR #1022, opened 2022-02-16
Confluence/remove grpc

confluence:
This PR removes the gRPC server and moves the scripting API to an endpoint in the HTTP server. The HTTP server has been renamed, and its different functions have been decoupled from one another so that they can be enabled and disabled independently. Commandline options have been adjusted.

This PR should be merged at the same time as the companion PRs in the [wrapper](https://github.com/idia-astro/carta-python/pull/31), [frontend](https://github.com/CARTAvis/carta-frontend/pull/1785), [protocol buffers](https://github.com/CARTAvis/carta-protobuf/pull/65), [controller](https://github.com/CARTAvis/carta-controller/pull/110) and [website](https://github.com/CARTAvis/CARTAvis.github.io/pull/18).


src/Session/SessionManager.cc
```diff
@@ -555,6 +543,39 @@ void SessionManager::RunApp() {
         .run();
 }
 
+bool SessionManager::EnableScripting() {
```
veggiesaurus: Not sure about this function name. I think `IsScriptngEnabled` makes more sense and is consistent with the rest of our codebase (`IsAuthenticated`, `IsValidFrontendFolder`, `IsValid`, `IsCompressedFits`)

veggiesaurus (changes requested):
one very minor comment from me, otherwise looks great.

src/Session/SessionManager.cc
confluence: This is actually unused (a relic of a previous design). I will remove it.

veggiesaurus (approved):

src/Main/ProgramSettings.h
```diff
@@ -34,15 +34,16 @@ struct ProgramSettings {
...
-    int grpc_port = -1;
     int omp_thread_count = OMP_THREAD_COUNT;
     int event_thread_count = 2;
     std::string top_level_folder = "/";
     std::string starting_folder = ".";
     std::string host = "0.0.0.0";
     std::vector<std::string> files;
     std::string frontend_folder;
-    bool no_http = false;
+    bool no_http = false; // Deprecated
```
jolopezl, line 44: What about explicitly marking this variable as deprecated so that we remember in the future to remove it?. 
```c++
[[deprecated]] bool no_http = false;
// or
bool no_http __attribute__((deprecated)) = false;
```
I prefer `[[deprecated]]` since it's more readable, and we already have [[maybe_unused]] in the codebase, so it's the same attribute style. Unfortunately, it will add a lot of clutter at compilation time, but it's a good reminder that improves code quality in the end since it forces us to keep our codebase clean.

It is also easy to remove the warning while we are developing. The following is enough:
```
cmake .. -DCMAKE_CXX_FLAGS="-Wno-deprecated-declarations"
```

jolopezl (changes requested):
Looks great!. I added one minor comment regarding deprecated variables. But more importantly, can I suggest changing the name of `SimpleHttpServer` to `HttpServer`?

src/Main/ProgramSettings.h
confluence: The renaming is a good idea; I will definitely do that. I would prefer not do do something that will add noise to compilation without a wider discussion -- I suggest making a separate issue to propose this.

confluence: During the renaming, I realised that the `#ifndef` guard names in the headers weren't consistently renamed during the recent file shuffle. I'll make a separate PR to fix these automatically using the style script.

jolopezl: > During the renaming, I realised that the `#ifndef` guard names in the headers weren't consistently renamed during the recent file shuffle. I'll make a separate PR to fix these automatically using the style script.

Sorry, my bad. I also spotted a problem with `Hdf5Loader.h` since it looks like it isn't catching the header guard. Thanks for adding that to the script; it's very useful.

CMakeLists.txt
```diff
@@ -245,7 +219,7 @@ set(SOURCE_FILES
         src/Session/OnMessageTask.cc
         src/Session/Session.cc
         src/Session/SessionManager.cc
-        src/SimpleFrontendServer/SimpleFrontendServer.cc
+        src/HttpServer/HttpServer.cc
```
jolopezl: I know it's tedious, but can we keep the alphabetical order?

jolopezl (approved):
Thanks, looks great!.

src/Main/ProgramSettings.h
confluence: I created issue #1023 for this.

jolopezl (approved):

Issue #845, opened 2021-06-18
add spatial and spectral range information in the file information


kswang1029:
cf casaviewer, the casaviewer provides range information as part of the file information including:
- R.A. range (or Longitude)
- Dec. range (or Latitude) 
- Frequency range
- Velocity range
- Stokes coverage

confluence:
This was fixed in #1042 .

PR #1026, opened 2022-02-22
Adding ranges information on the file browser

markccchiang:
#845, adding ranges information in the file preview. Including RA, Dec, Frequency, Velocity, and Stokes coverages. As shown below:
截圖 2022-02-22 上午10 37 32
 
@kswang1029 do we also need ranges information for compressed FITS files in the file preview? If so, I will do it in another separate PR.

### DIFF
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -1184,27 +1184,26 @@ void FileExtInfoLoader::FillCoordRanges(CARTA::FileInfoExtended& extended_info) 
 
     if (coord_system.hasDirectionCoordinate()) {
         auto direction_coord = coord_system.directionCoordinate();
-        casacore::Vector<double> ref_val = direction_coord.referenceValue();
-        if (ref_val.size() == 2) {
+        if (direction_coord.referenceValue().size() == 2) {
             casacore::Vector<int> direction_axes = coord_system.directionAxesNumbers();
-            casacore::Vector<double> pix(direction_axes.size(), 0);
-            casacore::Vector<double> world(pix.size());
+            casacore::Vector<double> pixels(direction_axes.size(), 0);
+            casacore::Vector<double> world(direction_axes.size(), 0);
             casacore::Vector<double> ra_range_vec(2);
             casacore::Vector<double> dec_range_vec(2);
             casacore::String units;
 
             // Get start world coord
-            direction_coord.toWorld(world, pix);
+            direction_coord.toWorld(world, pixels);
             ra_range_vec[0] = world[0];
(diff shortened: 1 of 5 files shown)


kswang1029:
> #845, adding ranges information in the file preview. Including RA, Dec, Frequency, Velocity, and Stokes coverages. As shown below: 截圖 2022-02-22 上午10 37 32
> 
> @kswang1029 do we also need ranges information for compressed FITS files in the file preview? If so, I will do it in another separate PR.

Yes, for compressed fits we need this set of info too. Good idea to support it in a separate PR.

veggiesaurus:
This is looking good. Do you think we can just use RA and DEC shorthands? I also wonder what we should do for situations where the image is in galactic coordinates. Should we not report the spatial range in GLON/GLAT?

kswang1029:
> This is looking good. Do you think we can just use RA and DEC shorthands? I also wonder what we should do for situations where the image is in galactic coordinates. Should we not report the spatial range in GLON/GLAT?

Depending on what CTYPE1/2/3 says we should have dynamic naming conventions.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -1175,56 +1176,65 @@ void FileExtInfoLoader::GetCoordNames(std::string& ctype1, std::string& ctype2,
...
+            x_range_vec[1] = world[0];
+            std::string x_end = direction_coord.format(units, casacore::Coordinate::DEFAULT, world[0], 0, true, true);
+            y_range_vec[1] = world[1];
+            std::string y_end = direction_coord.format(units, casacore::Coordinate::DEFAULT, world[1], 1, true, true);
+
+            // Set x and y coordinate names
+            if (axis_names(0) == "Right Ascension") {
+                axis_names(0) = "RA";
+            } else if (axis_names(0) == "Longitude") {
+                axis_names(0) = "GLON";
```
kswang1029: just LON and LAT. 

G means galactic. There is also ecliptic coordinate as ELON and ELAT. So let just try to display them as LON and LAT for simplicity.

kswang1029 (approved):
Tested and it looks good. The current implementation also fixes a bug in the casaviewer where the reported spatial coverage has one extra pixel offset. 👍

kswang1029:
@confluence For the support of compressed FITS, it could be in another PR if you prefer.

markccchiang:
> @confluence For the support of compressed FITS, it could be in another PR if you prefer.

For compressed FITS, the range information can be seen after opening the file. But it can not show on the preview stage. In order to keep the performance, I guess it is better not to open the file on the preview stage. We have to use another way to calculate the ranges from the header information.

pford (approved):

confluence (approved):

PR #1027, opened 2022-03-01
Show file list beam PA with correct value in deg

### LINKED ISSUE #1025: Beam position angle incorrectly reported and rendered as degrees instead of radians for CASA image
**Describe the bug**
For a `CASA` image created in `CASA` with `tclean`, the "file information" in the "file browser" dialog of `CARTA` reports the restoring beam position angle to be -1.44 deg while the "header" tab shows `BPA = -8.235030409149E+01`. Also, the beam ellipse when the image is loaded is using the position angle reported under "file information".

`imhead` in `CASA` shows that the position angle is -1.44 radians, so it looks like `CARTA` is assuming it is in degrees in some places.

**To Reproduce**
1. Click File -> Open image.
2. Click on the image.
3. Note the restoring beam position angle under "file information".
4. Click the "header" tab and note the different value.
6. Click Load and the beam ellipse is drawn with the "file information" position angle.

**Expected behavior**
The "file information" should report the correct position angle value with the appropriate units (either report radians like `imhead` does or convert value to degrees). The beam ellipse should be drawn based on the value and units of the position angle being consistent.

 - OS: MacOS 10.15.7
 - Browser: N/A (just running `CARTA` app directly in it's own window)
 - Version: 3.0.0-beta.2b

**Additional context**
One of my cubes where I am seeing this issue is fairly small (6 MB) and I can provide it if that would help.

Thanks.

pford:
Closes #1025 

Get single beam or median beam PA in casacore::Unit "deg".  Minor change.

### DIFF
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -1070,7 +1070,6 @@ void FileExtInfoLoader::AddBeamEntry(CARTA::FileInfoExtended& extended_info, con
     }
 
     if (!gaussian_beam.isNull()) {
-        casacore::Quantity beam_pos_angle(gaussian_beam.getPA());
         std::string beam_info = fmt::format("{:g}\" X {:g}\", {:g} deg", gaussian_beam.getMajor("arcsec"), gaussian_beam.getMinor("arcsec"),
             gaussian_beam.getPA(casacore::Unit("deg")));
 
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -1070,8 +1070,9 @@ void FileExtInfoLoader::AddBeamEntry(CARTA::FileInfoExtended& extended_info, con
     }
 
     if (!gaussian_beam.isNull()) {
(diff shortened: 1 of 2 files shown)


kswang1029:
@pford thanks for the fix. Now the file info shows the unit of the beam PA. Could you take one step further by converting rad to degree if the beam PA is defined in rad in the header? That is, we always use degree to display beam PA in the file info. Usually it is more intuitive to show beam PA in degrees (both numerically and rendered visually). As a note here: the displayed header in the FITS format also uses (implicitly) degree for beam PA unit. With this extra step, we will not need to modify the frontend rendering code as at the moment it takes the beam PA value and assumes it is in degrees.

markccchiang (approved):

confluence (approved):

pford:
@kswang1029 beam PA is now in deg

kswang1029 (changes requested):
@pford thanks for fixing the displayed beam PA in degree. I spotted that the frontend actually uses the values in beamTable for beam rendering. We will also need to apply the conversion there too as with the test image I still see the beam orientation is not rendered properly.
Screen Shot 2022-03-02 at 11 33 37

kswang1029 (approved):
Tested and ready to merge.

confluence (approved):

PR #1028, opened 2022-03-02
Include protobuf files and link homebrew libs to fix build errors on Mac

confluence:
After the recent removal of gRPC, the backend cannot be built on some M1 Mac systems without additional flags.

This PR adds an explicit command to include a protobuf directory which was previously included by the cmake config in the removed submodule. It also adds a link to the M1 homebrew library location.


kswang1029 (approved):
Confirmed with simply `cmake ..` the backend can be built on Intel Mac and M1 Mac successfully. Also did a simple test by loading an image, animating it and checking region spectral profile. All look good.

acdo2002 (approved):
Now in my Mac laptop, I only use `cmake ..` then it can build the backend

ajm-ska (approved):
Continues to work fine on the M1 macOS 12 system.

The ICD tests have actually passed even though it says "Some checks haven’t completed yet". I think it may be because the test suite took more than 1 hour to run because the CI systems were busy on another branch. This is a problem I need to address.

PR #1031, opened 2022-03-03
Add configurable option for carta-casacore path

jolopezl:
Add `CARTA_CASACORE_ROOT` variable to allow configuring the backend with `carta-casacore` in a different location. For example:

```
cmake3 .. -DCARTA_CASACORE_ROOT=$HOME/opt/carta-casacore
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,10 +111,15 @@ INCLUDE_DIRECTORIES(${HDF5_INCLUDE_DIR})
 ADD_SUBDIRECTORY(carta-protobuf)
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})
 
+if (NOT CARTA_CASACORE_ROOT)
+    set(CARTA_CASACORE_ROOT /opt/carta-casacore)
+endif()
+
 set(CARTA_CASACORE_INCLUDE_DIRS
-    /opt/carta-casacore/include
-    /opt/carta-casacore/include/casacore
-    /opt/carta-casacore/include/casacode)
+    ${CARTA_CASACORE_ROOT}/include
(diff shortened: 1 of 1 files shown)


confluence (approved):

ajm-ska (approved):
This works perfectly! It is a very useful addition.

Maybe it is out of the scope of this pull-request, but I was just wondering, would it be possible to add a FIND_PACKAGE or FIND_LIBRARY to check if the casacore and casacode libraries are present/accessible during "cmake"?
Otherwise we only discover that the files are not found halfway through the "make" process. I know casacore/casacode is not a traditional package so maybe it is not easy to do.

jolopezl:
> Maybe it is out of the scope of this pull-request, but I was just wondering, would it be possible to add a FIND_PACKAGE or FIND_LIBRARY to check if the casacore and casacode libraries are present/accessible during "cmake"?
> Otherwise we only discover that the files are not found halfway through the "make" process. I know casacore/casacode is not a traditional package so maybe it is not easy to do.

That would be another very useful addition. I think this is the starting point: https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-layout. We could try to experiment with that in the carta-casacore repository.

confluence:
Should we leave this for a separate PR, since the CMake extension for finding carta-casacore is probably something we should add to the carta-casacore repo? I know that something similar has been created for casacore; we can probably adapt that.

jolopezl:
> Should we leave this for a separate PR, since the CMake extension for finding carta-casacore is probably something we should add to the carta-casacore repo? I know that something similar has been created for casacore; we can probably adapt that.

Yes, we should leave everything else for a separate pr.

Issue #967, opened 2021-11-18
Migrate 2D image cache away from std::vector<float>


veggiesaurus:
We currently use a `std::vector<float>` for `Frame::_image_cache`. However, the ability to resize the image cache is not of any use, because a new cache is made for each separate image, and all image channels have the same 2D slice size. The main problem with using a `std::vector` is that we end up allocating memory _and_ initialising to zero before we fill it with values from disk. This means it will always be slower than something like a simple c-style array.

This is not a problem for small images, but for large images it can become a substantial portion of the initial image load time.

jolopezl:
@veggiesaurus, do we have any design constraint to resist introducing c-style arrays here in favor of a significant performance gain for large images?

veggiesaurus:
No, thats totally fine, as long as they are carefully cleaned up. We don't actually ever need to resize the image cache if I recall correctly

PR #1033, opened 2022-03-05
Migrate 2D image cache away from std::vector<float> using gsl::span (#967)

jolopezl:
Migrate 2D image cache away from `std::vector<float>` introducing `gsl::span` in several other places to avoid resizing `_image_cache` as a `std::vector`, then avoiding the initialization penalty.

- [ ] Other `std::vector`s can be removed?
- [ ] Check performance improvement.
- [ ] Check performance overhead of gsl::span.

This pull request is equivalent to #1034 without the need of using c-style arrays, but it introduces the user of [GSL](https://github.com/microsoft/GSL) that could be eventually removed if we move to C++20.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,12 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
     set(LINK_LIBS ${LINK_LIBS} stdc++fs)
 endif ()
 
+include(FetchContent)
+FetchContent_Declare(GSL
+        GIT_REPOSITORY "https://github.com/microsoft/GSL"
+        GIT_TAG "v3.1.0")
+FetchContent_MakeAvailable(GSL)
+
 # Include uWebSockets headers and build the uSockets lib
 include_directories(${CMAKE_SOURCE_DIR}/third-party/include)
 install_uWebSockets()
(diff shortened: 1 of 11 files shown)


veggiesaurus:
@jolopezl is gsl not the GNU Scientific Library? or do we now have _two_ dependencies with the same name?

jolopezl:
> @jolopezl is gsl not the GNU Scientific Library? or do we now have _two_ dependencies with the same name?

@veggiesaurus, unfortunately, it's a different library with the same name: [GSL: Guidelines Support Library](https://github.com/microsoft/GSL). However, it's one way to get `std::span` without moving forward to C++20. Please, don't consider this settled, I am not in favor of adding another dependency (despite that it's straightforward), but I like the capabilities of spans.

ajm-ska:
@jolopezl This branch is unable to build in the Ubuntu 18.04 CI Docker container:

```
  include could not find load file:
    FetchContent
CMake Error at CMakeLists.txt:160 (FetchContent_Declare):
  Unknown CMake command "FetchContent_Declare".
```

The default CMake in Ubuntu 18.04 is version 3.10.2, but looks like "FetchContent" was added in CMake 3.11.

As we will no longer be officially supporting Ubuntu 18.04 soon, I guess I can just manually install a newer CMake version in the 18.04 container for now. Maybe CMake 3.16.3, which is the default with Ubuntu 20.04, if that is OK with @confluence?

jolopezl:
Closing so that we don't add another dependency just to avoid using pointers directly.

PR #1034, opened 2022-03-05
Migrate 2D image cache away from std::vector<float> (#967)

jolopezl:
Migrate 2D image cache away from `std::vector<float>` using c-style arrays in several other places to avoid resizing `_image_cache` as a `std::vector`, then avoiding the initialization penalty.

- [x] Fix tests since these fail when using `-fsanitize=address` (update: all tests passing).
- [x] Other `std::vector`s can be removed? (update: I think so, should we?)
- [x] Check performance improvement. (update: a quick check using `hlsp_candels_hst_acs_gn-tot-30mas_f606w_v1.0_drz.fits` shows a mild improvement of a rough 10% in loading time, for smaller images the difference is negligible)

### DIFF
--- a/src/ImageStats/BasicStatsCalculator.h
+++ b/src/ImageStats/BasicStatsCalculator.h
@@ -32,10 +32,11 @@ class BasicStatsCalculator {
     T _min_val, _max_val;
     double _sum, _sum_squares;
     size_t _num_pixels;
-    const float* _data;
+    const T* _data;
+    size_t _data_size;
 
 public:
-    BasicStatsCalculator(const float* data, size_t data_size);
+    BasicStatsCalculator(const T* data, size_t data_size);
 
     void join(BasicStatsCalculator& other); // NOLINT
     void reduce(const size_t start, const size_t end);
--- a/src/ImageStats/BasicStatsCalculator.h
+++ b/src/ImageStats/BasicStatsCalculator.h
@@ -9,6 +9,7 @@
 
 #include <algorithm>
 
+
 namespace carta {
 
 template <typename T>
@@ -32,10 +33,10 @@ class BasicStatsCalculator {
     T _min_val, _max_val;
     double _sum, _sum_squares;
     size_t _num_pixels;
-    const std::vector<T>& _data;
+    const float* _data;
 
 public:
-    BasicStatsCalculator(const std::vector<T>& data);
+    BasicStatsCalculator(const float* data, size_t data_size);
 
     void join(BasicStatsCalculator& other); // NOLINT
     void reduce(const size_t start, const size_t end);
--- a/src/ImageStats/BasicStatsCalculator.h
+++ b/src/ImageStats/BasicStatsCalculator.h
@@ -9,7 +9,6 @@
 
 #include <algorithm>
 
-
 namespace carta {
 
 template <typename T>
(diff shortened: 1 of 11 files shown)


src/ImageStats/BasicStatsCalculator.h
```diff
@@ -32,10 +32,10 @@ class BasicStatsCalculator {
     T _min_val, _max_val;
     double _sum, _sum_squares;
     size_t _num_pixels;
-    const std::vector<T>& _data;
+    const float* _data;
 
 public:
-    BasicStatsCalculator(const std::vector<T>& data);
+    BasicStatsCalculator(const float* data, size_t data_size);
```
confluence: `data_size` is never used -- should it be removed?

Alternatively, should it be added to the object as a property? It's not clear to me why `reduce` takes a start and end parameter when it is only called from one place in the code (with zero and the full data size). This could take no parameters and use a size value set in the constructor.

confluence (changes requested):

src/ImageStats/BasicStatsCalculator.h
jolopezl: Yes, you are right, I might have automatically included the variable when moving to the pointer and size design, but it is unneeded here. I will drop what is unneeded here for the purpose of this pull request, but we should file a new issue for improving the design of the class if we want to.

jolopezl: To resolve this: I included `data_size` as a member and added a small range check to `BasicStatsCalculator::reduce`. I think it's ok doing so. I also realized we must use `const T*` instead of `const float*` since this is a template class.

src/ImageStats/BasicStatsCalculator.tcc
```diff
@@ -41,17 +43,22 @@ BasicStats<T>::BasicStats()
...
       _sum_squares(0),
       _num_pixels(0),
-      _data(data) {}
+      _data(data),
+      _data_size(data_size) {}
 
 template <typename T>
 void BasicStatsCalculator<T>::reduce(const size_t start, const size_t end) {
     size_t i;
+    if (start == end || end > _data_size) {
```
confluence: If the data is empty, this function will be called with a start of 0 and an end of 0, which I would expect to be a valid no-op if for some reason the data is empty.

~~On the other hand, if the end exceeds the bounds of the array, that will cause an invalid read, so I'd say we need a stronger precaution than a debug message.~~ Sorry; I misread this and didn't see the return!

I would check only for the second condition, set the end to the data size if it's triggered, and maybe upgrade the debug message to a warning (with a more specific message).

I don't feel that strongly about correcting the end automatically; if this is triggered it means an internal logic error in the code which should be fixed. So exiting is also fine -- but then we should exit with a more visible warning or error. Either way I don't think we should conflate the two conditions (and shouldn't warn for the first one at all).

Should we check if the start is greater than the end? That's also currently a no-op, but I can't see how that would be a valid use case.

confluence (changes requested):

src/ImageStats/BasicStatsCalculator.tcc
jolopezl: > I don't feel that strongly about correcting the end automatically; if this is triggered it means an internal logic error in the code which should be fixed.

I agree that we should never find this case during execution, but we should have an error in case of flawed logic and should be fixed.

confluence (approved):

PR #1042, opened 2022-03-10
Adding coordinate ranges information for compressed FITS files

markccchiang:
#845, adding coordinate ranges information for compressed FITS files on the file preview stage. Which includes RA, DEC, Frequency, Velocity ranges, and stokes coverage. They are calculated from the header information. These values can be verified with that after opening the compressed FITS file.


kswang1029 (changes requested):
tested with some test images and it works properly. There is one test fits.gz image resulting an incorrect range. Have passed the test image to @markccchiang for investigation.

kswang1029:
after having some discussion with @markccchiang, this PR can be merged if @confluence thinks okay. This can display the spatial range properly when images do not have sky rotation. With a sky rotation, the spatial range is tricky to derive and needs more thoughts (the spatial range from casa-imstat is questionable).

kswang1029 (approved):
sky rotation will be addressed in other PR. This one works properly with images without sky rotation.

confluence (approved):

PR #1049, opened 2022-03-11
Rwjs/unique sids

robsimmonds:
Create unique session ids even if reconnections are intermingled with new connections.

This builds on the session deletion PR from earlier today that I asked @confluence to review.

### DIFF
--- a/src/Session/SessionManager.cc
+++ b/src/Session/SessionManager.cc
@@ -4,9 +4,9 @@
    SPDX-License-Identifier: GPL-3.0-or-later
 */
 
+#include "SessionManager.h"
 #include "Logger/Logger.h"
 #include "OnMessageTask.h"
-#include "SessionManager.h"
 #include "ThreadingManager/ThreadingManager.h"
 #include "Util/Message.h"
 #include "Util/Token.h"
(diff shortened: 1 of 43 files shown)


src/Session/SessionManager.cc
```diff
@@ -51,9 +58,9 @@ void SessionManager::OnUpgrade(
         return;
     }
 
-    _session_number++;
-    // protect against overflow
-    _session_number = max(_session_number, 1u);
+    struct timeval tv;
```
veggiesaurus: we already use `uuid.h` for generating unique tokens. We shouldn't reinvent the wheel here

src/Session/SessionManager.cc
```diff
@@ -97,6 +104,7 @@ void SessionManager::OnDisconnect(WSType* ws, int code, std::string_view message
     uint32_t session_id = static_cast<PerSocketData*>(ws->getUserData())->session_id;
 
     // Delete the Session
+    _sessions[session_id]->DecreaseRefCount();
```
veggiesaurus: I think we should first check `_sessions.count(session_id)`

veggiesaurus (changes requested):

veggiesaurus:
@robsimmonds does this mean we should close #1048 without merging?

confluence:
> @robsimmonds does this mean we should close #1048 without merging?

@angus I think that we should first merge #1048, then merge dev into this, then merge this. I'm just waiting for CI to come back.

src/Session/SessionManager.cc
robsimmonds: That would mean using 128 bit numbers for these ids. I guess we can do that, but it will need more testing.

veggiesaurus: the alternative would be to use `https://www.cplusplus.com/reference/chrono/high_resolution_clock/now/` I guess.

veggiesaurus (approved):

confluence (approved):

PR #1053, opened 2022-03-15
Fix a bug of stokes animation when playing backward

markccchiang:
It is to solve the stuck problem of stokes animation when playing backward. This is one of the commits in the [draft PR 948](https://github.com/CARTAvis/carta-backend/pull/948). But I think it is better to merge this code change in a separate PR.


kswang1029 (approved):
tested. Playback modes backward and bouncing are now functional with the polarization axis. thanks @TienHao for finding this issue and @markccchiang for fixing this. 👍

confluence (approved):

Issue #150, opened 2019-04-04
2d image fitting


Kechil:
related frontend issue https://github.com/CARTAvis/carta-frontend/issues/1397

This is to provide users the ability to select a region and fit a 2D function such as 2D Gaussian to get feature properties. Note that full resolution image data need to be used as fitting input, instead of downsampled data.

Requirement document:
https://docs.google.com/document/d/1E9kL7FOhVtkDDit8VIYle61tU5PzSYXjnLpxZBCHliY/edit?usp=sharing

Implementation plan:
- [x] Basic single/multiple gaussian fitting and text outputs
- [x] Gaussian fitting with fixed initial values
- [x] Gaussian fitting of the current view in frontend
---- target for v3b3 ----
- [x] Output model/residual images
- [x] Estimate progress and listen for cancellation messages
- [ ] Gaussian fitting of selected regions
- [ ] support a background level in the fitting
- [ ] support different solvers in the fitting
- [X] ~also use beam deconvolved image as the input data for fitting~ moved to new issue https://github.com/CARTAvis/carta-backend/issues/1236

kswang1029:
Requirements:
- provide 2d gaussian function as a start.
- capability to define initial guess of fitting parameters 
- capability to fix part of fitting parameters
- provide fitting results in convenient units (e.g., gaussian center in pixel and wcs coordinates)
- provide fitting errors
- capability to fit multiple gaussians (could consider to have an upper limit, say 3 or 5) in a single shot
- export fitting results as a text file

Kechil:
v1.5

PR #1055, opened 2022-03-15
Yuhsuan/150 2d fitting

YuHsuan-Hwang:
This PR is for the first item of #150: basic single/multiple gaussian fitting and text outputs

Details:
* Added ImageFitter which set initial values and fit the current channel of the image using GSL library
* Nan pixels are ignored during the fitting
* Compute the current residual of each pixel in parallel during the fitting
* Return a result string showing fit parameters/errors and a log string showing details of the process

Corresponding PRs:
https://github.com/CARTAvis/carta-frontend/pull/1805
https://github.com/CARTAvis/carta-protobuf/pull/67
https://github.com/idia-astro/image-generator/pull/2


kswang1029 (changes requested):
overall this works nicely. 👍 Few comments:
1. I tried to generate an ellipse region to visualize the fitting result. There appears a 90 deg offset assuming FWHM_x is the first entry of the semi-axes in the region config dialog. I suggest we make this consistent. 
Screen Shot 2022-03-16 at 13 16 45

2. here is a summary of quick performance test (just monitering Mac activity monitor) of a single gaussian fit with different image sizes
Screen Shot 2022-03-16 at 13 59 52

It appears that the peak ram usage of the carta_backend while the image fitter is working is quite high comparing to the sizes of the input images. The ram usage of the carta_backend after the fitting is done seems to correlate with the peak ram usage. Might be worth to investigate this at some point. Also it takes a while if the image size is large(r). We could consider to have a visual warning at the frontend.

3. There might be memory leakage here when triggering fitting multiple times (seen with psrecord). This is using an image of 512x512 pixel.
test

4. question: do we always use full image to feed the image fitter? I suggest we have the frontend set a subimage determined by the image viewer (using the current viewable area) to feed the image fitter to improve performance.

src/ImageFitter/ImageFitter.cc
```diff
@@ -0,0 +1,227 @@
...
+
+#include <omp.h>
+
+using namespace carta;
+
+ImageFitter::ImageFitter(std::string unit) {
+    _image_unit = unit;
+
+    _fdf.f = FuncF;
+    _fdf.df = NULL; // internally computed using finite difference approximations of f when set to NULL
```
veggiesaurus: Use `nullptr` here and elsewhere in the code

veggiesaurus:
Please use something like valgrind to check your memory leaks, there's definitely something odd going on here. Perhaps you're not clearing up the GSL co py of the data?

YuHsuan-Hwang:
> overall this works nicely. 👍 Few comments:
> 
> 1. I tried to generate an ellipse region to visualize the fitting result. There appears a 90 deg offset assuming FWHM_x is the first entry of the semi-axes in the region config dialog. I suggest we make this consistent.

Do you suggest we change the definition for here or for the regions?

> 2. here is a summary of quick performance test (just monitering Mac activity monitor) of a single gaussian fit with different image sizes
> 
> 3. There might be memory leakage here when triggering fitting multiple times (seen with psrecord). This is using an image of 512x512 pixel.

Ok, I will investigate the memory usage.

> 4. question: do we always use full image to feed the image fitter? I suggest we have the frontend set a subimage determined by the image viewer (using the current viewable area) to feed the image fitter to improve performance.

We can implement this in a separate PR. Added to the implementation plan of the issue.

kswang1029:
> > overall this works nicely. 👍 Few comments:
> > 
> > 1. I tried to generate an ellipse region to visualize the fitting result. There appears a 90 deg offset assuming FWHM_x is the first entry of the semi-axes in the region config dialog. I suggest we make this consistent.
> 
> Do you suggest we change the definition for here or for the regions?
> 

I suggest we swap the definition of FWHM X and FWHM Y.


> > 2. here is a summary of quick performance test (just monitering Mac activity monitor) of a single gaussian fit with different image sizes
> > 3. There might be memory leakage here when triggering fitting multiple times (seen with psrecord). This is using an image of 512x512 pixel.
> 
> Ok, I will investigate the memory usage.
> 

👍 

> > 4. question: do we always use full image to feed the image fitter? I suggest we have the frontend set a subimage determined by the image viewer (using the current viewable area) to feed the image fitter to improve performance.
> 
> We can implement this in a separate PR. Added to the implementation plan of the issue.

👍

kswang1029:
Question about the error estimate: with a test image containing a gaussian with high S/N, it seems the errors are unexpectedly large. How do you compute the error here?
Screen Shot 2022-03-18 at 10 14 44

src/ImageFitter/ImageFitter.cc
```diff
@@ -0,0 +1,227 @@
...
+    gsl_multifit_nlinear_init(_fit_params, &_fdf, work);
+    gsl_blas_ddot(f, f, &_fit_status.chisq0);
+    int status = gsl_multifit_nlinear_driver(max_iter, xtol, gtol, ftol, print_iter ? Callback : NULL, NULL, &_fit_status.info, work);
+    gsl_blas_ddot(f, f, &_fit_status.chisq);
+    gsl_multifit_nlinear_rcond(&_fit_status.rcond, work);
+    gsl_vector_memcpy(_fit_params, y);
+
+    gsl_matrix* jac = gsl_multifit_nlinear_jac(work);
+    gsl_multifit_nlinear_covar(jac, 0.0, covar);
+    const double c = GSL_MAX_DBL(1, sqrt(_fit_status.chisq / (n - p)));
```
kswang1029: why do we need to use GSL_MAX_DBL to get the max of two double values? Here it is 1 and sqrt(_fit_status.chisq / (n - p)).

YuHsuan-Hwang:
> Question about the error estimate: with a test image containing a gaussian with high S/N, it seems the errors are unexpectedly large. How do you compute the error here?

Obtained from the the square root of diagonal elements of the covariance matrix. Please see https://www.gnu.org/software/gsl/doc/html/lls.html#overview for details of calculating the covariance matrix of unweighted data in GSL library.

src/ImageFitter/ImageFitter.cc
YuHsuan-Hwang: sqrt(_fit_status.chisq / (n - p)) is used when estimating errors of a poor fit. Please see https://www.gnu.org/software/gsl/doc/html/nls.html#exponential-fitting-example for example.

kswang1029: yes but why do we need to compare it to unity and pick up the greater one as c?

YuHsuan-Hwang: to check if the chi-squared value shows a poor fit

YuHsuan-Hwang:
> 2. here is a summary of quick performance test (just monitering Mac activity monitor) of a single gaussian fit with different image sizes
>    Screen Shot 2022-03-16 at 13 59 52

update: adjusted the algorithm to reduce done_ram

ajm-ska:
@YuHsuan-Hwang To save the need of making a separate branch and PR, would you be able to add:

> FIND_PACKAGE(GSL REQUIRED)

to say line 51 of the `CMakeLists.txt`?

It will check if gsl is present during the `cmake` stage so that you would not be notified if it happens to be missing minutes into the `make` stage. But mostly it will be necessary when creating the RHEL7 RPMS because it will allow us to use `-DGSL_INCLUDE_DIR=` and `-DGSL_LIBRARY=` flags to point it towards the newer custom installed gsl-2.5 and not the system's default gsl-1.15.

kswang1029:
@YuHsuan-Hwang when we fit a perfect gaussian image without noise, the GSL fitter reports it exceeds max iterations (it is 200 hard-coded) and returns an error. Do you think it is possible to just report the fitting results at niter=200? I suppose "reason for stopping" will tell us it exceeds max niter so that users will know and be cautious.

veggiesaurus:
> @YuHsuan-Hwang when we fit a perfect gaussian image without noise, the GSL fitter reports it exceeds max iterations 

@kswang1029 shouldn't it find a perfect fit quite easily?

src/ImageFitter/ImageFitter.cc
YuHsuan-Hwang: fixed.

YuHsuan-Hwang:
> @YuHsuan-Hwang To save the need of making a separate branch and PR, would you be able to add:
> 
> > FIND_PACKAGE(GSL REQUIRED)
> 
> to say line 51 of the `CMakeLists.txt`?
> 

@ajm-asiaa This is added.


> @YuHsuan-Hwang when we fit a perfect gaussian image without noise, the GSL fitter reports it exceeds max iterations (it is 200 hard-coded) and returns an error. Do you think it is possible to just report the fitting results at niter=200? I suppose "reason for stopping" will tell us it exceeds max niter so that users will know and be cautious.

@kswang1029 Adjusted to show a warning message and results instead of just an error.

Also adjusted error estimation to correctly calculate the errors of unweighted data

kswang1029:
> > @YuHsuan-Hwang when we fit a perfect gaussian image without noise, the GSL fitter reports it exceeds max iterations
> 
> @kswang1029 shouldn't it find a perfect fit quite easily?

It should. There are a number of conditions from gsl for convergence test. Somehow none is met except max_niter.

veggiesaurus:
> > 2. here is a summary of quick performance test (just monitering Mac activity monitor) of a single gaussian fit with different image sizes
> >    Screen Shot 2022-03-16 at 13 59 52
> 
> update: adjusted the algorithm to reduce done_ram

@kswang1029 can you repeat these tests?

kswang1029:
@YuHsuan-Hwang @veggiesaurus Here is a set of tests I tried with psrecord to track memory usage on macOS. Here are the steps for images with different sizes.
1. have a fresh backend and a fresh session
2. load the test image and configure the fitting initial guess
3. start psrecord and trigger fitting
4. once it is done, idle for few seconds
5. close the image  

Following are the results:
256x256
test_256x256

512x512
test_512x512

1024x1024
test_1024x1024

2048x2048
test_2048x2048

4096x4096
test_4096x4096

My observations:
1. for 512x512 and 1024x1024, even after the image is closed, we don't see memory is freed. Weird caching happening here? If we simply load an image (say 1GB in size) and close it directly, we see the memory usage drops back to the level of the initial state.
2. for 2048x2048 and 4096x4096, after the image is closed, we do see lots of memory is freed but it stays at few 100 MB levels. Weird caching?
3. the pattern of 256x256 is unlike 512x512 and 1024x1024 but alike 2048x2048 and 4096x4096
4. the fitting execution time and memory usage during the fitting stage increase exponentially(?) when the image size (a side) is doubled. 
5. the origin of the second pump of memory usage (right after the fitting is done) is suspicious.

YuHsuan-Hwang:
@kswang1029 Thanks for so many testings.
About the execution time, maybe we should compare the calculation time per iteration.
The second pump of memory usage is for calculating the covariance matrix. The pump disappears if we remove `gsl_multifit_nlinear_covar(jac, 0.0, covar);`. I'm not sure why we don't see the same pattern for 256x256 image.

kswang1029:
> @kswang1029 Thanks for so many testings.
> About the execution time, maybe we should compare the calculation time per iteration.
> The second pump of memory usage is for calculating the covariance matrix. The pump disappears if we remove `gsl_multifit_nlinear_covar(jac, 0.0, covar);`. I'm not sure why we don't see the same pattern for 256x256 image.

@YuHsuan-Hwang I tried to use scipy to fit 1024x1024 image, and it took ~4s. So there must be something odd going here.

kswang1029:
if we set the solver as `fdf_params.solver = gsl_multifit_nlinear_solver_cholesky;` in `int ImageFitter::SolveSystem()` (default solver is `gsl_multifit_nlinear_solver_qr`) the fitting time is reduced by ~2/3 on my laptop and comparable to `scipy.optimize.least_squares` (astropy has a wrapper of this). The final solutions + error estimates from both solvers (qr vs cholesky) are quite similar. But with cholesky, the backend pops up an error as 

> [2022-03-29 13:17:41.754] [CARTA] [error] gsl error: cholesky.c line645: matrix is not positive definite

more info about qr vs cholesky.c
https://www.gnu.org/software/gsl/doc/html/nls.html#c.gsl_multilarge_nlinear_solver.gsl_multifit_nlinear_solver_qr
https://www.gnu.org/software/gsl/doc/html/nls.html#c.gsl_multilarge_nlinear_solver.gsl_multifit_nlinear_solver_cholesky

YuHsuan-Hwang:
issues unrelated to memory usage and execution time are fixed.

kswang1029:
@YuHsuan-Hwang I spotted an issue when trying to fit three gaussians. 
Please try the attached image with the initial guess. Component 3 has odd solutions.
Screen Shot 2022-03-31 at 12 30 04
Screen Shot 2022-03-31 at 12 30 09
Screen Shot 2022-03-31 at 12 30 16

[Gaussian_triple.fits.zip](https://github.com/CARTAvis/carta-backend/files/8386017/Gaussian_triple.fits.zip)

Screen Shot 2022-03-31 at 12 26 56

kswang1029 (approved):
As the initial step of implementation, this is good to go. Here is an example with three-gaussian fit and the result is good.
Screen Shot 2022-04-02 at 09 45 42

src/ImageFitter/ImageFitter.cc
```diff
@@ -0,0 +1,231 @@
...
+            gsl_vector_get(_fit_errors, i * 6 + 2), _image_unit.size() > 0 ? fmt::format(" ({})", _image_unit) : "");
+        _results += fmt::format(
+            "FWHM X    = {:6f} +/- {:6f} (px)\n", gsl_vector_get(_fit_params, i * 6 + 3), gsl_vector_get(_fit_errors, i * 6 + 3));
+        _results += fmt::format(
+            "FWHM Y    = {:6f} +/- {:6f} (px)\n", gsl_vector_get(_fit_params, i * 6 + 4), gsl_vector_get(_fit_errors, i * 6 + 4));
+        _results += fmt::format(
+            "P.A.      = {:6f} +/- {:6f} (deg)\n", gsl_vector_get(_fit_params, i * 6 + 5), gsl_vector_get(_fit_errors, i * 6 + 5));
+        _results += "\n";
+    }
+}
```
confluence: I don't know if this has been discussed before or is planned for a future change, but would it not be cleaner to send raw component data in the response and format it into a string on the client side, instead of sending an arbitrary string formatted on the server side?

We could make a submessage for a value+error tuple (this could have a generic name, since it could potentially be reused elsewhere), and return a map of names:tuples in the response.

Is this too time-consuming to change now?

veggiesaurus: it would allow us to format things nicely on the frontend, so I do think it makes sense. But I also don't mind if this happens at a later date.

confluence (changes requested):
I'm upgrading the response format suggestion to a request for changes, following a discussion with @veggiesaurus. :)

src/ImageFitter/ImageFitter.cc
YuHsuan-Hwang: I planned to change this in the next PR because we will need to convert the values into world coordinate. But I think we can make the changes now since it's raised.

veggiesaurus: I think it makes sense to solve it now, because it's going to affect the protocol buffer messages, so that would mean _another_ version bump and coordinated PR for frontend, backend and carta-protobuf

YuHsuan-Hwang: @confluence @veggiesaurus fixed this issue. also changed `GaussianComponent` to doubles in order to show values with 6 digit precision in frontend.

confluence (approved):
I'm happy with this.

markccchiang (approved):

test/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ set(TEST_SOURCES
         TestHdf5Image.cc
         TestHistogram.cc
         TestIcd.cc
+        TestImageFitting
```
veggiesaurus: should this have a `.cc`?

veggiesaurus (approved):
one minor comment to check

test/CMakeLists.txt
YuHsuan-Hwang: fixed.

PR #1062, opened 2022-03-16
Fix pointer memory leaks

### LINKED ISSUE #1058: Memory leaks due to not cleaning up casacore objects
**Describe the bug**
Valgrind picks up a number of memory leaks when running the unit tests. Two of these are related to casacore objects that we're not cleaning up properly.

- `Session::782`: `casacore::CoordinateSystem* csys = _frames.at(file_id)->CoordinateSystem();`  
- `RegionHandler::1147`: `casacore::LCRegion* lcregion = ApplyRegionToFile(region_id, file_id, report_error);`

### LINKED ISSUE #1059: HDF5 file loader has memory leaks
**Describe the bug**
`Hdf5Loader::GetStatsDataTyped` creatse a `casacore::ArrayBase` raw pointer. This is fine, and we `delete` the pointer in `LoadStats2DBasic` and `LoadStats3DBasic` etc, but if an exception is thrown, this pointer is never deleted. We should make this a `unique_ptr<casacore::ArrayBase>` to get around this.

pford:
Closes #1058 
Closes #1059 

Fix memory leaks reported by valgrind:
- Frame::GetCoordinateSystem returns `std::shared_ptr<casacore::CoordinateSystem>`
- Region::GetImageRegion returns `std::shared_ptr<casacore::LCRegion>` for region converted to output coordinate system
- FileLoader::GetStatsData returns `std::unique_ptr<casacore::ArrayBase>` for HDF5 stats data

Also refactor FileLoader:
- casacore::IPosition instead of using IPos
- GetShape and GetCoordinateSystem to return requested object instead of bool


veggiesaurus:
@pford I'm still getting the `RegionHandler` error. I've attached the valgrind output, if that's any help.
image


[valgrind.zip](https://github.com/CARTAvis/carta-backend/files/8332942/valgrind.zip)

src/ImageData/FileLoader.cc
```diff
@@ -135,22 +135,12 @@ bool FileLoader::HasData(FileInfo::Data dl) const {
...
 }
 
-bool FileLoader::GetShape(IPos& shape) {
-    if (_image_shape.empty()) {
-        return false;
-    }
-
-    shape = _image_shape;
-    return true;
+casacore::IPosition FileLoader::GetShape() {
```
confluence, line 138: Should this be `IPos`, if we're trying to use that consistently? There are also a few other places where we're still using `casacore::IPosition`.

confluence: I'm not 100% sure if it makes sense to keep using `IPos`, when we don't use aliases for other casacore types. I would also be happy to remove these completely and use the full type name everywhere.

confluence (changes requested):

confluence (approved):

pford:
Updated title and description to include #1059

confluence:
Still approved after #1059 fix.

veggiesaurus:
Just running `valgrind` once more to double check, then I'll approve :slightly_smiling_face:

veggiesaurus (approved):

PR #1067, opened 2022-03-18
Calculate the spatial coordinate range for a rotated field

### LINKED ISSUE #1050: spatial range of a rotated field in the file info
as confirmed with @markccchiang, currently the spatial range of an image cube is derived using the bottom-left (BL) pixel and the top-right (TR) pixel. This works nicely if there is no sky rotation (mostly true with radio image cube). However, for image cubes from optical observations, there might be a non-zero sky rotation. When this happens, using the BL and TR pixels to derive the spatial range is not accurate. 

To have a better estimate of the spatial ranges in both ra/dec or lon/lat coordinates, we need to use the four corners of the images to get the world coordinates first. Then from the four ra (or lon) to fine the maximum range. Same applies to dec (or lat).

markccchiang:
Closes #1050. Taking into account the rotation of a spatial coordinate in the range calculation. The minimum and maximum values are calculated from four corners of an image plane.


kswang1029 (approved):
Tested with some images (also fits.gz) and it works fine.

confluence (approved):

PR #1068, opened 2022-03-18
768 Check file info parameters for duplicate request

### LINKED ISSUE #768: CARTA hangs when attempting double-click to load an image with long header
Not sure if it is the same issue and root cause as https://github.com/CARTAvis/carta-backend/issues/431 so I just create this new one. Feel free to close this one if it duplicates.

With the attached test fits file (caution: 20000+ long header entries) and v1.4,
on Mac, I see with one click, file info takes long long time to show (but it does show). With double click, however, the backend seems hanging with CPU 100% forever.

~~on Ubuntu VM, I see with one click, file info shows up instantly. With double click, the backend hangs with very low CPU usage.~~
edit by @veggiesaurus: Ubuntu issue is a separate (frontend issue)

[long_header_20000.fits.zip](https://github.com/CARTAvis/carta-backend/files/6106789/long_header_20000.fits.zip)

UPDATE: this seems to happen with the current dev too.

pford:
Closes #768 

Double-click on file list sends FileInfoRequest message twice to the backend.  Added struct to Session to check if same request sent within one minute.

I tried to build on MacOS 10.15 to test this but it did not like the new std::shared_ptr<float[]> data(new float[N]); ("no matching constructor")

### DIFF
--- a/src/Session/Session.cc
+++ b/src/Session/Session.cc
@@ -424,7 +424,6 @@ void Session::OnFileInfoRequest(const CARTA::FileInfoRequest& request, uint32_t 
     auto file = request.file();
     auto hdu = request.hdu();
     CARTA::FileInfoResponse response;
-    string message;
 
     if (!_requested_file_info.duplicate_request(directory, file, hdu)) {
         spdlog::debug("Responding to FileInfoRequest for {}/{} hdu {}.", directory, file, hdu);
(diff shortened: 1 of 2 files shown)


jolopezl:
> I tried to build on MacOS 10.15 to test this but it did not like the new std::shared_ptr<float[]> data(new float[N]); ("no matching constructor")

Is it possible that something went wrong locally? That shared_ptr specialization is supported in C++17 ([case 3: shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr)).

veggiesaurus:
I'm not convinced that this is the right approach. I think we should rather fix this on the frontend, because there are definitely legitimate reasons for getting file info multiple times

PR #1071, opened 2022-03-23
ajm/GitHub actions format check

### LINKED ISSUE #1061: Use Github Actions to run the carta-backend format check
Use GitHub Actions to perform the format check (running scripts/style.py) instead of using circleCI with the view of utilizing Github Action to do more Continuous Integration in future.

ajm-ska:
@veggiesaurus This was quite easy. It appears the default ubuntu-latest already has Python and clang-format installed by default [(see here)](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md)). So the yml file is as short and simple as possible.

In the 5th run, I checked if it is really working by temporarily messing up the format of the backend code, and I confirm it is working.

I put the backend code back the way it was, and have removed the `.circleci/config.yml` file.

I think the PR will give a failure on CircleCI until it is merged?


ajm-ska:
@veggiesaurus I have "unfollowed" the carta-backend through the CircleCI web interface, so I think that will make it stop trying to run on CircleCI from now on.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,11 @@
+name: ci
+on:
```
veggiesaurus: you can just use `on: push` here, to trigger on all branches. See [here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-events-to-trigger-workflows) for details.

veggiesaurus (changes requested):

.github/workflows/continuous_integration.yml
ajm-ska: Well spotted! Thanks! That is even simpler now!

PR #1072, opened 2022-03-23
Fix compatibility issue with macOS 10.15

jolopezl:
MacOS 10.15 comes with clang 11, which doesn't have shared_ptr array support (see https://en.cppreference.com/w/cpp/compiler_support, under C++17 library features). So, this covers that case just for that OS system. In all others, it should keep using the proposed shared_ptr array for performance gains.

It's important to remark that the current CI can't test this, but @pford can check since she spotted the problem in #1068. 

Another remark is that this is important for our build systems. Maybe @ajm-asiaa can make another cross-check using his macOS 10.15 VM.

These preprocessor directives can be removed once we drop support for macOS 10.15.

### DIFF
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -26,16 +26,10 @@
 #include "ImageStats/StatsCalculator.h"
 #include "Logger/Logger.h"
 
-#if defined(__APPLE__) && defined(__MACH__)
-#include <Availability.h>
-#endif
-
-#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
-#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
-// nothing to define here
+#if defined(__APPLE__) && defined(__MACH__) && defined(__clang__) && defined(__clang_major__) && (__clang_major__ < 12)
+// shared_ptr_array not supported in apple clang < 12
 #else
 #define HAS_SHARED_PTR_ARRAY
-#endif
 #endif
 
 static const int HIGH_COMPRESSION_QUALITY(32);
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -30,6 +30,14 @@
 #include <Availability.h>
 #endif
 
+#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
+// nothing to define here
+#else
+#define HAS_SHARED_PTR_ARRAY
+#endif
+#endif
+
 static const int HIGH_COMPRESSION_QUALITY(32);
 
 namespace carta {
@@ -1123,17 +1131,12 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
         } else {
             casacore::Slicer section = GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes);
             const auto N = section.length().product();
-#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
-#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
+#ifdef HAS_SHARED_PTR_ARRAY
+            std::shared_ptr<float[]> data(new float[N]); // zero initialization
+            if (GetSlicerData(section, data.get())) {
+#else
             std::vector<float> data(N);
             if (GetSlicerData(section, data.data())) {
-#else
-            std::shared_ptr<float[]> data(new float[N]); // zero initialization
-            if (GetSlicerData(section, data.get())) {
-#endif
-#else
-            std::shared_ptr<float[]> data(new float[N]); // zero initialization
-            if (GetSlicerData(section, data.get())) {
 #endif
                 cursor_value = data[0];
             }
@@ -1503,29 +1506,20 @@ bool Frame::FillSpectralProfileData(std::function<void(CARTA::SpectralProfileDat
                     count(_z_axis) = nz;
                     casacore::Slicer slicer(start, count);
                     const auto N = slicer.length().product();
-#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
-#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
+#ifdef HAS_SHARED_PTR_ARRAY
+                    std::shared_ptr<float[]> buffer(new float[N]);
+                    if (!GetSlicerData(slicer, buffer.get())) {
+                        return false;
+                    }
+                    // copy buffer to spectral_data
+                    memcpy(&spectral_data[start(_z_axis)], buffer.get(), nz * sizeof(float));
+#else
                     std::vector<float> buffer(N);
                     if (!GetSlicerData(slicer, buffer.data())) {
                         return false;
                     }
                     // copy buffer to spectral_data
                     memcpy(&spectral_data[start(_z_axis)], buffer.data(), nz * sizeof(float));
-#else
-                    std::shared_ptr<float[]> buffer(new float[N]);
-                    if (!GetSlicerData(slicer, buffer.get())) {
-                        return false;
-                    }
-                    // copy buffer to spectral_data
-                    memcpy(&spectral_data[start(_z_axis)], buffer.get(), nz * sizeof(float));
-#endif
-#else
-                    std::shared_ptr<float[]> buffer(new float[N]);
-                    if (!GetSlicerData(slicer, buffer.get())) {
-                        return false;
-                    }
-                    // copy buffer to spectral_data
-                    memcpy(&spectral_data[start(_z_axis)], buffer.get(), nz * sizeof(float));
 #endif
 
                     // update start z and determine progress
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1124,16 +1124,16 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
             casacore::Slicer section = GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes);
             const auto N = section.length().product();
 #ifdef __MAC_OS_X_VERSION_MIN_REQUIRED
-    #if __MAC_OS_X_VERSION_MIN_REQUIRED < 101100
+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 101100
             std::vector<float> data(N);
             if (GetSlicerData(section, data.data())) {
-    #else
-            std::shared_ptr<float[]> data(new float[N]); // zero initialization
-            if (GetSlicerData(section, data.get())) {
-    #endif
 #else
             std::shared_ptr<float[]> data(new float[N]); // zero initialization
-            if (GetSlicerData(section, data.get())) 
+            if (GetSlicerData(section, data.get())) {
+#endif
+#else
+            std::shared_ptr<float[]> data(new float[N]); // zero initialization
+            if (GetSlicerData(section, data.get()))
 #endif
                 cursor_value = data[0];
             }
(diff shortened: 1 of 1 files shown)


ajm-ska (approved):
This seems to work fine. I can build it in the macOS 10.15 VM. Images open OK, and all the Unit Tests pass. 

If the newly updated `Jenkinsfile-full` file was part of this branch, I could easily run the full ICD test suite through Jenkins (although not actually testing 10.15), but the PR hasn't been approved yet. So instead I just ran the full set of ICD tests against this branch on the RHEL8 server natively and all ICD tests pass.

src/Frame/Frame.cc
```diff
@@ -1119,8 +1123,18 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
         } else {
             casacore::Slicer section = GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes);
             const auto N = section.length().product();
+#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
```
veggiesaurus: There must be a neater way to do this 🤣

confluence: Would it not be better to add `&& __MAC_OS_X_VERSION_MIN_REQUIRED < 110000` to the outer conditional instead of having a nested conditional with two identical `else` bodies?

src/Frame/Frame.cc
```diff
@@ -1489,13 +1503,30 @@ bool Frame::FillSpectralProfileData(std::function<void(CARTA::SpectralProfileDat
                     count(_z_axis) = nz;
                     casacore::Slicer slicer(start, count);
                     const auto N = slicer.length().product();
+#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
```
confluence: As above; I would flatten this into a single conditional.

confluence (changes requested):

src/Frame/Frame.cc
veggiesaurus: I can't recall if it's possible, but doing some sort of `#define HAS_SHARED_PTR defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)` monstrosity might make the actual code more readable.

confluence: Good point; if we're going to collapse this into one check we can also do this.

pford (approved):
I can build on 10.15 with this branch.

src/Frame/Frame.cc
```diff
@@ -26,6 +26,18 @@
...
 
+#if defined(__APPLE__) && defined(__MACH__)
+#include <Availability.h>
+#endif
+
+#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
+#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
+// nothing to define here
+#else
+#define HAS_SHARED_PTR_ARRAY
```
confluence: Surely this should *also* be defined if you're not on a mac at all? And only *not* defined if you're on a mac and it's old? This leaves it undefined if you're not on a mac (which was not the case in the earlier code).

This is clearer if you collapse it into a single conditional (please check my negation logic):

```
#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 110000
// nothing to define here
#else
#define HAS_SHARED_PTR_ARRAY
#endif
```

confluence (changes requested):

jolopezl:
I found a way that doesn't depend on the additional header, and it works on macOS 10.15 (after just a quick check, so it needs a full check again, unfortunately, from people with access to those machines). This version does not require the additional header and explicitly covers the case where Clang is less than 12. We should keep in mind that this only explicitly covers the cases of our supported systems with the most default compiler configurations. I don't think we need to go beyond that.

src/Frame/Frame.cc
```diff
@@ -1119,8 +1125,13 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
         } else {
             casacore::Slicer section = GetImageSlicer(AxisRange(x), AxisRange(y), AxisRange(CurrentZ()), stokes);
             const auto N = section.length().product();
+#ifdef HAS_SHARED_PTR_ARRAY
             std::shared_ptr<float[]> data(new float[N]); // zero initialization
             if (GetSlicerData(section, data.get())) {
+#else
+            std::vector<float> data(N);
```
jolopezl: I am questioning myself now if we should try to keep the `shared_ptr` using:
```
std::shared_ptr<float> data(new float[N], [](int *v){
    delete[] v;
});
```
This is also discussed [here](Another option is also discussed [here](https://stackoverflow.com/a/13062069/2201117). Another nice alternative instead of the lambda function is passing `std::default_delete`:
```
std::shared_ptr<float> sp(new float[10], std::default_delete<float[]>());
```


I could try if I configure a VM and make another pull request for that. However, this will work in the meantime and affects one version only.

confluence (commented):
I'm happy to merge this once it's been re-tested.

confluence (approved):

src/Frame/Frame.cc
veggiesaurus: This seems like a good idea to me

pford:
Works on macOS 10.15 clang 11.0.3

pford (approved):

ajm-ska:
My macOS 10.15 actually has Clang 12 (from CommandLineTools)

> -- The C compiler identification is AppleClang 12.0.0.12000032
> -- The CXX compiler identification is AppleClang 12.0.0.12000032

It worked yesterday, but not today. I get many errors like this:
```
[ 66%] Building CXX object CMakeFiles/carta_backend.dir/src/Frame/Frame.cc.o
/Users/a10.15/carta-backend-fix-macos-101500-compatibility/src/Frame/Frame.cc:1129:38: error: no matching constructor for initialization of
      'std::shared_ptr<float []>'
            std::shared_ptr<float[]> data(new float[N]); // zero initialization
                                     ^    ~~~~~~~~~~~~
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/memory:3654:23: note: candidate constructor not viable: no known conversion from 'float *' to
      'std::nullptr_t' (aka 'nullptr_t') for 1st argument
    _LIBCPP_CONSTEXPR shared_ptr(nullptr_t) _NOEXCEPT;
                      ^
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/memory:3668:5: note: candidate constructor not viable: no known conversion from 'float *' to
      'const std::__1::shared_ptr<float []>' for 1st argument
    shared_ptr(const shared_ptr& __r) _NOEXCEPT;
    ^
```
From [here](https://trac.macports.org/wiki/XcodeVersionInfo#macOS10.15), macOS 10.15 officially supports up to Xcode 12.4 with clang-1200.0.32.29

I'm confused. Does it mean that even though this 10.15 system is using Clang 12, it doesn't have shared_ptr support?

I guess the problem is here:
```
#if defined(__APPLE__) && defined(__MACH__) && defined(__clang__) && defined(__clang_major__) && (__clang_major__ < 12)
// shared_ptr_array not supported in apple clang < 12
#else
#define HAS_SHARED_PTR_ARRAY
#endif
```

I tried the previous commit `f4f430929` and it works. So we probably need to define it as: ?
```
#if defined(__APPLE__) && defined(__MACH__) && defined(__MAC_OS_X_VERSION_MIN_REQUIRED)
#if __MAC_OS_X_VERSION_MIN_REQUIRED < 110000
// nothing to define here
#else
#define HAS_SHARED_PTR_ARRAY
#endif
#endif
```

confluence (changes requested):
I'm putting back my request for changes so that it doesn't get lost if we go back to the previous version.

veggiesaurus:
I think developers on 10.15 should still be using the latest compatible xcode. But once CARTA is _built_ using a shared ptr array, will it work on vanilla 10.15?

I do think @jolopezl's suggestion of simply specifying the destructor explicitly might be a cleaner solution. In fact, we might not even need to put this in a `#define`, as it will work just fine on newer compilers, won't it? And we can just remove the explicit destructor when we drop 10.15 support (once macOS 13 is out)

ajm-ska:
@veggiesaurus In the 10.15 VM, I deleted CommandLineTools with `sudo rm -rf /Library/Developer/CommandLineTools`. Now  there is no `clang` present, as I don't have Xcode installed either. 
I downloaded and started up the CARTA v3.0.0-beta.2b Electron version, and it works fine. 
Am I right that version uses the shared ptr array? If so, then it does work on vanilla 10.15.

veggiesaurus:
No, that still uses std vector. You'd need to build from dev and test there.

ajm-ska:
@veggiesaurus OK. I did a quick test. I built the latest 'dev' carta-backend on Big Sur, used dylibbundler to package it, copied it over and ran it on the vanilla 10.15 VM. It appears to run fine:

```
a10.15@acdcs-Mac bin % ./carta_backend 
[2022-03-25 16:47:02.054] [CARTA] [info] Writing to the log file: /Users/a10.15/.carta/log/carta.log
[2022-03-25 16:47:02.055] [CARTA] [info] /Users/a10.15/test/bin/./carta_backend: Version 3.0.0-beta.2c
[2022-03-25 16:47:02.058] [CARTA] [warning] Could not find CARTA frontend files in directory /Users/a10.15/test/bin/./../share/carta/frontend.
[2022-03-25 16:47:02.058] [CARTA] [warning] Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument.
[2022-03-25 16:47:02.058] [CARTA] [info] Listening on port 3002 with top level folder /, starting folder /Users/a10.15/test/bin. The number of OpenMP worker threads will be handled automatically.
^C[2022-03-25 16:47:04.986] [CARTA] [info] Exiting backend.
a10.15@acdcs-Mac bin % 
a10.15@acdcs-Mac bin % ./carta_backend --version
3.0.0-beta.2c
a10.15@acdcs-Mac bin % 
```

veggiesaurus:
> @veggiesaurus OK. I did a quick test. I built the latest 'dev' carta-backend on Big Sur, used dylibbundler to package it, copied it over and ran it on the vanilla 10.15 VM. It appears to run fine:
> 
> ```
> a10.15@acdcs-Mac bin % ./carta_backend 
> [2022-03-25 16:47:02.054] [CARTA] [info] Writing to the log file: /Users/a10.15/.carta/log/carta.log
> [2022-03-25 16:47:02.055] [CARTA] [info] /Users/a10.15/test/bin/./carta_backend: Version 3.0.0-beta.2c
> [2022-03-25 16:47:02.058] [CARTA] [warning] Could not find CARTA frontend files in directory /Users/a10.15/test/bin/./../share/carta/frontend.
> [2022-03-25 16:47:02.058] [CARTA] [warning] Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument.
> [2022-03-25 16:47:02.058] [CARTA] [info] Listening on port 3002 with top level folder /, starting folder /Users/a10.15/test/bin. The number of OpenMP worker threads will be handled automatically.
> ^C[2022-03-25 16:47:04.986] [CARTA] [info] Exiting backend.
> a10.15@acdcs-Mac bin % 
> a10.15@acdcs-Mac bin % ./carta_backend --version
> 3.0.0-beta.2c
> a10.15@acdcs-Mac bin % 
> ```

Ok, so what does this mean for this PR, @jolopezl ?

jolopezl:
> > @veggiesaurus OK. I did a quick test. I built the latest 'dev' carta-backend on Big Sur, used dylibbundler to package it, copied it over and ran it on the vanilla 10.15 VM. It appears to run fine:
> > ```
> > a10.15@acdcs-Mac bin % ./carta_backend 
> > [2022-03-25 16:47:02.054] [CARTA] [info] Writing to the log file: /Users/a10.15/.carta/log/carta.log
> > [2022-03-25 16:47:02.055] [CARTA] [info] /Users/a10.15/test/bin/./carta_backend: Version 3.0.0-beta.2c
> > [2022-03-25 16:47:02.058] [CARTA] [warning] Could not find CARTA frontend files in directory /Users/a10.15/test/bin/./../share/carta/frontend.
> > [2022-03-25 16:47:02.058] [CARTA] [warning] Failed to host the CARTA frontend. Please specify a custom location using the frontend_folder argument.
> > [2022-03-25 16:47:02.058] [CARTA] [info] Listening on port 3002 with top level folder /, starting folder /Users/a10.15/test/bin. The number of OpenMP worker threads will be handled automatically.
> > ^C[2022-03-25 16:47:04.986] [CARTA] [info] Exiting backend.
> > a10.15@acdcs-Mac bin % 
> > a10.15@acdcs-Mac bin % ./carta_backend --version
> > 3.0.0-beta.2c
> > a10.15@acdcs-Mac bin % 
> > ```
> 
> Ok, so what does this mean for this PR, @jolopezl ?

I think that if it works, that's fine. This change originated because we needed to build the backend un macOS 10.15 using Clang 11, that's the original Clang in macOS. It seems that upgrading it doesn't necessarily solve the problem, and I can't find at which exact version clang started to support the shared_ptr array.

**BUT**, I woke up this morning and realized that this change was much much more straightforward. We wanted those smart pointers to avoid initialization, but those pointers are never shared!!!, even better: there's no problem with `unique_ptr<T[]>` in Clang 11, so no compatibility issue, so just the wrong choice in the past.

confluence:
This looks *much* better. Can we re-re-test this?

jolopezl:
> This looks _much_ better. Can we re-re-test this?

I think this should work without any problems, we already use a `unique_ptr<T[]>` for `_image_cache`.

confluence (approved):

ajm-ska (approved):
I tested two scenarios:
1. Build on 10.15 with Clang 12.0.0 - carta-backend runs and unit tests pass.
2. Build on 10.15 with Clang 11.0.3 - carta-backend runs and unit tests pass.

I also built it on the RHEL8 server and ran the ICD tests. All tests passed. 

So everything seems to be OK now.

PR #1073, opened 2022-03-29
ajm/switching centos8 to almalinux

ajm-ska:
Although @jolopezl found a way to update the repositories in the CentOS8 Dockerfile due to CentOS8 now being EOL:
```
RUN dnf -y --disablerepo '*' --enablerepo=extras swap centos-linux-repos centos-stream-repos
RUN dnf -y distro-sync
```
maybe it is about time to switch over to AlmaLinux instead?

If you agree, I have modified the Dockerfile to use the `almalinux:latest` base image. 

AlmaLinux is open source and 1:1 binary compatible with RHEL8 (Another alternative could be Rocky Linux, but I chose AlmaLinux because we use the ALMA radio telescope array, even though AlmaLinux has absolutely no relation to ALMA at all).

As in the other Dockerfiles, it creates a development environment for the carta-backend to be built, and run, as well as allowing the unit-tests to be built and run (`cmake .. -Dtest=on`). It does not install the packages for using the Address Sanitizer debugging, if that is OK?

### DIFF
--- a/Dockerfiles/Dockerfile-AlmaLinux
+++ b/Dockerfiles/Dockerfile-AlmaLinux
@@ -1,14 +1,13 @@
-FROM centos:8.4.2105
+FROM almalinux:latest
 
 # Install the required packages
 RUN \
   dnf -y install epel-release && \
   dnf -y install 'dnf-command(config-manager)' && \
   dnf -y config-manager --set-enabled powertools && \
-  dnf -y install centos-release-openstack-victoria && \
   dnf -y update && \
   dnf -y install autoconf automake bison blas-devel bzip2 cfitsio-devel cmake curl-devel flex gcc \
-         gcc-c++ git git-lfs gsl-devel hdf5-devel lapack-devel \ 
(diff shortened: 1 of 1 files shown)


confluence (approved):
It looks like Alma Linux and Rocky Linux are currently completely interchangeable; we should keep an eye on the progress of both projects in case one emerges as a clear winner. But I see no reason not to go with the appropriately-named Alma for now. :)

jolopezl:
I think this is a nice change, I don't have any preference between AlmaLinux or RockyLinux, [both look good](https://wiki.almalinux.org/Comparison.html).

> As in the other Dockerfiles, it creates a development environment for the carta-backend to be built, and run, as well as allowing the unit-tests to be built and run (`cmake .. -Dtest=on`). It does not install the packages for using the Address Sanitizer debugging, if that is OK?

If something goes wrong with testing at the CI one would really want to debug properly. I understand that we might want to keep these images small (my laptop appreciates that!), but we also need lower-level debugging capabilities. What about making another Dockerfile for the whole `llvm` part? Maybe it isn't even that much space.

jolopezl (approved):

PR #1074, opened 2022-03-29
remove mentions of boost and add extra condition to link filesystem

ajm-ska:
This came about when I ran into trouble making a Flatpak. The GCC versions in Flatpak (11.2.0 and 10.2.0) were not able to link with `stdc++fs` for some reason:
```
/usr/lib/gcc/x86_64-unknown-linux-gnu/11.2.0/../../../../x86_64-unknown-linux-gnu/bin/ld: cannot find -lstdc++fs
```

Once I commented out the stdc++fs linking line in the CMakeLists.txt, it worked.

It is strange because, for example, GCC 10.2.0 in Ubuntu 21.10, works fine even with the linking line in CMakeLists.txt. 

It seems we do not need to explicitly link to `stdc++fs` since GCC version 9 https://gcc.gnu.org/gcc-9/changes.html :
`Using the types and functions in <filesystem> does not require linking with -lstdc++fs now.`

We still use GCC 8.5 in RHEL8 for example so the linking line is still needed. 

This PR makes two changes.

1. I propose to use `CMAKE_CXX_COMPILER_VERSION` as another condition. Although technically we should be able to reduce it to version 9.0, let's put it to 11.2.0 for now. Then it should only come into effect when using Flatpak (and likely the upcoming Ubuntu 22.04).
```
if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION LESS 11.2.0)
    set(LINK_LIBS ${LINK_LIBS} stdc++fs)
endif ()
```

2. As we are modifying that part of the CMakeLists.txt, we can use this opportunity to completely remove the rest of the mentions of boost::filesystem.

### DIFF
--- a/src/Util/FileSystem.h
+++ b/src/Util/FileSystem.h
@@ -7,12 +7,7 @@
 #ifndef CARTA_BACKEND__UTIL_FILESYSTEM_H_
 #define CARTA_BACKEND__UTIL_FILESYSTEM_H_
 
-#ifdef _BOOST_FILESYSTEM_
-#include <boost/filesystem.hpp>
-namespace fs = boost::filesystem;
-#else
 #include <filesystem>
 namespace fs = std::filesystem;
-#endif
 
 #endif // CARTA_BACKEND__UTIL_FILESYSTEM_H_
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,13 +146,7 @@ if (DisableContourCompression)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_DISABLE_CONTOUR_COMPRESSION_")
 endif (DisableContourCompression)
 
-option(UseBoostFilesystem "UseBoostFilesystem" OFF)
-
-if (UseBoostFilesystem)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_BOOST_FILESYSTEM_")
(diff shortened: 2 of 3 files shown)


ajm-ska:
Just for peace-of-mind, I will first test it on Ubuntu 21.10 and the Ubuntu 22.04 beta (Both use GCC 11.2.0)

src/Util/FileSystem.h
```diff
@@ -7,12 +7,7 @@
 #ifndef CARTA_BACKEND__UTIL_FILESYSTEM_H_
 #define CARTA_BACKEND__UTIL_FILESYSTEM_H_
 
-#ifdef _BOOST_FILESYSTEM_
```
veggiesaurus, line 10: Also need to edit `CommonTestUtilities.h`

veggiesaurus (changes requested):

src/Util/FileSystem.h
ajm-ska: Thanks!

ajm-ska:
It works fine on Ubuntu 21.10, but my existing Ubuntu 22.04 Docker container appears to have unrelated problems at the moment, so I am unable to test it right now. I will continue in the morning, so just in case you are waiting, I have turned this into a Draft PR.

ajm-ska:
OK. I'm confident that this works fine.

veggiesaurus (approved):

confluence (approved):

PR #1081, opened 2022-04-08
Fix ImageExprTest to pass Hdf5 test

pford:
Since the latest carta-casacore update containing a casacore::ImageOpener fix, LEL expressions for HDF5 images are successful. This PR moves the HDF5 test out of ImageExprFails and into Hdf5ImageExprTimesTwo.


ajm-ska (approved):
Thanks @pford This works fine.
I took the liberty to directly remove the `--gtest_filter=-ImageExprTest.ImageExprFails` filter in the`Jenkinsfile` and `Jenkinsfile-full` files that we were previously using to skip this test.
Now the CI will run all the Unit Tests.

PR #1082, opened 2022-04-11
Data type check in OpenFile to use LEL to open complex image

pford:
When user opens complex image with double-click or from command-line, check extended file info for data type and open file with LEL expression "AMPLITUDE(filename)" instead of usual loader.

Also refactored Message class a bit to use const references and vector iterators.


pford:
Will combine with reverted data type PR and resubmit.

PR #1079, opened 2022-04-04
Add data type to file info

pford:
As a first step for complex image support #520 , include data type in the extended file information computed_entries so that the frontend can use an LEL expression to open the image.

FileLoader has been removed from FileExtInfoLoader so that the complex image does not have to be opened from the loader, which currently fails.  Instead, the image is opened with casacore::ImageOpener, which returns an untyped casacore::LatticeBase*.  For non-float CASA images, a generic `ImageInterface<float>` is created from the image metadata only (shape, CoordinateSystem, units, MiscInfo, ImageInfo) so that ImageHeaderToFITS can continue to be used (requires `<float>`).  The data is ignored since it is not used here.

Determining the image's coordinate axes (x/y/z axes, spectral axis, stokes axis) has been moved to Util/Casacore.h so that it can be accessed from FileExtInfoLoader and FileLoader.  These values are returned and stored in a new struct `CoordinateAxes`.  Since FileExtInfoLoader no longer sets the stokes header values in FileLoader, these values are obtained from the image CoordinateSystem in the FileLoader.

PR #1085, opened 2022-04-13
Support complex data type

pford:
A simpler code change than failed PR #1079 and #1082 - keeps the loaders in FileExtInfoLoader to handle the complexity of opening images, e.g. compressed FITS.

When opening an image in the loaders, the native image data type is stored so that it can be shown in the file info.  When opening a complex image by filename is requested (e.g. double-click or open from command line), this data type is checked and the OpenFile message is converted to an OpenFile message with an LEL expression using amplitude.

Non-float CASA images (PagedImage) are converted to `TempImage<float>` with no data for file info after storing the data type from the original image.  If FileLoader::GetImage is called, the image type is checked for TempImage and fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [3.0.0-beta.3]
+
+### Added
+* Added data type to file info and open complex image with amplitude expression ([#520](https://github.com/CARTAvis/carta-backend/issues/520)).
+
 ## [3.0.0-beta.2]
 
 ### Changed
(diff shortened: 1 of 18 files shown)


confluence (approved):
I'm happy with the code; we just need @kswang1029 to test this.

jolopezl:
Works well with https://github.com/CARTAvis/carta-frontend/pull/1824. Thanks!

kswang1029:
@pford @confluence There are few issues picked up via the e2e tests:
1. for the file info of a HDF5 image, there should be an entry as "Has mipmaps = T" to indicate if the hdf5 image has mip maps or not. In this branch, this entry is missing.
2. computed images such as moment images and pv images are not appended automatically so we can not visualize them directly in the image viewer. Images appear to be generated successfully though.

pford:
>     1. for the file info of a HDF5 image, there should be an entry as "Has mipmaps = T" to indicate if the hdf5 image has mip maps or not. In this branch, this entry is missing.

The entry is at the bottom; as far as I can tell, this was the previous behavior.
mipmaps

>     2. computed images such as moment images and pv images are not appended automatically so we can not visualize them directly in the image viewer. Images appear to be generated successfully though.
Fixed.

kswang1029 (approved):
@pford thanks for addressing the issues. Computed images are now appended automatically after generation. 

Regarding the missing "Has mipmaps = T" entry, it is a bug in the e2e test. The reason why the test cannot pass is because the file info list is just too long to above the threshold so that not all entries in the view are rendered (hence the element does not exist and cannot be verified). "Has mipmaps = T" is just above the threshold so it is not generated and rendered until we scroll down. I have fixed the test and now the test can pass well. Apologies for this confusing human error.

Issue #1096, opened 2022-05-03
carta_backend built with ASAN flags crashes when opening a DS9 region file on M1 based Macs


ajm-ska:
**Describe the bug**
Our M1 Mac is unable to open DS9 region files when built with ASAN flags. It was first noticed through the ICD tests, but can be reproduced manually using the frontend. It is using the latest carta-casacore 3.4.0+6.5.0+2022.4.22.
This problem _only_ occurs on the M1 Mac with ASAN flags.

**To Reproduce**
1. Use an M1 Mac.
2. Build carta-backend using ASAN flags:
```
git clone https://github.com/CARTAvis/carta-backend.git
cd carta-backend
mkdir build
cd build
cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS='-O0 -g -fsanitize=address -fno-omit-frame-pointer' -DCMAKE_EXE_LINKER_FLAGS='-fsanitize=address' 
make -j 8
```
3. Start up the carta-backend using ASAN flags and an appropriate frontend. For example:
```
ASAN_OPTIONS=suppressions=/Users/acdc/carta-backend/debug/asan/myasan.supp LSAN_OPTIONS=suppressions=/Users/acdc/carta-backend/debug/asan/myasan-leaks.supp ASAN_SYMBOLIZER_PATH=/opt/homebrew/Cellar/llvm/13.0.1_1/bin/llvm-symbolizer ./carta_backend /images --top_level_folder /images --omp_threads 8 --no_database --verbosity=5 --frontend_folder=/Users/acdc/carta-frontend/build
``` 
4. Open the `M17_SWex.image` image.
5. File > Import Regions > `M17_SWex_handMakeRegions_world.reg`
6. carta_backend will crash with the following output: [ds9-reg-crash.txt](https://github.com/CARTAvis/carta-backend/files/8610251/ds9-reg-crash.txt)

(A zip file with `M17_SWex.image` and `M17_SWex_handMakeRegions_world.reg` is available for download here: [ds9-reg-crash-package.zip](https://github.com/CARTAvis/carta-backend/files/8610296/ds9-reg-crash-package.zip) )

**Expected behavior**
The carta_backend should not crash.

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]: macOS Monterey
 - Browser [e.g. chrome, safari, electron app]: Chrome
 - Browser version [e.g. 22]: 100
[...]

Issue #1066, opened 2022-03-17
carta_backend built with ASAN flags crashes when importing a specific region file.


ajm-ska:
**Describe the bug**
When we run the `DS9_REGION_EXPORT.test.ts` ICD test on macOS12, the carta_backend crashes. We can also trigger it manually.

**To Reproduce**
1. Build the carta_backend with the following flags:
```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS='-O0 -g -fsanitize=address -fno-omit-frame-pointer' -DCMAKE_EXE_LINKER_FLAGS='-fsanitize=address' 
```
2. Start the carta-backend with the appropriate ASAN flags (remember to adjust paths for your system). This way proves we are taking the current ASAN suppressions into account):
```
ASAN_OPTIONS=suppressions=/Users/ajm/carta-backend/debug/asan/myasan.supp LSAN_OPTIONS=suppressions=/Users/ajm/carta-backend/debug/asan/myasan-leaks.supp ASAN_SYMBOLIZER_PATH=llvm-symbolizer ./carta_backend --frontend_folder=/Users/ajm/carta-frontend/build --verbosity=5
```
or as normal:
```
./carta_backend --frontend_folder=/Users/ajm/carta-frontend/build --verbosity=5
```

3. Open the `M17_SWex.image` image file.

4. Open the `M17_SWex_handMadeRegions_world.reg` file available [here](https://alma.asiaa.sinica.edu.tw/_downloads/M17_SWex_handMadeRegions_world.reg)
 
5. carta_backend crashes.

**Expected behavior**
The carta_backend should not crash.

**Screenshots or videos**
The full crash log can be viewed here: [asan-region-crash.log](https://github.com/CARTAvis/carta-backend/files/8282636/asan-region-crash.log)
 

**Platform info (please complete the following information):**
- OS: macOS12 M1
- Browser: Firefox
- Version: 98

**Additional context**
If the carta_backend is built without the ASAN flags, i.e. only `cmake ..`, the crash does not occur.

acdo2002:
So far it occurs related to "export region" action, the error messages show between `[protocol] <== EXPORT_REGION` and `[protocol] ==> EXPORT_REGION_ACK`. 
Not related to the image type (happened in all three types: fits, image, and hdf5)
Can trigger manually by export region as "**world coordinates**" for both CRTF region file and DS9 region file. No error if I export region as "pixel coordinates"

update:
[...]

PR #1099, opened 2022-05-05
Fix changelog before release

confluence:
I have added what I think are all the relevant entries that were missing from the changelog. I have also moved an entry that I incorrectly added to the previous release, and changed the title of the current release section to `Unreleased` -- this should only be changed to the release version during the release.

I have omitted changes which affect only unit tests, compilation or our CI, since I don't think these are user-facing, but I did add a catch-all entry for all the ASAN fixes and memory leak fixes. I have also omitted fixes to changes within the same release (these should be considered part of the original change). Please let me know if you think I've missed something that should be added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Removed gRPC service and moved scripting interface to HTTP server ([#1022](https://github.com/CARTAvis/carta-backend/pull/1022)).
 * Added more fine-grained commandline flags to enable and disable functions of the HTTP server.
+* Optimised performance of image data cache ([#967](https://github.com/CARTAvis/carta-backend/issues/967)).
 * Added exit on timeout flag to Linux desktop launcher ([#989](https://github.com/CARTAvis/carta-backend/issues/989)).
 * Separated stdout and stderr logging ([#869](https://github.com/CARTAvis/carta-backend/issues/869)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,23 +4,44 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [3.0.0-beta.3]
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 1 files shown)


pford:
Should the DS9 import bug be added, or is it internal only (ASAN)?  If so, it was missed.

* Fixed DS9 import bug on MacOS ([#1066](https://github.com/CARTAvis/carta-backend/issues/1066)), ([#1096](https://github.com/CARTAvis/carta-backend/issues/1096)).

I guess this is part of the general ASAN entry.

pford (approved):
All my stuff is in.  Thanks @confluence !

jolopezl (changes requested):
Thanks! I am sorry for not adding this when due. I am missing the performance improvements in the image cache (#967, #1034).

confluence:
> I am missing the performance improvements in the image cache (#967, #1034).

I was on the fence about that, but you're right; if it's a performance optimisation, that's user-facing. I'll put it in.

jolopezl:
> > I am missing the performance improvements in the image cache (#967, #1034).
> 
> I was on the fence about that, but you're right; if it's a performance optimisation, that's user-facing. I'll put it in.

Thanks!

CHANGELOG.md
```diff
@@ -4,23 +4,45 @@ All notable changes to this project will be documented in this file.
...
+* Removed gRPC service and moved scripting interface to HTTP server ([#1022](https://github.com/CARTAvis/carta-backend/pull/1022)).
+* Added more fine-grained commandline flags to enable and disable functions of the HTTP server.
+* Optimised performance of image data cache ([#967](https://github.com/CARTAvis/carta-backend/issues/967)).
+* Added exit on timeout flag to Linux desktop launcher ([#989](https://github.com/CARTAvis/carta-backend/issues/989)).
+* Separated stdout and stderr logging ([#869](https://github.com/CARTAvis/carta-backend/issues/869)).
 
 ### Added
+* Added support for creating CASA LEL images dynamically ([#655](https://github.com/CARTAvis/carta-backend/issues/665)).
+* Added spatial and spectral range information to the file browser ([#845](https://github.com/CARTAvis/carta-backend/issues/845)).
 * Added computed stokes images and supported the analysis (profiles, contours, statistics, etc.) of them ([#433](https://github.com/CARTAvis/carta-backend/issues/433)).
```
kswang1029: stokes -> Stokes

jolopezl (approved):

PR #1103, opened 2022-05-06
Fix DS9 world region import

### LINKED ISSUE #1101: "box" region loaded from ds9 region file has incorrect angular size
**Describe the bug**
I have a rectangular ds9 region saved as a file. When I load the region into CARTA the size of the region is significantly smaller than expected and is also significantly smaller than the same region loaded into `ds9`.

**To Reproduce**
1. Use a region file the following contents:
```
# Region file format: DS9 astropy/regions
fk5
box(335.184002,-24.680416,0.000278,0.000278,0.000000) # name=0
```
2. File->"Open Image" to load FITS image including that coordinate in the FOV
3. File->"Import Regions" to load the region file specified in 1) above.

**Expected behavior**
A rectangular region centered on (RA,Dec)=(335.184002,-24.680416) with width and height of 0.000278deg should be showed on the image panel. The region panel of CARTA should show the same information for the region. This is the behavior of loading the region file in `ds9`.

The actual behavior is that a box is shown at the position, but the height and width are 0.000278arcsec. So this seems to be a unit interpretation problem. 

 - OS: MacOS 10.15.17 / Linux
 - Browser: Local CARTA install / Firefox 91.8.0esr 
 - Version: 3.0.0-beta.2b / 2.0.0

**Additional context**
This bug seems to be present in both v2.0.0 and v3.0.0-beta.2b, tested using the combinations listed above (i.e., 3.0.0-beta.2b on MacOS 10.15.17 with a local install, and 2.0.0 run from Linux with `--no_browser` and viewed remotely via Firefox).

pford:
Closes #1101 

For region definitions with wcs defined, import pure numbers (no unit) as degrees.

Specifically, box width/height and ellipse radii are deg (was arcsec, not sure why).  Still export these as explicit arcsec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [3.0.0-beta.3]
+## [Unreleased]
 
 ### Added
 * Added computed stokes images and supported the analysis (profiles, contours, statistics, etc.) of them ([#433](https://github.com/CARTAvis/carta-backend/issues/433)).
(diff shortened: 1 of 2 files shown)


confluence (approved):

kswang1029 (approved):
@pford thanks for addressing this issue. 👍

PR #1111, opened 2022-05-24
Tests for the vector fields calculator 

markccchiang:
These testing codes are separated and refactored from PR #1016.


kswang1029:
@veggiesaurus @confluence can we get this in first so that the ci can routinely run it as a guard for new commits? This one is pending for too long. New unit tests can follow your framework later on I think.

confluence:
> @veggiesaurus @confluence can we get this in first so that the ci can routinely run it as a guard for new commits? This one is pending for too long. New unit tests can follow your framework later on I think.

I am concerned about the amount of novel code added in these tests, and I need to review exactly how they interface with the source. But I agree that this has been in limbo for too long, and I will focus on it.

confluence:
This PR needs major revisions before it can be merged. Most of these tests are too high-level to be unit tests (which should be very low-level tests of individual functions in the component), many of them are outside the scope of testing the vector field calculator component (they should be in separate unit tests for other components), and the tests of this component should be checking its output against known reference data (rather than comparing its output to the output of the same algorithm applied to a different selection of data).

I think that it will be clearer what form these unit tests should take (and whether any changes should be made to the calculator code to support that) once we have discussed the unit testing strategy and documented a concrete set of guidelines for writing tests. In the meantime, I am going to convert this and PR #1112 (which has similar issues) to draft, so that it's clear that they have been deferred for now.

Issue #1013, opened 2022-01-27
per cube histogram gets calculated when switching to a new Stokes value


kswang1029:
**Describe the bug**
The render configuration widget and the contour rendering dialog have a histogram plot showing the per-channel histogram by default. Optionally users may switch to per-cube mode to see a histogram derived using the entire cube. 

When a 4D cube with multiple Stokes (eg Stokes IQUV) is loaded in CARTA and we request to see the per-cube histogram of Stokes I, we will see the backend starts to deliver partial per-cube histogram to the frontend. Once it is done, if we use the animator to switch to Stokes Q, for example, we will see the "Histogram" dropdown switches to per-channel (as expected) but the backend starts to calculate the _per-cube_ histogram of Stokes Q and once it is finished, the backend then calculates the per-channel histogram of the Stokes Q in the image viewer.  

**To Reproduce**
1. load a 4D cube with multiple Stokes and enable the JS console
2. request per-cube histogram
3. use the animator to switch to Stokes Q and monitor the JS console
4. we should see per-cube histogram being calculated first, then the per-channel histogram is calculated.

When the per-cube histogram is calculated we will see nothing happens from the GUI. But if we request a per-cube histogram when the backend is busy on calculating the unexpected per-cube histogram, we will see the progress bar shows up repeatedly. Sometimes this results in incorrect clip bounds (I saw 0,1).

**Expected behavior**
The expected behavior in step 4 should be that only the per-channel histogram is calculated. I guess the root cause is the logic behind SET_IMAGE_CHANNELS (where we can set a Stokes value via stokes parameter) and SET_HISTOGRAM_REQUIREMENTS (where we can request a per-channel or a per-cube histogram via region_id parameter)

**Screenshots or videos**
This is a screenshot of the JS console. The last REGION_HISTOGRAM_DATA carries the per-channel histogram. All the other REGION_HISTOGRAM_DATA before carry the unnecessary per-cube histogram data.

![Screen Shot 2022-01-27 at 12 53 08](https://user-images.githubusercontent.com/20819712/151294235-6191f989-6530-4eef-8d91-8fac7051d2d9.png)

**Platform info (please complete the following information):**
 - OS: only tested with macOS Monterey 
 - Browser: chrome
 - Version 97.0.4692.71

**Additional context**
N/A

markccchiang:
[...]

PR #1113, opened 2022-05-25
Do not calculate per-cube histogram when stokes is changed

markccchiang:
It is to solve #1013. The cube histogram should not be automatically calculated when stokes is changed.


kswang1029 (approved):
yes the issue is gone. 👍

acdo2002 (approved):
Follow the reproduce steps in #1013 

The current backend and frontend dev branch, the ICD sequences would be:
image

However in the branch and the corresponding frontend branch `mark/1013_avoid_auto_per_cube_histogram`.
The ICD sequences become:
image

The additional 
image
has been removed.

confluence (approved):

PR #1117, opened 2022-05-27
Confluence/http error codes

### LINKED ISSUE #1011: Fix error status codes in HTTP server
You can only set a status on a uWebSockets response once, and you have to set status first, then headers, then a body. To ensure that a status is set before headers are set, the uWebSockets function for setting headers automatically sets the status to 200 first. This means that any subsequent attempt to set a different status will have no effect.

In our server code, there are several places where we first set headers (through `AddNoCacheHeaders` or otherwise) and then attempt to set a status, which is silently failing. If the status is supposed to be 200 anyway, this makes no difference, but it's impossible for these functions ever to return an error status.

I found this while moving the scripting interface into the HTTP server, but I think that this needs to be fixed and tested in a separate PR -- it's possible that other parts of CARTA are depending on these functions always returning a 200 status, and that something will break once the error codes are restored.

### LINKED ISSUE #1115: More fine-grained HTTP error messages
Currently if scripting is disabled scripting requests receive a generic authentication error as a response. This is the same error that is returned in the case of an actual authentication failure, which is confusing. A more specific error message should be added. Analogous messages should be added for other server features which may be disabled.

This can probably be handled at the same time as #1011.

confluence:
Fixes #1011. Fixes #1115 (has a companion PR in the wrapper, which is independent of this merge). Also fixes the changelog to include some recent commits that we missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+* Enhanced image fitting performance by switching the solver from qr to cholesky ([#1114](https://github.com/CARTAvis/carta-backend/pull/1114)).
+* Made HTTP server return a different error code for disabled features ([#1115](https://github.com/CARTAvis/carta-backend/issues/1115)).
+
+### Fixed
+
+* Stopped calculating per-cube histogram unnecessarily when switching to a new Stokes value ([#1013](https://github.com/CARTAvis/carta-backend/issues/1013)).
+* Ensured that HTTP server returns error codes correctly ([#1011](https://github.com/CARTAvis/carta-backend/issues/1011)).
+
 ## [3.0.0-beta.3]
 
 ### Changed
--- a/src/HttpServer/HttpServer.cc
+++ b/src/HttpServer/HttpServer.cc
@@ -50,7 +50,7 @@ void HttpServer::RegisterRoutes() {
     if (_enable_scripting) {
         app.post("/api/scripting/action", [&](auto res, auto req) { HandleScriptingAction(res, req); });
     } else {
-        app.post("/api/scripting/action", [&](auto res, auto req) { Forbidden(res, req); });
+        app.post("/api/scripting/action", [&](auto res, auto req) { NotImplemented(res, req); });
     }
 
     if (_enable_database) {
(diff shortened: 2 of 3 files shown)


confluence:
@kswang1029 I added you because although this is a conceptually simple change we are now correctly returning error codes for certain possible error conditions, and it's possible that this breaks something because we haven't been returning them before. So we need to test this thoroughly before merging.

kswang1029:
@confluence could you provide one or two example use cases that would trigger different error codes? I am not very sure how I can test this thoroughly. 🤔 

If I use a wrong token, I see the following in the browser console:

```
PUT http://localhost:3002/api/database/preferences 403 (Forbidden)
WebSocket connection to 'ws://localhost:3002/?token=854159FF-DC80-4AF4-A93A-6AFCA933E7A3' failed:
```

confluence:
@kswang1029 I think your example would always have triggered an error correctly (because the authentication check is always performed before anything else). An example of something that should have changed is reading existing preferences if the preferences file is malformed and can't be parsed, or setting or clearing preferences with a malformed json string. These are mostly developer-facing rather than user-facing situations, but we may want to check e.g. what happens if a user edits the preferences file by hand and messes it up.

kswang1029:
> @kswang1029 I think your example would always have triggered an error correctly (because the authentication check is always performed before anything else). An example of something that should have changed is reading existing preferences if the preferences file is malformed and can't be parsed, or setting or clearing preferences with a malformed json string. These are mostly developer-facing rather than user-facing situations, but we may want to check e.g. what happens if a user edits the preferences file by hand and messes it up.

@confluence Thanks for the suggestions. I tried to remove `{` in the first line of my preferences.json and launched the backend and connected the URL. The backend console shows

`[2022-05-30 15:54:37.319] [CARTA] [warning] [json.exception.parse_error.101] parse error at line 2, column 14: syntax error while parsing value - unexpected ':'; expected end of input
` 
and the GUI shows the telemetry query dialog. Then my preferences.json is overwritten by the backend (kinda "recovered"). I don't see any messages in the frontend console. Do you think all these are the expected behaviors?

I also see a similar backend warning if I did the same thing to the layout.json

kswang1029:
@confluence @veggiesaurus If I set `"colormap": "XXX",` in the preferences.json, I will see a hidden colormap got adopted! 😂 The colorbar is empty though. I don't see any relevant messages from the backend console or the frontend console. Maybe this is not a relevant issue to this PR. 

Screen Shot 2022-05-30 at 16 04 02

I think it is the first colormap of https://github.com/CARTAvis/carta-frontend/blob/dev/src/static/allmaps.png

veggiesaurus (approved):

confluence:
@kswang1029 that looks like valid JSON, just with weird data. :) I think that's not relevant to this PR. Mostly I wanted to make sure we didn't introduce a crash in common functionality which previously had a silent request failure, but if nothing turns up in the E2E tests it's probably fine.

kswang1029:
> @kswang1029 that looks like valid JSON, just with weird data. :) I think that's not relevant to this PR. Mostly I wanted to make sure we didn't introduce a crash in common functionality which previously had a silent request failure, but if nothing turns up in the E2E tests it's probably fine.

ok. all e2e tests are passing.

confluence:
OK, I think that's safe enough for now. Will you approve the PR?

kswang1029 (approved):

Issue #1069, opened 2021-11-08
incorrect mean/std visualization in the render config widget


kswang1029:
![Screen Shot 2021-11-08 at 2 52 01 PM](https://user-images.githubusercontent.com/20819712/140696910-172441d3-a9e6-4b71-ac0f-8ea595886987.png)

confluence:
Has this been fixed by #1118?

kswang1029:
yes

PR #1118, opened 2022-05-30
Convert the pixel type to double when calculating sum squares

markccchiang:
The root cause of #1069 is the overflow of float multiplication. Converting the pixel type to double can solve this problem.


kswang1029 (approved):
confirmed that the issue is fixed and all e2e tests are passing. 👍

confluence (approved):

Issue #1014, opened 2022-02-04
Seg fault when using spectral widget on (older) gzipped data with VELO-LSR velocity units


jehibbard:
**Describe the bug**
CARTA v3-2b has a deg fault when I try to open a gzipped file of an older data cube (with CTYPE units VELO-LSR)

**To Reproduce**
1. Get the cube named "Tadpole-CARMA-CO-cube.fits" from https://www.cv.nrao.edu/~jhibbard/FITS/u10214/ 
2. Gzip it
3. open in in CARTA 3.0-2b
4. in the standard view. change the channel and draw a region
5. open a z-profile widget
6. CARTA will disconnect, and the following error is reported in the xterm:

/usr/local/bin/carta: line 14: 41986 Segmentation fault: 11  carta_backend "$@"

7. Conversely, if you have the zprofile widget already open and then draw a region, it again disconnects but now the error is something like:

carta_backend(41944,0x7000099ae000) malloc: *** error for object 0x7fe01804e590: pointer being freed was not allocated
carta_backend(41944,0x7000099ae000) malloc: *** set a breakpoint in malloc_error_break to debug
/usr/local/bin/carta: line 14: 41944 Abort trap: 6           carta_backend "$@"

**Expected behavior**
It should not disconnect, and display the spectra in the zprofile widget (as it does if the non-gzipped file is opened the same way)

**Screenshots or videos**
If applicable, add screenshots/videos to help explain your problem.

**Platform info (please complete the following information):**
 - OS: [e.g. iOS] macOS Catalina v10.15.7
 - Browser [e.g. chrome, safari] Safari v15.3
 - Version [e.g. 22] CARTA 3.0-2b

**Additional context**
Kuo-Song Wang said:
"This might have the same root cause as https://github.com/CARTAvis/carta-backend/issues/999 but it is good to submit a new one in the backend repo."

confluence:
Has this been fixed by #1119? Can we close it?

jehibbard:
Yes, I get the same error using v3-b3 and following steps 1-6 from my initial report.

jehibbard:
[...]

Issue #999, opened 2022-01-12
backend crash when requesting statistics from gz or fz FITS image


kswang1029:
**Describe the bug**
backend crash when requesting statistics from gz or fz FITS image

**To Reproduce**
1. Load one of the attached images
2. Enable the statistics widget
3. Should see backend crash

**Expected behavior**
backend should not crash

**Screenshots or videos**
n/a

**Platform info (please complete the following information):**
 - OS: macOS Monterey (intel)
 - Browser: chrome
 - Version:  97.0.4692.71

**Additional context**
test image:
https://drive.google.com/file/d/1RC7CzRIsgC9XSIco4flIc1FoOrTjc2Jf/view?usp=sharing
https://drive.google.com/file/d/1SUAGtcgEavjirm7yK87GU_1UaU5khMLU/view?usp=sharing

confluence:
Was this fixed by #1119?

kswang1029:
yes

PR #1119, opened 2022-06-03
Modify the compressed FITS image loader destruction

markccchiang:
I think it can solve #999 and #1014. When doing statistics calculations, several copies of an image loader are created and run in parallel. These copies are deleted when calculations are done. In CartaFitsImage, the close file action can only be executed when all of the CartaFitsImage objects are destroyed.


kswang1029:
@markccchiang confirmed this PR does not result in any regressions that the current e2e tests are capable to capture. Manual tests are on the way before approving the PR.

kswang1029 (approved):
tested manually with compressed fits and confirmed that region statistics, line/polyline spatial profiler, region histogram, pv generator and region spectral profile work and do not trigger crash.

pford:
Keeping track of copies sounds like a shared_ptr workaround.  But the real issue is several CartaFitsImages sharing a single fitsfile* pointer.  Maybe instead the copy should have a null _fptr and it will open the file when it needs to read something?

src/ImageData/CartaFitsImage.cc
```diff
@@ -54,15 +55,18 @@ CartaFitsImage::CartaFitsImage(const CartaFitsImage& other)
       _datatype(other._datatype),
       _has_blanks(other._has_blanks),
       _pixel_mask(nullptr),
-      _tiled_shape(other._tiled_shape) {
+      _tiled_shape(other._tiled_shape),
+      _is_copy(true) {
```
pford, line 59: Set _fptr(nullptr) instead?

pford (changes requested):

pford:
It occurred to me that opening the same gz file in another CartaFitsImage may decompress it into memory again.  Can you test this?  Your solution was probably better.

markccchiang:
> It occurred to me that opening the same gz file in another CartaFitsImage may decompress it into memory again. Can you test this? Your solution was probably better.

A quick test with `psrecord`. The memory usage for statistics calculations with multiple `_fptr` is 

test3

On the other hand, the memory usage for statistics calculations with a single `_fptr` is 

test2

So maybe it is better to use the original method?

markccchiang:
> It occurred to me that opening the same gz file in another CartaFitsImage may decompress it into memory again. Can you test this? Your solution was probably better.

A quick test with `psrecord`. The memory usage for statistics calculations with multiple `_fptr` is 

test3

On the other hand, the memory usage for statistics calculations with a single `_fptr` is 

test2

So maybe it is better to use the original method?

pford:
Yes, please put it back to the tested version.  Sorry about that, I forgot about decompression!

pford (approved):

PR #1125, opened 2022-06-14
Add some wrappers to construct protobuf messages

### LINKED ISSUE #960: Use wrappers to construct protocol buffer messages where possible
Some wrappers for constructing protocol buffer messages in a more shorthand way were recently added to `Util/Message`. There are many places in the code where we could be using these wrappers instead of the more verbose multi-line syntax for filling in one field at a time (apart from the cases where we deliberately construct the message in multiple stages).

markccchiang:
Closes #960. Add some wrappers for constructing protocol buffer messages. In order to avoid the verbose multi-line syntax for filling in a protobuf message and reduce the main source code.


src/Frame/Frame.cc
```diff
@@ -753,7 +746,9 @@ bool Frame::FillRegionHistogramData(
         if (!GetStokesTypeIndex(histogram_config.coordinate, stokes)) {
             continue;
         }
-        histogram_data.set_stokes(stokes);
+
+        // create and fill region histogram data message
+        CARTA::RegionHistogramData histogram_data = Message::RegionHistogramData(file_id, region_id, z, stokes, 1.0);
```
veggiesaurus: any thoughts on whether we should use `auto` here? It should allow us to get rid of a bunch of `#includes` :shrug: @confluence @pford

confluence: That's a good idea -- the type should be unambiguous from the wrapper name, so it shouldn't affect readability.

kswang1029:
no regression is observed from the e2e tests.

src/Region/CrtfImportExport.cc
```diff
@@ -417,10 +417,7 @@ RegionState CrtfImportExport::ImportAnnSymbol(casacore::CountedPtr<const casa::A
 
         // Set control points
         std::vector<CARTA::Point> control_points;
-        CARTA::Point point;
-        point.set_x(pixel_coords[0]);
-        point.set_y(pixel_coords[1]);
-        control_points.push_back(point);
+        control_points.push_back(Message::Point(pixel_coords[0], pixel_coords[1]));
```
confluence: Given how frequently we construct a `Message::Point` from a `std::vector<casacore::Quantity>` or a `casacore::Vector<casacore::Double>`, I wonder if it would be worthwhile to add overloaded `Point` functions for these use cases (which take one of these vectors as a parameter, and have optional parameters for the x and y indices, which default to 0 and 1). Then we could e.g. write this as just `Message::Point(pixel_coords)`.

src/Region/Ds9ImportExport.cc
```diff
@@ -503,34 +497,24 @@ RegionState Ds9ImportExport::ImportEllipseRegion(std::vector<std::string>& param
...
         if (_pixel_coord) {
-            CARTA::Point point;
-            point.set_x(param_quantities[0].getValue());
-            point.set_y(param_quantities[1].getValue());
-            control_points.push_back(point);
-            point.set_x(param_quantities[2].getValue());
-            point.set_y(param_quantities[3].getValue());
-            control_points.push_back(point);
+            control_points.push_back(Message::Point(param_quantities[0].getValue(), param_quantities[1].getValue()));
+            control_points.push_back(Message::Point(param_quantities[2].getValue(), param_quantities[3].getValue()));
```
confluence: Because we sometimes use different indices, the utility functions would need optional parameters -- then this could be written as
```
control_points.push_back(Message::Point(param_quantities));
control_points.push_back(Message::Point(param_quantities, 2, 3));
```

src/Region/Ds9ImportExport.cc
```diff
@@ -503,34 +497,24 @@ RegionState Ds9ImportExport::ImportEllipseRegion(std::vector<std::string>& param
...
                 return region_state;
             }
 
             // bmaj, bmin
-            CARTA::Point point;
-            point.set_x(WorldToPixelLength(param_quantities[2], 0));
-            point.set_y(WorldToPixelLength(param_quantities[3], 1));
-            control_points.push_back(point);
+            control_points.push_back(
+                Message::Point(WorldToPixelLength(param_quantities[2], 0), WorldToPixelLength(param_quantities[3], 1)));
```
confluence, line 517: It's probably not worth doing anything about this case; we can keep it like this.

confluence (approved):
I have a suggestion for adding some utility functions, but it's not strictly necessary. In general this all looks good.

src/Region/CrtfImportExport.cc
veggiesaurus: I like that idea. Like this?
```c++
Message::Point(const std::vector<casacore::Quantiy>& input, int x_index = 0, int y_index = 1);
```

confluence: Yup. Just for these two vector types; I didn't see any other cases that I think warrant additional simplification.

src/Region/RegionHandler.cc
```diff
@@ -2559,19 +2540,15 @@ RegionState RegionHandler::GetTemporaryRegionState(std::shared_ptr<casacore::Coo
     if (corner.empty()) {
         return RegionState();
     }
-    point.set_x(corner[0]);
-    point.set_y(corner[1]);
-    control_points[0] = point;
+    control_points[0] = Message::Point(corner[0], corner[1]);
```
confluence: For consistency I think we should add one more utility function for `std::vector<double>`.

src/Util/Message.cc
```diff
@@ -107,13 +109,27 @@ CARTA::AddRequiredTiles Message::AddRequiredTiles(
...
+    return point;
+}
+
+CARTA::Point Message::Point(const std::vector<casacore::Quantity>& input, int x_index, int y_index) {
+    CARTA::Point point;
+    point.set_x(input[x_index].getValue());
+    point.set_y(input[y_index].getValue());
+    return point;
+}
+
```
confluence: These should be calling the function that takes two floats, not duplicating its internals.

confluence (changes requested):

src/Util/Message.cc
```diff
@@ -107,13 +109,34 @@ CARTA::AddRequiredTiles Message::AddRequiredTiles(
...
+CARTA::Point Message::Point(const std::vector<double>& input, int x_index, int y_index) {
+    CARTA::Point point;
+    point.set_x((float)input[x_index]);
+    point.set_y((float)input[y_index]);
+    return point;
+}
+
```
confluence: These are still duplicating the internals of `Message::Point(float x, float y)`. You should be *calling* that function from inside these functions instead (and I think that if you do that you won't need the explicit casts).

confluence (changes requested):

confluence (approved):

kswang1029 (approved):
all e2e tests are passing. 👍

PR #1126, opened 2022-06-14
Added 2022 to copyright years

confluence:
Should we switch to a year range (2018 - 2022) instead of continuing to add individual years? A range seems to be widely accepted if the years are contiguous.

### DIFF
--- a/scripts/style.py
+++ b/scripts/style.py
@@ -48,7 +48,7 @@ class Test:
 
 class Header(Test):
     HEADER = """/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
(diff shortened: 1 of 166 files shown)


veggiesaurus:
That sounds sensible to me

pford:
I think we should switch to a range now rather than keep adding to the list, which is already getting long.

https://www.gnu.org/licenses/gpl-howto.html
`For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.`

pford (approved):

Issue #1083, opened 2022-04-12
support beam info and beam rendering from an AIPS fits cube


kswang1029:
example AIPS fits images can be obtained from https://www2.mpia-hd.mpg.de/THINGS/Data.html

For such AIPS fits images, the beam info is not represented in header as BMAJ, BMIN, and BPA. It adopted a non-standard representation using the HISTORY keyword. The "standard" AIPS form is (all in degrees)

`HISTORY AIPS   CLEAN BMAJ=  3.8158E-03 BMIN=  3.4928E-03 BPA= -40.24 = `

We should still try to look for BMAJ, BMIN, and BPA to set up beam info. If those are not available, then look for that specific HISTORY to guess the beam shape for the file info (hence the frontend can render the beam properly).

[HO_II_NA_MOM0_THINGS.FITS.zip](https://github.com/CARTAvis/carta-backend/files/8469578/HO_II_NA_MOM0_THINGS.FITS.zip)

detailed discussions in https://docs.google.com/document/d/1qUm02JOLxOoG_4p_lDJpKRj7z4tnwsEh5GoOxASKNSs/edit#

Let's do 1 and 2 as described in the doc

The frontend counterpart is https://github.com/CARTAvis/carta-frontend/issues/2164

kswang1029:
@markccchiang do you have time to look into this?

markccchiang:
> @markccchiang do you have time to look into this?

Ok, I will take a look at it.

pford:
This is supported in casacore, but for #935 we removed the HISTORY beam. (See FitsLoader::RemoveHistoryBeam)

My comments in #935 indicated it was for AIPS images according to the casacore::FITSImage header comments.

kswang1029:
@jott3077 made this request to support AIPS fits cube. Deriving beam shape from history is really not ideal but we have no other choices. #935 is an example demonstrating it can lead to bad results. 

Maybe we will need to warn users in the file info if the beam info is derived from history 🤔️

veggiesaurus:
This is rather frustrating, because history is really not the appropriate place for this information, and the user was quite right to ask us to ignore it. If we add this back, we may produce misleading results, surely?

kswang1029:
[...]

Issue #935, opened 2021-10-08
FITS header extraction of BMAJ, BMIN, BPA


kswang1029:
reported by a user:

If the header does not contain BMAJ, BMIN, BPA as keys but these strings are stored as part of the HISTORY context, the backend still picks them up. The backend should ignore any keys from HISTORY.

> SIMPLE  =                    T /                                                
BITPIX  =                  -32 /                                                
NAXIS   =                    3 /                                                
NAXIS1  =                  585 /                                                
NAXIS2  =                  587 /                                                
NAXIS3  =                    1 /                                                
EXTEND  =                    T /Tables following main image                     
BLOCKED =                    T /Tape may be blocked                             
OBJECT  = '        '           /Source name                                     
TELESCOP= '        '           /                                                
INSTRUME= '        '           /                                                
OBSERVER= '        '           /                                                
DATE-OBS= '1987-05-26'         /Obs start date YYYY-MM-DD                       
DATE-MAP= '2017-08-14'         /Last processing date YYYY-MM-DD                 
BSCALE  =    1.00000000000E+00 /REAL = TAPE * BSCALE + BZERO                    
BZERO   =    0.00000000000E+00 /                                                
BUNIT   = 'JY/BEAM '           /Units of flux                                   
EQUINOX =      1.950000000E+03 /Epoch of RA DEC                                 
DATAMAX =      7.697948813E-01 /Maximum pixel value                             
DATAMIN =     -5.964934826E-03 /Minimum pixel value                             
CTYPE1  = 'RA---SIN'           /                                                
CRVAL1  =    8.28809244186E+01 /                                                
CDELT1  =     -2.777777845E-04 /                                                
CRPIX1  =      2.980000000E+02 /                                                
CROTA1  =      0.000000000E+00 /                                                
CTYPE2  = 'DEC--SIN'           /                                                
CRVAL2  =    2.19817385450E+01 /                                                
CDELT2  =      2.777777845E-04 /                                                
CRPIX2  =      2.750000000E+02 /                                                
CROTA2  =      0.000000000E+00 /                                                
CTYPE3  = 'FREQ    '           /                                                
CRVAL3  =    1.41000000000E+09 /                                                
CDELT3  =      1.000000000E+00 /                                                
CRPIX3  =      1.000000000E+00 /                                                
CROTA3  =      0.000000000E+00 /                                                
HISTORY --------------------------------------------------------------------    
[...]

PR #1127, opened 2022-06-14
1063 fix info/load time for FITS with long history

### LINKED ISSUE #1063: Loading FITS Images from NED Hangs CARTA
Using CARTA v1.4 server find that trying to load a FITS image from NED, specifically the FITS image at http://ned.ipac.caltech.edu/uri/NED::Image/fits/2009A+A...506.1123H/NGC_0253:I:vla+eff_6_cm_I_30arcs:H2009, hangs CARTA.  Never completes loading image, with spinning arrow display trying to show file information.

pford:
Closes #1063 

Find number of headers using cfitsio; if > 10,000 use CartaFitsImage which will read all headers into one list `_all_header_strings` and non-HISTORY headers into another list `_image_header_strings`.  Using the image headers list to set up CartaFitsImage (particularly the coordinate system) speeds this up considerably.  This has no effect for an image with a large number of headers which are not HISTORY.  All headers, including history, are shown in the file browser.

NOTE the image does not use beam information in the HISTORY headers - see #935 and #1083 (same as previous behavior, which removes the history beam when casacore::FITSImage is used).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * Stopped calculating per-cube histogram unnecessarily when switching to a new Stokes value ([#1013](https://github.com/CARTAvis/carta-backend/issues/1013)).
 * Ensured that HTTP server returns error codes correctly ([#1011](https://github.com/CARTAvis/carta-backend/issues/1011)).
+* Fixed slow loading of FITS image with large number of HISTORY headers ([#1063](https://github.com/CARTAvis/carta-backend/issues/1063)).
 
 ## [3.0.0-beta.3]
 
--- a/src/ImageData/CartaFitsImage.cc
+++ b/src/ImageData/CartaFitsImage.cc
@@ -274,10 +274,10 @@ void CartaFitsImage::SetUpImage() {
     // Read headers into string
     int nheaders(0);
     std::string header;
-    GetFitsHeaders(nheaders, header);
+    GetFitsHeaderString(nheaders, header);
 
     // Headers as String vector to pass to converter
-    casacore::Vector<casacore::String> header_strings = FitsHeaderStrings(nheaders, header);
+    SetFitsHeaderStrings(nheaders, header);
 
     casacore::Record unused_headers;
     casacore::LogSink sink;
@@ -289,7 +289,7 @@ void CartaFitsImage::SetUpImage() {
     try {
         // Set coordinate system
         coord_sys = casacore::ImageFITSConverter::getCoordinateSystem(
-            stokes_fits_value, unused_headers, header_strings, log, 0, _shape, drop_stokes);
+            stokes_fits_value, unused_headers, _image_header_strings, log, 0, _shape, drop_stokes);
     } catch (const casacore::AipsError& err) {
         if (err.getMesg().startsWith("TabularCoordinate")) {
             // Spectral axis defined in velocity fails if no rest freq to convert to frequencies
@@ -342,7 +342,7 @@ void CartaFitsImage::SetUpImage() {
     }
 }
 
-void CartaFitsImage::GetFitsHeaders(int& nheaders, std::string& hdrstr) {
+void CartaFitsImage::GetFitsHeaderString(int& nheaders, std::string& hdrstr) {
     // Read header values into single string, and store some image parameters.
     // Returns string and number of keys contained in string.
     // Throws exception if any headers missing.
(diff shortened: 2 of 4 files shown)


kswang1029 (commented):
@pford yes this improves the file info generation time very significantly. I tried the case with ~<10000 header entries (9900 HISTORY entries) and it took a while to get the file info generated. I wonder if we should set the threshold from 10000 to 2000 or 5000 for example. I tested those two cases and the elapsed time was reasonable in my point of view. What would be drawback if we set a lower threshold?

veggiesaurus:
> @pford yes this improves the file info generation time very significantly. I tried the case with ~<10000 header entries (9900 HISTORY entries) and it took a while to get the file info generated. I wonder if we should set the threshold from 10000 to 2000 or 5000 for example. I tested those two cases and the elapsed time was reasonable in my point of view. What would be drawback if we set a lower threshold?

The only additional performance issue I think this would introduce is actually counting the number of headers, but that's probably not too bad. cfitsio does this by reading 80 characters at a time IIRC

confluence (approved):

markccchiang (approved):

pford:
> > @pford yes this improves the file info generation time very significantly. I tried the case with ~<10000 header entries (9900 HISTORY entries) and it took a while to get the file info generated. I wonder if we should set the threshold from 10000 to 2000 or 5000 for example. I tested those two cases and the elapsed time was reasonable in my point of view. What would be drawback if we set a lower threshold?
> 
> The only additional performance issue I think this would introduce is actually counting the number of headers, but that's probably not too bad. cfitsio does this by reading 80 characters at a time IIRC

The drawback to using a threshold is that users would have to use another tool besides CARTA to get the full history of what processing has been done to obtain the image.  I do not know how important this is to users, so it is a user/science team decision on feature vs performance.  I think if we truncate the history we would need to let users know that it is incomplete.

The headers are already counted in this PR using cfitsio `fits_get_hdrspace` to force using CartaFitsImage over casacore::FITSImage when the number of headers is high.

kswang1029:
> > > @pford yes this improves the file info generation time very significantly. I tried the case with ~<10000 header entries (9900 HISTORY entries) and it took a while to get the file info generated. I wonder if we should set the threshold from 10000 to 2000 or 5000 for example. I tested those two cases and the elapsed time was reasonable in my point of view. What would be drawback if we set a lower threshold?
> > 
> > 
> > The only additional performance issue I think this would introduce is actually counting the number of headers, but that's probably not too bad. cfitsio does this by reading 80 characters at a time IIRC
> 
> The drawback to using a threshold is that users would have to use another tool besides CARTA to get the full history of what processing has been done to obtain the image. I do not know how important this is to users, so it is a user/science team decision on feature vs performance. I think if we truncate the history we would need to let users know that it is incomplete.
> 
> The headers are already counted in this PR using cfitsio `fits_get_hdrspace` to force using CartaFitsImage over casacore::FITSImage when the number of headers is high.

I am confused. With an image having 20000 HISTORY, I can still see all history entries in the file info. 🤔 Test images are attached.
[set_fits_long_history.zip](https://github.com/CARTAvis/carta-backend/files/8953965/set_fits_long_history.zip)

pford:
> I am confused. With an image having 20000 HISTORY, I can still see all history entries in the file info. 🤔 Test images are attached. [set_fits_long_history.zip](https://github.com/CARTAvis/carta-backend/files/8953965/set_fits_long_history.zip)

All of the headers are saved to be shown in the file info, but only the non-history headers are used to set up the image CoordinateSystem.  This step was the biggest performance hit when all headers were used.

kswang1029:
> > I am confused. With an image having 20000 HISTORY, I can still see all history entries in the file info. 🤔 Test images are attached. [set_fits_long_history.zip](https://github.com/CARTAvis/carta-backend/files/8953965/set_fits_long_history.zip)
> 
> All of the headers are saved to be shown in the file info, but only the non-history headers are used to set up the image CoordinateSystem. This step was the biggest performance hit when all headers were used.

I see. I think it is okay to just use non-HISTORY entries to set up image CoordinateSystem. Normally information in HISTORY should have zero contribution to build up coordinate system I suppose (except the aips beam patch in casacore(?))@veggiesaurus @pford @confluence would you agree?

confluence:
@pford @kswang1029 I'm not sure if we're all on the same page -- my understanding is that KS is suggesting using a smaller threshold of total headers (smaller than 10K) to trigger the optimisation, not introducing a threshold for truncating the history headers in the viewer.

Is there any reason not to do this for *all* files? Is there a significant performance overhead for files with fewer headers?

confluence:
@kswang1029 I made my last comment before seeing yours. I see no reason not to do this optimisation for all files, if there is no performance drawback and the history headers are not used to set up the CoordinateSystem.

kswang1029:
> @pford @kswang1029 I'm not sure if we're all on the same page -- my understanding is that KS is suggesting using a smaller threshold of total headers (smaller than 10K) to trigger the optimisation, not introducing a threshold for truncating the history headers in the viewer.
> 
> Is there any reason not to do this for _all_ files? Is there a significant performance overhead for files with fewer headers?

@confluence i think what @pford mentioned is if we have a header with 100 normal entries and 10000 history entries, only the 100 normal entries are used to build up coordiantesystem. If it is a header with 100 normal entries and 9899 history entries, all 9999 entries are used to set up coordiantesystem which takes time to go through all one by one (and most of them are ineffective in that task).

veggiesaurus:
> I see. I think it is okay to just use non-HISTORY entries to set up image CoordinateSystem. Normally information in HISTORY should have zero contribution to build up coordinate system I suppose 

We may get some push back from users, but I am fiercely against using `HISTORY` and `COMMENT` headers for anything other than plain text information for the users, so I agree

pford:
> @pford @kswang1029 I'm not sure if we're all on the same page -- my understanding is that KS is suggesting using a smaller threshold of total headers (smaller than 10K) to trigger the optimisation, not introducing a threshold for truncating the history headers in the viewer.
> 
> Is there any reason not to do this for _all_ files? Is there a significant performance overhead for files with fewer headers?

@confluence thank you for the explanation, I misunderstood it as a threshold for the number of headers to read.

Since the loader and its image are used for extended file info as well as for image access after it is loaded, using this for all files would eliminate the use of casacore::FITSImage entirely (unless I add a flag to use CartaFitsImage for the file browser only).  The casacore version is the default since it has been tested and used more extensively.

kswang1029 (approved):
having 2000 as threshold is good and appears fast in my test env. Also tested fits.gz and it works properly.

PR #1128, opened 2022-06-15
Yuhsuan/150 fov fitting

YuHsuan-Hwang:
This PR is for the third item in #150: image fitting with field of view.

Details:

With the fov info sent from the frontend, the backend creates a temporary region to obtain the region data. The image fitter applies an offset to the center positions of Gaussian components when fitting the region. The temporary region is removed after the fitting task is done.

The advantage of this approach is that we support fitting the current fov of a rotated image (usually a spatially matched image) without implementing complicated calculation. In addition, the users can reproduce the fitting result by saving a subimage of the fov region (with info provided in the frontend). The reproduce steps will be simpler after we support image fitting with created regions in the future (probably only few-lines change to support this).

Using the region data requires slightly more memory usage than using the existed image cache, but the patterns of the CPU and momery profiles are similar. Please check the link for details of the performance tests:
https://docs.google.com/document/d/1orLw1Szpy8awVvypkAdAwhkS1u9lYuxkiOah-9hROrE/edit?usp=sharing

Update: also fixed incorrect pixel number in `FuncF()` when fitting images with nan pixels

Corresponding PRs:
https://github.com/CARTAvis/carta-frontend/pull/1885
https://github.com/CARTAvis/carta-protobuf/pull/74

~Still working on:~
- [X] ~add unit tests~
- [X] ~send protobuf PR~


kswang1029:
@YuHsuan-Hwang this draft PR works fine as far as I can tell. It works with matched image as well. @veggiesaurus @confluence @pford please proceed with code review. 🙂

YuHsuan-Hwang:
Update: added unit tests. also fixed incorrect pixel number in `FuncF()` when fitting images with nan pixels

src/ImageFitter/ImageFitter.cc
```diff
@@ -13,28 +13,38 @@
...
-    _fdf.n = _fit_data.n;
 
     // avoid GSL default error handler calling abort()
     gsl_set_error_handler(&ErrorHandler);
 }
 
-bool ImageFitter::FitImage(
-    float* image, const std::vector<CARTA::GaussianComponent>& initial_values, CARTA::FittingResponse& fitting_response) {
+bool ImageFitter::FitImage(size_t width, size_t height, float* image, const std::vector<CARTA::GaussianComponent>& initial_values,
+    CARTA::FittingResponse& fitting_response, size_t offset_x, size_t offset_y) {
```
pford, line 28: I would prefer keeping fitting_response at the end of the parameters since it is an output value but the offsets are input values.

test/TestImageFitting.cc
```diff
@@ -60,38 +102,34 @@ class ImageFittingTest : public ::testing::Test {
...
-        std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(file_path));
-        std::unique_ptr<TestFrame> frame(new TestFrame(0, loader, "0"));
-
-        CARTA::FittingResponse fitting_response;
-        std::unique_ptr<carta::ImageFitter> image_fitter(new carta::ImageFitter(frame->GetWidth(), frame->GetHeight()));
-        bool success = image_fitter->FitImage(frame->GetImageCacheData(), _initial_values, fitting_response);
+        return file_path;
+    }
 
+    void CompareResults(CARTA::FittingResponse fitting_response, bool success, std::string failed_message) {
```
pford: Use const references for fitting_response and string parameters

pford (changes requested):

test/TestImageFitting.cc
YuHsuan-Hwang: This is fixed

src/ImageFitter/ImageFitter.cc
YuHsuan-Hwang: @pford The offset parameters are optional parameters (with default zeros), so they are added at the end.

pford (approved):

kswang1029 (approved):

confluence (approved):

kswang1029:
pending approval before all e2e test passing again

kswang1029 (approved):
via e2e tests, no regression is observed in [4da68be](https://github.com/CARTAvis/carta-backend/pull/1128/commits/4da68be8641b487248608d88c0879338d486e317)

veggiesaurus (approved):

PR #1131, opened 2022-06-17
Recognize the FITS gzip file from ALMA archive

### LINKED ISSUE #1130: images as fits.gz from ALMA Science Archive not recognized by CARTA
**Describe the bug**
Reported by a v3 beta tester. Images as fits.gz downloaded from the ALMA Science Archive are not recognized by CARTA. If we unzip it and re-zip it (in my case with the gzip in macOS "Apple gzip 353.100.22"), the new fits.gz file can be recognized.

test images are attached
[test_by_mac_gzip.fits.gz](https://github.com/CARTAvis/carta-backend/files/8918988/test_by_mac_gzip.fits.gz)
[test_from_alma_archive.fits.gz](https://github.com/CARTAvis/carta-backend/files/8918993/test_from_alma_archive.fits.gz)

**To Reproduce**
1. download the example images above (or download such from the ALMA Science Archive directly)
2. check the file list and we will only see one image showing up

**Expected behavior**
The fits.gz images from the ALMA science archive should be recognized by CARTA.

 - Browser [e.g. chrome, safari, electron app]: v3b3 electron app
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v3b3 electron app
 - Frontend branch [e.g. dev, v3b2 release]: v3b3 electron app

markccchiang:
Closes #1130. It seems the magic number of fits.gz from ALMA archive is different from that of ordinary FITS gzip files. Although they are the same file type. @kswang1029 could you please check does it can also recognize the other fits.gz files from ALMA archive? Thank you!

### DIFF
--- a/src/Util/File.cc
+++ b/src/Util/File.cc
@@ -25,7 +25,7 @@ uint32_t GetMagicNumber(const std::string& filename) {
 
 bool IsCompressedFits(const std::string& filename) {
     // Check if gzip file, then check .fits extension
-    if (IsGzFile(GetMagicNumber(filename))) {
+    if (IsGzMagicNumber(GetMagicNumber(filename))) {
         fs::path gz_path(filename);
         std::string extension = gz_path.stem().extension().string();
         return HasSuffix(extension, ".fits");
@@ -34,7 +34,7 @@ bool IsCompressedFits(const std::string& filename) {
     return false;
 }
 
-bool IsGzFile(uint32_t magic_number) {
+bool IsGzMagicNumber(uint32_t magic_number) {
     std::string hex_string = fmt::format("{:#x}", magic_number);
     return (hex_string.length() > 4) && (hex_string.substr(hex_string.length() - 4) == "8b1f");
 }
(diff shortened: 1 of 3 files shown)


kswang1029:
uh, nice catch. Does this imply with different brands of gzip, the magic number is different?

markccchiang:
> uh, nice catch. Does this imply with different brands of gzip, the magic number is different?

According to [this link](https://docs.fileformat.com/compression/gz/), the magic number of gz file is `1f 8b`. The other number corresponds to the compression method (usually `08` for DEFLATE) and file flags. The header of the ordinary gz file is `1f8b 0808` and the header of ALMA is `1f8b 0800`. I guess the only difference between them is the file flag. So maybe we can identify gz files by only checking whether it has `1f8b` number.

kswang1029 (approved):
checked with alma gz images and some others in the test image pool. All working. No regression is observed from e2e tests.

src/Util/File.cc
```diff
@@ -34,6 +34,11 @@ bool IsCompressedFits(const std::string& filename) {
     return false;
 }
 
+bool IsGzFile(uint32_t magic_number) {
```
confluence: This is a confusing name for a function which takes a magic number as a parameter (not a file name or file object). I would suggest `IsGzMagicNumber` for more clarity.

confluence (changes requested):
Other than the function name, I think this is fine.

confluence (approved):

PR #1139, opened 2022-06-27
Fix CartaFitsImage mask for floating point data

pford:
Bug found with e2e testing failure after merge of PR to fix long history headers by using CartaFitsImage.  The image stats were incorrect, with the wrong number of pixels and all other stats set to NaN.

All values were being used for the calculations since there was no mask from the BLANK header.  Only integer FITS images have a BLANK header with a value representing undefined pixels, but float and double images use NaN values. CartaFitsImage now creates the pixel mask from float/double data values which are finite (not NaN or infinite).


confluence (approved):

kswang1029 (approved):
all e2e tests are passing now. I have done some manual tests and it works fine as far as I can tell. 👍

PR #1145, opened 2022-07-11
Fix matched polygon pixel count

### LINKED ISSUE #1108: error in pixel count of a matched polygon region
**Describe the bug**
When there are two images with identical WCS header matched spatially, a polygon region may give inconsistent pixel count as seen with the statistics widget on the matched image. Here is "inconsistent" does not mean they have to be identical on both image as CARTA adopted polygon resampling again in order to perform region reprojection on matched image so it is possible to have few pixel differences. The issue here is the difference appears too large as expected.

**To Reproduce**
1. download the test images at https://drive.google.com/file/d/1uv2f-quEVsf1r9Imt_SwNS6EFv1NvHmO/view?usp=sharing
2. load image01.fits and image02.fits and match them spatially
3. load the region file polygon.crtf
4. load two statistics widget
5. configure one to show image01.fits and the region explicitly and configure the other one to show image02.fits and the region explicitly too
6. check the NumPixels entry and we see quite different pixel counts

**Expected behavior**
The expected NumPixels from both images should be identical theoretically but it is fine to have just few pixels differences (understandable because of the resampling process). 

**Screenshots or videos**
Screen Shot 2022-05-17 at 12 38 24

 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v3b3 release
 - Frontend branch [e.g. dev, v3b2 release]: v3b3 release

pford:
Closes #1108 

casacore LCPolygon sets the mask for each y in the region's bounding box to the x value of the intercept of the line segment if not horizontal or the x value of the next point if horizontal.  Setting many points along the line resulted in many horizontal line segments which added pixels not included in reference image.

In this fix, for a matched polygon with an approximately horizontal line segment, only the points nearest an even y-pixel are selected in that segment.

### DIFF
--- a/src/Region/Region.cc
+++ b/src/Region/Region.cc
@@ -381,7 +381,7 @@ std::shared_ptr<casacore::LCRegion> Region::GetImageRegion(int file_id, std::sha
                 spdlog::debug("Using direct region conversion for {}", RegionName(region_state.type));
             } else {
                 // Use polygon approximation of reference region to translate to another image
-                spdlog::debug("Using polygon approximation for rotbox or distorted region {}", RegionName(region_state.type));
+                spdlog::debug("Using polygon approximation for matched {} region", RegionName(region_state.type));
                 lc_region = GetAppliedPolygonRegion(file_id, output_csys, output_shape);
 
                 // Cache converted polygon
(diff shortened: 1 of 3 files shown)


kswang1029:
@pford tested with a rectangle with a rotation of 1 degree and a polygon with sides ~horizontal. Pixel counts are not identical but very close so this is okay and understandable. For an oblate ellipse rotated by 1 degree, we see large errors. The visualization of the sampled pixels is at the bottom-right panel. 

Screen Shot 2022-07-12 at 12 36 41

pford:
@kswang1029 I applied the horizontal segment fix to the matched ellipse polygon approximation.

kswang1029 (approved):
all e2e tests are passing. Manual test results are promising. 👍 Ready for code review @pford @confluence

confluence (approved):

src/Region/Region.cc
```diff
@@ -1311,3 +1356,55 @@ bool Region::ConvertWorldToPixel(std::vector<casacore::Quantity>& world_point, s
...
+
+        // Line connecting points ~horizontal - keep point nearest integral pixel
+        int pixel_y = static_cast<int>(this_y);
+
+        if (!ValuesNear(this_y, float(pixel_y))) {
+            // Skip point not near pixel
+            continue;
+        }
+
+        if ((static_cast<int>(next_y) == pixel_y) && ((this_y - pixel_y) > (next_y - pixel_y))) {
```
markccchiang, line 1391: I am just wondering is it better using `fabs(this_y - pixel_y) > fabs(next_y - pixel_y)` than `(this_y - pixel_y) > (next_y - pixel_y)`?

pford: pixel_y is the integer part of the float values y and next _y, so these subtraction values will always be positive.

markccchiang (approved):

PR #1150, opened 2022-07-15
578 check CASA image read-lock to prevent hang

### LINKED ISSUE #578: CARTA freezes when file browser is opened during file change
**System:** Linux, Firefox 79.0-1 via carta.idia.ac.za

**Issue description:** Opening the file browser in a directory where a file is being written freezes CARTA the moment the file is trying to be opened.

**To reproduce:**
 1. Opening a CASA image file in CARTA.
 2. Delete the file on the disk.
 3. Re-create the CARA image with different image parameters (only tested with same file size).
 4. While the file is being written open the file browser and access the file.

**Workaround: In CARTA:** Reload the website

pford:
Closes #578 

When requesting file info for a CASA image which is being written, the process waits for a read-lock in order to construct the PagedImage in CasaLoader (to convert to FITS headers), causing carta to hang.  In this fix, casacore::LockFile is used to check whether the image can be read-locked; if not, file info fails and the error message indicates the image is locked by another process.

I tried creating the PagedImage with the AutoNoReadLocking option but it still waited to acquire the lock.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed slow loading of FITS image with large number of HISTORY headers ([#1063](https://github.com/CARTAvis/carta-backend/issues/1063)).
 * Fixed the DS9 import bug with region properties ([#1129](https://github.com/CARTAvis/carta-backend/issues/1129)).
 * Fixed incorrect pixel number when fitting image with nan pixels ([#1128](https://github.com/CARTAvis/carta-backend/pull/1128)).
-* Fixed file list hang when file is being written ([#578](https://github.com/CARTAvis/carta-backend/pull/578)).
+* Fixed errors on loading images via LEL ([#1144](https://github.com/CARTAvis/carta-backend/issues/1144)).
+* Fixed the DS9 import bug with no header line ([#1064](https://github.com/CARTAvis/carta-backend/issues/1064)).
+* Fixed file info hang when a file is being written ([#578](https://github.com/CARTAvis/carta-backend/pull/578)).
 
 ## [3.0.0-beta.3]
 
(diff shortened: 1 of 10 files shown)


kswang1029:
all e2e tests are passing as expected. 

I tried the following test:
1) load a CASA image (80000x80000, 24G) on a laptop with 16GB RAM. It takes a long while as my laptop is swapping a lot.
2) once the image is loaded in CARTA (seeing the image is rendered), I delete the image
3) then from another directory I copy the same image to the same directory that CARTA has accessed. This takes several seconds on my laptop with SSD.
4) while the copy is in progress, I request an image file list. I see the spinner shows up, then a file list including the image being copied with a smaller file size (17.2 GB, guessing this is a snapshot of the increasing size).
5) while copying is still in progress, I request the file info and it can be generated successfully.

Not sure if all the above steps and results make sense to you @pford @confluence ? I also repeated the test with v3b3 macOS electron app. The behavior is the same. 🤔 so I am not quite sure what I tried can emulate the use case in #578

confluence (approved):

pford:
@kswang1029 I used the CASA task `importfits`, which converts a FITS image to CASA.  importfits locks the CASA image until it is complete, which would hang the file info waiting for the lock.

kswang1029:
@pford that's a clever way! 
I tried your branch and v3b3, while importfits is running, with both versions, I can still quickly get a file list from a directory where the casa image is being converted from FITS. The "casa image" is not recognized as a casa image yet and is recognized as a usual folder so we can navigate in and see logtable, mask0, etc. I think I cannot reproduce the original hanging issue. 🤦🏻‍♂️

kswang1029:
@pford wait I am going to try again. I missed a critical step. 🤦🏻🤦🏻🤦🏻

kswang1029 (approved):
now I can see the difference. However, with v3b3, I do see the spinner in the file info pane  when the casa image is locked. Once the casa image is ready, if I click another image and click the casa image again, I will see the file info. So CARTA is not hanging forever in my case. 

Anyway, I think that this improvement does prevent the forever run of file info generation. 👍 

ps. e2e tests are all passing.

pford:
carta does not hang forever, just until the other process releases the lock, which could be a while for a large image.  The user is stuck with the spinner until the file info is shown (carta "freezes" in the issue description) and cannot select another file.

PR #1153, opened 2022-07-19
ajm/jenkins slack notifications

ajm-ska:
An ongoing issue has been with `Jenkinsfile-full`. The Jenkinsfile-full builds the carta-backend, runs the unit tests, and runs the full-ICD test suite, on all 7 supported platforms. It takes ~45 minutes and uses a lot of local computational resources. It is not practical to run on every commit. Therefore we had decided to only run it once a day on the 'dev' branch. The issue has been that, even though the Jenkins job finishes, the build status remains 'pending' on Github indefinitely. I just can't figure out why. The Jenkins log says it sends out the final status message, but Github seemingly fails to receive it for some unknown reason.

This PR is an attempt to overcome the issue.

I propose to use the [Jenkins Slack Notification plugin](https://plugins.jenkins.io/slack/). Instead of reporting the build status of Jenkinsfile-full to Github, it reports the build status to Slack. We can join the `#jenkins-build-status` Slack channel (and mute it if it is annoying). If something changes in the carta-backend that breaks the build or unit tests on the platforms that we don't cover with the basic `Jenkinsfile`, or breaks the ICD tests, we will notice immediately.

Here are some notes:

- To reduce the number of messages posted, it will only post a message if any of the 7x carta-backend builds or 7x Unit Test runs fail. It will not post a success for each one, but we will still get one global 'success' message at the very end.
- At first, I wanted it to report the platform of the ICD test that failed. That didn't work out. It seems when using a matrix configuration (which we need to use otherwise the Jenkinsfile becomes too long to parse) then if one stage fails, all consecutive stages are marked as failures, even if they pass. This seems to be a [known issue](https://issues.jenkins.io/browse/JENKINS-61405). The easiest solution is to put the post command outside the matrix loop. It will report one global 'success' or one global 'unstable'. We can simply click the URL to see which stage failed and check the detailed log output, something that we would need to do anyway.
- If a backend build or unit test fails, it will be reported as a failure. If an ICD test does not pass, I decided to mark it 'unstable' because most times, that's what it is. The ICD tests usually pass after running a second time, or they need to be updated due to legitimate backend changes. I know this is not ideal. Either way, we always check, and if there is a serious issue introduced, we will report it. Also, @acdo2002's upcoming RxJS re-write of the ICD tests may help the ICD tests to pass more reliably. Not to mention, the ongoing work to migrate some ICD tests to Unit Tests. So over time, this may become less of an issue. Another advantage is we can also conveniently identify the issue from the colour of the Slack message; red will mean the backend or unit tests failed, yellow will mean an ICD test failed.
- The job on Jenkins to run the Jenkinsfile-full needs to be a Pipeline, and not a Multi-branch pipeline. For a Pipeline job, we can stop it from reporting the build status to Github, but I can't for a Multi-branch pipeline. In order to get the Branch ID from a Pipeline job, I needed to set it as a global variable in the Groovy context. Otherwise, we could have simply used the ${env.BRANCH_ID} variable that is available by default for a Multi-branch pipeline. Also, we need to apply the `Check out to specific local branch` setting in the Pipeline job, otherwise `BRANCH_ID` will only report `HEAD` and not the branch name. (Note: The main `Jenkinsfile` runs through a Multi-branch pipeline job).
- I have also updated the main `Jenkinsfile` to use slackSend. It still reports the build status of every commit to Github, but now also reports to Slack. I plan to use slackSend in the carta-frontend Jenkinsfile too. Similarly, to reduce the number of messages sent to Slack, `Jenkinsfile` only reports one global success or one global failure. 
- On the `#jenkins-build-status` Slack channel, for the main Jenkinsfile, we can see the identifier 'carta-backend' followed by the status (Success or Failure), the branch name, the commit ID, and a URL direct to the build in the Jenkins Blue Ocean interface. For the daily Jenkinsfile-full build, I have given it the name `Full ICD test suite`. You can see an example in the attached image below:
Screen Shot 2022-07-19 at 2 17 42 PM

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,7 @@
 pipeline {
     agent none
     environment {
-        COMMIT_ID = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
+        COMMIT_ID=''
     }
     options {
         preserveStashes() 
@@ -15,6 +15,9 @@ pipeline {
                     }
                     steps {
                         catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
+                            script {
+                                COMMIT_ID = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
+                            }
                             sh "uname -a"
                             sh "lsb_release -a"
                             sh "git submodule update --init --recursive"
@@ -127,7 +130,7 @@ pipeline {
             slackSend color: 'good', message: "Success - ${env.BRANCH_NAME} ${COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
         }
         failure {
-             slackSend color: 'danger', message: "Failed - ${env.BRANCH_NAME} ${COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
+             slackSend color: 'danger', message: "Failed - ${env.BRANCH_NAME} ${env.COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,8 @@
 pipeline {
     agent none
+    environment {
+        COMMIT_ID = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
+    }
     options {
         preserveStashes() 
     }
@@ -121,10 +124,10 @@ pipeline {
     }
     post {
         success {
-            slackSend color: 'good', message: "Success - ${env.BRANCH_NAME} ${env.GIT_REVISION:0:7} (<${env.RUN_DISPLAY_URL}|open>)";
+            slackSend color: 'good', message: "Success - ${env.BRANCH_NAME} ${COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
         }
         failure {
-             slackSend color: 'danger', message: "Failed - ${env.BRANCH_NAME} ${env.GIT_REVISION:0:7} (<${env.RUN_DISPLAY_URL}|open>)";
+             slackSend color: 'danger', message: "Failed - ${env.BRANCH_NAME} ${COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,12 +1,3 @@
-void setBuildStatus(String message, String state) {
-  step([
-      $class: "GitHubCommitStatusSetter",
-      reposSource: [$class: "ManuallyEnteredRepositorySource", url: "https://github.com/CARTAvis/carta-backend"],
-      contextSource: [$class: "ManuallyEnteredCommitContextSource", context: "ci/jenkins/build-status"],
-      errorHandlers: [[$class: "ChangingBuildStatusErrorHandler", result: "UNSTABLE"]],
-      statusResultSource: [ $class: "ConditionalStatusResultSource", results: [[$class: "AnyBuildResult", message: message, state: state]] ]
-  ]);
-}
 pipeline {
     agent none
     options {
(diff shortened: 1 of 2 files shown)


veggiesaurus:
Just quoting what I said about the companion PR in the frontend (https://github.com/CARTAvis/carta-frontend/pull/1919):

> I don't have a problem in general with this PR, but I don't think introducing another moving part (the jenkins-slack connection) is the right thing to do. We should rather fix the underlying issue or request help fixing it from a devops specialist.

veggiesaurus:
@ajm-asiaa can you file a separate issue for tracking the lack of updating on github, and then we can merge this in? That way we won't lose track of the fact that the github updates are still a problem

ajm-ska:
@veggiesaurus Besides the Slack notifications solving the problem with the full ICD test suite, I took @kswang1029's suggestion and also got curl status posts working in another branch: [ajm/jenkins-curl-test](https://github.com/CARTAvis/carta-backend/tree/ajm/jenkins-curl-test). The syntax is like this: 
```
    post {
        success {
            withCredentials([string(credentialsId: 'acdc-jenkins-token', variable: 'TOKEN')]) {
            sh("""curl -X POST -H "Accept: application/vnd.github+json" \
            -H "Authorization: token $TOKEN" https://api.github.com/repos/CARTAvis/carta-backend/statuses/${COMMIT_ID_LONG} \
            -d "{\\"state\\":\\"success\\",\\"target_url\\":\\"${env.RUN_DISPLAY_URL}\\",\\"description\\":\\"Success\\",\\"context\\":\\"testing curl\\"}" """)
            }
        }
```
This way it appears guaranteed that Github will receive the message. I was just doing some extra investigating because the log file says: 
```
Warning: A secret was passed to "sh" using Groovy String interpolation, which is insecure.
		 Affected argument(s) used the following variable(s): [TOKEN]
		 See https://jenkins.io/redirect/groovy-string-interpolation for details.
```
It seems `withCredentials([string(credentialsId: ... ` is the [right way](https://www.jenkins.io/doc/pipeline/steps/credentials-binding/) but I just need to use triple single quotes instead of triple double quotes! Although when I did that, the other Groovy variables were not interpreted correctly. It is almost there.

Anyway, when I figure that out, I was planning to put the curl post method into this branch because I made further improvements to the Jenkinsfile-full as noted in the first message in this PR. I'll try to finish it shortly.

Regarding the Slack notification plugin, I still like the idea. It doesn't take anything away from what we were doing before. For the carta-backend, for example, the Jenkinsfile would still build the backend and run the unit tests on the 3 platforms, then post the build status/checks on Github. Everything would continue to be the same as before. The Slack notification plugin is only a bonus feature for the benefit of myself and/or the test team for us to easily keep an eye on the status of builds. Yes, I know email notifications are an alternative, but I prefer the idea of concise unobtrusive colour-coded messages in the Slack channel that I can quickly glance at the time of my choosing. I never intended to force all developers to use Slack notifications. I only added you to the channel to help your review of the PR.

veggiesaurus:
> Regarding the Slack notification plugin, I still like the idea. It doesn't take anything away from what we were doing before. For the carta-backend, for example, the Jenkinsfile would still build the backend and run the unit tests on the 3 platforms, then post the build status/checks on Github. Everything would continue to be the same as before. The Slack notification plugin is only a bonus feature for the benefit of myself and/or the test team for us to easily keep an eye on the status of builds. Yes, I know email notifications are an alternative, but I prefer the idea of concise unobtrusive colour-coded messages in the Slack channel that I can quickly glance at the time of my choosing. I never intended to force all developers to use Slack notifications. I only added you to the channel to help your review of the PR.

Ok, that's fine :+1:

veggiesaurus (approved):

confluence (approved):

ajm-ska:
There are still two more things to do before merging:
1. I need to adjust some settings in the Jenkins GUI for the Jenkinsfile-full (Full ICD test suite) pipeline job to suppress the default Github Status checks from being sent. Otherwise, we will probably continue to get the "In progress" notification here on Github. I'm fairly sure I had the right settings before. Instead of Github Status checks, we will just rely on the status sent via curl.
2. I still have a mistake in the curl command! I will need another commit.

PR #1158, opened 2022-07-25
Angus/1902 parsing error

### LINKED ISSUE #1157: Editing of preferences causes parsing error and prevents application of some preferences
I have recently noticed that whenever I edit any preference through the frontend dialog, I get this error in the Firefox console:

```
XML Parsing Error: not well-formed
Location: http://192.168.0.98:3003/api/database/preferences
Line Number 1, Column 1:
```

Additionally, while some preferences are applied as normal (e.g. turning event logging on and off), some (e.g. overlay settings) have no effect. They _are_ saved correctly, and are applied the next time I open CARTA.

I have reproduced this on Ubuntu Focal using both the `carta` and `carta-beta` packages, after clearing my user directory so that a brand new one is created (so the issue is not an old preferences file). I have also reproduced this in the latest dev backend + frontend built from source.

In Chromium I do not see any error output in the browser console, but I see the same symptoms when I try to change the overlay settings.

veggiesaurus:
Closes #1157. All that was missing was the `Content-type: application/json` header on a number of responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed the getstat error on generated image ([#1148](https://github.com/CARTAvis/carta-backend/issues/1148)).
 * Fixed file info hang when a CASA image is locked ([#578](https://github.com/CARTAvis/carta-backend/pull/578)).
 * Fixed region export failure when no write permission ([#1133](https://github.com/CARTAvis/carta-backend/pull/1133)).
+* Fixed HTTP response codes when returning response to PUT requests ([#1157](https://github.com/CARTAvis/carta-backend/issues/1157)).
 
 ## [3.0.0-beta.3]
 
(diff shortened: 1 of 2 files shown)


confluence:
Is there a frontend component? When I test this in Firefox, I no longer get the parsing errors, but I still get the incorrect behaviour (the preferences are not applied immediately; only when I close the session and open a new session). Specific examples: changing the global WCS overlay colour; showing / hiding the colorbar. In general I think all the WCS overlay settings are affected.

veggiesaurus:
> Is there a frontend component? When I test this in Firefox, I no longer get the parsing errors, but I still get the incorrect behaviour (the preferences are not applied immediately; only when I close the session and open a new session). Specific examples: changing the global WCS overlay colour; showing / hiding the colorbar. In general I think all the WCS overlay settings are affected.

there's no frontend component. IS it not that you're changing the _default_ properties, which are not applied to the current view? This is a matter that @kswang1029 can weigh in on :) (not a bug in itself)

confluence:
Hmm. This may be user error, but I think I'm missing something. I'm changing the values in the `WCS and Image Overlay` tab. The values in the `Vector Overlay Configuration` are clearly labelled as defaults, and there's a different place to configure the vector overlay style for the current image (from the view menu), so that makes sense. But if the WCS / image overlay settings are supposed to be analogous, 1) I think the labels should reflect this more clearly, and 2) where are the controls to change the overlay colours for the current image?

confluence:
Never mind; I just found them! That explains the issue.

confluence (approved):
I have confirmed that this fixes the parsing error. My other observation about some preferences not being applied was based on a misunderstanding; everything seems to be working correctly.

PR #1162, opened 2022-07-27
Add directory in region list if listing all files or not image

### LINKED ISSUE #1159: [critical for v3-stable release] cannot navigate through file system when trying to load a region file
**Describe the bug**
This is considered as a regression which I did not pick up from https://github.com/CARTAvis/carta-backend/pull/1142. Apologies...

With #1142, only region files are listed in the directory. However, if we need to navigate to other parent folders with the breadcrumbs, there is no way to access folders therein as they are filtered out.

**To Reproduce**
1. try to load a region file in a parent directory with the file browser
2. no directory is visible from the file list

**Expected behavior**
We still need to show directories so users can navigate through the file system to load a region file.

 - Browser version [e.g. 22]: 103

pford:
Closes #1159 

List directories in region file list if

- listing all files (could be CASA/MIRIAD image but listed as directory)
- checking contents/extension and image type of directory is unknown.


confluence (approved):

kswang1029 (approved):
all current e2e tests are passing. Manual tests are fine. 👍

PR #1163, opened 2022-07-27
Fix the crash when trying to load an unsupported image file

### LINKED ISSUE #1161: [critical for v3-stable release] backend crash when trying to load an unsupported image file
**Describe the bug**
Usually this should not happen via the GUI as the default file list parser only include supported image formats in the resulting file list. However, in the preferences now we have an option to show _all_ files so the file list can have all other unsupported files such as pdf, png, etc. Then if we single click on one, the file info shows it is unsupported as expected. However, if we double-click on one, the backend crashes...

**To Reproduce**
1. set the "file list" dropdown menu in the "Global" tab of the preferences dialog to "all files"
2. navigate to a folder with unsupported image formats, such as png or pdf
3. single-click on one and backend should return an error message and the frontend shows it in the file info
4. double-click on one and the backend should crash

**Expected behavior**
Backend should not crash in this case and it should also throw an error to the frontend.

 - Browser version [e.g. 22]: 103

markccchiang:
Closes #1161. The backend will cache a file loader as a null pointer if the file format is unsupported. I added a check whether a file loader is a null pointer before checking its updated time.


confluence (approved):

kswang1029 (approved):
all current e2e tests are passing.

Tried to load various kinds of unsupported files in CARTA and CARTA behave correctly without crashing. 👍

PR #1165, opened 2022-07-28
PR template

### LINKED ISSUE #1116: Create a PR template
We need a PR template similar to the one added to the frontend, so that we don't keep forgetting to update the changelog. The checklist should also include linking issues and companion PRs in the other components.

confluence:
Closes #1116. This is a simple PR template based on the frontend PR template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,12 @@
+**Description**
+
+* What is implemented or fixed? Mention the linked issue(s), if any.
+* How does this PR solve the issue? Give a brief summary.
+* Are there any companion PRs (frontend, protobuf)?
+* Is there anything else that testers should know (e.g. exactly how to reproduce the issue)?
+
+**Checklist**
+
+- [ ] changelog updated / no changelog update needed
+- [ ] e2e test passing / added corresponding fix
+- [ ] protobuf updated to the latest dev commit / no protobuf update needed
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,3 +10,5 @@
 - [ ] changelog updated / no changelog update needed
 - [ ] e2e test passing / added corresponding fix
 - [ ] protobuf updated to the latest dev commit / no protobuf update needed
+- [ ] added reviewers and assignee
+- [ ] added ZenHub estimate, milestone, and release


kswang1029 (commented):
Just wonder if we need a check field for ICD test as well (one platform should be sufficient)? @ajm-asiaa ideas?

ajm-ska:
Here are my thoughts:
 
- The current full ICD test suite takes about 45 minutes to run. That includes building a carta-backend first. We run the 7 platforms in parallel, so in principle running on one platform compared to all 7 platforms should not make much difference.
- We found it wasn't feasible to run the full ICD test suite on all 7 platforms for every single commit (that is why we decided to run it once a day on the 'dev' branch), but we might be able to make the ICD tests run on a PR only.
- If we run on a PR, we could reduce the computational load by running only on 1 platform, perhaps on the M1 MacMini as that would be the fastest platform (I'm not sure of the total run-time on the M1 MacMini at the moment but it could be ~35 minutes).
- The majority of ICD test failures are due to carta-backend code changes and/or protobuffer updates that break the existing ICD tests. Therefore, the ICD tests would need to be updated accordingly. So any PR would require coordination with the test team (@acdo2002) first.
- A further complication would be the need to have separate branches on the ICD-test repository containing updated tests for each PR as well as for the existing 'dev' branch. It would also require coordination on merging. It feels like things could get complicated with merge conflicts!
- With the previous 2 points in mind, I still prefer the current setup where we only run the ICD tests once a day (0800 Taiwan time) on the 'dev' branch. We would catch any problems introduced to the 'dev' branch the day before and have time to see if it is only a matter of updating the ICD tests, or if a legitimate issue/bug has been introduced.

kswang1029:
> Here are my thoughts:
> 
> * The current full ICD test suite takes about 45 minutes to run. That includes building a carta-backend first. We run the 7 platforms in parallel, so in principle running on one platform compared to all 7 platforms should not make much difference.
> * We found it wasn't feasible to run the full ICD test suite on all 7 platforms for every single commit (that is why we decided to run it once a day on the 'dev' branch), but we might be able to make the ICD tests run on a PR only.
> * If we run on a PR, we could reduce the computational load by running only on 1 platform, perhaps on the M1 MacMini as that would be the fastest platform (I'm not sure of the total run-time on the M1 MacMini at the moment but it could be ~35 minutes).
> * The majority of ICD test failures are due to carta-backend code changes and/or protobuffer updates that break the existing ICD tests. Therefore, the ICD tests would need to be updated accordingly. So any PR would require coordination with the test team (@acdo2002) first.
> * A further complication would be the need to have separate branches on the ICD-test repository containing updated tests for each PR as well as for the existing 'dev' branch. It would also require coordination on merging. It feels like things could get complicated with merge conflicts!
> * With the previous 2 points in mind, I still prefer the current setup where we only run the ICD tests once a day (0800 Taiwan time) on the 'dev' branch. We would catch any problems introduced to the 'dev' branch the day before and have time to see if it is only a matter of updating the ICD tests, or if a legitimate issue/bug has been introduced.

Ok. I recalled the complexity now. I am open to my (bad?) idea. Once v3 is out, I will keep improving e2e tests.

pford:
Should we add Milestone and Estimate?

pford:
> Should we add Milestone and Estimate?

I meant to the checklist!

confluence:
> I meant to the checklist!

Oh! Yes, that seems wise. :)

confluence:
Done. I also added an item for the reviewers and assignee.

pford (approved):

kswang1029 (approved):

veggiesaurus (approved):

PR #1181, opened 2022-08-04
Refactor the timer for performance measurements

### LINKED ISSUE #1180: Refactor the timer for performance measurements
The timer used to measure performance in the code caused me a lot of reading noise. It would be better to refactor this part of the code in a shorter syntax.

markccchiang:
Closes #1180 

**Description**

* Refactor the code for the performance timer.
* Create a new class `PerfTimer` and reuse it in the performance measurements.
* No companion PRs (frontend, protobuf), backend issue only.
* @acdo2002 This performance timer should work as before. Please check if the time report in the performance log is correct. Thank you!

**Checklist**

- [x] changelog updated / ~~no changelog update needed~~
- [x] protobuf updated to the latest dev commit / no protobuf update needed


confluence (changes requested):
The `Timer` class as it is currently implemented was designed to be instantiated once and used to store and calculate multiple measurements. Creating multiple instances of this class (and adding a wrapper which hides a hardcoded measurement name) is not the best way to integrate this code.

I have discussed this with @veggiesaurus, who originally wrote `Timer`, and we agree that it's unnecessarily complex. It was added for debugging purposes, currently isn't used anywhere except the unit test that tests it, and perhaps shouldn't be used as designed, for thread safety reasons. The way that you are currently creating multiple timer instances is correct and we should keep that idea -- but the implementation can be simplified a lot:

1. Remove the old `Timer` class (and its test) completely
2. Put the functionality that you are replacing in this PR directly into your `PerfTimer` class, and rename it to `Timer`.

So the new simple `Timer` class should just store a single starting time, and have an `Elapsed` method which creates an end timestamp and returns a time delta.

I would also suggest creating a `TimeDelta` struct with `ms` and `us` members, and returning that from `Elapsed`, so that you can access the measurement in both milliseconds and microseconds without repeatedly using a hardcoded multiplication factor. Apart from that (and the removal of the old `Timer`) I think the refactoring is fine.

src/DataStream/Contouring.cc
```diff
@@ -234,9 +235,8 @@ void TraceContours(float* image, int64_t width, int64_t height, double scale, do
     }
 
     if (spdlog::get(PERF_TAG)) {
-        auto t_end_contours = std::chrono::high_resolution_clock::now();
-        auto dt_contours = std::chrono::duration_cast<std::chrono::microseconds>(t_end_contours - t_start_contours).count();
-        auto rate_contours = width * height / (double)dt_contours;
+        auto dt_contours = t.Elapsed(Timer::ms);
+        auto rate_contours = width * height / t.Elapsed(Timer::us);
```
confluence: Surely every time you call `Elapsed` you recalculate the value using a different end time? This is why I was suggesting returning a struct from `Elapsed`:

```
auto dt = t.Elapsed();
auto dt_contours = dt.ms;
auto dt_rate_contours = width * height / dt.us;
```

The `TimeDelta` struct constructor can set the `ms` and `us` properties from a `us` parameter.

(Since the `ms` value is derived from the original `us` value, and we use it in every instance that we use a timer here, I don't think it's necessary to do anything more complicated to avoid calculating it every time we construct a `TimeDelta`.)

confluence (changes requested):

confluence (approved):

acdo2002 (approved):
Now it looks the time zone changes to the UTC.
I tested several ICD tests locally (include current and rxjs version), they are all fine.

kswang1029 (approved):
I did a test with dev and this branch by loading the same image to see the performance logs. The two look similar.

PR #1183, opened 2022-08-22
Yuhsuan/150 fixed fitting

YuHsuan-Hwang:
**Description**
This PR is for the second item of #150: fixing initial values

* `SetInitialValues()` checks the booleans sent from frontend and stores the fitting value vector index of the Gaussian component parameters in`fit_values_indexes`. `-1` is stored if the parameter is fixed, indicating that the initial value should be used instead.
* `GetGaussianParams()` checks `fit_values_indexes` and returns the values that should be used for a Gaussian component.

Tested that fitting with no parameters fixed remains the same result.

corresponding frontend PR: https://github.com/CARTAvis/carta-frontend/pull/1984

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * Added a check of averaging width when calculating line/polyline spatial profiles or PV images ([#1174](https://github.com/CARTAvis/carta-backend/issues/1174)).
+* Added support for image fitting with fixed parameters ([#150](https://github.com/CARTAvis/carta-backend/issues/150)).
 
 ### Fixed
 * Fixed issues with AIPS velocity axis by restoring previous casacore headers ([#1771](https://github.com/CARTAvis/carta-frontend/issues/1771)).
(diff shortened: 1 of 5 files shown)


confluence (approved):

veggiesaurus (approved):

markccchiang (approved):

kswang1029:
a quick comment: for the center position, if it is fixed, we dont need to show the '(arcsec)' string
Center X        = 6:12:54.3777367749 (arcsec) (fixed)
Center Y        = 17:59:14.1574465497 (arcsec) (fixed)

kswang1029:
@YuHsuan-Hwang a quick question: if we have two gaussian components, with the current implementation, we cannot have one gaussian with all parameters fixed. Is this intended? Can we make it more flexible so as long as there is one parameter free regardless how many components are, we can tigger the image fit?

YuHsuan-Hwang:
> a quick comment: for the center position, if it is fixed, we dont need to show the '(arcsec)' string Center X = 6:12:54.3777367749 (arcsec) (fixed) Center Y = 17:59:14.1574465497 (arcsec) (fixed)

> @YuHsuan-Hwang a quick question: if we have two gaussian components, with the current implementation, we cannot have one gaussian with all parameters fixed. Is this intended? Can we make it more flexible so as long as there is one parameter free regardless how many components are, we can tigger the image fit?

Yes, I should have make it more flexible. These two are frontend issues; fixed in the frontend PR.

kswang1029 (approved):
this looks good. all e2e tests are passing. 👍

PR #1193, opened 2022-09-07
ajm/self hosted GitHub runner

### LINKED ISSUE #1192: Improvement: Use Github Actions for carta-backend CI
Our current carta-backend CI is handled by Jenkins. The Jenkins instance runs locally on a server at ASIAA and sends the jobs out to local nodes at ASIAA. We can not guarantee that the Jenkins service and nodes are online and functional 24 hours a day, 365 days a year.

An improvement would be to let Github Actions handle the CI instead. 

Although it would not be practical to use the free [Github-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) for building the carta-backend due to limited resources (e.g. only 2 cores per VM), we could instead use [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners). 

The self-hosted runners would be running locally at ASIAA so we could still not guarantee availability 24 hours a day 365 days a year, but perhaps we could have additional self-hosted runners available at another site i.e. at IDIA. Then perhaps a last resort fall-back could be the slow cloud based Github-hosted runners in the unlikely event that no self-hosted runners are available.

ajm-ska:
**Description**

The old `Jenkinsfile` has been removed and [Github Actions self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) will handle the CI. 

As a reminder, our basic backend CI consisted of: 
- Format check
- Ubuntu 20.04 backend build + unit tests on Ubuntu 20.04
- AlmaLinux 8 backend build + unit tests on AlmaLinux 8
- macOS12 M1 backend build + unit tests on macOS12

The format check originally ran on the Github Actions cloud server and continues to do so. Now, if the format check fails, the backend builds will not proceed.

**About the unit tests**

I originally wanted the backend builds and unit tests to run as separate "jobs" because if a backend build passed and a unit test failed, we would immediately know that only the unit test is the problem. As far as I can see, the only way to do this is to save the unit tests folder and restore it for the next job as "workflow artifacts". The saving process uploads it to Github and then downloads it for the next job. This process seems to add about 3 minutes to the entire job which is not ideal. 

Instead, I'm just running the backend builds and unit tests in the same job. Now if either the backend build _or_ unit tests fail, the job/platform on which it runs (Ubuntu-Focal, macOS-12, AlmaLinux-8) will be marked as a failure. The developer will need to click in to look at the details to determine which stage failed. Although maybe that is OK because they would need to do so anyway to see why something failed.

Another reason why I wanted unit tests to be separate jobs is that the detailed results would then only be one click away in the jobs list on the left-hand-side menu. However, we can instead conveniently use the `dorny/test-reporter` action to do the same thing.

The macOS unit tests on M1 often fail randomly for unknown reasons, so I'm trying the `nick-fields/retry` action. If the unit tests fail it should wait 60 seconds between tries and then try to re-run them up to 3 times only. Hopefully, that will help. By only attempting to re-run 3 times, it should not cover up any real issues that may be introduced to the unit tests in future.

**About code coverage**

I really wanted to highlight the code coverage so that developers will always keep its importance in mind. After some experimentation, I think the `irongut/CodeCoverageSummary` action is the best because it can be made to only show the "line rate" ("line rate" is higher than the alternative "branch rate"). Plus, we can use the `LouisBrunner/checks-action` action to post the results as a separate job on the left-hand-side menu.

The only downside is I think the code coverage report slows everything down. At present, the Code Coverage report is handled as part of the Ubuntu-Focal job. Running unit tests built with the code coverage flags seems to add about 3 minutes to the run time. 

An alternative is that we could generate a Code Coverage report for a Pull Request only. What do you think @veggiesaurus?

I would also like to experiment with the `marocchino/sticky-pull-request-comment` action that would post the coverage report as a comment on Github when a Pull Request is made.

**About the self-hosted runners**

macOS-12 is running locally on an M1 Mac Mini. It has everything installed to build the carta-backend and run the unit tests with ASAN flags. Its label is `macOS-12-M1`.

For the Linux self-hosted runners, I decided to use Docker. Therefore the Linux self-hosted runners only need to have Docker available to the user that is running the actions-runner service (Just in case IDIA want to add some self-hosted runners). I decided the labels we should use are "Linux", "X64", and "Docker". I made two new Docker containers; `carta/backend-builder-ubuntu-focal` and `carta/backend-builder-almalinux-8`. They are capable of building the carta-backend and running the unit tests with ASAN flags. I decided to use our "carta" Docker Hub account. Perhaps we could use "carta" account for developer-related Docker containers only, and the "cartavis" account for user-related Docker containers. 

So far I added two Linux self-hosted runners at ASIAA. They happen to be called "almalinux-8.5" and "ubuntu-20.04" as that is the OS they are running. But that is not important. Now that we are using Docker containers, the Ubuntu 20.04 server could do the AlmaLinux CI build and vice versa.

To make it as fast as the old Jenkins CI, I have the Linux backend builds set to use 32 cores. I was thinking about `-j` to use all available cores so that it could possibly work on the 2-core Github-hosted runners. But now I don't think it is possible: [The machine can be either a GitHub-hosted runner or a self-hosted runner](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on). So it seems we can only run on self-hosted runners.

**Slack**

Finally, after some experimentation, I'm using `baijunyao/action-slack-notify` to also post the build status to Slack. I know there has been a dislike to using Slack, but I personally prefer it compared to email notifications. It will not bother anybody. Only myself (or the build-team) need to be part of the Slack channel so that we can easily monitor for any problems. All other developers can see the build status through Github and emails.

**Checklist**

- [x] e2e test passing / added corresponding fix - e2e should not be affected.
- [x] added ZenHub estimate, milestone, and release - not affected


ajm-ska:
I'm sorry it required so many commits. There was quite a lot of trial and error because I couldn't know if it worked or not until I pushed the changes. Plus I'm not familiar with YAML and so made a lot of syntax mistakes.

kswang1029:
What does that “health” column mean? Below a threshold it is a ❌?

ajm-ska:
> What does that “health” column mean? Below a threshold it is a ❌?

Yes. We can only set one lower and upper threshold. I have the lower one set to 38% and the upper set to 75%. Anything under 38% will show the X. Anything between 38% to 75% will show a ➖. Anything above 75% will show a ✔. 

With the Code Coverage Action that I am using, we are not able to hide the different sections (ImageData, ImageFitter etc.) to only show the overall percentage. It must show that full table. But maybe that is a good thing and will encourage developers to improve the code coverage.

veggiesaurus (approved):

Issue #1178, opened 2022-08-04
Support loading swapped-axes image cubes (eg RA-Channel-Dec) (backend)


kswang1029:
user feedback from v3 beta test

This is to support rotated cubes such as RA-Channel-Dec.

The backend needs to be able to recognize such cubes and generate proper file info.

The companion frontend issue is https://github.com/CARTAvis/carta-frontend/issues/1953

markccchiang:
@kswang1029 Do we have the sample file of the rotating cube to do the test?

markccchiang:
@kswang1029 Is it OK to load a swapped-axes cube and render it as `ra-dec` image for now? The backend needs to tell the frontend real axis numbers of ra, dec, spectral, and stokes axes. So the frontend can render a `ra-dec` image with the correct coordinate system (in AST setting). I have created a backend branch [mark/1178_support_swapped_axes_images](https://github.com/CARTAvis/carta-backend/tree/mark/1178_support_swapped_axes_images) and a frontend branch [mark/support_swapped_axes_images](https://github.com/CARTAvis/carta-frontend/tree/mark/support_swapped_axes_images) for this task. While the region spectral profile, PV, and moments images need to fix too. Because the default ra and dec axis numbers are 0 and 1 in the backend. Considering to create another backend issue to fix it. To render a rotating cube as `ra-channel` or `dec-channel` images would be another issue too after the previous job is done.

kswang1029:
> @kswang1029 Is it OK to load a swapped-axes cube and render it as `ra-dec` image for now? The backend needs to tell the frontend real axis numbers of ra, dec, spectral, and stokes axes. So the frontend can render a `ra-dec` image with the correct coordinate system (in AST setting). I have created a backend branch [mark/1178_support_swapped_axes_images](https://github.com/CARTAvis/carta-backend/tree/mark/1178_support_swapped_axes_images) and a frontend branch [mark/support_swapped_axes_images](https://github.com/CARTAvis/carta-frontend/tree/mark/support_swapped_axes_images) for this task. While the region spectral profile, PV, and moments images need to fix too. Because the default ra and dec axis numbers are 0 and 1 in the backend. Considering to create another backend issue to fix it. To render a rotating cube as `ra-channel` or `dec-channel` images would be another issue too after the previous job is done.

The main goal is to let the backend load a cube with the axes in any order and let the frontend render the first two axes as x and y and being able to animate in z.

kswang1029:
done via #1194

[...]

PR #1194, opened 2022-09-08
Mark/1178 supports loading swapped-axes image cubes

markccchiang:
**Description**

Support loading swapped axes image cubes #1178.

Fix the render axes to [0, 1], i.e., the first two axes from the header. The original render axes setting is direction axes, if any.

~No need to change the protobuf. But it seems the AST plot settings have to be modified in the frontend, in order to plot RA-channel or DEC-channel coordinates.~
Companion PRs: protobuf [PR 77](https://github.com/CARTAvis/carta-protobuf/pull/77) and frontend [PR 2011](https://github.com/CARTAvis/carta-frontend/pull/2011).

This is the new feature. We may need sample image cubes with swapped axes for testing.

**Checklist**

- [x] protobuf updated to the latest dev commit ~/ no protobuf update needed~


kswang1029:
@markccchiang the file info for a PV image needs a fix. The offset axis is with the "NA" label. It should be "Spatial" I guess?

Screen Shot 2022-11-30 at 09 46 57


In fact maybe we can simply display CTYPE1/2/3/4, so

> Shape = [256, 256, 480, 1] (X, Y, Spectral, Stokes)

becomes

> Shape = [256, 256, 480, 1] (RA, DEC, FREQ, STOKES)

or

> Shape = [256, 256, 480, 1] (X: RA, Y: DEC, Z: FREQ, W: STOKES)


for example.

kswang1029:
@markccchiang another file info issue: for a rotated cube, there is a repeated (and incorrect) "frequency range" info.

Screen Shot 2022-11-30 at 09 58 35

src/ImageData/FileLoader.cc
```diff
@@ -290,9 +302,11 @@ std::vector<int> FileLoader::GetRenderAxes() {
     if (_image_shape.size() > 2) {
         // Normally, use direction axes
```
pford: As I understand the new feature, x=0 and y=1 always (unless PV), regardless of location of direction axes.  So I don't think we need this block or `get_dir_axes` parameter (not used anyway, always false).  As I recall this block was a workaround because the frontend would not support loading RA-Channel-Dec image as Channel vs RA, so it forced Dec vs RA.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -680,41 +701,92 @@ void FileExtInfoLoader::AddShapeEntries(CARTA::FileInfoExtended& extended_info,
...
+            axis_types[spectral_axis] = name;
+        }
+    }
+
+    if (stokes_axis > -1) {
+        axis_types[stokes_axis] = "STOKES";
+    }
+
+    // For PV image
+    if (spectral_axis == 0 && axis_types[1] == "NA") {
```
pford: Can have CASA image with [OFFSET, STOKES, FREQ]

pford (changes requested):

kswang1029:
@markccchiang found an exception for UV image. The CTYPE1 is UU and CTYPE2 is VV but it is displayed as NA.

Screen Shot 2022-12-14 at 13 05 51

kswang1029:
@markccchiang here is another issue about axis identification. The CTYPE3 here is VOPT but it is recognized as FREQ in the file info. 

Screen Shot 2022-12-14 at 13 11 56
Screen Shot 2022-12-14 at 13 12 06

I suggest we display the context in the parentheses as (CTYPE1, CTYPE2, CTYPE3, CTYPE4) as defined in the image header.
`Shape = [800, 800, 70, 1] (RA, DEC, FREQ, STOKES)`

src/FileList/FileExtInfoLoader.cc
```diff
@@ -680,41 +688,76 @@ void FileExtInfoLoader::AddShapeEntries(CARTA::FileInfoExtended& extended_info,
...
+    auto* axes_numbers_info = extended_info.mutable_axes_numbers();
+    // Change to 1-based axis indices
+    axes_numbers_info->set_dir_x(direction_axes[0] + 1);
+    axes_numbers_info->set_dir_y(direction_axes[1] + 1);
+    axes_numbers_info->set_spectral(spectral_axis + 1);
+    axes_numbers_info->set_stokes(stokes_axis + 1);
+    axes_numbers_info->set_depth(depth_axis + 1);
+
+    // Set axis types with respect to axis numbers 0~3
+    std::vector<std::string> axis_types(4, "NA");
```
pford: I wonder if we could pass in the CTYPE names instead of iterating through all of the header entries.  If not compressed FITS, `_loader->GetCoordinateSystem()->worldAxesNames()` and for compressed FITS you just iterated through the header entries in AddInitialComputedEntries so capture these values.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -680,41 +688,76 @@ void FileExtInfoLoader::AddShapeEntries(CARTA::FileInfoExtended& extended_info,
...
+                axis_types[1], axis_types[2], axis_types[3]);
             break;
     }
     auto shape_entry = extended_info.add_computed_entries();
     shape_entry->set_name("Shape");
     shape_entry->set_value(shape_string);
     shape_entry->set_entry_type(CARTA::EntryType::STRING);
 
-    if (chan_axis >= 0) {
+    if (depth_axis >= 0) {
```
pford: Shouldn't this be `spectral_axis` since it is "Number of channels"? If the image is RA-Channel-Dec (as in issue title) then depth is Dec axis.  To support images with axes in any order, the backend should distinguish between channels and depth (spectral axis vs. z axis), and so should the frontend.

pford (changes requested):

src/ImageData/FileLoader.cc
pford: You removed this GetRenderAxes function and set render_axes={0, 1}.  But a CASA PV image can have [OFFSET, STOKES, FREQ] in which case render axes are [0, 2].  Stokes is never a rendered axis.

src/FileList/FileExtInfoLoader.cc
pford: or `image->coordinates()->worldAxesNames()`

pford (approved):

pford:
Oops, this was my fault.  For some reason casacore uses Freq for spectral worldAxisNames instead of the actual header value.  Sorry @markccchiang  you will have to change it back to iterating through header entries.

kswang1029:
to handle the general cases like
(0:ra, 1:dec, 2:stokes, 3:channel) ==> 24 combinations
(0:ra, 1:dec, 2:channel, 3:stokes) ==> 24 combinations
(0:ra, 1:dec, 2:channel) ==> 6 combinations
(0:ra, 1:dec, 2:stokes) ==> 6 combinations

we need to incorporate a general rule in the current PR to ignore stokes (if it exists) first, then from the rest, we pick up the 1st as x-axis and the 2nd as y-axis. The rest (if it exists) is z-axis (iterable via the animator). stokes (if it exists) remains as the polarization axis.

kswang1029:
FYI. Working on fixing and add basic e2e tests for this new feature now. Once done I will approve this PR.

kswang1029:
@markccchiang one minor issue spotted in the file info.

Name = gaussian_array_large_GALACTIC_2031.image
Data type = floatShape = [4, 746, 10, 746] (STOKES, GLON, FREQ, GLAT)
Number of channels = 10
Number of polarizations = 4
Coordinate type = Longitude, Frequency
Projection = SIN
Image reference pixels = [374, 6]
Image reference coords = [`-132.27.25.3488`, 1.4204e+09 Hz]
Image ref coords (deg) = [`-132.457 deg`, 1.4204e+09 Hz]

longitude is [0 deg, 360 deg].

kswang1029 (approved):
feature-wise this is looking good for the current scope - supporting image visualization of a swapped-axes cube. 👍  Please proceed code review.

src/ImageData/FileLoader.cc
```diff
@@ -175,9 +175,11 @@ std::shared_ptr<casacore::CoordinateSystem> FileLoader::GetCoordinateSystem(cons
     return std::make_shared<casacore::CoordinateSystem>();
 }
 
-bool FileLoader::FindCoordinateAxes(casacore::IPosition& shape, int& spectral_axis, int& z_axis, int& stokes_axis, std::string& message) {
+bool FileLoader::FindCoordinateAxes(casacore::IPosition& shape, std::vector<int>& direction_axes, int& spectral_axis, int& stokes_axis,
```
confluence: I would use "spatial" instead of "direction" in all these variables and comments, for consistency with our existing terminology.

confluence (changes requested):
I think that this is OK; I would just like to rename these variables. We have previously referred to the X and Y axes in the code as "spatial" rather than "direction" axes, and I think it will be less confusing if we stick to that naming convention.

src/ImageData/FileLoader.cc
kswang1029: just a IIRC note, in casacore it is called directional axis. This may be why. 🙂

confluence: If only we could all decide on one standard! :smile:

kswang1029 (approved):
looking good. new e2e tests are implemented.

confluence (approved):

PR #1198, opened 2022-09-16
updating all Dockerfiles plus CMakeLists.txt for carta-gsl on el7

### LINKED ISSUE #1197: update the Dockerfiles
The carta-backend Dockerfiles should be updated.

We should have a Dockerfile for Ubuntu 22.04. 

AlmaLinux 9 is now available, so we need to split the old` Dockerfile-AlmaLinux` (which uses almalinux:latest) into two. 

We are starting to switch over to the new [CARTAVis Copr RPM repository](https://copr.fedorainfracloud.org/coprs/cartavis/carta/) so should adjust the RedHat based Dockerfiles accordingly.

ajm-ska:
**Description**

The Dockerfiles have been updated. All Dockerfiles are capable of building the current carta-backend with ASAN flags and the Unit Tests. I have checked all six, and the carta-backend runs plus all Unit Tests run and pass in each one.

I have added `Dockerfile-ubuntu-22.04`. Currently the line to add the kern-7 PPA (for up-to-date casacore-data) is commented out because they currently do not produce 'jammy' packages. I know the carta-backend will still work fine, but casacore-data may not be up to date. I put a note saying as much at the top of the Dockerfile. Once the Kernsuite team produce 'jammy' packages, we can update the Dockerfile again.

I have split the old `Dockerfile-AlmaLinux` into two files. When we created it first, we were using `almalinux:latest` because there was only AlmaLinux 8 (Redhat 8). But now there is AlmaLinux 9.

I have converted the three RedHat-based Dockerfiles (CentOS7, AlmaLinux8, and AlmaLinux9) to use our new [CARTAvis Copr RPM repo](https://copr.fedorainfracloud.org/coprs/cartavis/carta) rather than our old https://packages.cartavis.org/. I have put all required packages on our Copr repo so that each one can build the carta-backend with ASAN flags and run the Unit Tests.

CentOS7 is slightly problematic because it now requires a newer GSL (gsl-2.5) rather than the default version (gsl-1.15). At the time, there was concern about simply creating a gsl-2.5 RPM and installing it in the normal location because it could potentially conflict with other packages such as Inkscape that were built against the older default gsl-1.15 version. So it was decided we should call our newer RPM `carta-gsl` and install it to a custom location, namely `/opt/carta-gsl`. Furthermore, to ensure it could not interfere with default gsl, we do not install pkgconfig information with it. 

This has been perfectly fine when creating the carta-backend el7 RPMs because we can simply add additional build flags to the SPEC file. e.g. `-DCMAKE_CXX_FLAGS="-I/opt/carta-gsl/include" DCMAKE_CXX_STANDARD_LIBRARIES="-L/opt/carta-gsl/lib"`

However, explaining this special requirement to potential el7 users would be a bit complicated. So in order to keep it as simple as `cmake ..`, I propose to modify the carta-backend `CMakeLists.txt` file as follows:

```
FIND_PACKAGE(GSL QUIET)
if (GSL_FOUND)
    message(STATUS "Found gsl using find_package")
else ()
    if(EXISTS /opt/carta-gsl)
        message(STATUS "Found carta-gsl for el7")
        INCLUDE_DIRECTORIES(/opt/carta-gsl/include)
        LINK_DIRECTORIES(/opt/carta-gsl/lib)
    else ()
        message(FATAL_ERROR "Could not find gsl")
    endif ()
endif()
```

This conveniently deals with the special case of carta-gsl on el7 and has no effect on any other platform. I have tested it, and it seems to work fine.

I think that is the easiest way. An alternative would be additional work to re-do the carta-gsl RPM in order to get it to install a special `carta-gsl.pc` alongside the normal `gsl.pc` in the default pkconfig search folder. However, the carta-backend CMakeLists.txt would still need to be modified in order to ignore GSL only on el7 and only pick up the special CARTA-GSL. That seems a bit more complicated.

I should mention the only caveat of using Copr, is that there is no support for el7 aarch64.

We should also merge @veggiesaurus' `angus/minimal_dockerfile` branch.

**Checklist**

- [x] e2e test should not be affected

### DIFF
--- a/Dockerfiles/Dockerfile-ubuntu-22.04
+++ b/Dockerfiles/Dockerfile-ubuntu-22.04
@@ -1,0 +1,32 @@
+FROM ubuntu:22.04
+# Note: kern PPA has been temporarily disabled until kernsuite provide 'jammy' packages. 
+# Until then, the 'casacore-data' package may not be fully up-to-date.
+
+# Install the basic packages
+RUN \
+  apt-get update && \
+  apt-get -y upgrade && \
+  DEBIAN_FRONTEND=noninteractive \
+  apt-get install -y apt-utils autoconf bison build-essential cmake curl fftw3-dev flex gcc g++ \
+    gdb gfortran git git-lfs libgmock-dev googletest libblas-dev libcfitsio-dev \
+    libgsl-dev libgtest-dev libhdf5-dev liblapack-dev libncurses-dev \
+    libprotobuf-dev libpugixml-dev libreadline-dev libssl-dev libstarlink-ast-dev \
+    libtool libxml2-dev libxslt1-dev libzstd-dev pkg-config protobuf-compiler \
+    python3-numpy python3-astropy software-properties-common unzip wcslib-dev wget uuid-dev
+
+# Get carta dependencies carta-casacore, fits2idia, and zfp from cartavis-team PPA and casacore-data from KERN
+RUN \
+  #apt-add-repository -y -s ppa:kernsuite/kern-7 && \
+  add-apt-repository -y ppa:cartavis-team/carta-testing && \
+  apt-get update && \
+  apt-get -y install carta-casacore-dev fits2idia libzfp-dev
+
+# Forward port so that the webapp can properly access it from outside of the container
+EXPOSE 3002
+
+ENV HOME /root
+WORKDIR /root
+
+# Overwrite this with 'CMD []' in a dependent Dockerfile
+CMD ["bash"]
+
--- a/Dockerfiles/Dockerfile-ubuntu-22.04
+++ b/Dockerfiles/Dockerfile-ubuntu-22.04
@@ -16,7 +16,6 @@ RUN \
 
 # Get carta dependencies carta-casacore, fits2idia, and zfp from cartavis-team PPA and casacore-data from KERN
 RUN \
-  #apt-add-repository -y -s ppa:kernsuite/kern-7 && \
   add-apt-repository -y ppa:cartavis-team/carta-testing && \
   apt-get update && \
   apt-get -y install carta-casacore-dev fits2idia libzfp-dev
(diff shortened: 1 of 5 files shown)


Dockerfiles/Dockerfile-ubuntu-22.04
```diff
@@ -0,0 +1,32 @@
...
+  apt-get install -y apt-utils autoconf bison build-essential cmake curl fftw3-dev flex gcc g++ \
+    gdb gfortran git git-lfs libgmock-dev googletest libblas-dev libcfitsio-dev \
+    libgsl-dev libgtest-dev libhdf5-dev liblapack-dev libncurses-dev \
+    libprotobuf-dev libpugixml-dev libreadline-dev libssl-dev libstarlink-ast-dev \
+    libtool libxml2-dev libxslt1-dev libzstd-dev pkg-config protobuf-compiler \
+    python3-numpy python3-astropy software-properties-common unzip wcslib-dev wget uuid-dev
+
+# Get carta dependencies carta-casacore, fits2idia, and zfp from cartavis-team PPA and casacore-data from KERN
+RUN \
+  #apt-add-repository -y -s ppa:kernsuite/kern-7 && \
```
confluence: I would completely remove these references to KERN from here -- the note at the top is sufficient. The KERN PPA targetting Jammy will have a different release version, so there's no reason to keep KERN-7 commented out here.

confluence (approved):
I would prefer to remove the commented out KERN lines apart from the note at the top (since this version of KERN will never be used for Jammy).

Other than that, I think this is fine -- it's more correct and consistent to install our custom gsl to `/opt` with the usual name than to install it to the default path with a different name. We already use `/opt` for our casacore.

Dockerfiles/Dockerfile-ubuntu-22.04
```diff
@@ -16,7 +16,6 @@ RUN \
 
 # Get carta dependencies carta-casacore, fits2idia, and zfp from cartavis-team PPA and casacore-data from KERN
```
confluence: The end of this comment should also be removed.

confluence (changes requested):

Dockerfiles/Dockerfile-ubuntu-22.04
ajm-ska: Well spotted

confluence (approved):

veggiesaurus (approved):

Issue #1109, opened 2022-05-18
Improve unit testing strategy


confluence:
There are two ongoing issues which are making it difficult to add and maintain effective unit tests for all parts of the codebase:

1. It is difficult to isolate and test many functions because of access specifiers or because of the complex state that they depend on. As a result, we are testing a subset of functions, and sometimes writing overly complex novel code in unit tests to set up all the required state (sometimes replicating code in the source which we should be testing). This means that some of our unit tests are too high-level, and that sometimes crashes and errors are introduced in unit test code which are never triggered in the source.
2. We don't have a consistent strategy for creating reference data to test correctness of calculations. In some cases we are testing algorithms in the source using similar or the same algorithms in the unit tests.

I think that we should:

1. Investigate more complex features of the Google Test library which we are already using (and possibly consider switching to / adding another library if necessary) and figure out how to make sure that every function in the source can have a unit test which effectively tests its behaviour in isolation. This will most probably require changes to the source to allow unit tests to access source class internals -- this must not have an adverse impact on performance, and possibly should only be enabled at compile-time if tests are enabled.
2. Identify one or more independent libraries which can be used to generate data to test correctness. I would prefer not to add more dependencies to the tests on libraries which are not used in the source -- it may make more sense to decouple these, and maintain some test data (ideally very small synthetic files) in a separate repository (which can be added as a submodule).

This should be an epic, with separate sub-issues for these two tasks.

PR #1200, opened 2022-09-26
Tests for the Frame class using mock objects

### LINKED ISSUE #1189: Add TileCache tests with mock objects
~This is part of #1109. Add tests for the `Frame` class which use mock objects for all the complex child objects that Frame depends on (the file loader, tile cache, etc.). This should also be a proof of concept of the kinds of modifications which need to be made to the source to allow mock objects to be substituted in the tests.~

Following the discussion at the F2F, instead of starting with the large and complex `Frame` class we should instead tackle a much smaller and more achievable task as a proof of concept -- the tile cache classes, which were already partially refactored in the frame test PR. After this is done, we can gradually apply the same process to the rest of the code.

confluence:
**Description**

This implements #1189, which is part of the unit testing epic #1109.

This is a proof of concept of using [mock objects](https://google.github.io/googletest/gmock_for_dummies.html) to enable effective testing of classes in the backend codebase which have complex subcomponents. Each function in the class should have one or more unit tests which test **only** the correctness of the function, mocking out any functionality of the subcomponents.

Mocks implemented in this draft PR so far: `FileLoader` and `TileCache`. Only a few `Frame` unit tests have been added to this draft so far; more mock implementations may be added later as required.

Any class in the codebase which needs to be mocked needs to be refactored into an interface and an implementation -- the mock class can then inherit from the interface. Mocking non-virtual methods is not recommended, and we should avoid it.

To allow mock subcomponents to be injected into the classes being tested when they are initialised inside the class (not passed in as parameters), I propose creating factory functions with different implementations in the real code and in the test code (only one of which will be linked at a time). So far I have done this with `TileCache` (see `src/Util/Factories.cc` and `test/Factories.cc`) -- this is a very simple implementation which makes certain assumptions; more complicated implementations could be added for other components.

It is possible to create a subclass which selectively mocks specific functions -- this allows a test of a function which calls another function in the same class to isolate and test only the behaviour of the calling function without having to account for the real behaviour of the called function. In this unit test, custom half-mock subclasses are created inside tests and mock only the functions which will be used in that test. A more complete Frame mock should be created later for unit tests which need to use Frame as a subcomponent (e.g. tests of the Session class).

In addition to changes required for implementing the mock objects, this PR includes several minor fixes for issues discovered during the writing of the unit tests (added / changed operators and other struct features, slight refactoring to remove redundancy or inconsistency, etc.).

There are no companion PRs.

**Checklist**

- [ ] e2e test passing / added corresponding fix


confluence:
As discussed in the F2F, I will also add annotations to the functions at the same time as the unit tests (once #1215 is complete) and start a discussion about refactoring some parts of `Frame` into other components (to be done in a separate PR).

confluence:
Following the discussion at the F2F, I'm going to start with a smaller, more self-contained set of classes (probably the tile cache) to produce a simpler initial proof of concept. The frame class is so large and complex that it probably needs to be refactored substantially before / while tests are added.

Once a complete example exists, we can create individual issues for adding unit tests and documentation to different parts of the code.

I will close this PR and the associated issue once I have created a new issue for the new approach.

confluence:
I edited #1189 instead of creating a new issue, and am now closing this PR (I'm not deleting the branch yet because some of the work in it can be reused).

PR #1209, opened 2022-10-18
Use existing region ids when resuming session

pford:
Fix setting new region ids on resume.

Added check that region id is NEW_REGION_ID (-1, sent from frontend) before assigning new id.

No

Create then delete regions, then create again (so that region ids do not start at 1).  Restart backend, resume session, and check that profiles/stats work where expected for region id and type.  Export regions and check that all were exported to file.

**Checklist**

- [x] changelog updated


kswang1029 (approved):
yes, confirmed that after the fix, region analytics work fine in a resumed session. Before the fix, region analytics do not work in a resumed session.

No regression is observed from e2e tests.
Will try to enhance the e2e tests accordingly.

confluence (approved):

Issue #1210, opened 2022-10-19
Backend creates new region IDs when resuming session


pford:
A conditional added so that a region id is not created for a temporary region causes assigned region ids to be re-created.

PR #1211, opened 2022-10-21
Fix matched region export crashes

pford:
Fix error/crash when exporting region to matched image.

Fixes critical bugs #1205 and #1208.  Also contains resume session fix in PR #1209 for issue #1210 , needed while testing fix.

Usually regions are converted to a matched image with WCRegion::toLCRegion, but this fails when the images have different coordinate systems. Instead, the wcs control points are converted to the matched image, which exposed bugs in this code path.  The fixes include: (1) set wcs control points for conversion if not already set, (2) fix array data type conversions with assignment of values not arrays, and (3) determine data type (float/double) for pixel values in region Record in order to read the values successfully.

No

Export regions from a matched image with a direction coordinate system which differs from the reference image, for example the test images in #1205 .

**Checklist**

- [ ] changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Fixed issues with AIPS velocity axis by restoring previous casacore headers ([#1771](https://github.com/CARTAvis/carta-frontend/issues/1771)).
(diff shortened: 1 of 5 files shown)


kswang1029 (approved):
Verified with a different set of test images with different coordinate systems and the fix works without crash. No regression is observed from the current e2e tests. Will try to enhance the e2e tests accordingly.

confluence:
I've merged #1209; this needs a merge to clean up the conflicts.

confluence (approved):

PR #1214, opened 2022-11-14
Angus/1213 no runtime config

### LINKED ISSUE #1213: Add the no-runtime-config flag back into dev and v3 branches
**Describe the bug**
NRAO archive needs to be able to serve CARTA from a backend build, but without the API/database stuff. The runtime config stuff is currently tied to whether the backend is opened with `--no_frontend`. 

I made a branch for v2.0 that solved this issue (`archived/angus/no_runtime_config`), but we would need to update this a bit for v3.0.

veggiesaurus:
**Description**

Closes #1213 by adding the `no_runtime_config` commandline argument. Workaround for the NRAO archive.

**Checklist**

- [x] changelog updated / ~~no changelog update needed~~

### DIFF
--- a/src/HttpServer/HttpServer.cc
+++ b/src/HttpServer/HttpServer.cc
@@ -24,7 +24,8 @@ const std::string success_string = json({{"success", true}}).dump();
 uint32_t HttpServer::_scripting_request_id = 0;
 
 HttpServer::HttpServer(std::shared_ptr<SessionManager> session_manager, fs::path root_folder, fs::path user_directory,
-    std::string auth_token, bool read_only_mode, bool enable_frontend, bool enable_database, bool enable_scripting)
+    std::string auth_token, bool read_only_mode, bool enable_frontend, bool enable_database, bool enable_scripting,
+    bool enable_runtime_config)
     : _session_manager(session_manager),
       _http_root_folder(root_folder),
       _auth_token(auth_token),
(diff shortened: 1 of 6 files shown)


src/HttpServer/HttpServer.cc
```diff
@@ -80,10 +82,14 @@ void HttpServer::RegisterRoutes() {
...
-    res->writeHeader("Content-Type", "application/json");
-    res->end(runtime_config.dump());
+    if (_enable_runtime_config) {
+        json runtime_config = {{"apiAddress", "/api"}};
+        res->writeStatus(HTTP_200);
+        res->writeHeader("Content-Type", "application/json");
+        res->end(runtime_config.dump());
+    } else {
+        res->writeStatus(HTTP_200)->end();
+    }
```
confluence: I think this should be enabled or disabled in `RegisterRoutes`, not here, and return 501 (not implemented) for consistency with other disabled features (would that require a change in the calling code?).

If there's a good reason for it to return 200 anyway, I think we should still handle this logic in `RegisterRoutes` and make a different default handler (called something like `QuietSuccess`).

confluence (changes requested):

src/HttpServer/HttpServer.cc
veggiesaurus: There's a good reason to return 200, because it just means there's no backend-mandated config. I can add it to a separate handler

confluence (approved):

confluence:
This has been tested by @kswang1029, and the flag does cause saved preferences to be ignored.

Issue #1216, opened 2022-11-17
[performance] loading casa region text file with 20000 regions is slow and takes lots of RAM


kswang1029:
based on the test with psrecord, it appears that loading casa region text file with 20000 regions is slow and takes lots of RAM. Loading ds9 region text file is more efficient as a comparison. However, ds9 seems performing even better e2e-wise. 

![Screen Shot 2022-11-17 at 12 59 24](https://user-images.githubusercontent.com/20819712/202360384-0b47cce9-739a-45c1-a78d-fcdaaec2b733.png)

UPDATE: test files https://drive.google.com/file/d/1CFVrkYcthNZlXucqEAVQbbYfYowNI4X6/view?usp=share_link

pford:
@kswang1029 please attach the region file or a link to it

kswang1029:
> @kswang1029 please attach the region file or a link to it

due to file size, sending you via slack

pford:
@YuHsuan-Hwang @veggiesaurus would it be better to send regions in batches with a progress field rather than in one ImportRegionAck message?  Something like 1000 at a time?  Or do you think the frontend time is acceptable once it gets a response?

veggiesaurus:
> @YuHsuan-Hwang @veggiesaurus would it be better to send regions in batches with a progress field rather than in one ImportRegionAck message? Something like 1000 at a time? Or do you think the frontend time is acceptable once it gets a response?

I think that would be useful. However, in order to do that, we would need a new message type in the ICD. At the moment, we have single-request, single-response, and data streams. We'd need something like a single-request, many-responses approach.

I have an idea of how to do this, but we will need to formalise it a bit.

Issue #340, opened 2019-09-04
Support image annotation text file (backend)


pford:
Supported in v1.4:
- [x] name
- [x] color
- [x] line width
- [x] dashed line

Requirements for v4:
https://docs.google.com/document/d/1YCB_lIVH8ftJcNS0gliE2uHmuiowGkV6DQYJlfteJUM/edit#

Must support:

- [x] SetRegion
- [x] ExportRegion
- [x] ImportRegion
- [x] ResumeSession
- [x] Import/Export from matched image

kswang1029:
@pford @veggiesaurus with this doc https://docs.google.com/document/d/1_eo3kmzHrMGDLuYmbMi6RQHytf0NiGyR_igRlBFjssc/edit I tried to find common properties of the casa region format and the ds9 region format. Please have a look. There are properties that are supported in one but not the other. Perhaps what we can do in CARTA is that we only support common properties as listed in the document. Please feel free to share your thoughts. : )

veggiesaurus:
Since we currently only support region colours, line thicknesses and dash length on the frontend, I'd suggest we start with just those three properties. The colour could be sent as it is defined in the file, as a `string`, and then interpreted in the frontend (either hex, an RGB string or a standard color). The thickness should be in pixels. The only hitch is that the frontend UI currently supports even dash lengths (e.g. 5 on, 5 off), rather than an arbitrary array of dashes. We could still send those to the frontend, though, and once we have support for arbitrary dash lengths, they'd work.

veggiesaurus:
@pford can we implement the three important parameters (colour, line width and dash) for 1.4?

veggiesaurus:
@pford for 1.4 we just need the above three (colour, line width and dash). For the rest, we can wait until 1.5 or later

pford:
@veggiesaurus SET_REGION has the region parameters as separate fields, and I could just add the style parameters.  But IMPORT_REGION_ACK and RESUME_SESSION.ImageProperties contain RegionProperties submessage arrays consisting of region_id and RegionInfo, so I would need to add the style parameters to RegionInfo.  What do you think about changing SET_REGION to use RegionInfo as a single field?  I think we can keep the SET_REGION.region_id field; RegionProperties was added to send a list of regions.

veggiesaurus:
@pford are you saying we have something like this?

```
message SetRegion {
    sfixed32 region_id = 1;
    RegionInfo region_info = 2;
}
```
[...]

PR #1220, opened 2022-11-22
Remove casa CRTF region file parser for imports

### LINKED ISSUE #1219: Remove imageanalysis parser for CRTF imports
Due to performance and incompatibility issues, remove use of casa imageanalysis CRTF **import** implementation and use CARTA implementation only.  This will improve the CRTF performance issue noted in #1216 .

With the addition of annotation regions in CARTA in v4.0, some regions which are not supported in CRTF (compass and ruler) are exported with a preceding `#` so that the CRTF file remains supported by other packages such as casaviewer, astropy, etc.  The `#` usually indicates a comment line and is skipped by the casa CRTF parser.  Since these lines must be checked in CARTA for an annotation region definition, we can no longer use imageanalysis to import CRTF files since there will be no indication that these regions were skipped.

The imagenanalysis implementation has slower performance because, as part of the import to create casa annotation regions, (1) pixel coordinates are converted to world coordinates (direction coordinate only, else fails) and (2) a composite region of all regions in the file is created.  However, these features are not required or used by CARTA.  In addition, any import failure in the file causes the entire import to fail, which was one motivation for the CARTA parser implementation as a workaround.

For now, imagenalysis continues to be used for CRTF region **export** for formatting, particularly world coordinate strings.

pford:
Closes #1219 .  Use internal implementation for annotation support and performance.

**Description**

* What is implemented or fixed?
When annotation regions are added for #340 , some regions not supported by CRTF externally will need to be imported back into CARTA.  This can only happen using the internal CRTF parser, since the casa CRTF parser will ignore these lines which start with `#`.  As an added bonus, the internal parser can improve performance of importing 20000 regions by 8-10x (#1216 ).

* How does this PR solve the issue?
Removes casa::RegionTextFile from CrtfImportExport import constructor so that the internal parser is always used.  Supporting functions also removed.

No, backend only.

Load various CRTF files and ensure that all regions are imported.

**Checklist**

- [x] changelog updated
- [x] ~e2e test passing~, e2e test fixed

### DIFF
--- a/src/Region/CrtfImportExport.cc
+++ b/src/Region/CrtfImportExport.cc
@@ -321,7 +321,7 @@ void CrtfImportExport::ProcessFileLines(std::vector<std::string>& lines) {
         } else if (region.find("poly") != std::string::npos) {
             region_state = ImportAnnPolygonLine(parameters, coord_frame);
         } else if (region == "global") {
-            ImportGlobalParameters(properties);
+            _global_properties = properties;
         } else {
             _import_errors.append(region + " not supported.\n");
         }
@@ -333,18 +333,6 @@ void CrtfImportExport::ProcessFileLines(std::vector<std::string>& lines) {
             // Set RegionProperties and add to list
             RegionProperties region_properties(region_state, region_style);
             _import_regions.push_back(region_properties);
-        }
-    }
-}
-
-void CrtfImportExport::ImportGlobalParameters(std::unordered_map<std::string, std::string>& properties) {
-    // Import properties from global file line to map
-    std::vector<std::string> global_keywords = {"coord", "range", "frame", "corr", "veltype", "restfreq", "linewidth", "linestyle",
-        "symsize", "symthick", "color", "font", "fontsize", "fontstyle", "usetex", "label", "labelcolor", "labelpos", "labeloff"};
-
-    for (auto& property : properties) {
-        if (std::find(global_keywords.begin(), global_keywords.end(), property.first) != global_keywords.end()) {
-            _global_properties[property.first] = property.second;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


pford:
Note imageanalysis annotation regions are still used for region export, for proper formatting.  Its use should be reviewed separately.

kswang1029 (commented):
The crtf region import performance is boosted. 🚀


Spotted an e2e test failure due to image coordinate precision. In the dev branch, image coordinate is trimmed/rounded(?) to the first decimal (so the e2e test is designed accordingly). Now with the new parser, that trimming/rounding issue is gone. So I will need to revise the e2e test accordingly. Please allow me some time to perform additional user tests. Overall this PR looks quite promising!

src/Region/CrtfImportExport.cc
```diff
@@ -742,13 +337,25 @@ void CrtfImportExport::ProcessFileLines(std::vector<std::string>& lines) {
     }
 }
 
-casacore::String CrtfImportExport::GetRegionDirectionFrame(std::unordered_map<std::string, std::string>& properties) {
-    casacore::String dir_frame;
+void CrtfImportExport::ImportGlobalParameters(std::unordered_map<std::string, std::string>& properties) {
+    // Import properties from global file line to map
+    std::vector<std::string> global_keywords = {"coord", "range", "frame", "corr", "veltype", "restfreq", "linewidth", "linestyle",
```
confluence: Can we change this to a `std::unordered_set` and use its `find` member function? It would avoid the repeated linear lookup.

confluence (changes requested):

src/Region/CrtfImportExport.cc
pford: On second thought, I think we should just save the map without checking for keywords.  This was previously a map for the imageanalysis implementation

confluence (approved):

kswang1029 (approved):
this should be fine as far as I can tell. 👍 One e2e test is fixed due to the increased precision of parsed image coordinate so when it is converted to angular size, the value for assertion changed.

PR #1223, opened 2022-12-02
Move the vector field calculation under the Frame

### LINKED ISSUE #1235: backend crash when requesting vector overlay data from computed polarization component
**Describe the bug**
This is a bug report from @jott3077.

When we load a Stokes cube and set the intensity source as "current image" and angular source as "computed PA", then generate the vector overlay on the image, if we use the animator to switch to one of the computed polarization component (eg total polarization intensity), the backend will crash. The macOS crash report is attached below.

> Thread 3 Crashed:
0   libsystem_kernel.dylib        	    0x7ff81372e00e __pthread_kill + 10
1   libsystem_pthread.dylib       	    0x7ff8137641ff pthread_kill + 263
2   libsystem_c.dylib             	    0x7ff8136afd24 abort + 123
3   libc++abi.dylib               	    0x7ff813720082 abort_message + 241
4   libc++abi.dylib               	    0x7ff813711245 demangling_terminate_handler() + 242
5   libobjc.A.dylib               	    0x7ff81360de19 _objc_terminate() + 104
6   libc++abi.dylib               	    0x7ff81371f4a7 std::__terminate(void (*)()) + 8
7   libc++abi.dylib               	    0x7ff813721d05 __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 27
8   libc++abi.dylib               	    0x7ff813721ccc __cxa_throw + 116
9   libcasa_casa.6.dylib          	       0x110914d54 casacore::Slicer::inferShapeFromSource(casacore::IPosition const&, casacore::IPosition&, casacore::IPosition&, casacore::IPosition&) const + 612
10  libcasa_lattices.6.dylib      	       0x11114bdc7 casacore::LatticeRegion::LatticeRegion(casacore::Slicer const&, casacore::IPosition const&) + 167
11  libcasa_lattices.6.dylib      	       0x110f9a73c casacore::SubLattice<float>::setRegion(casacore::Slicer const&) + 60
12  libcasa_lattices.6.dylib      	       0x110fb28cb casacore::SubLattice<float>::SubLattice(casacore::MaskedLattice<float> const&, casacore::Slicer const&, casacore::AxesSpecifier) + 251
13  libcasa_images.6.dylib        	       0x110d0bf80 casacore::SubImage<float>::SubImage(casacore::ImageInterface<float> const&, casacore::Slicer const&, casacore::AxesSpecifier, bool) + 144
14  carta_backend                 	       0x10ee12a27 carta::FileLoader::GetSlice(casacore::Array<float, std::__1::allocator<float> >&, carta::StokesSlicer const&) + 343
15  carta_backend                 	       0x10ed96674 carta::Frame::GetSlicerData(carta::StokesSlicer const&, float*) + 84
16  carta_backend                 	       0x10eda63fb carta::Frame::GetDownsampledRasterData(std::__1::vector<float, std::__1::allocator<float> >&, int&, int&, int, int, CARTA::ImageBounds&, int) + 587
17  carta_backend                 	       0x10edaeca3 carta::VectorFieldCalculator::DoCalculations(std::__1::function<void (CARTA::VectorOverlayTileData&)> const&) + 2611
18  carta_backend                 	       0x10ef8cc1c carta::Session::SendVectorFieldData(int) + 556
19  carta_backend                 	       0x10ef8bfe7 carta::Session::OnSetImageChannels(CARTA::SetImageChannels const&) + 823
20  carta_backend                 	       0x10ef8143d carta::SetImageChannelsTask::execute() + 301
21  carta_backend                 	       0x10efe4050 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, carta::ThreadManager::StartEventHandlingThreads(int)::$_0> >(void*) + 176
22  libsystem_pthread.dylib       	    0x7ff8137644e1 _pthread_start + 125
23  libsystem_pthread.dylib       	    0x7ff81375ff6b thread_start + 15

**To Reproduce**
1. load a Stokes cube and enable the vector overlay dialog
2. set the intensity source to "current image" and generator the vector overlay
3. use the animator to switch to Stokes Q. This works fine.
4. use the animator to switch to one of the computed component, such as total polarization intensity, the backend will crash

**Expected behavior**
no crash in step 4.

 - Browser version [e.g. 22]: v3-stable electron app
 - Backend branch [e.g. dev, v3b2 release]: v3-stable
 - Frontend branch [e.g. dev, v3b2 release]: v3-stable

markccchiang:
**Description**

This is the code refactoring for the vector field calculation. Also, fix #1235.

As @pford suggested in the f2f meeting, it is better that the task of the vector field calculation is called and protected by the Frame. 

No.

No need to change the ICD test.

**Checklist**

- [x] ~changelog updated /~ no changelog update needed


src/Frame/VectorFieldCalculator.cc
```diff
@@ -12,7 +12,7 @@
 
 namespace carta {
 
-bool VectorFieldCalculator::DoCalculations(VectorFieldCallback& callback) {
+bool DoVectorFieldCalculation(Frame* frame, const std::function<void(CARTA::VectorOverlayTileData&)>& callback) {
     if (!(frame && frame->IsValid())) {
```
pford: Since Frame calls this function, Frame should do any needed checks (call a check function in VectorFieldCalculator), lock the mutex, then pass in anything needed in DoVectorFieldCalculation such as VectorFieldSettings, instead of passing in itself.

pford (changes requested):

src/Frame/VectorFieldCalculator.cc
markccchiang: Does it mean moving the code block in `DoVectorFieldCalculation` into Frame's function `CalculateVectorField`?

pford: I was thinking in `Frame::CalculateVectorField`:
* if frame is not valid return false
* if vector_field_settings are ok: lock mutex, call `DoVectorFieldCalculation(vector_field_settings, ...)`, unlock mutex, return result
* if vector_field_settings are not ok: clear settings, set response message, return true

I may have missed some details, but this is the idea.

pford: I don't think you need a check function in VectorFieldCalculator just to check the settings (stokes_intensity and stokes_angle), unless there is more to it than that

src/Frame/Frame.cc
```diff
@@ -2334,4 +2331,216 @@ bool Frame::GetDownsampledRasterData(
...
+            settings.compression_type, settings.compression_quality);
+        empty_response.set_progress(1.0);
+        callback(empty_response);
+        return true;
+    }
+
+    return DoVectorFieldCalculation(settings, callback);
+}
+
+bool Frame::DoVectorFieldCalculation(
```
pford: This is a very long function.  See what you can move into smaller functions, or into VectorField if it does not involve `GetDownsampledRasterData` (for example, the calculations starting with the lambda functions do not need to be in Frame)

src/Frame/Frame.cc
```diff
@@ -2334,4 +2329,221 @@ bool Frame::GetDownsampledRasterData(
...
+            std::string stokes = one.first;
+            if (stokes_flag[stokes] &&
+                !GetDownsampledRasterData(stokes_data[stokes], width, height, _z_index, stokes_indices[stokes], bounds, mip)) {
+                return false;
+            }
+        }
+
+        // Calculate the current stokes as polarized intensity or polarized angle
+        std::vector<float> current_stokes_data;
+        if (current_stokes_as_pi || current_stokes_as_pa) {
```
pford, line 2396: There are a lot of `if` statements from here.  Can we put all the `if (current_stokes_as_*)` stuff in this block, or even use `Frame::FillDownsampledData` for both cases (no stokes or current stokes)?

And then have `if (calculate_pi)` and `if (calculate_pa)` blocks.  I think you could even move the calculations to VectorField, something like `if (calculate_pi) { CalculatePolarizedIntensity(tile_pi, tiles[i], stokes_data, settings, width, height); }` which calculates pi, applies transforms, and fills tile data.  Same for calculate_pa.

My concern, as we talked about at the F2F, is that the backend has very large classes (including Frame) and we should move as much code as possible to other files that do the work once we have the data.

pford (approved):

acdo2002:
@markccchiang 
I have tested this open PR branch with my currently developing Vector Overlay ICD tests.
All of the tests are passed, no obvious regression performance issue found compare to the dev branch. (6 test cases run ~26sec).
The testing platform include Ubuntu 22.04, Mac 11.6, and Mac 12.4.

kswang1029 (approved):
A set of e2e tests is implemented with the dev backend+dev frontend branches and then applied to this branch+dev frontend. With additional manual tests, no regression is observed as far as I can tell. 👍

src/DataStream/VectorField.cc
```diff
@@ -0,0 +1,171 @@
...
+            FillTileData(
+                tile_pa, tile.x, tile.y, tile.layer, mip, width, height, stokes_data["CUR"], compression_type, compression_quality);
+        }
+    }
+
+    // Calculate PI and PA using stokes data I, Q or U
+    // Lambda function to apply a threshold
+    auto apply_threshold = [&](float i, float result) {
+        return ((std::isnan(i) || (!std::isnan(threshold) && (i < threshold))) ? FLOAT_NAN : result);
+    };
```
confluence: This should be moved out into its own function, to match `Valid`, etc..

src/DataStream/VectorField.cc
```diff
@@ -0,0 +1,171 @@
...
+        std::vector<float> pi;
+        pi.resize(width * height);
+
+        // Lambda function to calculate PI, errors are applied
+        auto calc_pi = [&](float q, float u) {
+            if (!std::isnan(q) && !std::isnan(u)) {
+                return ((float)std::sqrt(std::pow(q, 2) + std::pow(u, 2) - (std::pow(q_error, 2) + std::pow(u_error, 2)) / 2.0));
+            }
+            return FLOAT_NAN;
+        };
```
confluence: This should be moved out into its own function, to match `Valid`, etc..

src/DataStream/VectorField.cc
```diff
@@ -0,0 +1,171 @@
...
+    CARTA::CompressionType compression_type = settings.compression_type;
+    float compression_quality = settings.compression_quality;
+    int stokes_intensity = settings.stokes_intensity;
+    int stokes_angle = settings.stokes_angle;
+    double q_error = settings.q_error;
+    double u_error = settings.u_error;
+    bool calculate_pi = settings.calculate_pi;
+    bool calculate_pa = settings.calculate_pa;
+    bool current_stokes_as_pi = settings.current_stokes_as_pi;
+    bool current_stokes_as_pa = settings.current_stokes_as_pa;
```
confluence: The only reason to do this is to avoid the verbosity of using `settings.*` throughout this function. But this could be avoided entirely if this function were a member of the same struct, instead of taking the struct as a parameter.

I recommend that you turn the `VectorFieldSettings` struct into a `VectorField` class, and move all these functions inside it. You will then be able to refer to the data members inside the member functions with bare names.

confluence (changes requested):

src/DataStream/VectorField.cc
confluence: I was not suggesting that you *nest* the settings struct *inside* a new class; I was suggesting that you *convert* the settings struct *into* a class, and move the functions into that class. This function is directly manipulating properties of the struct, which is why it should be a member *of the same struct*. There only needs to be one layer -- a `VectorField` class which contains all the data members and all the functions directly. It's fine for this class to have a `ClearSettings` function which clears the data members.

(Do not abbreviate "settings" to "sets"; "sets" has a different meaning and this is confusing.)

Converting the struct to a class is mostly a convention (if something contains non-trivial functions as well as data, it makes more sense conceptually to declare it as a class).

The other functions are less coupled to the struct members, but they are all related to each other and to the same functionality, so the cleanest way to structure this is to group them all in the same class.

confluence (approved):
I'm happy with this code structure. @kswang1029 I think we need another test after the code changes.

kswang1029:
no regression is seen from the e2e tests. :+1: I will perform some manual tests then approve this.

kswang1029 (approved):
@markccchiang this looks good, concluded from manual tests and e2e tests. 👍 This also fixed the crash issue #1235 

@confluence feel free to merge this in : ) Apologies for the delay.

Issue #1215, opened 2022-11-16
Set up automatic generation of documentation from annotations


confluence:
Set up a framework for generating documentation from annotations in the code (probably Doxygen format) and publishing it on Read The Docs.

PR #1225, opened 2022-12-06
Add Doxygen documentation

confluence:
This (mostly) implements #1215. The original plan was to use Breathe to integrate Doxygen documentation with Sphinx and publish it on Read The Docs. However, after testing various options, I have concluded that Breathe is not currently fit for purpose -- it converts only a small subset of the Doxygen output, presents it in a much less usable format, and is very buggy (our complex codebase triggers a lot of errors and warnings that are not easy to resolve).

There is no compelling reason for us to use it, so I suggest that we don't.  If we really want to put up this documentation on RTD, there is apparently a way of bypassing Sphinx entirely and side-loading a static Doxygen website -- but it would be simpler to host this documentation on GitHub Pages, and it's possible to do this and trigger automatic rebuilds using GitHub Actions.

Plain Doxygen is very useful out of the box, even with most of the codebase undocumented. It automatically parses a lot of information out of the bare code, and generates various useful diagrams which can be used to visualise relationships between components. I have enabled a lot of options which are off by default, like exposing private class members -- because this is internal documentation for developers, it makes sense to show more internal structure and not just the public API. This is particularly useful for making the diagrams more complete. I have also added a [modern stylesheet](https://jothepro.github.io/doxygen-awesome-css) which makes the generated HTML look a lot better.

~Some elements must be documented explicitly to be picked up by Doxygen (e.g. functions which are not inside classes).~ There is an option for extracting all entities, even if they're undocumented. I think it's sensible to keep this on for now.

I suggest putting Doxygen comments only in header files, for consistency and to avoid duplication. This seems to be the standard. I believe that Doxygen will merge documentation for the same objects, so we have the option of documenting the API in headers and implementation specifics in implementation files if this makes sense (although perhaps that level of information belongs in inline comments).

Steps to build and view the docs locally:

1. Install Doxygen and Graphviz (I have tested with Doxygen 1.9.5, 1.9.6 and 1.8.17 and saw no major differences)
2. Make sure you update submodules to get the stylesheet
3. `cd docs`
4. `doxygen`
5. Open `html/index.html` in your browser (Firefox and Chrome both work)
6. If you want to test dark mode, you need to host the files on a webserver.

Currently set to draft, because I want to

- [X] Try adding more documentation comments to different components
- [X] See how expensive it is to generate call dependency graphs<sup>1</sup>
- [X] Add a landing page
- [X] Proof of concept integration of arbitrary extra pages (markdown is now supported)<sup>2</sup>

1. Call and caller graphs are a bit slow to generate the first time, but after that only the necessary files are updated. The difference should also be negligible if the docs are generated automatically on merges to `dev`.  I think it's a good idea to include these (at least for now, to aid in refactoring).
2. Instead of creating a separate set of documentation somewhere else (e.g. the repository wiki) for human-written information like code style, design decisions, etc., we can integrate them into these documents. I made a `static` directory to store these files, and to demonstrate the concept I started a page for developer FAQs and tips. A current limitation of the markdown support is that all pages are added at the top level (of the sidebar), but Doxygen also supports an older static file format which allows the creation of a hierarchy of sub-pages. 

Once this is ready for review, I will also

- [X] Replicate this whole setup in the `fits2idia` repository to test out the GitHub Pages deployment.

**Checklist**

- [X] e2e test passing / ~added corresponding fix~


kswang1029 (approved):
no regression is observed from e2e tests. 

with macOS Monterey, doxygen 1.9.5, and graphviz 2.50.0 (both installed via conda), I can generate the doc smoothly. The context is amazing. 👍

YuHsuan-Hwang (approved):
I can also successfully generate the document 👍
I think the info in FAQ is very useful. Maybe good to also add them to `README.md` or `contributing.md`

markccchiang (approved):

PR #1227, opened 2022-12-07
Fix c&p error in tile cache key hash function

confluence:
This is a one-character fix for a logic error that I found while writing demo documentation. This tile cache key hash function should be combining the hashes for its x and y coordinates, but it's actually using the x coordinate twice. This means that all keys with the same x coordinate will have colliding hashes.

This may be having a detrimental effect on the performance of Y profiles (for HDF5 files, which use the tile cache), but the total number of tiles / keys in use at once may be too low for the impact to be significant. If there is a noticeable difference, it may be worthwhile to mention this in the changelog, otherwise it's probably too minor a change (we could add an aggregate bugfix item before release).

- [x] ~changelog updated~ / no changelog update needed
- [X] added ZenHub estimate, milestone, and release

### DIFF
--- a/src/Cache/TileCacheKey.h
+++ b/src/Cache/TileCacheKey.h
@@ -24,7 +24,7 @@ namespace std {
 template <>
 struct hash<carta::TileCacheKey> {
     std::size_t operator()(const carta::TileCacheKey& k) const {
-        return std::hash<int32_t>()(k.x) ^ (std::hash<int32_t>()(k.x) << 1);
+        return std::hash<int32_t>()(k.x) ^ (std::hash<int32_t>()(k.y) << 1);
     }
 };
 } // namespace std


pford (approved):

kswang1029 (approved):
this should be fine as far as I can tell by loading a large fits image and its hdf5 version and playing around image zoom/pan, cursor spatial profiler, line/polyline spatial profiler, and contouring. All e2e tests are passing.

As for performance impact, I am not sure how to test it. Any ideas?

confluence:
> As for performance impact, I am not sure how to test it. Any ideas?

We could repeatedly request full-resolution profiles from a large file, and see if there's a difference in the difference between the speed of X and Y profiles before and after this change. But it's probably not necessary to spend a lot of time on it now -- I suspect that reading non-contiguous data from disk for the Y profiles will dwarf any impact the slower cache lookups may have had.

I'm going to merge this as-is.

PR #1240, opened 2023-02-20
Angus/1185 state save

### LINKED ISSUE #1185: Initial support for saving/restoring of workspaces (backend)
Associated frontend issue https://github.com/CARTAvis/carta-frontend/issues/1272

This a feature that CARTA can save a snapshot the the state in the GUI and restore it later. Snapshots will be termed "workspaces"

More details will be provided.

veggiesaurus:
**Description**
Closes #1185. Implements initial functionality required for workspaces. There is plenty still left to do on this feature, but for beta1, we aim to have the core functionality working. 

The backend doesn't actually need to do too much in order to support this. The main things I've added are:
- CORS headers for the API calls
- Ability to simply _list_ objects, rather than loading them all. This means we can refresh the list on demand without opening each file

Companion PR of https://github.com/CARTAvis/carta-frontend/pull/2084

**Checklist**


confluence (approved):

pford (approved):

veggiesaurus:
@confluence I think https://github.com/CARTAvis/carta-frontend/pull/2084 is close to getting merged in, can you have another look at this branch to make sure we're all good?

src/HttpServer/HttpServer.cc
```diff
@@ -56,16 +61,28 @@ void HttpServer::RegisterRoutes() {
...
         app.get("/api/database/snippets", [&](auto res, auto req) { HandleGetObjects("snippet", res, req); });
+        app.get("/api/database/snippet/:name", [&](auto res, auto req) { HandleGetObject("snippet", res, req); });
         app.put("/api/database/snippet", [&](auto res, auto req) { HandleSetObject("snippet", res, req); });
         app.del("/api/database/snippet", [&](auto res, auto req) { HandleClearObject("snippet", res, req); });
+
+        app.get("/api/database/list/workspaces", [&](auto res, auto req) { HandleGetObjectList("workspace", res, req); });
+        app.get("/api/database/workspaces", [&](auto res, auto req) { HandleGetObjects("workspace", res, req); });
+        app.get("/api/database/workspace/:name", [&](auto res, auto req) { HandleGetObject("workspace", res, req); });
+        app.put("/api/database/workspace", [&](auto res, auto req) { HandleSetObject("workspace", res, req); });
+        app.del("/api/database/workspace", [&](auto res, auto req) { HandleClearObject("workspace", res, req); });
```
confluence, line 85: Now that I'm looking at this again, I don't love the use of magic hardcoded strings here, and would prefer it if we created constants for these (or maybe an enum would be cleaner). But this is a minor issue which predates this PR, so we could fix it later.

src/HttpServer/HttpServer.cc
```diff
@@ -461,6 +580,8 @@ bool HttpServer::WriteObjectFile(const std::string& object_type, const std::stri
             obj["$schema"] = CARTA_LAYOUT_SCHEMA_URL;
         } else if (object_type == "snippet") {
             obj["$schema"] = CARTA_SNIPPET_SCHEMA_URL;
+        } else if (object_type == "workspace") {
+            obj["$schema"] = CARTA_WORKSPACE_SCHEMA_URL;
```
confluence, line 588: Perhaps this should be a lookup in a `SCHEMA_URL` unordered map. I think you won't need an explicit hash function if you change the strings to enum values (since C++14 `std::hash` can be used on enums directly, so I believe that the default Just Works now).

src/HttpServer/HttpServer.cc
veggiesaurus: can I suggest we make this and the suggestion below a separate issue that we solve after code freeze?

confluence: Works for me; I'd rather not mess around with a working PR now. :)

kswang1029 (approved):

Issue #829, opened 2021-06-09
custom histogram calculations


kswang1029:
At the moment a histogram is computed with range and n_bin derived automatically. To make the histogram widget be more useful for analysis, custom
- n_bin
- bin_width
- range
need to be configurable.

related to frontend: https://github.com/CARTAvis/carta-frontend/issues/1488

veggiesaurus:
I think this might already actually be supported on the backend

edit: no, currently the histogram config only supports `num_bins`, not the range itself

kswang1029:
and need ICD changes I suppose

veggiesaurus:
> and need ICD changes I suppose

yes, non-breaking ones though, and probably pretty trivial

kswang1029:
Done with #1254

PR #1254, opened 2023-04-14
Mark/829 customize histogram calculations

markccchiang:
**Description**

Customized the histogram calculations as described in #829.

Modified the protobuf message `HistogramConfig` and added options for the number of bins or pixel bounds which will be applied in histogram calculations.

Yes. The companion [frontend PR](https://github.com/CARTAvis/carta-frontend/pull/2141) and [protobuf PR](https://github.com/CARTAvis/carta-protobuf/pull/83) are shown on the hyperlinks.

The ICD tests related to the histogram calculations need to be changed.

**Checklist**

- [x] protobuf updated to the latest dev commit ~/ no protobuf update needed~

### DIFF
--- a/src/Cache/RequirementsCache.h
+++ b/src/Cache/RequirementsCache.h
@@ -67,7 +67,7 @@ struct HistogramConfig {
     bool fixed_num_bins;
     int num_bins;
     bool fixed_bounds;
-    Bounds<float> bounds;
+    HistogramBounds bounds;
 
     HistogramConfig() : coordinate("z"), channel(CURRENT_Z), fixed_num_bins(false), num_bins(AUTO_BIN_SIZE), fixed_bounds(false) {}
 
@@ -77,17 +77,17 @@ struct HistogramConfig {
           fixed_num_bins(config.fixed_num_bins()),
           num_bins(config.num_bins()),
           fixed_bounds(config.fixed_bounds()),
-          bounds(Bounds<float>(config.bounds())) {}
+          bounds(HistogramBounds(config.bounds())) {}
 
     bool operator!=(const HistogramConfig& rhs) const {
         return (coordinate != rhs.coordinate) || (channel != rhs.channel) || (fixed_num_bins != rhs.fixed_num_bins) ||
                (fixed_bounds != rhs.fixed_bounds) || (num_bins != rhs.num_bins) || (bounds != rhs.bounds);
     }
 
-    Bounds<float> GetBounds(const BasicStats<float>& stats) const {
+    HistogramBounds GetBounds(const BasicStats<float>& stats) const {
         float min = fixed_bounds ? bounds.min : stats.min_val;
         float max = fixed_bounds ? bounds.max : stats.max_val;
-        return Bounds<float>(min, max);
+        return HistogramBounds(min, max);
     }
 };
 
@@ -113,7 +113,7 @@ struct HistogramCache {
         stats = stats_;
     }
 
-    bool GetHistogram(int num_bins_, const Bounds<float>& bounds, Histogram& histogram_) {
+    bool GetHistogram(int num_bins_, const HistogramBounds& bounds, Histogram& histogram_) {
         if (histograms.count(num_bins_)) {
             const auto& hist = histograms.at(num_bins_);
             if (bounds == hist.GetBounds()) {
--- a/src/Cache/RequirementsCache.h
+++ b/src/Cache/RequirementsCache.h
@@ -95,6 +95,16 @@ struct HistogramConfig {
     }
 };
 
+struct HistogramBounds {
+    float min;
+    float max;
+
+    HistogramBounds(const HistogramConfig& config, const BasicStats<float>& stats) {
+        min = config.fixed_bounds ? config.min_val : stats.min_val;
+        max = config.fixed_bounds ? config.max_val : stats.max_val;
+    }
+};
+
 struct RegionHistogramConfig {
     std::vector<HistogramConfig> configs;
 };
--- a/src/Cache/RequirementsCache.h
+++ b/src/Cache/RequirementsCache.h
@@ -90,6 +90,12 @@ struct HistogramConfig {
           fixed_bounds(config.fixed_bounds()),
           min_val(config.bounds().min()),
           max_val(config.bounds().max()) {}
+
+    bool operator!=(const HistogramConfig& rhs) const {
+        return (coordinate != rhs.coordinate) || (channel != rhs.channel) || (fixed_num_bins != rhs.fixed_num_bins) ||
+               (fixed_bounds != rhs.fixed_bounds) || (num_bins != rhs.num_bins) || !AreEqual(bin_width, rhs.bin_width) ||
+               !AreEqual(min_val, rhs.min_val) || !AreEqual(max_val, rhs.max_val);
+    }
 };
 
 struct RegionHistogramConfig {
--- a/src/Cache/RequirementsCache.h
+++ b/src/Cache/RequirementsCache.h
@@ -65,9 +65,19 @@ struct HistogramConfig {
     std::string coordinate;
     int channel;
     int num_bins;
+    bool fixed_bounds;
+    float min_val;
+    float max_val;
 
-    HistogramConfig() {}
-    HistogramConfig(const std::string& coordinate, int chan, int bins) : coordinate(coordinate), channel(chan), num_bins(bins) {}
+    HistogramConfig() : coordinate("z"), channel(CURRENT_Z), num_bins(AUTO_BIN_SIZE), fixed_bounds(false), min_val(0), max_val(0) {}
+
+    HistogramConfig(const CARTA::SetHistogramRequirements_HistogramConfig& config)
+        : coordinate(config.coordinate()),
+          channel(config.channel()),
+          num_bins(config.num_bins()),
+          fixed_bounds(config.fixed_bounds()),
+          min_val(config.bounds().min()),
+          max_val(config.bounds().max()) {}
 };
 
 struct RegionHistogramConfig {
@@ -75,7 +85,7 @@ struct RegionHistogramConfig {
 };
 
 struct HistogramCache {
-    BasicStats<float> stats;
+    BasicStats<float> stats;                       // Statistics data without fixed bounds
     std::unordered_map<int, Histogram> histograms; // key is num_bins
 
     HistogramCache() {}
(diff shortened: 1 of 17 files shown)


kswang1029:
this looks good. please proceed code review.

kswang1029 (approved):
this looks fine. no regression is seen from e2e tests.

acdo2002:
I checked this backend PR + protobuf PR, it won't affect any current ICD tests both on MacOS and Ubuntu.
They are all working fine with current ICD-RxJS tests.
However the new sub-message (HistogramConfig) has been added in the /stream/region_histogram.proto
Additional ICD test about this sub-message is required!

src/Cache/RequirementsCache.h
```diff
@@ -64,18 +64,53 @@ struct CacheIdHash {
 struct HistogramConfig {
     std::string coordinate;
     int channel;
+    bool fixed_num_bins;
     int num_bins;
+    bool fixed_bounds;
+    float min_val;
+    float max_val;
```
confluence: This should be a `HistogramBounds` object, not two separate values. That will simplify a lot of things.

src/Util/Image.cc
```diff
@@ -29,3 +29,7 @@ bool IsComputedStokes(int stokes) {
 bool IsComputedStokes(const std::string& stokes) {
     return IsComputedStokes(StokesValues[StokesStringTypes[stokes]]);
 }
+
+bool AreEqual(float val1, float val2) {
+    return fabs(val1 - val2) < 1e-6;
+}
```
confluence: This should be a function on the `HistogramBounds` struct, it should compare both sets of bounds at once, and it should probably be renamed to `Close` or just `Equal`.

src/Cache/RequirementsCache.h
```diff
@@ -64,18 +64,53 @@ struct CacheIdHash {
...
+          fixed_num_bins(config.fixed_num_bins()),
+          num_bins(config.num_bins()),
+          fixed_bounds(config.fixed_bounds()),
+          min_val(config.bounds().min()),
+          max_val(config.bounds().max()) {}
+
+    bool operator!=(const HistogramConfig& rhs) const {
+        return (coordinate != rhs.coordinate) || (channel != rhs.channel) || (fixed_num_bins != rhs.fixed_num_bins) ||
+               (fixed_bounds != rhs.fixed_bounds) || (num_bins != rhs.num_bins) || !AreEqual(min_val, rhs.min_val) ||
+               !AreEqual(max_val, rhs.max_val);
```
confluence: Here the two `AreEqual` comparisons should be replaced by one comparison (which calls the revised function on the bounds object).

src/Cache/RequirementsCache.h
```diff
@@ -64,18 +64,53 @@ struct CacheIdHash {
...
+    }
+};
 
-    HistogramConfig() {}
-    HistogramConfig(const std::string& coordinate, int chan, int bins) : coordinate(coordinate), channel(chan), num_bins(bins) {}
+struct HistogramBounds {
```
confluence: The comparison function should be here, and take another bounds object as a parameter.

src/Cache/RequirementsCache.h
```diff
@@ -92,10 +127,13 @@ struct HistogramCache {
         stats = stats_;
     }
 
-    bool GetHistogram(int num_bins_, Histogram& histogram_) {
+    bool GetHistogram(int num_bins_, float min_val, float max_val, Histogram& histogram_) {
```
confluence: The parameter should be a single bounds object.

src/Cache/RequirementsCache.h
confluence: This struct should also be moved from here to `ImageStats/Histogram`, and used in the `Histogram` object in addition to the two separate bounds values:

* The underlying `_min_val` and `_max_val` values are used internally in calculations, so they should be kept as they are.
* The `GetMinVal` and `GetMaxVal` getters are used in one of the `Histogram` constructors, so they're also needed.
* Apart from that, `GetMinVal` and `GetMaxVal` are used in `ConsistencyCheck` and in a test to perform a comparison which is a duplicate of what is being introduced here -- those places should be updated to use the new bounds comparison.
* A `GetBounds` function should be added to `Histogram` which returns a `HistogramBounds` constructed from the two bounds values
* `HistogramBounds` needs a constructor which just takes two floats

src/Cache/RequirementsCache.h
```diff
@@ -92,10 +127,13 @@ struct HistogramCache {
...
         if (histograms.count(num_bins_)) {
-            histogram_ = histograms.at(num_bins_);
-            return true;
+            const auto& hist = histograms.at(num_bins_);
+            if (AreEqual(hist.GetMinVal(), min_val) && AreEqual(hist.GetMaxVal(), max_val)) {
```
confluence: Once `Histogram` is adjusted to have a `GetBounds` which returns a single bounds object, the same single comparison can be done here.

src/Frame/Frame.cc
```diff
@@ -871,15 +867,15 @@ bool Frame::GetBasicStats(int z, int stokes, BasicStats<float>& stats) {
...
+bool Frame::GetCachedImageHistogram(int z, int stokes, int num_bins, const HistogramBounds& bounds, Histogram& hist) {
     // Get image histogram results from cache
     int cache_key(CacheKey(z, stokes));
     if (_image_histograms.count(cache_key)) {
         // get from cache if correct num_bins
         auto results_for_key = _image_histograms[cache_key];
 
         for (auto& result : results_for_key) {
-            if (result.GetNbins() == num_bins) {
+            if (result.GetNbins() == num_bins && AreEqual(result.GetMinVal(), bounds.min) && AreEqual(result.GetMaxVal(), bounds.max)) {
```
confluence: The same change can be made here.

src/Frame/Frame.cc
```diff
@@ -888,12 +884,12 @@ bool Frame::GetCachedImageHistogram(int z, int stokes, int num_bins, Histogram&
...
 }
 
-bool Frame::GetCachedCubeHistogram(int stokes, int num_bins, Histogram& hist) {
+bool Frame::GetCachedCubeHistogram(int stokes, int num_bins, const HistogramBounds& bounds, Histogram& hist) {
     // Get cube histogram results from cache
     if (_cube_histograms.count(stokes)) {
         for (auto& result : _cube_histograms[stokes]) {
             // get from cache if correct num_bins
-            if (result.GetNbins() == num_bins) {
+            if (result.GetNbins() == num_bins && AreEqual(result.GetMinVal(), bounds.min) && AreEqual(result.GetMaxVal(), bounds.max)) {
```
confluence: And here.

src/ImageStats/StatsCalculator.cc
```diff
@@ -23,13 +23,12 @@ void CalcBasicStats(BasicStats<float>& stats, const float* data, const size_t da
     stats = mm.GetStats();
 }
 
-Histogram CalcHistogram(int num_bins, const BasicStats<float>& stats, const float* data, const size_t data_size) {
-    if ((stats.min_val == std::numeric_limits<float>::max()) || (stats.max_val == std::numeric_limits<float>::min()) || data_size == 0) {
+Histogram CalcHistogram(int num_bins, const HistogramBounds& bounds, const float* data, const size_t data_size) {
+    if ((bounds.min == std::numeric_limits<float>::max()) || (bounds.max == std::numeric_limits<float>::min()) || data_size == 0) {
```
confluence: These two comparisons should go inside another function on the `HistogramBounds` object, perhaps called `Invalid`.

src/ImageStats/StatsCalculator.cc
```diff
@@ -23,13 +23,12 @@ void CalcBasicStats(BasicStats<float>& stats, const float* data, const size_t da
...
         // empty / NaN region
         return Histogram(1, 0, 0, data, data_size);
-    } else {
-        return Histogram(num_bins, stats.min_val, stats.max_val, data, data_size);
     }
+    return Histogram(num_bins, bounds.min, bounds.max, data, data_size);
```
confluence: The `Histogram` constructor should probably take a single bounds object as a parameter (it can access the individual values to set them).

confluence (changes requested):
This is mostly fine, but I would like the `HistogramBounds` struct to be used in more places, and for comparisons of the bounds to be moved into functions on this struct. This will greatly simplify some repeated code and make use of these values more consistent.

src/Util/Image.cc
confluence: We should also use `std::numeric_limits<float>::epsilon()` instead of a hard-coded error value. (I moved this comment; I had it in the wrong place -- sorry!)

confluence (approved):
Happy to merge. Good idea to create a generic `Bounds` object. 👍

kswang1029:
@confluence please wait for a final e2e test and wait for the companion frontend PR plus protobuf PR. This requires a coordinated merge.

confluence:
> @confluence please wait for a final e2e test and wait for the companion frontend PR plus protobuf PR. This requires a coordinated merge.

Sure; I'm not planning to merge now. :)

kswang1029 (approved):
looking good and no regression from e2e tests 👍

acdo2002:
I test this current PR and find it will break one ICD-RxJS test. 
About the swapped image. 
Open “HH211_IQU-swap-rfds.image” (https://drive.google.com/drive/folders/1xJbRMcmq-6ka_WCgUMyiwMcodhQP45fg?usp=sharing) the first returned REGION_HISTOGRAM_DATA 
should be like (current carta-backend dev branch)
Screen Shot 2023-04-28 at 3 24 45 PM
Because the image at the first channel is all NaN.

However in the current PR, I found it shows
Screen Shot 2023-04-28 at 3 33 29 PM
after open the image, the first returned REGION_HISTOGRAM_DATA  contains some values.
It should not have any value if the whole image values are all NaN

kswang1029:
> I test this current PR and find it will break one ICD-RxJS test. About the swapped image. Open “HH211_IQU-swap-rfds.image” (https://drive.google.com/drive/folders/1xJbRMcmq-6ka_WCgUMyiwMcodhQP45fg?usp=sharing) the first returned REGION_HISTOGRAM_DATA should be like (current carta-backend dev branch) Screen Shot 2023-04-28 at 3 24 45 PM Because the image at the first channel is all NaN.
> 
> However in the current PR, I found it shows Screen Shot 2023-04-28 at 3 33 29 PM after open the image, the first returned REGION_HISTOGRAM_DATA contains some values. It should not have any value if the whole image values are all NaN

good catch. looks like the binWidth and firstBinCenter are crazy numbers.

kswang1029:
@markccchiang @acdo2002 @confluence on another thought about the issue above, the current behavior is not harmful to users at all I think. Those big numbers are understandable in this case. So not sure if we need to address this (at least from users perspective). If not, we can revise ICD tests accordingly.

markccchiang:
> I test this current PR and find it will break one ICD-RxJS test. About the swapped image. Open “HH211_IQU-swap-rfds.image” 

@acdo2002 This problem is due to the different types of statistics (min/max) values and histogram bounds. When the image statistics values are unavailable, they are assigned to extreme values of `float` type. However, the type of histogram bounds is `double`, which will cause a problem when checking their validity. The reason I use `double` type as the histogram bounds from configuration is to avoid the imprecision conversion from requested protobuf messages.

acdo2002:
I have confirmed the current PR, the ICD-RxJS tests are all passed on Ubuntu22.04, MacOS13, and MacOS11.
The weird histogramData ICD messages in the swapped image has been disappeared.

Issue #1259, opened 2023-05-02
[regression] PV generator is not consistent between fits and hdf5


acdo2002:
**Describe the bug**
Running the ICD-RxJS test for PR #1253 and found the `PV_GENERATOR_HDF5_COMPARED_FITS.test.ts` is failed, indicating the PV generator images of fits and hdf5 are not conisitent

**To Reproduce**
1. Open `HD163296_CO_2_1.fits` ([link](https://drive.google.com/drive/folders/1Wi1WxXk6gg-aQQXhXTTEN8m8vM4bWMcX?usp=sharing)) and set a line region: controlPoints: [{ x: 79, y: 77 }, { x: 362, y: 360 }], rotation: 135
2. Create the first PV image (width:3) for `HD163296_CO_2_1.fits` and save this image as `HD163296_CO_2_1_pv-fits-pv-preview.image`
3. Open `HD163296_CO_2_1.hdf5`, spatially and spectrally matched to `HD163296_CO_2_1.fits`.
4. Create the second PV image (width:3) for `HD163296_CO_2_1.hdf5` and save this image as `HD163296_CO_2_1_pv-hdf5-pv-preview.image`
5. Append image using LEL: HD163296_CO_2_1_pv-fits-pv-preview.image` - `HD163296_CO_2_1_pv-hdf5-pv-preview.image`, then you can see the residual (i.e. difference)

**Expected behavior**
On the backend dev commit : `f5c9379c`
the LEL subtraction should be all "0" image, like this
<img width="1364" alt="Screen Shot 2023-05-02 at 3 17 09 PM" src="https://user-images.githubusercontent.com/34430648/235616024-7e27bfd8-3ba7-4957-bfbb-abe1ebfeb7c2.png">

**Screenshots or videos**
However in this PR #1253 , there is a residual in the LEL subtraction
<img width="1367" alt="Screen Shot 2023-05-02 at 3 17 21 PM" src="https://user-images.githubusercontent.com/34430648/235616186-26c73209-2d67-49d5-a163-fc99066e3492.png">

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]: macOS 11 & 13
 - Browser [e.g. chrome, safari, electron app]:  chrome
 - Browser version [e.g. 22]: Version 111.0.5563.110
 - Backend branch [e.g. dev, v3b2 release]: v4-beta (due to https://github.com/CARTAvis/carta-backend/pull/1253)
[...]

PR #1267, opened 2023-05-16
Fix HDF5 PV generator regression failure

pford:
**Description**

#1259 
The HDF5 PV image had duplicate profiles from the Hdf5Loader cache. The cache was deleted only for region id = TEMP_REGION_ID.  However, for parallel profiles each region now has a unique id decremented from the temp id, so the profile was being cached and reused when the same id was assigned later.  Now the cache is deleted if the region id <= TEMP_REGION_ID. Also removed distortion check output for matched regions, causing too much debug output; now the log only reports that the polygon approximation is being used.
No.
See description in issue: create PV images from identical FITS and HDF5 source images, then compare them using a LEL expression.

**Checklist**

- [ ] e2e test passing / added corresponding fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 * Prevent the installation of pugixml library files ([#1261](https://github.com/CARTAvis/carta-backend/issues/1261)).
 * Fixed spatial profile for polyline in widefield image ([#1258](https://github.com/CARTAvis/carta-backend/issues/1258)).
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):
this appears fine with some manual tests. no regression from e2e tests. @acdo2002 could you repeat your test to see if the problem is fixed?

acdo2002 (approved):
Yes, the regression problem solved in this PR.
The ICD-RxJS are all passed now. Once this PR merged into the dev, I will unblock the part of ICD tests.
And manual check the same PV cut in fits and hdf5 is identical, because two PV cut images subtracted are all 0.
Screen Shot 2023-05-17 at 5 02 04 PM

confluence (approved):

PR #1255, opened 2023-04-27
Fix FITS header and data errors

pford:
**Description**

* What is implemented or fixed?
Fixes #1233 and #1265 
* How does this PR solve the issue? 

1. CompressedFits: fix header parsing error for "COMMENT ================================"
2. CartaFitsImage: read rescaled 16-bit int data correctly with cfitsio (using equivalent bitpix=float)
3. FileExtInfoLoader: fix unit string ("degrees") in FITS image header unrecognized by casacore (regression failure)

* Are there any companion PRs (frontend, protobuf)?
No
* Is there anything else that testers should know (e.g. exactly how to reproduce the issue)?
Open compressed fits.gz image attached to issue 1233 and FITS image attached to issue 1265.  Images should be opened and displayed properly.

**Checklist**

- [x] changelog updated / ~no changelog update needed~
- [x] e2e test passing / ~added corresponding fix~
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] added reviewers and assignee
- [x] added ZenHub estimate, milestone, and release

PR #1269, opened 2023-05-23
Fixes the case-sensitive of reading BUNIT from header

### LINKED ISSUE #1187: BUNIT map header appears to be case sensitive
**Describe the bug**
We have received a helpdesk ticket from a user who has an issue with a header in their image files.  `BUNIT = 'Jy/Beam'` results in the statistics widget showing NaN for FluxDensity. The user discovered that if they modify the header to `BUNIT = 'Jy/beam'`, then CARTA correctly displays the FluxDensity value. The user says the same problem occurs in both CARTA v2.0 and v3.0. 

**To Reproduce**
I have not personally tried it, but perhaps modify the BUNIT header in a FITS file from 'Jy/Beam' to 'Jy/beam' and check the FluxDensity value in the statistics widget each time.

**Expected behavior**
I assume it should not be case sensitive and any combination of `Jy/Beam`, `Jy/Beam` or `JY/BEAM` should work?

markccchiang:
**Description**

Closes #1187. 

It appears that the casacore is case-sensitive to BUNIT setting reads from the file header. To solve it, CARTA will convert any case combinations such as `Jy/Beam`, `jy/beam`, or `JY/BEAM`, etc. to `Jy/beam` while opening a file. 

No.

The statistics calculation for `FluxDensity` is now available for any files with the case combinations of BUNIT `Jy/beam`.

**Checklist**

### DIFF
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -92,18 +92,30 @@ void FileLoader::OpenFile(const std::string& hdu) {
 
     // Normalize the upper/lower cases of BUNIT string from header
     if (_image) {
-        casacore::String bunit = _image->units().getName();
+        std::string bunit = _image->units().getName();
         bunit.erase(std::remove(bunit.begin(), bunit.end(), '"'), bunit.end());
 
-        if (!bunit.empty()) {
-            auto prefix = bunit.at(0, 1);
-            if (prefix == 'M' || prefix == 'm') {
-                casacore::String sub_bunit = bunit.at(1, bunit.size() - 1);
-                NormalizeUnit(sub_bunit);
-                _image->setUnits(prefix + sub_bunit);
-            } else {
-                NormalizeUnit(bunit);
-                _image->setUnits(bunit);
+        if (bunit.size() > 1) {
+            std::string prefix = bunit.substr(0, 1);
+            bool concat(false);
+            if (prefix == "M" || prefix == "m") {
+                bunit = bunit.substr(1, bunit.size() - 1);
+                concat = true;
+            }
+
+            std::string lower_bunit = bunit;
+            std::transform(lower_bunit.begin(), lower_bunit.end(), lower_bunit.begin(), ::tolower);
+            std::unordered_set<std::string> lower_bunits = {"jy/beam", "jypb", "jy beam-1", "jy beam^-1"};
+            std::string norm_bunit("Jy/beam");
+
+            if (lower_bunits.count(lower_bunit) && bunit != norm_bunit) {
+                casacore::Unit new_bunit;
+                if (concat) {
+                    new_bunit.setName(prefix + norm_bunit);
+                } else {
+                    new_bunit.setName(norm_bunit);
+                }
+                _image->setUnits(new_bunit);
             }
         }
     }
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -89,7 +89,24 @@ bool FileLoader::CanOpenFile(std::string& /*error*/) {
 
 void FileLoader::OpenFile(const std::string& hdu) {
     AllocateImage(hdu);
-    NormalizeBunit();
+
+    // Normalize the upper/lower cases of BUNIT string from header
+    if (_image) {
+        casacore::String bunit = _image->units().getName();
+        bunit.erase(std::remove(bunit.begin(), bunit.end(), '"'), bunit.end());
+
+        if (!bunit.empty()) {
+            auto prefix = bunit.at(0, 1);
+            if (prefix == 'M' || prefix == 'm') {
+                casacore::String sub_bunit = bunit.at(1, bunit.size() - 1);
+                NormalizeUnit(sub_bunit);
+                _image->setUnits(prefix + sub_bunit);
+            } else {
+                NormalizeUnit(bunit);
+                _image->setUnits(bunit);
+            }
+        }
+    }
 }
 
 typename FileLoader::ImageRef FileLoader::GetImage(bool check_data_type) {
@@ -873,15 +890,6 @@ double FileLoader::CalculateBeamArea() {
     }
 
     return info.getBeamAreaInPixels(-1, -1, _coord_sys->directionCoordinate());
-}
-
-void FileLoader::NormalizeBunit() {
-    if (_image) {
-        auto bunit = _image->units().getName();
-        bunit.erase(std::remove(bunit.begin(), bunit.end(), '"'), bunit.end());
-        NormalizeUnit(bunit);
-        _image->setUnits(bunit);
-    }
 }
 
 bool FileLoader::GetStokesTypeIndex(const CARTA::PolarizationType& stokes_type, int& stokes_index) {
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -873,11 +873,12 @@ double FileLoader::CalculateBeamArea() {
 void FileLoader::ResetBunit() {
     auto image = GetImage();
     if (image) {
+        std::unordered_map<std::string, std::string> bunits = {{"jy/beam", "Jy/beam"}, {"mjy/beam", "mJy/beam"}};
         std::string bunit = image->units().getName();
         std::string lower_bunit = image->units().getName();
         std::transform(lower_bunit.begin(), lower_bunit.end(), lower_bunit.begin(), [](unsigned char c) { return std::tolower(c); });
-        if (lower_bunit == "jy/beam" && bunit != "Jy/beam") {
-            casacore::Unit new_units("Jy/beam");
+        if (bunits.count(lower_bunit) && bunit != bunits[lower_bunit]) {
+            casacore::Unit new_units(bunits[lower_bunit]);
             image->setUnits(new_units);
         }
     }
--- a/src/ImageData/FileLoader.cc
+++ b/src/ImageData/FileLoader.cc
@@ -877,17 +877,10 @@ double FileLoader::CalculateBeamArea() {
 
 void FileLoader::NormalizeBunit() {
     if (_image) {
-        std::string bunit = _image->units().getName();
+        auto bunit = _image->units().getName();
         bunit.erase(std::remove(bunit.begin(), bunit.end(), '"'), bunit.end());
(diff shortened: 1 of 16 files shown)


kswang1029:
Could you also check the case with mJy/beam?

ie milli-Jy (10^-3)

If I remember correctly some data cube use such unit.

ajm-ska:
@markccchiang I have attempted to test this by modifying the BUNIT in a FITS file.
The only two combinations that I find not to work in the current 'dev' branch are:

1. `Jy/Beam` 
2. `jy/beam`.

In this branch, `Jy/Beam` now works. 
Screenshot 2023-05-23 at 3 15 53 PM

However, `jy/beam` does not seem to work.
Screenshot 2023-05-23 at 3 02 30 PM

I am testing it correctly?
Although I know the log file explains that it should not work: 

```
2023-05-23 07:05:32	INFO		FITS unit "Jy/Beam" does not conform to the FITS standard.
2023-05-23 07:05:32	INFO	+	Correct units are always lower case except when derived from a name.
2023-05-23 07:05:32	INFO	+	Please use "beam" instead of "Beam", "pixel" instead of "Pixel".
2023-05-23 07:06:15	WARN		FITS unit "jy/beam" unknown to CASA - will treat it as non-dimensional.
2023-05-23 07:06:15	WARN	+	 See http://fits.gsfc.nasa.gov/fits_standard.html for a list of valid units.
```

`JY/BEAM` and `Jy/beam` combinations work fine.

confluence:
It looks like the all-lowercase string is being processed by casacore at an earlier stage, before it can be corrected by this code. The normalization would have to be performed before this parsing happens in order to be effective.

markccchiang:
> However, `jy/beam` does not seem to work. 

@ajm-asiaa Looks like the all-lowercase string is being processed by the casacore with additional `"` characters which are not in my consideration. It is fixed now and please recheck it. Thank you!

src/ImageData/FileLoader.cc
```diff
@@ -870,6 +870,23 @@ double FileLoader::CalculateBeamArea() {
     return info.getBeamAreaInPixels(-1, -1, _coord_sys->directionCoordinate());
 }
 
+void FileLoader::ResetBunit() {
```
confluence: Please rename this to `NormalizeBunit`. Apart from that I think this is good to go (after a re-test).

confluence (changes requested):

confluence (approved):

ajm-ska (approved):
I confirm that the four case combination scenarios work now so that the `FluxDensity` can be calculated:

Screenshot 2023-05-25 at 7 33 21 AM
Screenshot 2023-05-25 at 7 33 50 AM
Screenshot 2023-05-25 at 7 34 18 AM
Screenshot 2023-05-25 at 7 34 37 AM

Also, the four case combination scenarios in "milli" units work too:
Screenshot 2023-05-25 at 7 47 32 AM
Screenshot 2023-05-25 at 7 48 00 AM
Screenshot 2023-05-25 at 7 48 35 AM
Screenshot 2023-05-25 at 7 48 56 AM
(You can ignore that the numerical values are not different because all I did was modify the text of the BUNIT value in the header).

kswang1029 (changes requested):
the file info of the complex image cannot be generated as spotted by e2e tests. also it cannot be loaded. @markccchiang any idea?

confluence:
I have noticed that Pam's current PR for fixing various FITS header issues ([#1255](https://github.com/CARTAvis/carta-backend/pull/1255)) contains a more generic function for normalizing multiple header values -- I have asked in that PR if these functions can be consolidated into one, so I'm holding off on merging this until Pam can respond.

confluence (changes requested):
I have just merged #1255. Does fixing _this_ issue actually require changing the value of the units in the image object, or only in the header (as in the unit normalization function in #1255)? If only a change to the header is required, it would make more sense to extend that normalization function to cover these additional BUNIT cases.

(If we do need to change the units in the image object, then I think that we should add a more generic common function call to the top-level `OpenFile` functions, but before we do that I would like to clarify whether we can just change the header.)

kswang1029:
> I have just merged #1255. Does fixing _this_ issue actually require changing the value of the units in the image object, or only in the header (as in the unit normalization function in #1255)? If only a change to the header is required, it would make more sense to extend that normalization function to cover these additional BUNIT cases.
> 
> (If we do need to change the units in the image object, then I think that we should add a more generic common function call to the top-level `OpenFile` functions, but before we do that I would like to clarify whether we can just change the header.)

Just a note that related: we do intensity unit conversion (eg Jy/beam -> K) for region spectral profile at frontend by referring to BUNIT. Region statistics such as flux density is from the backend.

markccchiang:
> I have just merged #1255. Does fixing _this_ issue actually require changing the value of the units in the image object, or only in the header (as in the unit normalization function in #1255)? If only a change to the header is required, it would make more sense to extend that normalization function to cover these additional BUNIT cases.
> 

From my understanding so far, it requires changing the BUNIT string in the image object in order to calculate the `FluxDensity`. May I ask for @pford's opinion on whether this fix is a proper way or not?

src/ImageData/FileLoader.cc
```diff
@@ -870,6 +875,15 @@ double FileLoader::CalculateBeamArea() {
...
 
+void FileLoader::NormalizeBunit() {
+    if (_image) {
+        auto bunit = _image->units().getName();
+        bunit.erase(std::remove(bunit.begin(), bunit.end(), '"'), bunit.end());
+        NormalizeUnit(bunit);
+        _image->setUnits(bunit);
+    }
+}
+
```
confluence: Now that you are only calling this from one place (the base file loader's `OpenFile`) I don't think it's necessary to split this into a function anymore. I suggest that you put this code directly in `OpenFile`.

Did you re-test all the capitalization cases? If I understand @pford's comments correctly, the `NormalizeUnit` function won't correct all of them, so some custom handling may be required.

Overall I think that this is OK now, but I will also wait for @pford to have a look.

pford:
> > I have just merged #1255. Does fixing _this_ issue actually require changing the value of the units in the image object, or only in the header (as in the unit normalization function in #1255)? If only a change to the header is required, it would make more sense to extend that normalization function to cover these additional BUNIT cases.
> 
> From my understanding so far, it requires changing the BUNIT string in the image object in order to calculate the `FluxDensity`. May I ask for @pford's opinion on whether this fix is a proper way or not?

The unit headers were changed only in CARTA::FileInfoExtended header_entries. You are correct, the image object itself must be changed (it is not changed on disk).   The image object is passed to casacore::ImageStatistics and its units (casacore::ImageInterface::units()) are checked to determine whether to calculate flux density.  So setting the normalized unit with _image->setUnits must be done.

src/ImageData/FileLoader.cc
pford: Unless I missed something, `NormalizeUnit` will fix all cases.  The input string is converted to all caps (*before* checking), then checked (all valid units will pass as all caps), then converted `fromFITS` to the proper mixed- or lower-case.

pford (approved):

src/ImageData/FileLoader.cc
confluence: Oh, good. Then this should be fine. I'll ask @ajm to re-test to verify, and then we can merge this.

confluence (approved):

ajm-ska (approved):
I guess it is OK as the`FluxDensity` is calculated in every scenario.

What I notice now is that the Pixel Unit is modified when the File Browser displays it in the File Information:
The variations `JY/BEAM`, `jy/beam`, `Jy/Beam` are changed to `Jy/beam`. 
The "milli" variations `mJY/BEAM`, `mjy/beam`, `mJY/Beam` are all changed to `MJY/beam`.
(It seems a bit odd, but guess the user would always interpret it to be "milli" as it would be improbable to have mega-Jansky values!)

@kswang1029 better check with the e2e tests just incase they are tripped up by the File Information changing the "case" of the Pixel Units.

kswang1029:
> I guess it is OK as the`FluxDensity` is calculated in every scenario.
> 
> What I notice now is that the Pixel Unit is modified when the File Browser displays it in the File Information: The variations `JY/BEAM`, `jy/beam`, `Jy/Beam` are changed to `Jy/beam`. The "milli" variations `mJY/BEAM`, `mjy/beam`, `mJY/Beam` are all changed to `MJY/beam`. (It seems a bit odd, but guess the user would always interpret it to be "milli" as it would be improbable to have mega-Jansky values!)
> 
> @kswang1029 better check with the e2e tests just incase they are tripped up by the File Information changing the "case" of the Pixel Units.

MJy/beam and mJy/beam are different things. We need to fix this.

ajm-ska:
> MJy/beam and mJy/beam are different things. We need to fix this.

@kswang1029 I forgot that I didn't have a real Milli-Jansky image to test. I only modified the BUNIT entry by prefixing an "m". So I'm not sure if that could be part of the problem?

kswang1029:
there are some images that cannot be loaded in CARTA with [62c2ce3](https://github.com/CARTAvis/carta-backend/pull/1269/commits/62c2ce3962be153bdb433d5ed94cbd791959a3ac) as picked up by e2e tests. Checking with @markccchiang via slack.

pford:
We need to fix NormalizeUnit, which is used for the headers, then it can be used here as well. Preserve the leading m/M and capitalize the rest in NormalizeUnit.  Unless there is a unit with milli in the middle?

markccchiang:
> We need to fix NormalizeUnit, which is used for the headers, then it can be used here as well. Preserve the leading m/M and capitalize the rest in NormalizeUnit. Unless there is a unit with milli in the middle?

@pford Thank you for fixing the NormalizeUnit! I have merged your branch and extended it to reset the valid BUNIT.

ajm-ska:
I confirm that the `FluxDensity` is now successfully calculated in all scenarios, and the Pixel Units are converted to either `Jy/beam` or `mJy/beam` in each scenario.

confluence:
I have merged in @pford's PR. Please merge in dev and resolve the merge conflict.

kswang1029 (approved):
Manual tests with some artificial cubes with unit variants are fine as far as I can tell. No regression from e2e tests.

PR #1273, opened 2023-06-05
Check whether the FITS beam table consists of 64-bit floats

### LINKED ISSUE #1166: CARTA crashes when using beam tables with 64-bit floats
An issue was discovered by @amirkazemim where certain FITS cubes he was producing with a custom script would not open on the IDIA CARTA server. CARTA would crash trying to read the FITS file in file browser, after which it would display a "Connection lost" error. Amir reduced the issue to the beam table being written as 64-bit floats (the astropy default), whereas when 32-bit beam tables were used, they opened just fine in CARTA. The cubes which have beam tables as 64-bit floats open fine in ds9, and were converted fine to IDIA's HDF5 schema. Therefore it seems like this might be a CARTA bug.

I was hoping one of the CARTA developers could reproduce the issue, but if not, please let us know and we can produce a cube.

markccchiang:
**Description**

Closes #1166.

Using `cfitsio` to check whether the FITS beam table consists of 64-bit floats. If so, open the file via `CartaFitsImage` object. Otherwise, open the file with `casacore::FITSImage`.

No.

The crash could be avoided when opening the FITS file with beams table consisting of 64-bit floats.

**Checklist**

- [ ] e2e test passing ~/ added corresponding fix~

### DIFF
--- a/src/ImageData/FitsLoader.h
+++ b/src/ImageData/FitsLoader.h
@@ -238,24 +238,36 @@ bool FitsLoader::Is64BitBeamTable(const std::string& filename) {
 
     // Get BEAMS columns data type
     int casesen(CASEINSEN), colnum(0);
+#if defined(__APPLE__)
+    char ttype, tform;
+#else
     char* ttype = (char*)malloc(sizeof(char*));
     char* tform = (char*)malloc(sizeof(char*));
+#endif
     char tunit, nulstr, tdisp;
     long tbcol;
     double scale, zero;
 
     auto close_file = [&]() {
         fits_close_file(fptr, &status);
+#if defined(__APPLE__)
+#else
         free(ttype);
         free(tform);
+#endif
     };
 
     std::vector<std::string> keys = {"BMAJ", "BMIN", "BPA", "CHAN", "POL"};
     for (auto& key : keys) {
         status = 0;
         fits_get_colnum(fptr, casesen, key.data(), &colnum, &status);
+#if defined(__APPLE__)
+        fits_get_acolparms(fptr, colnum, &ttype, &tbcol, &tunit, &tform, &scale, &zero, &nulstr, &tdisp, &status);
+        if (tform == 'K' || tform == 'D') { // 'K' is 64-bit long signed integer, and 'D' is double precision float
+#else
         fits_get_acolparms(fptr, colnum, ttype, &tbcol, &tunit, tform, &scale, &zero, &nulstr, &tdisp, &status);
         if (*tform == 'K' || *tform == 'D') { // 'K' is 64-bit long signed integer, and 'D' is double precision float
+#endif
             close_file();
             spdlog::warn("BEAMS table consists of 64-bit parameters. Convert these values to 32-bit.");
             return true;
--- a/src/ImageData/FitsLoader.h
+++ b/src/ImageData/FitsLoader.h
@@ -32,7 +32,7 @@ private:
 
     int GetNumHeaders(const std::string& filename, int hdu);
     void RemoveHistoryBeam(unsigned int hdu_num);
-    bool Is64BitBeamTable(const std::string& filename);
+    bool Is64BitBeamsTable(const std::string& filename);
 };
 
 FitsLoader::FitsLoader(const std::string& filename, bool is_gz) : FileLoader(filename, "", is_gz) {}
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):
Tested with macOS with apple silicon and intel processors. 👍 no regression observed from e2e tests.

src/ImageData/FitsLoader.h
```diff
@@ -207,6 +213,71 @@ void FitsLoader::RemoveHistoryBeam(unsigned int hdu_num) {
...
+        free(tform);
+#endif
+    };
+
+    std::vector<std::string> keys = {"BMAJ", "BMIN", "BPA", "CHAN", "POL"};
+    for (auto& key : keys) {
+        status = 0;
+        fits_get_colnum(fptr, casesen, key.data(), &colnum, &status);
+#if defined(__APPLE__)
+        fits_get_acolparms(fptr, colnum, &ttype, &tbcol, &tunit, &tform, &scale, &zero, &nulstr, &tdisp, &status);
```
pford: Would fits_get_coltype() be easier, then check if typecode is TDOUBLE?  Rather than malloc/free char* arguments.

int fits_get_coltype / ffgtcl
      (fitsfile *fptr, int colnum, > int *typecode, long *repeat, long *width, int *status)

pford (changes requested):

pford:
CartaFitsImage::ReadBeamsTable reads these columns as TFLOAT.  Shouldn't this be fixed as well, to determine the column type then read the values as TFLOAT or TDOUBLE?  casacore::Quantity expects a double value with a string unit so that is no problem.

markccchiang:
> CartaFitsImage::ReadBeamsTable reads these columns as TFLOAT. Shouldn't this be fixed as well, to determine the column type then read the values as TFLOAT or TDOUBLE? casacore::Quantity expects a double value with a string unit so that is no problem.

It seems in function `fits_read_col` if the reading data type is inconsistent with the writing data type of an array we declared, an error or crash would happen. So maybe it is ok to keep the original way that we convert all beam parameters to the `float` type(?)

pford:
> > CartaFitsImage::ReadBeamsTable reads these columns as TFLOAT. Shouldn't this be fixed as well, to determine the column type then read the values as TFLOAT or TDOUBLE? casacore::Quantity expects a double value with a string unit so that is no problem.
> 
> It seems in function `fits_read_col` if the reading data type is inconsistent with the writing data type of an array we declared, an error or crash would happen. So maybe it is ok to keep the original way that we convert all beam parameters to the `float` type(?)

I am suggesting that we determine the column datatype (fits_get_coltype) then read the data as that type, rather than losing precision converting double to float and back to double.
```
// for bmaj
std::string bmaj_name("BMAJ");
fits_get_colnum(fptr, casesen, bmaj_name.c_str(), &colnum, &status);
fits_get_coltype(fptr, colnum, &datatype, &repeat, &width, &status);

std::vector<float> bmaj_f;
std::vector<double> bmaj_d;

if (datatype == TFLOAT) {
    bmaj_f.resize(nrow);
    fits_read_col(fptr, datatype, colnum, firstrow, firstelem, nrow, fnulval, bmaj_f.data(), &anynul, &status);
} else if (datatype == TDOUBLE) {
    bmaj_d.resize(nrow);
    fits_read_col(fptr, datatype, colnum, firstrow, firstelem, nrow, fnulval, bmaj_d.data(), &anynul, &status);
}
```
Better to put this in a method, something like `GetColumnData(fitsfile* fptr, const std::string& name, int nrow, int& datatype, std::vector<float>& values_f, std::vector<double>& values_d)` where the returned datatype tells you which returned vector to use to set the Quantities (or do not return datatype and just check which vector is not empty).  Then call the method for "BMAJ", "BMIN", "BPA".

markccchiang:
> Better to put this in a method, something like `GetColumnData(fitsfile* fptr, const std::string& name, int nrow, int& datatype, std::vector<float>& values_f, std::vector<double>& values_d)` where the returned datatype tells you which returned vector to use to set the Quantities (or do not return datatype and just check which vector is not empty). Then call the method for "BMAJ", "BMIN", "BPA".

Thank you for the suggestion! I refactored this part of the code in a slightly different way, added a check for the data type, and got a data array accordingly. Hope it is easier to read.

pford (approved):

confluence (approved):

kswang1029 (approved):
looks good. no regression from e2e tests.

PR #1276, opened 2023-06-28
Add an option to cache the whole cube image

markccchiang:
**Description**

#1286 
This is the trial to cache the whole cube image if its memory space is under a specific quota. Users can set the memory quota via the command line `--full_image_cache_size_available` when starting the backend. For example, `./carta_backend --full_image_cache_size_available 1000` means at most 1000 **MB** of memory space for the backend can be used. If the memory of a cube image we open is under 1000 MB, it will cache all the cube image data. The upper limit of full image cache is defined as 90% of users' computer ram capacity. If users set the full image cache that exceeds the upper limit, the backend will reset it to the upper limit. By default, the full image cache is 0 MB. This function will not apply to HDF5 images (only available for CASA or FITS files). The cube image cache will be used in the following calculations:

- [x] image rendering, including animation
- [x] cube histogram calculations
- [x] contours
- [x] vector field calculations
- [x] spatial profiles
- [x] spectral profiles for cursor/point region
- [x] spectral profiles for the other region types, including their statistical calculations
- [x] PV image generator
- [x] above functions for computed stokes data
:construction: image moments (This item will do in a separate PR after this PR is merged)

Create an `ImageCache` struct under the `Frame` class and use it to deal with the channel or cube image data cache, which is obtained from the casacore.

No.

The performances of cached cube images are expected to be better than without the cache (i.e., the current dev branch). Especially when accessing spectral profiles or data in the z-direction.

**Checklist**

- [ ] e2e test passing / ~added corresponding fix~


ajm-ska:
@markccchiang Just to mention, I am aware the macOS12 unit tests are failing here. They consistently failed 4 times in a row when I made Github Actions rerun them. When I let Github Actions re-run the previous commit, it passed the first time. 

The error is: `Error: Timeout of 120000ms hit`.

I have tried building and running it manually with the Address Sanitizer flags as that is what Github Actions has always been doing. The runtime is quite variable: 152.68s, 65.16s, 144.22s.
To make things more confusing, I tried the previous commit (`b7fb274`) and the Unit Tests run times appear to be more consistent but all longer than 2 minutes: 152.19s, 151.23s, 150.13s. So it seems that it was very lucky that the Github Actions somehow passed twice on the previous commit, (and maybe lucky it passed on earlier commits too)!

I figured a better test would be to compare it with the 'dev' branch. I get: 90.50s, 91.35s, 91.51s. 

Therefore, something in this branch has slowed down the Units Tests runtime and they can no longer consistently finish within 2 minutes. It seems to only be an issue on macOS12, which is an original M1 Mac. 

I have not gone so far as to identify which commit in your branch has caused the Unit Tests to slow down because first I guess we need to know if this is considered an actual issue? 
It may show up later in the performance testing.
The one 65.16s runtime looks great, but unfortunately, it is not consistent.

If it is not considered an issue, you can get around the Github Actions failure by increasing the timeout in `.github/workflows/continuous_integration.yml` line 130: `timeout_minutes: 2` to perhaps 3.

markccchiang:
> @markccchiang Just to mention, I am aware the macOS12 unit tests are failing here. They consistently failed 4 times in a row when I made Github Actions rerun them. When I let Github Actions re-run the previous commit, it passed the first time.
> 
> The error is: `Error: Timeout of 120000ms hit`.
> 

@ajm-asiaa Thank you for the reminding and investigating! You are right. This is due to the sample image files I generated are too large. Because I was doing performance tests for the functionalities of cached cube images in unit tests. I will reduce the size of sample image files. So this problem will be solved without changing the timeout setting.

kswang1029:
@markccchiang here are a few findings I observed so far:
1. if we have two images loaded and matched, when we enable multiple-profile plot mode in the spectral profiler to see region spectra, the backend will crash.
2. If we load three images (20, 20, and 90 MB), from the backend log we see (I set `--reserved_memory 4`)

> [2023-07-18 04:28:28.210] [CARTA] [info] 4 GB of reserved memory are available.
> [2023-07-18 04:28:28.213] [CARTA] [info] Cache the whole cube image data.
> [2023-07-18 04:28:28.287] [CARTA] [info] 3 GB of reserved memory are available.
> [2023-07-18 04:28:28.673] [CARTA] [info] Cache the whole cube image data.
> [2023-07-18 04:28:28.681] [CARTA] [info] 2 GB of reserved memory are available.
> [2023-07-18 04:28:28.935] [CARTA] [info] Cache the whole cube image data.
> [2023-07-18 04:28:28.943] [CARTA] [info] 1 GB of reserved memory are available.

This seems not right. The actually RAM usage from the macOS activity monitor shows the right amount.
3. In the backend log, maybe we should show more digits to display the amount of available memory (eg 2.34 GB)
4. When we launch carta_backend with `--reserved_memory` flag, we should also add [info] log too.


Will have more tests soon.

kswang1029:
@markccchiang Overall, this does improve UX significantly in many I/O intensive features. @veggiesaurus please have a test as well if possible. We also need a strategy on RAM management for the controller deployment. Currently it is the backend flag `--reserved_memory` doing the trick. Not sure about the controller side. 🤔

veggiesaurus:
We might also need the ability to control the _total_ amount of memory that cached cubes can take up. For example, if I load 10 images each with 10 GB, would that allow it, and use up 100 GB of memory?

kswang1029:
> We might also need the ability to control the _total_ amount of memory that cached cubes can take up. For example, if I load 10 images each with 10 GB, would that allow it, and use up 100 GB of memory?

@markccchiang could you check if the current setup is for the total amount of caching memory, not for individual image? I also found that we need to improve the [info] logs when loading images that can or cannot be cached. The use case could be I reserved 4 GB as image cache. Then I load the following images in order
1. a.fits 3 GB
2. b.fits 2 GB
3. c.fits 500 MB
In this case, a.fits will be cached. b.fits will not. c.fits will be cached. In any case, we will need to have [info] log to be more  informative on whether or not the image being loaded will be cached or not.

markccchiang:
> @markccchiang could you check if the current setup is for the total amount of caching memory, not for individual image? I also found that we need to improve the [info] logs when loading images that can or cannot be cached. The use case could be I reserved 4 GB as image cache. Then I load the following images in order

The current setup is for the total amount of caching memory (not for individual images). But I need to change the precision of reserved memory from +/- 1 GB to < +/- 1 MB, as does for the [info] log.

kswang1029:
@markccchiang @ajm-asiaa I wonder for the standalone deployment (ie electron app), are we able to expose `--reserved_memory` flag if we start CARTA via an alias in commend line? Like
carta --reserved_memory 4

ajm-ska:
> @markccchiang @ajm-asiaa I wonder for the standalone deployment (ie electron app), are we able to expose `--reserved_memory` flag if we start CARTA via an alias in commend line? Like carta --reserved_memory 4

@kswang1029 Yes. That would work! Our Electron version can now accept any extra flags in the command line and will pass them to the carta_backend. Same with the AppImage versions.

kswang1029:
@markccchiang one minor suggestion: when the image can be cached, we have a [info] log
```
[2023-07-19 02:52:17.605] [CARTA] [info] 4.000 GB of reserved memory are available.
[2023-07-19 02:52:17.607] [CARTA] [info] Cache the whole cube image data.
[2023-07-19 02:52:18.794] [CARTA] [info] 1.178 GB of reserved memory are available.
```

I suggest if the image cannot be cached within the available reserved memory, we also have a [info] log like
```
[2023-07-19 02:52:17.605] [CARTA] [info] 4.000 GB of reserved memory are available.
[2023-07-19 02:52:17.607] [CARTA] [info] Image too large. Not cache the whole cube image data.
[2023-07-19 02:52:18.794] [CARTA] [info] 4.000 GB of reserved memory are available.
```

kswang1029:
@veggiesaurus how do we support this new feature for the controller case? Can the controller read a table of user id and its size of dedicated memory cache, and when the controller starts a carta_backend for the user, `--reserved_memory` can be set accordingly?

kswang1029:
@markccchiang it looks like we can only set an integer for --reserved_memory, can it be a positive float? (need to ignore values <= 0, meaning no reserved memory).

> --reserved_memory 1.5
> [2023-07-19 11:10:37.938] [critical] Argument ‘1.5’ failed to parse

veggiesaurus:
> @veggiesaurus how do we support this new feature for the controller case? Can the controller read a table of user id and its size of dedicated memory cache, and when the controller starts a carta_backend for the user, `--reserved_memory` can be set accordingly?

perhaps it makes sense to make the reserved memory MB and integer 🤔

markccchiang:
> perhaps it makes sense to make the reserved memory MB and integer 🤔

Both integer or float types of reserved memory are ok with me. From users' point of view, they might prefer setting this number with a floating point for flexibility. @kswang1029 what do you think of it?

kswang1029:
> > perhaps it makes sense to make the reserved memory MB and integer 🤔
> 
> Both integer or float types of reserved memory are ok with me. From users' point of view, they might prefer setting this number with a floating point for flexibility. @kswang1029 what do you think of it?

My original intention of using float is that the default unit is GB. So we need float to have a finer control. But if we choose MB, then integer should be just fine (and maybe simpler for parsing?)

kswang1029:
@markccchiang e2e tests are passing, implying basic features should work as usual. Please convert this into a formal PR and please proceed code review while manual tests are in progress. Thanks.

kswang1029 (approved):
this looks good as far as I can tell. no regression from e2e tests (even if whole cube is cached). please proceed code review.

pford (changes requested):
Without tagging specific code lines, I find this new struct confusing and trying to do too much.  A good start would be to limit this only to the whole image.  When loading data, determine if the image data size meets the memory requirements; then set a Frame flag and fill either the ImageCache data map or the Frame data cache for the current channel/stokes only.  When the Frame needs data, it will check this flag to see which cache to access.

Then ImageCache will only store data cubes, which should simplify the data map key.  It looks like the image stokes cubes are loaded with a key >= 0, and computed stokes is loaded with key -1 so there is at most one computed stokes loaded; is this correct?  If so, the cache size calculation (CubeImageSize) needs to account for this additional cube, or check the memory again before storing the computed cube? For clarity, use COMPUTED_STOKES rather than CURRENT_CHANNEL_STOKES for the key, and record its type as computed_stokes rather than cur_stokes.  Then you can check whether the computed stokes type is cached when needed (GetValue, GetImageCacheData, spectral data) rather than recomputing it.

To me, "cube" refers to the 3D data for a single stokes (width * height * depth) but an image with a stokes axis is 4D and can have  up to 4 cubes; so I find the naming using "cube image" (cube_image_cache, CubeImageSize) confusing.

markccchiang:
> Without tagging specific code lines, I find this new struct confusing and trying to do too much. A good start would be to limit this only to the whole image. When loading data, determine if the image data size meets the memory requirements; then set a Frame flag and fill either the ImageCache data map or the Frame data cache for the current channel/stokes only. When the Frame needs data, it will check this flag to see which cache to access.
> 
> Then ImageCache will only store data cubes, which should simplify the data map key. It looks like the image stokes cubes are loaded with a key >= 0, and computed stokes is loaded with key -1 so there is at most one computed stokes loaded; is this correct? If so, the cache size calculation (CubeImageSize) needs to account for this additional cube, or check the memory again before storing the computed cube? For clarity, use COMPUTED_STOKES rather than CURRENT_CHANNEL_STOKES for the key, and record its type as computed_stokes rather than cur_stokes. Then you can check whether the computed stokes type is cached when needed (GetValue, GetImageCacheData, spectral data) rather than recomputing it.
> 
> To me, "cube" refers to the 3D data for a single stokes (width * height * depth) but an image with a stokes axis is 4D and can have up to 4 cubes; so I find the naming using "cube image" (cube_image_cache, CubeImageSize) confusing.

I see. I will change this PR to draft and refactor the codes. And will reopen it after the refactoring work is done.

kswang1029 (approved):
re-tested and this should be fine as far as I can tell.

src/Cache/ImageCacheCalculator.h
```diff
@@ -0,0 +1,34 @@
...
+#include "Util/Image.h"
+
+#include <casacore/casa/Arrays/IPosition.h>
+#include <casacore/lattices/Lattices/ArrayLattice.h>
+
+#include <vector>
+
+namespace carta {
+
+class ImageCacheCalculator {
```
confluence: All of the functions on this class take a cache object as the first parameter and perform operations on that object's state. This strongly indicates that this should not be a separate class -- all of this functionality should be encapsulated in the cache object itself (which should be a class rather than a struct).

I see that this was the original design, and you have recently changed it -- it's not clear to me why this change was made.

src/Frame/Frame.h
```diff
@@ -220,6 +222,11 @@ class Frame {
         std::vector<float>& data, int& downsampled_width, int& downsampled_height, int z, int stokes, CARTA::ImageBounds& bounds, int mip);
     bool CalculateVectorField(const std::function<void(CARTA::VectorOverlayTileData&)>& callback);
 
+    float ReservedMemory() const; // MB
```
confluence: Unclear naming. This should be called something like "RequiredMemory", or perhaps "MemorySize".

src/Main/ProgramSettings.cc
```diff
@@ -192,7 +192,8 @@ void ProgramSettings::ApplyCommandLineSettings(int argc, char** argv) {
         ("enable_scripting", "enable HTTP scripting interface", cxxopts::value<bool>())
         ("files", "files to load", cxxopts::value<std::vector<string>>(positional_arguments))
         ("no_user_config", "ignore user configuration file", cxxopts::value<bool>())
-        ("no_system_config", "ignore system configuration file", cxxopts::value<bool>());
+        ("no_system_config", "ignore system configuration file", cxxopts::value<bool>())
+        ("m,reserved_memory", "reserved memory per session", cxxopts::value<int>(), "<MB>");
```
confluence: Both the name and the description should be more specific to the actual functionality. Perhaps something like "full_image_cache", and "maximum amount of memory (in MB) to use for full image cache".

("reserve" implies to me that the full amount will be preallocated.)

confluence:
I'm in the process of making a more detailed code review, but would it not make more sense for the memory limit to be global, not per-session, and tracked in the session manager rather than the session? Otherwise the user may open multiple browser sessions and exceed the limit (since it is not global).

confluence:
In future, perhaps we should unify the APIs of the per-channel image cache, the full image cache, and the tile cache in a single interface -- but that is probably outside the scope of this PR.

src/Main/ProgramSettings.cc
```diff
@@ -268,6 +269,10 @@ saving regions or generated images).
     
 'no_user_config' and 'no_system_config' may be used to ignore the user and 
 global configuration files, respectively.
+
+'reserved_memory' defines the amount of memory per session can be used to cache
+the image data. By default, it is 0, in the unit of MB. If the defined amount of
+memory is above a certain cube image data, then the backend will cache it all.
```
confluence: This should say something like:
```
'full_image_cache' allows small images to be cached fully in memory up to a global size limit, which can improve the performance of various tasks. This option defines the total amount of memory to reserve for this cache, in MB (by default it is set to zero, and the cache is disabled). When an image is opened, it will be cached in full if it fits within the remaining available cache memory, otherwise it will be cached in smaller portions as needed.
```

The exact text is dependent on the resolution of the global limit issue which I raised in a separate comment.

src/Frame/Frame.cc
```diff
@@ -2403,4 +2424,102 @@ bool Frame::DoVectorFieldCalculation(const std::function<void(CARTA::VectorOverl
...
+    return false;
+}
+
+float Frame::GetValue(int x, int y, int z, int stokes) {
+    if (_cube_image_cache_valid) {
+        return ImageCacheCalculator::GetValue(_cube_image_cache, x, y, z, stokes, _width, _height);
+    }
+    return _channel_image_cache[_width * y + x];
+}
+
```
confluence: These functions have confusing names and signatures. They're all called `Get*`, but the first two load a value into an output parameter only if the cube cache is in use and return a boolean success flag, and the last always returns a value. The names don't indicate that the functions have anything to do with caches.

I would suggest `LoadCachedPointSpectralData` and `LoadCachedRegionSpectralData` for the first two functions. The third one can keep its current name.

src/Frame/Frame.cc
```diff
@@ -2403,4 +2424,102 @@ bool Frame::DoVectorFieldCalculation(const std::function<void(CARTA::VectorOverl
...
+    GetStokesTypeIndex("I", _cube_image_cache.stokes_i, mute_err_msg);
+    GetStokesTypeIndex("Q", _cube_image_cache.stokes_q, mute_err_msg);
+    GetStokesTypeIndex("U", _cube_image_cache.stokes_u, mute_err_msg);
+    GetStokesTypeIndex("V", _cube_image_cache.stokes_v, mute_err_msg);
+
+    // Get beam area
+    _cube_image_cache.beam_area = _loader->CalculateBeamArea();
+}
+
+float* Frame::GetImageCache(int z, int stokes) {
```
confluence: This function name is confusing. I suggest `GetImageData`.

src/Frame/Frame.cc
```diff
@@ -2403,4 +2424,102 @@ bool Frame::DoVectorFieldCalculation(const std::function<void(CARTA::VectorOverl
...
+        stokes = _stokes_index;
+    }
+
+    if (_cube_image_cache_valid) {
+        return _cube_image_cache.GetChannelImageCache(z, stokes, _width, _height);
+    }
+    return _channel_image_cache.get();
+}
+
+bool Frame::LoadImageCacheData(int stokes) {
```
confluence: This function name is confusing. I was going to suggest `FillImageCache`, but that already exists and calls this function -- it's unclear to me why these are two separate functions. Can this functionality be consolidated into the existing `FillImageCache`?

confluence (commented):
I have made a few specific comments. My general suggestions would be:

1. Merge `ImageCacheCalculator` back into the cache class
2. Create a class for the per-channel cache which has the same API (inherits from a shared interface)
3. Move most of the cache logic out of Frame itself and into these classes -- each cache object should know how to load data from the file, how to return a value, how to return a block of data, etc..
4. Whenever the frame needs to access data, it should first get the appropriate cache (e.g. through a `GetDataCache` function on the frame, which would contain all the logic required to pick the correct cache -- if we use different caches in different situations, we could use optional parameters to this function to flag situations when a different cache should be used)...
5. ...and then the API functions should be called on this cache object. At this point Frame shouldn't care which cache is being used. The cache class should internally handle loading new data from disk if it's necessary, etc..

This model could later be extended to include the tile cache, which currently has a lot of special handling, but that can be done later. For now it would be good to unify the APIs of the channel cache and cube cache, and simplify how they are used.

pford:
Some additional suggestions:

- Set the image data shape parameters in the constructor rather than passing height, width, depth to functions.
- If there is not enough memory to load all of the stokes cubes, would it help to load the current stokes cube into the cube cache, similar to the channel cache?
- Move channel/stokes checks into the classes, rather than external caller having to validate before requesting data (DataExist, ChannelImageCacheValid, ValidateChannelImageCache, InvalidateChannelImageCache -- these are too specific to the cache type for an interface anyway).  If this check is missed, the backend could crash or return invalid data.  The cache should know which channel/stokes are loaded and return true/value/pointer (ok) or false/NAN/nullptr (fail), then the caller can use the return value to decide the next action.
- CubeImageCache::GetValue should check if _computed_stokes_channel_data can be used before calculating these values.  Have you checked the performance of calling the GetValue function for each element in a for loop vs. retrieving the value(s) by index (and computing stokes if necessary) inline?
- For readability and clarity, use parentheses in calculations, for example `size_t idx = (width * height * z) + (width * y) + x;`

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,64 @@
...
+*/
+
+#include "ImageCache.h"
+
+#include "Logger/Logger.h"
+#include "Util/Stokes.h"
+
+namespace carta {
+
+ImageCache::ImageCache(ImageCacheType type)
```
confluence: This constructor should not have an explicit type parameter, and should not be called directly. This parent class should have a static factory method which performs all the logic to decide what kind of cache should be created, and returns the appropriate subclass object.

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,64 @@
...
+}
+
+float ImageCache::GetValue(int x, int y, int z, int stokes, size_t width, size_t height) {
+    return FLOAT_NAN;
+}
+
+bool ImageCache::LoadCachedPointSpectralData(
+    std::vector<float>& profile, int stokes, PointXy point, size_t width, size_t height, size_t depth) {
+    return false;
+}
```
confluence: All these functions which have no default implementations and have to be implemented in subclasses should be pure functions instead of having empty implementations like this.

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+bool ImageCache::DataExist(int stokes) const {
+    return false;
+}
+
+void ImageCache::ValidateChannelImageCache() {}
+
+void ImageCache::InvalidateChannelImageCache() {}
```
confluence: These functions are specific to the image cache and should only be in the image cache. External code should not be accessing them directly, because no components outside the cache class should have to be aware of which cache implementation is being used.

src/Frame/Frame.cc
```diff
@@ -83,8 +84,42 @@ Frame::Frame(uint32_t session_id, std::shared_ptr<FileLoader> loader, const std:
...
+        } else if (FULL_IMAGE_CACHE > 0) {
+            spdlog::info("Image too large ({:.0f} MB). Not cache the whole image data.", MemorySizeOfWholeImage());
+        }
+    }
+
+    if (!_image_cache) {
+        // By default, the image cache is per channel, if it is not assigned as for the whole image data
+        _image_cache = std::make_unique<ChannelImageCache>();
+    }
+
```
confluence: * None of this logic should be in Frame. This should all be in a static factory function in the parent cache class (interface). The cache class should have an internal reference to the loader (which can be passed in from here).

* `FULL_IMAGE_CACHE` should be renamed to something like `FULL_IMAGE_CACHE_SIZE_AVAILABLE`, and it should be a static property on the parent cache class.

* Both incrementing and decrementing this variable should be protected by a mutex.

src/Frame/Frame.cc
```diff
@@ -364,38 +399,61 @@ bool Frame::SetCursor(float x, float y) {
...
+        // Fill cube image cache
+        if (!_image_cache->DataExist(stokes) && !IsComputedStokes(stokes)) {
+            Timer t;
+            if (!load_image_data_to_cache(ALL_Z)) {
+                return false;
+            }
+            auto dt = t.Elapsed();
+            spdlog::performance("Load {}x{}x{} image to cache in {:.3f} ms at {:.3f} MPix/s", _width, _height, _depth, dt.ms(),
+                (float)(_width * _height * _depth) / dt.us());
+        }
```
confluence: Again, all of this logic (this entire function) should be encapsulated in the cache classes. Each cache type should be responsible for its own logic (so for example you wouldn't have a check to see if it's a cube cache -- the cube cache code would only be in the cube cache class).

src/Frame/Frame.cc
```diff
@@ -574,7 +631,7 @@ bool Frame::GetRasterTileData(std::shared_ptr<std::vector<float>>& tile_data_ptr
     if (mip > 1 && !IsComputedStokes(_stokes_index)) {
         // Try to load downsampled data from the image file
         loaded_data = _loader->GetDownsampledRasterData(tile_data, _z_index, _stokes_index, bounds, mip, _image_mutex);
-    } else if (!_image_cache_valid && _loader->UseTileCache()) {
+    } else if (!_image_cache->ChannelImageCacheValid() && _loader->UseTileCache()) {
```
confluence: This function should be renamed to something like `CachedChannelDataAvailable`.

confluence: There's already an `ImageCacheAvailable` which can be reused for this -- but it should be moved into the cache classes, and it should have default parameter values so that it checks availability of the current z and stokes if it is called with no arguments.

src/Frame/Frame.cc
```diff
@@ -2434,4 +2482,53 @@ bool Frame::DoVectorFieldCalculation(const std::function<void(CARTA::VectorOverl
...
+float Frame::GetValue(int x, int y, int z, int stokes) {
+    return ImageCacheAvailable(z, stokes) ? _image_cache->GetValue(x, y, z, stokes, _width, _height) : FLOAT_NAN;
+}
+
+bool Frame::ImageCacheAvailable(int z, int stokes) const {
+    return (_image_cache->Type() == ImageCacheType::Channel && z == _z_index && stokes == _stokes_index &&
+               _image_cache->ChannelImageCacheValid()) ||
+           _image_cache->Type() == ImageCacheType::Cube;
+}
+
```
confluence: None of these functions should be in `Frame`. All this logic should be inside the cache classes. Instead of these explicit type checks, each cache subclass should have its own specific logic.

src/Main/Main.cc
```diff
@@ -24,6 +24,15 @@
...
 
+float FULL_IMAGE_CACHE = 0; // MB
+
+int GetTotalSystemMemory() {
+    long pages = sysconf(_SC_PHYS_PAGES);
+    long page_size = sysconf(_SC_PAGE_SIZE);
+    unsigned long long memory_in_bytes = pages * page_size; // Bytes
+    return (int)(memory_in_bytes / 1e6);                    // MB
+}
+
```
confluence: This variable should be a static property on the cache parent class. This function should either be moved to the same file as the cache, or moved to an appropriate location in `Util` (maybe a new `System` or `Memory` file?).

src/Main/Main.cc
```diff
@@ -161,10 +170,23 @@ int main(int argc, char* argv[]) {
...
+                if (settings.full_image_cache > memory_upper_limit) {
+                    spdlog::warn("Full image cache {} MB is greater than the system upper limit {} MB, reset it to {} MB.",
+                        settings.full_image_cache, memory_upper_limit, memory_upper_limit);
+                    settings.full_image_cache = memory_upper_limit;
+                }
+
+                // Set the global variable for full image cache
+                FULL_IMAGE_CACHE = settings.full_image_cache;
+                start_info += fmt::format(" Total amount of full image cache {} MB.", FULL_IMAGE_CACHE);
+            }
```
confluence: This should be moved to a static function in the parent cache class.

src/Session/Session.cc
```diff
@@ -638,13 +644,15 @@ void Session::DeleteFrame(int file_id) {
     if (file_id == ALL_FILES) {
         for (auto& frame : _frames) {
             frame.second->WaitForTaskCancellation(); // call to stop Frame's jobs and wait for jobs finished
-            frame.second.reset();                    // delete Frame
+            FULL_IMAGE_CACHE += frame.second->MemorySize();
```
confluence: This should be done by a static function on the parent cache class, and should be protected by a mutex.

src/Session/Session.cc
```diff
@@ -638,13 +644,15 @@ void Session::DeleteFrame(int file_id) {
...
+            frame.second.reset(); // delete Frame
         }
         _frames.clear();
         _image_channel_mutexes.clear();
         _image_channel_task_active.clear();
     } else if (_frames.count(file_id)) {
         _frames[file_id]->WaitForTaskCancellation(); // call to stop Frame's jobs and wait for jobs finished
+        FULL_IMAGE_CACHE += _frames[file_id]->MemorySize();
```
confluence: As above.

confluence (changes requested):
All of the cache logic should be encapsulated inside the cache classes. The parent cache class should have static functions and properties which manage the global information (how much cache space is available) and decide what type of cache should be created when a frame requests a cache object.

It should be unnecessary ever to check for an explicit cache type once the cache object has been created. Functionality which differs between the two subclasses (channel and cube) should be contained in overridden functions.

The external calling code (e.g. inside Frame) shouldn't interact with the cache class internals, and shouldn't even be aware of the cache's type -- it should only check if channel data is available, request channel data, etc., using generic functions. The channel cache class should be responsible for loading the data for a new channel if this is necessary.

Rearranging the code like this will allow us to consolidate all the cache-related logic in one appropriate place, and to simplify Frame a little.

markccchiang:
I have renamed the command line parameter `full_image_cache_size_available` for the memory space available to cache the full image data. For example, `./carta_backend --full_image_cache_size_available 1000` means at most 1000 MB of memory space for the backend can be used to cache the full image data. A new struct `ImageState` is created under the Frame to store the image parameters, including a cube image size, axes indices, current channel, and current stokes. A new class `LoaderHelper` is created under the Frame, which stores the shared pointer of the `FileLoader` subclass and `ImageState`. It provides general or customized methods for `ImageCache` subclasses. The `ImageCache` subclasses handle the logic of loading or updating image data through methods from the `LoaderHelper`. They also handle the calculations of cached image data for spatial or spectral profiles. A new ImageCache subclass `StokesImageCache` is created to cache a single stokes cube image data if the memory quota is not enough to cache all stokes cubes, or if an image has only one stoke. I am still checking the code. When finished, I will change the label of this PR from `awaiting code changes` to `awaiting code review`.

confluence:
> I have renamed the command line parameter `full_image_cache_size_available` for the memory space available to cache the full image data.

I will wait for all the changes before reviewing, but please change this to `full_image_cache_size` -- "available" implies the amount of memory that is *left* from the starting size, which is why it's appropriate in the name of the internal variable, but the preference exposed to the user should not include this word, because it represents the full starting size.

markccchiang:
I have renamed the command line parameter `full_image_cache_size` as the memory quota for caching the cube image data. I also renamed the ImageCache subclasses `FullImageCache` and `CubeImageCache`. `FullImageCache` is used to cache the full image data with multiple stokes, and `CubeImageCache` is used to cache the full image data with single stokes. The major difference between them is the type of data container. One uses a map of the float pointer, and another uses a float pointer. The latter is supposed to be a little bit more efficient than using a map of the float pointer container. @kswang1029 suggested not to cache a single stokes cube data for a multiple stokes image. Because it would confuse users when performing multiple stokes analysis.

kswang1029:
UPDATE(2023/12/12): with [11cac6f](https://github.com/CARTAvis/carta-backend/pull/1276/commits/11cac6fe64438747105aabbfedcd4abaa3668d4e) I re-run e2e tests with different cache sizes (0, 10, 100, 1000, 2000, 4000, 8000 MB) and did not see regressions. With manual tests, I tried all region analytics and did not find regressions as far as I could tell.

>       --no_user_config          ignore user configuration file
>       --no_system_config        ignore system configuration file
>       --full_image_cache_size <MB>
>                                 maximum amount of memory (in MB)
> 

We might need to reword this. \<MB> is not needed as the description part should be enough. We may have the description like "maximum amount of memory for full image caching in MB"

@pford @confluence please proceed code review. @acdo2002 please help on manual testing.

kswang1029:
UPDATE(20231220): now tested the case of Stokes hypercube. The hypercube is cached and computed polarization quantities (such as polarization intensity) are not cached and are still computed on-the-fly as usual. 👍

src/Cache/ImageCache.h
```diff
@@ -0,0 +1,71 @@
...
+#include <casacore/casa/Arrays/IPosition.h>
+#include <casacore/lattices/Lattices/ArrayLattice.h>
+
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+// Global variable
+extern float FULL_IMAGE_CACHE_SIZE_AVAILABLE; // MB
+extern std::mutex FULL_IMAGE_CACHE_SIZE_AVAILABLE_MUTEX;
```
confluence: There is no need for these to be global. They should only ever be accessed directly from the `ImageCache` class, so it makes sense for them to be static properties on that class.

src/Frame/ImageState.h
confluence, line 1: It's unclear to me what the advantage is of adding this struct. Elsewhere in the code, these basic properties are being accessed on the Frame class (e.g. in RegionHandler and in Session). Adding another layer here and using it in a single place is both adding complexity and is inconsistent. The cache could store a shared pointer to its frame and use it to access these properties on the frame object.

src/Frame/LoaderHelper.h
confluence, line 1: This class is also an unnecessary additional layer. Its functions are either aliases for properties on the ImageState object (which I suggest that you also eliminate, and use the Frame class directly instead), or should be moved directly to the cache (or perhaps the frame or the loader if they're generic enough -- but moving all of this into the cache object to start with probably makes sense). For example, the code in the 'FillXXXCache' functions should be moved into the `UpdateChannelImageCache` functions.

confluence (changes requested):
This is looking a lot better, but I strongly suggest eliminating both the `LoaderHelper` class (and moving its functionality directly into the cache classes) and the `ImageState` struct (and accessing its functionality directly on the Frame class, which is consistent with other parts of the code).

kswang1029:
UPDATE(2024/02/19): based on [16d57d6](https://github.com/CARTAvis/carta-backend/pull/1276/commits/16d57d6bfc19db8cc9b91ba1924d4560dd0041e6), all e2e tests are passing (tried 8 GB cache)

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,337 @@
...
+std::unique_ptr<ImageCache> ImageCache::GetImageCache(Frame* frame, std::shared_ptr<FileLoader> loader, std::mutex& image_mutex) {
+    if (!(loader->UseTileCache() && loader->HasMip(2))) {
+        auto width = frame->Width();
+        auto height = frame->Height();
+        auto depth = frame->Depth();
+        auto num_stokes = frame->NumStokes();
+
+        if (depth > 1) {
+            auto full_image_memory_size = ImageCache::ImageMemorySize(width, height, depth, num_stokes);
+            if (_full_image_cache_size_available >= full_image_memory_size) {
```
confluence, line 37: The available cache size mutex should be locked before this check is made. The lock should be removed from the `FullImageCache` and `CubeImageCache` constructors and added here.

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,337 @@
...
+    std::unique_lock<std::mutex> ulock(_image_mutex);
+    bool data_ok = _loader->GetSlice(tmp, stokes_slicer);
+    _loader->CloseImageIfUpdated();
+    ulock.unlock();
+    return data_ok;
+}
+
+double ImageCache::GetBeamArea() {
+    return _loader->CalculateBeamArea();
+}
```
confluence: Is there any reason to add this wrapper function? This is used in the constructor initialiser list, and in both places I think the loader function could be called.

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+        spdlog::error("Unknown point spatial profile config: {}", config);
+    }
+}
+
+bool ImageCache::IsValid() const {
+    return _valid;
+}
+
+void ImageCache::AssignFullImageCacheSizeAvailable(int& full_image_cache_size_available, std::string& msg) {
```
confluence: Since this function is only used once to set the initial cache size, I suggest renaming it to `SetFullImageCacheSize`.

src/Cache/ChannelImageCache.h
```diff
@@ -0,0 +1,48 @@
...
+#include <casacore/casa/Arrays/IPosition.h>
+#include <casacore/lattices/Lattices/ArrayLattice.h>
+
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+namespace carta {
+
+class ChannelImageCache : public ImageCache {
```
confluence: I would suggest renaming this class to `ChannelCache` (and also updating the function names).

src/Cache/CubeImageCache.h
```diff
@@ -0,0 +1,49 @@
...
+#include <casacore/casa/Arrays/IPosition.h>
+#include <casacore/lattices/Lattices/ArrayLattice.h>
+
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+namespace carta {
+
+class CubeImageCache : public ImageCache {
```
confluence: I would suggest renaming this class to `StokesCache` for clarity ("cube" is ambiguous; it could be used to refer to a whole 4D image).

src/Cache/ChannelImageCache.cc
```diff
@@ -0,0 +1,78 @@
...
+
+    _channel_image_cache_valid = true;
+    return true;
+}
+
+void ChannelImageCache::SetImageChannels(int z, int stokes) {
+    _channel_image_cache_valid = false;
+    _frame->SetCurrentZ(z);
+    _frame->SetCurrentStokes(stokes);
+}
```
confluence: This function should not be updating the channel and stokes on the frame object. The calling function in `Frame` should update those values directly, as before. This function should be called something like `UpdateValidity`, and *only* update the validity flag on the cache object (if appropriate).

src/Frame/Frame.cc
```diff
@@ -327,25 +290,24 @@ bool Frame::SetImageChannels(int new_z, int new_stokes, std::string& message) {
...
     } else {
-        if ((new_z != _z_index) || (new_stokes != _stokes_index)) {
-            bool z_ok(CheckZ(new_z));
-            bool stokes_ok(CheckStokes(new_stokes));
+        if (ZStokesChanged(new_z, new_stokes)) {
+            bool z_ok(ValidZ(new_z));
+            bool stokes_ok(ValidStokes(new_stokes));
             if (z_ok && stokes_ok) {
-                _z_index = new_z;
-                _stokes_index = new_stokes;
```
confluence: This behaviour should be restored.

src/Frame/Frame.cc
```diff
@@ -327,25 +290,24 @@ bool Frame::SetImageChannels(int new_z, int new_stokes, std::string& message) {
...
-
-                // invalidate the image cache
-                InvalidateImageCache();
+                bool write_lock(true);
+                queuing_rw_mutex_scoped cache_lock(&_cache_mutex, write_lock);
+                _image_cache->SetImageChannels(new_z, new_stokes);
```
confluence: This function should be renamed (as described above), so this call should just be updating the cache validity.

src/Cache/ImageCache.cc
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    // Get image size
+    _width = _frame->Width();
+    _height = _frame->Height();
+    _depth = _frame->Depth();
+    _num_stokes = _frame->NumStokes();
+}
+
+StokesSlicer ImageCache::GetImageSlicer(const AxisRange& x_range, const AxisRange& y_range, const AxisRange& z_range, int stokes) {
```
confluence: I suggest moving this function back to the `Frame` class.

src/Frame/Frame.cc
```diff
@@ -365,37 +327,9 @@ bool Frame::SetCursor(float x, float y) {
 }
 
 bool Frame::FillImageCache() {
```
confluence: I suggest renaming this to `UpdateChannelCache`.

src/Frame/Frame.h
```diff
@@ -296,14 +321,10 @@ class Frame {
...
-    //    std::vector<float> _image_cache; // image data for current z, stokes
-    long long int _image_cache_size;
-    std::unique_ptr<float[]> _image_cache;
-    bool _image_cache_valid;       // cached image data is valid for current z and stokes
-    queuing_rw_mutex _cache_mutex; // allow concurrent reads but lock for write
-    std::mutex _image_mutex;       // only one disk access at a time
-    bool _cache_loaded;            // channel cache is set
-    TileCache _tile_cache;         // cache for full-resolution image tiles
+    std::unique_ptr<ImageCache> _image_cache; // Image cache for current z and stokes, or for the whole image
+    queuing_rw_mutex _cache_mutex;            // allow concurrent reads but lock for write
```
confluence: This mutex should be a property on the cache -- the cache functions should access it internally; the frame should not access it directly before calling the functions.

If there are any cache functions which sometimes need to lock the cache for writing and sometimes do not, a write lock flag should be passed to them as a parameter. But I don't think that there are any cases like this.

src/Cache/ImageCache.h
```diff
@@ -0,0 +1,78 @@
...
+    bool GetStokesTypeIndex(const string& coordinate, int& stokes_index, bool mute_err_msg);
+    bool TileCacheAvailable();
+
+    virtual float* GetChannelData(int z, int stokes) = 0;
+    virtual float GetValue(int x, int y, int z, int stokes) const = 0;
+
+    virtual bool LoadCachedPointSpectralData(std::vector<float>& profile, int stokes, PointXy point) = 0;
+    virtual bool LoadCachedRegionSpectralData(const AxisRange& z_range, int stokes, const casacore::ArrayLattice<casacore::Bool>& mask,
+        const casacore::IPosition& origin, std::map<CARTA::StatsType, std::vector<double>>& profiles) = 0;
+    virtual bool CachedChannelDataAvailable(int z, int stokes) const = 0;
```
confluence: I suggest removing `Cached` from these names.

src/Frame/Frame.h
```diff
@@ -17,6 +17,8 @@
 #include <shared_mutex>
 #include <unordered_map>
 
+#include "Cache/ChannelImageCache.h"
+#include "Cache/FullImageCache.h"
```
confluence: Are these imports still required? I think the `ImageCache` header should be sufficient.

src/Frame/Frame.h
```diff
@@ -266,6 +279,25 @@ class Frame {
     // For vector field calculation
     bool DoVectorFieldCalculation(const std::function<void(CARTA::VectorOverlayTileData&)>& callback);
 
+    // For image cache
+    float* GetImageData(int z = CURRENT_Z, int stokes = CURRENT_STOKES);
+    bool ImageCacheAvailable(int z = CURRENT_Z, int stokes = CURRENT_STOKES) const;
```
confluence: Instead of defaulting to `CURRENT_Z` and `CURRENT_STOKES`, which requires converting these constants to `_z` and `_stokes` whenever the function is called without parameters, I suggest explicitly defining two versions of each of these functions: one with no parameters, which calls the other with `_z` and `_stokes`, and one with two required parameters.

I believe that in all the cases where these functions are called with parameters, they are always called with real z and stokes values, never with `CURRENT_Z` or `CURRENT_STOKES` -- so these conversions should not be necessary at all and can be removed once this change is made.

src/Cache/ChannelImageCache.cc
```diff
@@ -0,0 +1,78 @@
...
+
+bool ChannelImageCache::LoadCachedRegionSpectralData(const AxisRange& z_range, int stokes,
+    const casacore::ArrayLattice<casacore::Bool>& mask, const casacore::IPosition& origin,
+    std::map<CARTA::StatsType, std::vector<double>>& profiles) {
+    return false;
+}
+
+float ChannelImageCache::GetValue(int x, int y, int z, int stokes) const {
+    return CachedChannelDataAvailable(z, stokes) ? _channel_data[(_width * y) + x] : FLOAT_NAN;
+}
```
confluence: The high-level Frame functions that use the cache already call `CachedChannelDataAvailable` before fetching data from the cache. These internal cache functions (`GetValue`, `GetChannelData`) should not be making the same check _again_. This is especially a problem for this function, which is called inside a loop in the main cache class.

src/Cache/CubeImageCache.cc
```diff
@@ -0,0 +1,113 @@
...
+}
+
+float* CubeImageCache::GetChannelData(int z, int stokes) {
+    return CachedChannelDataAvailable(z, stokes) ? _stokes_data.get() + (_width * _height * z) : nullptr;
+}
+
+float CubeImageCache::GetValue(int x, int y, int z, int stokes) const {
+    size_t idx = (_width * _height * z) + (_width * y) + x;
+    return CachedChannelDataAvailable(z, stokes) ? _stokes_data[idx] : FLOAT_NAN;
+}
```
confluence: It should also not be necessary for these implementations to call `CachedChannelDataAvailable`. (`GetValue` is called by the spectral data functions below, which are called without a check, but they both perform the check internally.)

src/Frame/Frame.h
```diff
@@ -99,14 +100,23 @@ class Frame {
...
-    int CurrentStokes();
-    int SpectralAxis();
-    int StokesAxis();
+    casacore::IPosition OriginalImageShape() const; // Image shape from the original file
+    void SetCurrentZ(int z);
+    void SetCurrentStokes(int stokes);
+    void CheckCurrentZ(int& z) const;
+    void CheckCurrentStokes(int& stokes) const;
+    bool IsCurrentChannel(int z, int stokes) const;
+    bool IsCurrentStokes(int stokes) const;
```
confluence: As discussed in the comment about eliminating the `CURRENT_...` default parameters, the two `Check...` functions should not be necessary, and should be eliminated. After that, the two `IsCurrent...` functions will also no longer be necessary -- the cache functions which use them can directly refer to `CurrentZ()` and `CurrentStokes()` instead.

confluence (changes requested):
A few attributes should be moved, some redundant checks should be eliminated, and some classes and functions should be renamed for clarity, but I think this is close to being ready for merging. There are also some merge conflicts which need to be resolved.

src/Cache/StokesCache.h
```diff
@@ -0,0 +1,49 @@
...
+    bool LoadPointSpectralData(std::vector<float>& profile, int stokes, PointXy point) override;
+    bool LoadRegionSpectralData(const AxisRange& z_range, int stokes, const casacore::ArrayLattice<casacore::Bool>& mask,
+        const casacore::IPosition& origin, std::map<CARTA::StatsType, std::vector<double>>& profiles) override;
+    bool ChannelDataAvailable(int z, int stokes) const override;
+
+    bool UpdateChannelCache(int z, int stokes) override;
+    void UpdateValidity(int stokes) override;
+
+private:
+    bool FillCubeImageCache(std::unique_ptr<float[]>& stokes_data, int stokes);
```
confluence: This should be renamed to `FillStokesCache`.

test/TestCubeImageCache.cc
```diff
@@ -0,0 +1,772 @@
+/* This file is part of the CARTA Image Viewer: https://github.com/CARTAvis/carta-backend
```
confluence: This test file should be renamed (probably to `TestImageCaches`), and the class and test names should be updated to match the new names in the code.

src/Frame/Frame.h
```diff
@@ -99,15 +99,18 @@ class Frame {
...
+    size_t Width() const;                           // length of x axis
+    size_t Height() const;                          // length of y axis
+    size_t Depth() const;                           // length of z axis
+    size_t NumStokes() const;                       // if no stokes axis, number of stokes = 1
+    int XAxis() const;
+    int YAxis() const;
+    int ZAxis() const;
+    int StokesAxis() const;
+    int CurrentZ() const;
+    int CurrentStokes() const;
```
confluence: It's fine to add all of these getters for consistency, even if though some of these values aren't currently accessed from outside the frame class, but it isn't necessary to use them *inside* the frame class. It's fine to use the bare properties (e.g. `_width` rather than `Width()`). This will also greatly simplify the diff of the frame class.

src/Frame/Frame.cc
```diff
@@ -209,55 +231,55 @@ StokesSlicer Frame::GetImageSlicer(const AxisRange& x_range, const AxisRange& y_
...
+    if (XAxis() >= 0) {
         int start_x(x_range.from), end_x(x_range.to);
 
         // Normalize x constants
         if (start_x == ALL_X) {
             start_x = 0;
         }
         if (end_x == ALL_X) {
-            end_x = _width - 1;
+            end_x = Width() - 1;
```
confluence: This is an example of where the properties should still be used (rather than the getters). All of these little changes should be reverted.

There are a few places in the original code where getters are used instead of properties, and for consistency I think that we should update them, but I would prefer to do that in a separate PR (to keep unrelated changes out of this PR).

src/Frame/Frame.cc
```diff
@@ -369,37 +393,8 @@ bool Frame::SetCursor(float x, float y) {
...
-    return true;
-}
-
-void Frame::InvalidateImageCache() {
-    bool write_lock(true);
-    queuing_rw_mutex_scoped cache_lock(&_cache_mutex, write_lock);
-    _image_cache_valid = false;
+bool Frame::UpdateChannelCache() {
+    return _image_cache->UpdateChannelCache(CurrentZ(), CurrentStokes());
 }
```
confluence: Since this is now a one-liner, and only used in a few places in Frame, I would prefer to remove this wrapper function and call the function on the cache directly.

src/Cache/ChannelCache.cc
```diff
@@ -0,0 +1,84 @@
...
+bool ChannelCache::LoadRegionSpectralData(const AxisRange& z_range, int stokes, const casacore::ArrayLattice<casacore::Bool>& mask,
+    const casacore::IPosition& origin, std::map<CARTA::StatsType, std::vector<double>>& profiles) {
+    return false;
+}
+
+float ChannelCache::GetValue(int x, int y, int z, int stokes) {
+    bool write_lock(false);
+    queuing_rw_mutex_scoped cache_lock(&_cache_mutex, write_lock);
+    return _channel_data[(_width * y) + x];
+}
```
confluence: This function is used in two different ways: from Frame, to access a single value (where this lock is required), and inside a loop in other cache functions (where this lock should not be used; the calling function should acquire the lock once).

But this is part of a broader issue with the way that the cache functions are currently used from frame, which I will describe in more detail below.

src/Cache/StokesCache.cc
```diff
@@ -0,0 +1,124 @@
...
+    if (!mask.shape().empty() && ChannelDataAvailable(ALL_Z, stokes) && !IsComputedStokes(stokes)) {
+        auto get_value = [&](size_t idx) { return _stokes_data[idx]; };
+        DoStatisticsCalculations(z_range, mask, origin, _beam_area, get_value, profiles);
+        return true;
+    }
+    return false;
+}
+
+bool StokesCache::ChannelDataAvailable(int z, int stokes) const {
+    return (z == ALL_Z) && (stokes == _frame->CurrentStokes()) && _stokes_image_cache_valid;
```
confluence: Surely the `(z == ALL_Z)` check here is redundant? If the stokes matches, data is available for any z value.

src/Frame/Frame.cc
```diff
@@ -1075,16 +1052,13 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
     float cursor_value_with_current_stokes(0.0);
 
     // Get the cursor value with current stokes
-    if (_image_cache_valid) {
-        bool write_lock(false);
-        queuing_rw_mutex_scoped cache_lock(&_cache_mutex, write_lock);
-        cursor_value_with_current_stokes = _image_cache[(y * _width) + x];
-        cache_lock.release();
+    if (ImageCacheAvailable()) {
+        cursor_value_with_current_stokes = _image_cache->GetValue(x, y, CurrentZ(), CurrentStokes());
```
confluence: There are several places in Frame where the cache is used like this. It doesn't make sense to check first if the cache is valid and then acquire the cache lock and get the data. The check should be performed after the lock is acquired, in the same function as the data retrieval.

The correct thing to do is probably to refactor all instances like this so that the public function on the cache returns a boolean and sets an input/output parameter instead, checks internally whether the data is available, and returns false if it isn't. Some cache functions already have this kind of interface, but it's not consistent.

In this specific case, there should be two separate cache functions for getting a value: a public one, which acquires the lock and checks if the data is available, and a protected internal function which just fetches the data.

It would also be helpful to unify the behaviour of functions of data which fetch data only if it exists in the cache, and functions which update the cache if necessary and then fetch data. Currently the frame explicitly updates the channel cache in several places. This should be the internal responsibility of the cache -- the frame should only *invalidate* the cache when the channel or stokes changes. Functions for fetching data from the cache should have an `update` flag which determines whether the cache should be updated on demand when that function is called. The frame would then set that flag appropriately.

src/Frame/Frame.cc
```diff
@@ -1245,39 +1226,25 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
                             have_profile = true;
                         }
                     } else { // Use image cache to return full resolution data or prepare data for decimation
+                        get_spatial_profile_from_cache(CurrentStokes());
+                    }
+                } else {
+                    // Required stokes is not the current stokes or the stokes needs to be computed
+                    if (ImageCacheAvailable(CurrentZ(), stokes)) {
+                        get_spatial_profile_from_cache(stokes);
```
confluence: `ImageCacheAvailable` is called a second time here inside `get_spatial_profile_from_cache`. But this is an example of logic which should be refactored.

confluence (changes requested):
This is a partial review, which includes some general remarks about the current structure of the code, and ways in which it could be refactored to divide responsibility between these classes more consistently.

PR #1290, opened 2023-08-04
Fix save image filepath

### LINKED ISSUE #1252: Save image bug with filepath
On the [IDIA CARTA server](https://carta.idia.ac.za) (running v3.0.1), I open one of the [THINGS](https://www2.mpia-hd.mpg.de/THINGS/Data.html) cubes (from `/carta_share/public/jcollier/THINGS/`), create a moment map or PV diagram, and then File --> Save image. In the file browser, when I try to save the image, a red box pops up with the following error msg: "Cannot save image type from loader.Directory: /carta_share/public/collier/THINGS does not exist and cannot be created". 

I then go back to the root directory and navigate back to the same directory, and save again, which is successful.

Even if I try to overwrite a file, a dialog pops up to mention the file exists and asks if I want to overwrite it, but attempting to save/overwrite it causes the same issue.

pford:
Saving an image using the file list fails when the frontend requests a file list for a relative path starting with '/'.  The file list is returned correctly with the same path (still prepended by '/') in the FileListResponse, then the save fails for this path.  A different implementation is used for path handling in file list (casacore) and save file (std::filesystem).  std::filesystem treats the requested folder as an absolute path and does not use the top folder, but casacore appends the requested folder to the top folder as commanded.

FileListHandler should be refactored but I did not want to do it now, so close to a release.  I did move repeated code to the common function and renamed variables referring to the "root" folder which is now the "top" folder.

* What is implemented or fixed?
Closes #1252 
This PR returns a file path in the file list response which is relative to the top folder and does not start with '/' so that the save file path is correct.  It also removes a misleading error message "Cannot save image type from loader." which is returned when the loader is not for an LEL image (which has a special way to save) but then saves the image in the normal way for other image types.
No
See issue #1252 for steps to reproduce: set a top level folder which is not /.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Support animation playback with matched images in multi-panel view ([#1860](https://github.com/CARTAvis/carta-frontend/issues/1860)).
+
 ### Fixed
 * Prevent the installation of pugixml library files ([#1261](https://github.com/CARTAvis/carta-backend/issues/1261)).
 * Fixed spatial profile for polyline in widefield image ([#1258](https://github.com/CARTAvis/carta-backend/issues/1258)).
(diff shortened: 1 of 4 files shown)


confluence (approved):

kswang1029 (approved):
tested the use case and it works now (not v4-beta). no regression from e2e tests. 👍

PR #1294, opened 2023-08-14
Yuhsuan/1277 fitting error

### LINKED ISSUE #1277: image fitting error too small
**Describe the bug**

> Gaussian fitting with 1 component(s)
summary from method 'trust-region/levenberg-marquardt':
number of iterations = 8
function evaluations = 76
Jacobian evaluations = 0
reason for stopping  = small step size
initial |f(x)|       = 8.847404581429e-05
final |f(x)|         = 8.567777841149e-05
initial cost         = 7.827656782749e-09
final cost           = 7.340681713529e-09
residual variance    = 5.195103831230e-12
final cond(J)        = 1.142824649002e+07

As a common example of fitting results using real image data where noise level is much lower and unity, the reduced chi-square (ie 5.195103831230e-12) is too small, resulting an overly-underestimated fitting error for each free parameter. 

**To Reproduce**
1. load an image where noise level is much less then 1, (for example, 1e-6 Jy/beam)
2. try image fitting and see the fitting log, as well as the derived errors for fitting parameters

**Expected behavior**
For a gaussian-like source, the reduced chisquare of the fitting result should be ~ 1.

 - OS [e.g. macOS Monterey]: macOS Ventura
 - Browser version [e.g. 22]: 104
 - Backend branch [e.g. dev, v3b2 release]: dev, v4-beta
 - Frontend branch [e.g. dev, v3b2 release]: dev, v4-beta

YuHsuan-Hwang:
**Description**

1. Fixed #1277 by implementing error calculations based on [Condon (1997)](https://ui.adsabs.harvard.edu/abs/1997PASP..109..166C/abstract), which is also used by [casa imfit](https://casadocs.readthedocs.io/en/latest/api/tt/casatasks.analysis.imfit.html).
    * When the beam table is empty, we assume the samples are independent and apply formulas in Condon (1997) section 2 eqn 21.
    * When the beam table is not empty, we assume there is correlated noise and apply formulas in Condon (1997) section 3 eqn 41 and 42.
    * If the beam table is not empty but the beam of the current plane is not found, we apply the first beam.
    * If the background offset is a free parameter during fitting, background offset error is calculated by the covariance matrix.

2. Changed from unweighted fitting to weighted fitting with the information of image standard deviation.
    * Please note that this will change the iteration number and the fitting results and thus the execution time.

3. Added calculation of integrated flux in the fitting result. This requires frontend branch https://github.com/CARTAvis/carta-frontend/pull/2229 and protobuf branch https://github.com/CARTAvis/carta-protobuf/pull/88.
    * Applied Condon (1997) section 2 eqn 21 and section 3 eqn 42 for the calculation.
    * Applicable only when image units are exactly "Jy/pixel" or "Jy/beam." The calculated flux unit is Jy.
    * When both amplitude and FWHM sizes of a component are fixed, the integrated flux becomes fixed.

    * Obtained beam size, image unit, ~and image std (of the entire image or regions)~ in `Frame.FitImage`.
    * Obtained Median Absolute Deviation (MAD) as image std in `ImageFitter.SolveSystem`.
    * Applied weighted fitting (requires image std) in `ImageFitter.SolveSystem`.
    * Added error calculation (requires beam size and image std) and integrated flux calculation (requires image unit) in `ImageFitter.CalculateErrors`.

    * https://github.com/CARTAvis/carta-frontend/pull/2229
    * https://github.com/CARTAvis/carta-protobuf/pull/88

* ~ToDo~
- [x] ~code maintenance: add documentation in `ImageFitter.h`.~

**Checklist**

- [ ] protobuf updated to the latest dev commit / ~no protobuf update needed~


src/Frame/Frame.cc
```diff
@@ -1773,17 +1803,32 @@ bool Frame::FitImage(const CARTA::FittingRequest& fitting_request, CARTA::Fittin
                 return false;
             }
 
+            float image_std;
+            std::vector<CARTA::StatsType> required_stats = {CARTA::StatsType::Sigma};
+            bool per_z = false;
+            std::map<CARTA::StatsType, std::vector<double>> stats_map;
```
confluence: This should probably be `unordered_map`, but I see that it's `map` in the existing code. There are a few other places where we're using `map`; this should probably be raised in a separate PR (I don't think there are any cases where we actually need an ordered map -- ensuring that we use `unordered_map` may improve performance, but `map` uses less memory, so perhaps this warrants further discussion).

confluence (approved):

markccchiang (approved):

pford (approved):

src/Frame/Frame.cc
YuHsuan-Hwang: @confluence Yes, `GetRegionStats` requires `map`. Looks like we don't set up a long list of stats type, so it's fine to use `map` for now.
Do you prefer I add a TODO in `GetRegionStats` or file a code maintenance issue?

confluence: I think a more general code maintenance issue would be good.

YuHsuan-Hwang: Filed #1295.

YuHsuan-Hwang:
New changes:
* Added documentation in `ImageFitter.h`.
* Replaced std of the image / region with Median Absolute Deviation (MAD) calculation for weighted fitting and error calculation.
* Fixed incorrect center error calculation.

pford (approved):

confluence (approved):

kswang1029 (approved):

kswang1029:
this is looking good. I have adjusted e2e tests accordingly.

src/ImageFitter/ImageFitter.cc
```diff
@@ -223,26 +240,112 @@ int ImageFitter::SolveSystem(CARTA::FittingSolverType solver) {
...
+            center_y_err =
+                sqrt(sq_center_major_err * pow(cos(pa * DEG_TO_RAD), 2.0) + sq_center_minor_err * pow(sin(pa * DEG_TO_RAD), 2.0));
+            amp_err = sqrt(amp * amp * 2.0 / rho_square_1);
+            fwhm_x_err = sqrt(fwhm_x * fwhm_x * 2.0 / rho_square_2);
+            fwhm_y_err = sqrt(fwhm_y * fwhm_y * 2.0 / rho_square_3);
+            const double tmp = fwhm_x * fwhm_y / (fwhm_x * fwhm_x - fwhm_y * fwhm_y);
+            pa_err = sqrt(4.0 * tmp * tmp / rho_square_3) * 180.0 / M_PI;
+
+            if (_unit == "Jy/beam") {
+                const double flux = 2 * M_PI * fwhm_x * fwhm_y * SQ_FWHM_TO_SIGMA * amp / _beam_size / _beam_size;
```
kswang1029: need to divide a solid angle conversion factor as pi / 4.0 / ln(2.0) ~ 1.1330900354567985

`flux = 2 * M_PI * fwhm_x * fwhm_y * SQ_FWHM_TO_SIGMA * amp / _beam_size / _beam_size / (M_PI / 4.0 / log(2.0))`

acdo2002 (approved):

Issue #1296, opened 2023-08-18
Add deployment information to telemetry platform information


crocka:
Added ``DEPLOYMENT_TYPE`` during building of different packages, so we should add the ``DEPLOYMENT_TYPE`` to the platformString that is sent during connection to the telemetry server.

confluence:
Implemented in #1300.

PR #1300, opened 2023-08-22
Carli/1296 add deployment to telemetry platform info

crocka:
**Description**
This PR solves #1296, by sending the ``--DDEPLOYMENT_TYPE`` flag from CMAKE, the carta-backend option ``--controller_deployment``, and the env variable ``CARTA_DOCKER_DEPLOYMENT`` set in Dockerfiles, to the frontend through telemetry platform string.

**Checklist**

- [x] e2e test passing / added corresponding fix


kswang1029 (approved):
I simply run e2e tests and no regression is spotted as expected.

confluence (approved):

crocka:
@kswang1029 @confluence  Sorry I forgot to push the changes that sets the deployment string to "unknown". Please take a look again.

confluence (approved):
Still looks good to me.

crocka:
@ajm-asiaa Suggested we add an env variable in the docker container to identify docker deployments. Environmental variables in docker container are pretty safe from changes.

ajm-ska:
Just to add, I suggested it because, for the Docker containers on cartavis/carta Docker Hub, we use @veggiesaurus's dockerfile_minimal to create them. That Dockerfile simply installs the Debian carta-backend. So if we were to use the `cartavis/carta` Docker container, telemetry would report it to be "deb" (as well as the platform being Linux and ubuntu 20.04.6 LTS)

src/Util/App.cc
```diff
@@ -82,3 +82,27 @@ std::string GetReleaseInformation() {
 #endif
     return std::string("Platform information not available");
 }
+
+std::string ExecuteCommand(const char* command) {
```
confluence: I would rename this to `OutputOfCommand`.

src/Session/Session.cc
```diff
@@ -419,10 +419,17 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
         platform_string_map["deployment"] += fmt::format(" ({})", std::getenv("USER_DEPLOYMENT"));
     }
 
+    string arch = ExecuteCommand("uname -m");
+    string version = ExecuteCommand("uname -r");
```
confluence: This is the version of the Linux kernel, not the version of the distro release. I don't know if this corresponds to something useful on macOS, but it's probably not what we're looking for on Linux.

If you're using `/etc/os-release` to get the Linux distro type, you can use `VERSION` or `VERSION_ID` from there instead (depending on what level of detail you want).

src/Session/Session.cc
```diff
@@ -419,10 +419,17 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
...
+
 #if __APPLE__
     platform_string_map["platform"] = "macOS";
+    string macOSVersion = ExecuteCommand("grep -A1 \"ProductUserVisibleVersion\" /System/Library/CoreServices/SystemVersion.plist | tail -n 1 | awk -F'[<>]' '{print $3}'");
+    platform_string_map["backendPlatformInfo"] = fmt::format("{{distro: macOS, version: {}, arch: {}}}", macOSVersion, arch);
 #else
     platform_string_map["platform"] = "Linux";
+    string linuxVariant = ExecuteCommand("grep -oP '(?<=^ID=)[^\"]+' /etc/os-release");
+    platform_string_map["backendPlatformInfo"] = fmt::format("{{distro: Linux, variant: {}, version: {}, arch: {}}}", linuxVariant, version, arch);
```
confluence: You should not be using `[^\"]+` in the regex -- that will fail entirely if there are any quotes in the field (and there probably will be in `VERSION` and `VERSION_ID`).

Terminology issues: `macOS` and `Linux` are not distros; they're operating systems. `Ubuntu` (for example) is not a variant; it's a distro. "Version" here should probably be "release".

However, I have just checked the existing code, and it's not clear to me why it's necessary to add most of this, since we already appear to be including the entire `/etc/os-release` file above in `"release_info"`, and the platform in `"platform"`. Is this not just duplicating what we already have? The only new piece of information is the architecture, which we could put in a separate field called `"architecture"`.

confluence (changes requested):

Dockerfiles/Dockerfile-ubuntu-22.04
```diff
@@ -24,6 +24,7 @@ RUN \
 EXPOSE 3002
 
 ENV HOME /root
+ENV USER_DEPLOYMENT=docker
```
confluence: Since this is an environment variable, I would suggest a more specific name here, like `CARTA_USER_DEPLOYMENT`.

confluence (changes requested):

ajm-asiaa:
@crocka @confluence I'm sorry about causing confusion here.

src/Session/Session.cc
```diff
@@ -408,6 +408,22 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
...
+    platform_string_map["deployment"] = "unknown";
+#endif
+
+    if (std::getenv("CARTA_USER_DEPLOYMENT")) {
+        platform_string_map["deployment"] += fmt::format(" ({})", std::getenv("CARTA_USER_DEPLOYMENT"));
+    }
+
+    string arch = OutputOfCommand("uname -m");
+
+    std::cout << VERSION_ID << "   " << std::endl;
```
confluence: Please remove this debugging output.

src/Session/Session.cc
```diff
@@ -408,6 +408,22 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
...
+    platform_string_map["architecture"] = fmt::format("{}", arch);
```
confluence: `arch` is already a string -- this formatting step isn't necessary.

src/Session/Session.cc
```diff
@@ -408,6 +408,22 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
...
+    platform_string_map["deployment"] = DEPLOYMENT_TYPE;
+#else
+    platform_string_map["deployment"] = "unknown";
+#endif
+
+    if (std::getenv("CARTA_USER_DEPLOYMENT")) {
+        platform_string_map["deployment"] += fmt::format(" ({})", std::getenv("CARTA_USER_DEPLOYMENT"));
+    }
+
+    string arch = OutputOfCommand("uname -m");
```
confluence: Please use `std::string` here.

confluence (changes requested):

confluence (approved):

src/Main/Main.cc
```diff
@@ -86,6 +86,10 @@ int main(int argc, char* argv[]) {
             }
         }
 
+        if (settings.controller_build) {
+            setenv("CARTA_USER_DEPLOYMENT", "controller", 1);
```
confluence: This should not be reusing (and overwriting) the same environment variable that is being used to detect a docker deployment. These should be saved to separate telemetry entries.

It should also not be setting an environment variable here at all -- this value should be saved internally in a variable which can then be used in the `Session` code -- I suggest setting a static variable on the session class (you can see how this is done for the exit timeouts in `Main.cc`).

src/Main/ProgramSettings.cc
```diff
@@ -197,6 +197,7 @@ void ProgramSettings::ApplyCommandLineSettings(int argc, char** argv) {
     options.add_options("Deprecated and debug")
         ("debug_no_auth", "accept all incoming WebSocket connections on the specified port(s) (not secure; use with caution!)", cxxopts::value<bool>())
         ("no_runtime_config", "do not send a runtime config object to frontend clients", cxxopts::value<bool>())
+        ("controller_build", "used when the backend is hosted for carta-controller", cxxopts::value<bool>())
```
confluence: This should be renamed to `controller_deployment` -- this is not a kind of backend build.

`hosted for` should be `launched by`.

src/Session/Session.cc
```diff
@@ -408,6 +408,20 @@ void Session::OnRegisterViewer(const CARTA::RegisterViewer& message, uint16_t ic
...
     platform_string_map["release_info"] = GetReleaseInformation();
+
+#ifdef DEPLOYMENT_TYPE
+    platform_string_map["deployment"] = DEPLOYMENT_TYPE;
+#else
+    platform_string_map["deployment"] = "unknown";
+#endif
+
+    if (std::getenv("CARTA_USER_DEPLOYMENT")) {
+        platform_string_map["deployment"] += fmt::format(" ({})", std::getenv("CARTA_USER_DEPLOYMENT"));
```
confluence: I would recommend adding this information as a completely separate boolean `docker_deployment` entry, and also adding a separate boolean `controller_deployment` entry for the controller deployment.

confluence (changes requested):

Dockerfiles/Dockerfile-AlmaLinux8
```diff
@@ -24,6 +24,7 @@ RUN \
 EXPOSE 3002
 
 ENV HOME /root
+ENV DOCKER_DEPLOYMENT=docker
```
confluence: This should have a `CARTA_` prefix. The point of adding the prefix was to avoid adding a generically named variable into a huge namespace over which we have no control.

src/Main/Main.cc
```diff
@@ -86,6 +86,12 @@ int main(int argc, char* argv[]) {
             }
         }
 
+        if (settings.controller_deployment) {
+            setenv("CONTROLLER_DEPLOYMENT", "controller", 1);
```
confluence: You should not be setting and then reading an environment variable in order to convey this information from one part of the code to a different part of the code. This should be done internally. We already have a mechanism for using static variables on the Session class to save certain settings (various timeouts) which can then be used inside session functions. Please use the same method to save this value.

Dockerfiles/Dockerfile-AlmaLinux8
confluence: This should also be a boolean flag with a value of 1 (the current value of `docker` doesn't make sense with the new name).

confluence (changes requested):

src/Main/Main.cc
crocka: Thank you so much for your patience reviewing my code! I fixed it according to your requirements.

confluence: You are still setting or unsetting an environment variable here. This should be removed.

src/Session/SessionManager.cc
```diff
@@ -88,7 +88,7 @@ void SessionManager::OnConnect(WSType* ws) {
     // create a Session
     std::unique_lock<std::mutex> ulock(_sessions_mutex);
     _sessions[session_id] = new Session(ws, loop, session_id, address, _settings.top_level_folder, _settings.starting_folder,
-        _file_list_handler, _settings.read_only_mode, _settings.enable_scripting);
+        _file_list_handler, _settings.read_only_mode, _settings.enable_scripting, _settings.controller_deployment);
```
confluence: While this does work, and we do already pass some settings values into sessions like this, I would prefer not to change the signature of the session constructor for a property that is only used in telemetry. Please use the same method that we use to set the exit timeouts: make the variable *static*, and set it using a *static function* on the Session class. This is already being done with the two exit timeout properties in Main, after the settings are parsed (not per session in SessionManager).

(These other existing settings should probably also be removed from this constructor and set as static properties, but that is outside the scope of this PR -- I will file a separate code maintenance issue for that.)

confluence (changes requested):

src/Main/ProgramSettings.cc
```diff
@@ -197,6 +197,7 @@ void ProgramSettings::ApplyCommandLineSettings(int argc, char** argv) {
     options.add_options("Deprecated and debug")
         ("debug_no_auth", "accept all incoming WebSocket connections on the specified port(s) (not secure; use with caution!)", cxxopts::value<bool>())
         ("no_runtime_config", "do not send a runtime config object to frontend clients", cxxopts::value<bool>())
+        ("controller_deployment", "used when the backend is hosted for carta-controller", cxxopts::value<bool>())
```
confluence: Please update the description here. "hosted for" should be "launched by".

confluence (changes requested):
Minor outstanding issue with the option description. Other than that, everything looks good now.

src/Main/ProgramSettings.cc
crocka: Wow, thank you for the quick feedback.

confluence (approved):

confluence:
@kswang1029 I am happy with the code changes. This may need an end-to-end re-test of the whole feature, since there have been multiple changes in what is sent to the frontend.

kswang1029 (approved):
I tried two platforms: arm-based macOS and x86_64-based Ubuntu. The results are

macOS

> [Telemetry] [uuid=3b621120-3dbd-11ee-979f-65b1bb659aa5, sessionId=6b8ab830-46cc-11ee-9490-d5d21522061f] connection {"serverFeatureFlags":0,"platformInfo":{"deployment":"unknown","release_info":"ProductName:\t\tmacOS\nProductVersion:\t\t13.4.1\nProductVersionExtra:\t(c)\nBuildVersion:\t\t22F770820d\n","architecture":"arm64","platform":"macOS"}}

Ubuntu

> [Telemetry] [uuid=d9609da0-7465-11ed-979f-65b1bb659aa5, sessionId=4ca92ea0-46cd-11ee-8846-c17b5db8d13a] connection {"serverFeatureFlags":0,"platformInfo":{"release_info":"PRETTY_NAME=\"Ubuntu 22.04.3 LTS\"\nNAME=\"Ubuntu\"\nVERSION_ID=\"22.04\"\nVERSION=\"22.04.3 LTS (Jammy Jellyfish)\"\nVERSION_CODENAME=jammy\nID=ubuntu\nID_LIKE=debian\nHOME_URL=\"https://www.ubuntu.com/\"\nSUPPORT_URL=\"https://help.ubuntu.com/\"\nBUG_REPORT_URL=\"https://bugs.launchpad.net/ubuntu/\"\nPRIVACY_POLICY_URL=\"https://www.ubuntu.com/legal/terms-and-policies/privacy-policy\"\nUBUNTU_CODENAME=jammy\n","platform":"Linux","deployment":"unknown","architecture":"x86_64"}}

The requirements should be met, along with some extra info from the ubuntu telemetry. 👍

YuHsuan-Hwang (approved):

ajm-ska (approved):
I confirm this still works.

When doing a Docker deployment, I see new entry `"docker_deployment":"true"` in the telemetry.
In addition, when doing a carta-controller deployment, I see new entry `"controller_deployment":"true"` in the telemetry. 
So everything seems to be perfect!

Issue #1314, opened 2023-09-15
Symlinked paths are dereferenced


confluence:
When we follow a symlink in the file browser, we are now redirected to the symlinked directory -- this is unexpected behaviour, which was not present in 4.0.0-beta.1. On the IDIA servers it is present in a snapshot package that I built on August 28, so the change must have been introduced before that point. I believe that this may have been a side effect of [this PR](https://github.com/CARTAvis/carta-backend/pull/1290).

confluence:
Fixed in #1317.

Issue #1284, opened 2023-07-28
Misleading error message if incorrect directory provided in open file request


confluence:
The GUI file browser ensures that files and directories sent to the backend in open file requests are real, but this restriction can be bypassed through the scripting interface. If an attempt is made to open a file in a non-existent directory, the request fails cleanly with an error message (as expected), but the error message reports that the _file name_ does not exist (which is misleading). The error should refer to the directory instead.

Steps to reproduce:

1. Launch frontend and backend >= 4.0 release (with scripting enabled); install latest Python wrapper; launch Python interpreter
2.
```
from carta.session import Session
session = Session.interact("PUT URL WITH TOKEN HERE", PUT_SESSION_ID_HERE)
img = session.open_image("bad_directory/somefile.fits")
```

Expected behaviour: error dialog reports that `bad_directory` does not exist.

Actual behaviour: error dialog reports that `somefile.fits` does not exist.

pford:
@confluence is the path expected to be absolute, relative to the root folder, or relative to the starting folder?

confluence:
The path sent to the backend is relative to the root folder. The `bad_directory` in my example is relative to the *current* directory, and will be transformed into the full path relative to the root, but the backend doesn't have to care about that -- it's fine if it just prints the full path in the error message (that will also make it clearer to the user how their path was expanded and where they went wrong).

confluence:
Fixed in #1317.

PR #1317, opened 2023-10-05
Fix scripting interface and symlink issues

### LINKED ISSUE #1283: Backend crashes attempting to load LEL expression with non-existent directory as parameter
**Describe the bug**

The GUI file browser ensures that the files and directories sent in open file requests are real, but this restriction can be bypassed through the scripting interface.

Directories can either be included in the file paths used in a LEL expression, or a base directory can be specified in the directory parameter. If a non-existent directory is used in the expression, the backend fails cleanly and reports that the file doesn't exist, but if the non-existent directory is provided in the directory parameter, the backend crashes, as reported in [this PR comment](https://github.com/CARTAvis/carta-python/pull/108#issuecomment-1655235598). I have reproduced this on Ubuntu 20.04 (it causes a segmentation fault).

**To Reproduce**

1. Install the Python wrapper ([PR #108](https://github.com/CARTAvis/carta-python/pull/108) is about to be merged into dev); launch Python3 interpreter
2. Launch the backend and frontend (>= 4.0 release needed for latest wrapper; enable scripting flag in backend); look in backend terminal output for URL and session ID
3.
```
from carta.session import Session
session = Session.interact("PUT URL WITH TOKEN HERE", PUT_SESSION_ID_HERE)
img = session.open_LEL_image("any expression, like image1.fits + image2.fits", directory="not_a_real_directory")
```
4. Backend will crash. It should not crash if the same bad directory is included in the file path(s) in the expression.

**Expected behavior**

The backend should fail cleanly and report that the directory does not exist.

pford:
**Description**
 Fix file list issues concerning invalid directories through scripting interface and symlinks which were dereferenced to the symlink directory path
Fixes #1283 #1284  #1314
The directory path is checked before adding the filename for an appropriate error message. The symlink path uses the requested path rather than the absolute path of the actual location for the parent and directory in the response.
No
Set invalid directory in scripting interface session.open_image and session.open_LEL_image, should get error that directory (not file) does not exist. In file browser, select symlinked directory and check directory breadcrumbs, should be previous path with directory added.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+* Fixed scripting interface and symlink directory issues ([#1283](https://github.com/CARTAvis/carta-frontend/issues/1283), [#1284](https://github.com/CARTAvis/carta-frontend/issues/1284), [#1314](https://github.com/CARTAvis/carta-frontend/issues/1314)).
+
 ## [4.0.0]
 
 ### Changed
--- a/src/FileList/FileListHandler.cc
+++ b/src/FileList/FileListHandler.cc
@@ -55,6 +55,9 @@ void FileListHandler::GetRelativePath(std::string& folder) {
             folder = folder.substr(1); // remove leading '/'
         }
     }
+    if (folder.empty()) {
+        folder = ".";
+    }
 }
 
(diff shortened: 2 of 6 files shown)


loveluthien (approved):
The backend won't crash and show the correct error messages when opening non-existing directories. The symbolic link doesn't expand.

confluence (approved):
This looks good to me. I have tested that it fixes all three issues.

kswang1029 (approved):

PR #1319, opened 2023-10-12
Yuhsuan/1318 fix fitting mad

### LINKED ISSUE #1318: Image fitting: MAD calculation of images with nans
**Describe the bug**
(Found by @acdo2002) The MAD (median absolute deviation) calculation becomes lower when more additional nans are included in the fitting data. When std is 0 (with a large number of nans), the fitting fails.

**Screenshots or videos**
image

**To Reproduce**
1. Fit the entire image and the fitting fails.

**Expected behavior**
The MAD calculation does not change when additional nans are included (and the fitting result does not change).
Fitting the entire image does not fail.

 - OS [e.g. macOS Monterey]: macOS Big Sur
 - Browser version [e.g. 22]: 117
 - Backend branch [e.g. dev, v3b2 release]: dev, v4
 - Frontend branch [e.g. dev, v3b2 release]: dev, v4

**Additional context**
[Download test image](https://drive.google.com/file/d/1WhyovU0w1mwiP1FoLiZb9SEK5Ro0S6iE/view?usp=sharing)

YuHsuan-Hwang:
**Description**

  Fixed #1318.
  Exclude nans when calculating MAD (median absolute deviation). Instead of using the raw image data, create a tmp double array with image data excluding nans for the calculation.
* Other changes:
  Minor adjustment of the unit test - extend the area of FOV fitting to get a more precise result.
* Tested that:
  The fitting result and MAD does not change when additional nans are included.
  Fitting the entire test image does not get MAD 0 and does not fail.
  Also checked that the calculation time and the memory pattern do not change.

**Checklist**

- [x] e2e test passing / corresponding fix added / ~new e2e test created~

### DIFF
--- a/src/ImageFitter/ImageFitter.cc
+++ b/src/ImageFitter/ImageFitter.cc
@@ -47,7 +47,7 @@ bool ImageFitter::FitImage(size_t width, size_t height, float* image, double bea
     _create_residual_data = create_residual_image;
     _progress_callback = progress_callback;
 
-    CalculateNanNum();
+    CalculateNanNumAndStd();
     SetInitialValues(initial_values, background_offset, fixed_params);
 
     // avoid SolveSystem crashes with insufficient data points
@@ -141,7 +141,7 @@ void ImageFitter::StopFitting() {
     _fit_data.stop_fitting = true;
 }
 
-void ImageFitter::CalculateNanNum() {
+void ImageFitter::CalculateNanNumAndStd() {
     std::vector<double> data_notnan;
 
     _fit_data.n_notnan = _fit_data.n;
--- a/src/ImageFitter/ImageFitter.cc
+++ b/src/ImageFitter/ImageFitter.cc
@@ -143,6 +143,7 @@ void ImageFitter::StopFitting() {
 
 void ImageFitter::CalculateNanNumAndStd() {
     std::vector<double> data_notnan;
+    data_notnan.reserve(_fit_data.n);
 
     _fit_data.n_notnan = _fit_data.n;
     for (size_t i = 0; i < _fit_data.n; i++) {
--- a/src/ImageFitter/ImageFitter.cc
+++ b/src/ImageFitter/ImageFitter.cc
@@ -142,12 +142,18 @@ void ImageFitter::StopFitting() {
 }
 
 void ImageFitter::CalculateNanNum() {
+    std::vector<double> data_notnan;
+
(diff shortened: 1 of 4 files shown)


pford (approved):

markccchiang (approved):

acdo2002 (approved):

kswang1029:
@YuHsuan-Hwang if the input image does not have any NaN pixels, do we expect to see slightly different fitting results in this PR comparing to dev?

I see the solution is to have a double array to exclude NaN. Would it be sensible to refer to BITPIX for the array declaration? Not sure this is the root cause for the above question?

src/ImageFitter/ImageFitter.cc
```diff
@@ -141,13 +141,19 @@ void ImageFitter::StopFitting() {
     _fit_data.stop_fitting = true;
 }
 
-void ImageFitter::CalculateNanNum() {
+void ImageFitter::CalculateNanNumAndStd() {
+    std::vector<double> data_notnan;
```
confluence, line 145: I would suggest calling `reserve` on this vector (with the original data size, which is the upper limit) to avoid possible reallocations as it is filled up.

(What is the size of the data that we're duplicating here -- one full image channel? Is there a significant memory usage impact when the image is large?)

confluence (approved):

YuHsuan-Hwang:
@kswang1029 
> if the input image does not have any NaN pixels, do we expect to see slightly different fitting results in this PR comparing to dev?

Yes, there are slightly changes because my previous workaround triggers some preprocessing to the array.

> I see the solution is to have a double array to exclude NaN. Would it be sensible to refer to BITPIX for the array declaration? Not sure this is the root cause for the above question?

Double array is required for the GSL function.

src/ImageFitter/ImageFitter.cc
YuHsuan-Hwang: Thanks for the comment. `reserve` is added.

The size of the data is an image channel. Yes, it can be large with large images, but it should be temporary and the workspace `gsl_multifit_nlinear_workspace` used in the fitting process is larger than it.

YuHsuan-Hwang:
Debug log "MAD = ..." is added so that testing is easier.

loveluthien (approved):
It works well.

kswang1029 (approved):
looks fine. corresponding e2e tests are updated. no other regression seen from e2e tests.

PR #1321, opened 2023-10-16
Store the global settings in a singleton class

### LINKED ISSUE #1302: Code maintenance: accessing global settings properties from session objects
We currently pass config settings values into session objects in two completely different ways:

1. [Using static functions to set static properties on the Session class](https://github.com/CARTAvis/carta-backend/blob/aa5df4829d1e627fb6694cfc701ef88a0d39ea4d/src/Main/Main.cc#L55C11-L55C11):
```
if (settings.wait_time >= 0) {
    Session::SetExitTimeout(settings.wait_time);
}

if (settings.init_wait_time >= 0) {
    Session::SetInitExitTimeout(settings.init_wait_time);
}
```

2. [Passing them into the Session constructor](https://github.com/CARTAvis/carta-backend/blob/aa5df4829d1e627fb6694cfc701ef88a0d39ea4d/src/Session/SessionManager.cc#L90C5-L91C83): 
```
_sessions[session_id] = new Session(ws, loop, session_id, address, _settings.top_level_folder, _settings.starting_folder,
        _file_list_handler, _settings.read_only_mode, _settings.enable_scripting);
```

2 doesn't really make sense. These properties don't change per session; they're global properties which remain unchanged for the lifetime of the backend. They should all be static.

Rather than duplicating most or all settings values as static properties on Session, however, I think it would be cleaner for us to turn `ProgramSettings` itself into a singleton which is initialised in Main and can be imported from any part of the code and used to access the settings properties directly.

markccchiang:
**Description**

Closes #1302.

Creating a singleton class `Global` that stores the `ProgramSettings` object for global settings. The `ProgramSettings` object from `Global` is initialized in the Main, and its values can be accessed directly by any objects via the `Global` static functions without duplicating them.

No.

This PR is the code maintenance. So everything is normal and not changed in the tests. 

**Checklist**


kswang1029 (approved):
No regression from e2e tests. If there are specific tests that need to be performed, please let me know. Otherwise, please proceed code review. 👍

src/Main/GlobalValues.h
```diff
@@ -0,0 +1,88 @@
...
+#ifndef CARTA_BACKEND_SRC_MAIN_GLOBALVALUES_H_
+#define CARTA_BACKEND_SRC_MAIN_GLOBALVALUES_H_
+
+#include "ProgramSettings.h"
+
+#include <iostream>
+
+namespace carta {
+
+class GlobalValues {
```
confluence: Adding these additional `GlobalValues` and `Global` classes is unnecessarily verbose.

All that is necessary is to add static functions directly to the `ProgramSettings` class: one to initialise the settings instance (from Main), and one to retrieve the settings instance (to be used wherever the settings are needed).

The getters are also unnecessary. Once the settings instance is retrieved, the attributes can be accessed directly, as they are now (e.g. `settings.help` in Main).

src/Session/SessionManager.cc
```diff
@@ -87,8 +88,8 @@ void SessionManager::OnConnect(WSType* ws) {
 
     // create a Session
     std::unique_lock<std::mutex> ulock(_sessions_mutex);
-    _sessions[session_id] = new Session(ws, loop, session_id, address, _settings.top_level_folder, _settings.starting_folder,
-        _file_list_handler, _settings.read_only_mode, _settings.enable_scripting);
+    _sessions[session_id] = new Session(ws, loop, session_id, address, Global::TopLevelFolder(), Global::StartingFolder(),
+        _file_list_handler, Global::ReadOnlyMode(), Global::EnableScripting());
```
confluence: The settings shouldn't be passed in as parameters to Session here; they should be read from the global settings instance inside the session code.

confluence (changes requested):

src/Main/GlobalValues.h
markccchiang: Initially, I tried inserting static functions to set and retrieve setting values. But found that all variables and functions from the `ProgramSettings` have to become `static` type. Turing the ProgramSettings to a singleton class also increases the complexity of the original code, plus the need to modify its testers (If I did not miss anything). A relatively simple way is to wrap the ProgramSettings in a new singleton object. We can avoid the risk of modifying original codes and clearly distinguish their functionalities. I have combined the `Global` and `GlobalValues` as a new `Global` class to set and retrieve global settings. Hope it looks likely not so wordy.

src/Main/Global.h
```diff
@@ -13,75 +13,69 @@
 
 namespace carta {
 
-class GlobalValues {
+class Global {
```
confluence: Please refer to my previous comment -- there is no need for this wrapper class at all. This functionality for creating and retrieving a singleton instance should be directly in the `ProgramSettings` class. The getters are unnecessary (the properties on the settings object can be accessed directly, as before).

confluence (changes requested):

src/Main/ProgramSettings.h
```diff
@@ -31,6 +31,11 @@
 
 namespace carta {
 struct ProgramSettings {
+    static ProgramSettings& GetInstance() {
+        static ProgramSettings global_values;
+        return global_values;
```
confluence: I suggest renaming `global_values` to `settings` here.

src/Main/Main.cc
```diff
@@ -46,14 +46,13 @@ int main(int argc, char* argv[]) {
         sigaction(SIGINT, &sig_handler, nullptr);
 
         // Main
-        carta::ProgramSettings settings(argc, argv);
+        auto settings = ProgramSettings::GetInstance() = std::move(carta::ProgramSettings(argc, argv));
```
confluence: This is a little awkward. I would suggest adding an `Initialise` static method to `ProgramSettings` which initialises the instance internally, and calling it here (something like: `ProgramSettings::Initialise(argc, argv)`).

src/Main/Main.cc
```diff
@@ -79,13 +78,8 @@ int main(int argc, char* argv[]) {
...
 
-        if (settings.wait_time >= 0) {
-            Session::SetExitTimeout(settings.wait_time);
-        }
-
-        if (settings.init_wait_time >= 0) {
-            Session::SetInitExitTimeout(settings.init_wait_time);
-        }
+        Session::SetExitTimeout();
+        Session::SetInitExitTimeout();
```
confluence: In these two specific cases, the previous behaviour was cleaner and should be restored.

src/Session/Session.cc
```diff
@@ -100,17 +101,13 @@ bool Session::_controller_deployment = false;
...
       _loop(loop),
       _id(id),
       _address(address),
-      _top_level_folder(top_level_folder),
-      _starting_folder(starting_folder),
-      _table_controller(std::make_unique<TableController>(_top_level_folder, _starting_folder)),
-      _read_only_mode(read_only_mode),
-      _enable_scripting(enable_scripting),
+      _table_controller(std::make_unique<TableController>(
+          ProgramSettings::GetInstance().top_level_folder, ProgramSettings::GetInstance().starting_folder)),
```
confluence: These properties could be read from the global settings object inside the `TableController` constructor.

src/Session/Session.cc
```diff
@@ -362,7 +368,7 @@ bool Session::FillFileInfo(
     // Resolve filename and fill file info submessage
     bool file_info_ok(false);
 
-    fullname = GetResolvedFilename(_top_level_folder, folder, filename, message);
+    fullname = GetResolvedFilename(ProgramSettings::GetInstance().top_level_folder, folder, filename, message);
```
confluence: `Session` uses the settings in multiple places. To avoid repeating this verbose syntax, I would suggest adding a `ProgramSettings& _settings` to this class, and setting it to the instance in the constructor. It will then be possible to refer to e.g. `_settings.top_level_folder`.

confluence (changes requested):

src/Logger/Logger.cc
```diff
@@ -6,16 +6,14 @@
...
 
 namespace carta {
 namespace logger {
 
-static bool log_protocol_messages(false);
-
-void InitLogger(bool no_log_file, int verbosity, bool log_performance, bool log_protocol_messages_, fs::path user_directory) {
-    log_protocol_messages = log_protocol_messages_;
-
+void InitLogger() {
```
confluence, line 18: I suggest adding a `_settings` property to this class as well.

src/Session/SessionManager.cc
```diff
@@ -531,7 +531,8 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
             auto t_session = session->GetLastMessageTimestamp();
             auto t_now = std::chrono::high_resolution_clock::now();
             auto dt = std::chrono::duration_cast<std::chrono::seconds>(t_now - t_session);
-            if ((_settings.idle_session_wait_time > 0) && (dt.count() >= _settings.idle_session_wait_time)) {
+            if ((ProgramSettings::GetInstance().idle_session_wait_time > 0) &&
+                (dt.count() >= ProgramSettings::GetInstance().idle_session_wait_time)) {
```
confluence: I suggest restoring the `_settings` property to this class. Since it currently only uses one setting, we could also just add that as a property directly, but this is more consistent and future-proof (we may use other settings in the future).

src/Table/TableController.cc
```diff
@@ -185,7 +183,7 @@ void TableController::OnFilterRequest(
 
 void TableController::OnFileListRequest(
     const CARTA::CatalogListRequest& file_list_request, CARTA::CatalogListResponse& file_list_response) {
-    fs::path root_path(_top_level_folder);
+    fs::path root_path(ProgramSettings::GetInstance().top_level_folder);
```
confluence: I suggest adding a `_settings` property to this class as well (same rationale as for the other classes).

test/TestMain.cc
```diff
@@ -49,10 +46,10 @@ int main(int argc, char** argv) {
...
     }
 
-    verbosity = result["verbosity"].as<int>();
-    no_log = result["no_log"].as<bool>();
-    log_performance = result["log_performance"].as<bool>();
-    log_protocol_messages = result["log_protocol_messages"].as<bool>();
+    ProgramSettings::GetInstance().verbosity = result["verbosity"].as<int>();
+    ProgramSettings::GetInstance().no_log = result["no_log"].as<bool>();
+    ProgramSettings::GetInstance().log_performance = result["log_performance"].as<bool>();
+    ProgramSettings::GetInstance().log_protocol_messages = result["log_protocol_messages"].as<bool>();
```
confluence: We can avoid this verbosity by creating a local reference variable.

confluence (changes requested):
I think that this is good, and almost ready to be merged -- but the other classes which use settings should also have a property with a reference to the object (like `Session` does), so that they don't have to call `GetInstance` every time to access it (and can use more shorthand syntax).

src/Table/TableController.h
```diff
@@ -32,7 +33,7 @@ struct TableViewCache {
 
 class TableController {
 public:
-    TableController(const std::string& top_level_folder, const std::string& starting_folder);
+    TableController(ProgramSettings& settings);
```
confluence: You should not be passing this in as a parameter; you should be setting it in the initializer list, the same way you're doing it in `Session`.

src/Logger/Logger.cc
```diff
@@ -6,16 +6,14 @@
...
 
 namespace carta {
 namespace logger {
 
-static bool log_protocol_messages(false);
-
-void InitLogger(bool no_log_file, int verbosity, bool log_performance, bool log_protocol_messages_, fs::path user_directory) {
-    log_protocol_messages = log_protocol_messages_;
-
+void InitLogger(const ::carta::ProgramSettings& settings) {
```
confluence: My mistake; I didn't realise in my previous review that this was a collection of functions and not a class.

This should be refactored to be closer to the original solution: restore the static `log_protocol_messages` variable and set it in this function using the value from the settings object. But don't pass the settings object in as a parameter; call `GetInstance` once at the start of the function to create a local reference. The remaining functions should be reverted to use the static variable, as before.

confluence (changes requested):

src/Logger/Logger.cc
```diff
@@ -6,15 +6,15 @@
...
 
 namespace carta {
 namespace logger {
 
-static bool log_protocol_messages(false);
-
-void InitLogger(bool no_log_file, int verbosity, bool log_performance, bool log_protocol_messages_, fs::path user_directory) {
-    log_protocol_messages = log_protocol_messages_;
+void InitLogger() {
+    auto& settings = ProgramSettings::GetInstance();
```
confluence, line 20: In this function, after this line, set the static variable `log_protocol_messages` that was there in the original code (copying the value from this settings object), so that the other functions (which only use that one setting value) can use this variable instead of fetching the whole settings object.

src/Logger/Logger.cc
```diff
@@ -109,7 +109,8 @@ void InitLogger(bool no_log_file, int verbosity, bool log_performance, bool log_
 }
 
 void LogReceivedEventType(const CARTA::EventType& event_type) {
-    if (log_protocol_messages) {
+    auto& settings = ProgramSettings::GetInstance();
+    if (settings.log_protocol_messages) {
```
confluence: This function only uses that one boolean flag -- so the previous behaviour should be restored. That flag can be set inside the init function, as before.

src/Logger/Logger.cc
```diff
@@ -120,7 +121,8 @@ void LogReceivedEventType(const CARTA::EventType& event_type) {
 }
 
 void LogSentEventType(const CARTA::EventType& event_type) {
-    if (log_protocol_messages) {
+    auto& settings = ProgramSettings::GetInstance();
+    if (settings.log_protocol_messages) {
```
confluence: As above -- the original behaviour should be restored.

confluence (changes requested):

confluence:
I think that this is conceptually good now; I just want to be careful about the lifetime of the settings variable (since there are some places where the reference is being accessed where previously objects were using copied data). @pford am I right that the variable is created in the body of the `try` in the main function, and will be destroyed when the function exits that block? Should we *declare* it at the top of the function with the shared pointers? I'm concerned by the possibility that other threads may be accessing it during shutdown (or may in the future).

confluence (changes requested):
I suggest that you remove the `_settings` references that I previously suggested (in `Session` and `TableController`), and restore the individual settings attributes on these objects. These attributes should, however, be set in the object constructors using the singleton settings object (and *not* passed in separately as parameters, as before -- that part should not be reverted).

This will avoid possible issues with the lifetime of the settings object, and I think also make it clearer that the values are set once and not expected to change.

I think it will be cleaner to follow the same pattern as the `InitLogger` function, and temporarily create a reference inside the constructor, rather than repeat the `GetInstance` call in the initializer list. So for example in the `Session` constructor you should do this:

```
auto& settings = ProgramSettings::GetInstance();
_top_level_folder = settings.top_level_folder
(...)
```

...rather than repeating `_top_level_folder(ProgramSettings::GetInstance().top_level_folder)` for each of these properties.

confluence (approved):

pford (approved):

kswang1029 (approved):

PR #1322, opened 2023-10-17
Improve file IDs for generated images

### LINKED ISSUE #1224: Better way to assign file IDs for generated in-memory images
When an image is opened or appended, the frontend assigns a file ID for the image.  But when a request is made for a generated image, the file ID in the request is for the source image from which the new image is generated.  The backend calculates a file ID for the new, generated image using the input file ID, a multiplier determined by the generated image type, and some offset.  Some requests generate multiple images so multiple file IDs are needed.

These file ID calculations should be simplified and if possible coordinated between the image generators (MomentGenerator, PvGenerator, and ImageFitter when model and residual image options are added).  The new file IDs must not duplicate any file IDs already in use, unless the new image is replacing existing images of the same type.

There is also a need to determine what type of image it is, preferably not from its file ID being in a certain range.

pford:
**Description**

Closes #1224 

Increment file ID from last opened image (saved by Session with each Frame saved to the frames map), rather than multiplier applied to input file ID.  Multipliers and GeneratedImage file_id have been removed.

CARTAvis/carta-frontend#2275, CARTAvis/carta-protobuf#91

Create generated images (moments, PV, image fitting) and ensure images are opened and behave as expected.

**Checklist**

- [x] protobuf updated to the latest dev commit / ~no protobuf update needed~


pford:
@YuHsuan-Hwang with the change here to increment the last used file ID by 1, all file IDs are positive so I had to remove the check for a negative file ID from ImageFitter (previously for PV image or model/residual image).  Is this check critical?  In my testing, the fit for these images succeeded with valid settings.

YuHsuan-Hwang:
@pford 
> with the change here to increment the last used file ID by 1, all file IDs are positive so I had to remove the check for a negative file ID from ImageFitter (previously for PV image or model/residual image). Is this check critical? In my testing, the fit for these images succeeded with valid settings.

I think it's ok to remove them. In previous approach we get extremely large IDs when multiplying 1000 several times, and tile encoding fails.

However, frontend checks generated images by ID < 0 and ID > 1000, so we will need changes in the frontend.

pford:
> @pford
> 
> > with the change here to increment the last used file ID by 1, all file IDs are positive so I had to remove the check for a negative file ID from ImageFitter (previously for PV image or model/residual image). Is this check critical? In my testing, the fit for these images succeeded with valid settings.
> 
> I think it's ok to remove them. In previous approach we get extremely large IDs when multiplying 1000 several times, and tile encoding fails.
> 
> However, frontend checks generated images by ID < 0 and ID > 1000, so we will need changes in the frontend.

Just wondering, why does the frontend need to know the images are generated?  Are they treated differently than other images? One idea is that instead of incrementing a positive ID, the generated IDs could be negative (starting after ALL_FILES -1 constant) and decrement each time.

But my thought is that there are better ways to identify image types than by file ID.  The frontend receives these images in a response message (MomentResponse, PvResponse, FittingResponse) so at that point it could flag them as generated.  Or explicitly set a boolean generated_image field in OpenFileAck in these responses?  Or add an enum if the frontend needed to distinguish between moment/pv/fitting images?

YuHsuan-Hwang:
> Just wondering, why does the frontend need to know the images are generated? Are they treated differently than other images? One idea is that instead of incrementing a positive ID, the generated IDs could be negative (starting after ALL_FILES -1 constant) and decrement each time.

It's used for resuming sessions and saving workspaces. The generated images are not resumed/saved.

> But my thought is that there are better ways to identify image types than by file ID. The frontend receives these images in a response message (MomentResponse, PvResponse, FittingResponse) so at that point it could flag them as generated. Or explicitly set a boolean generated_image field in OpenFileAck in these responses? Or add an enum if the frontend needed to distinguish between moment/pv/fitting images?

Yes, I agree we don't change the backend current approach. I can add a flag in frontend and replace the logic easily.

kswang1029 (approved):
Looks good. Tried manual tests with combination of LEL, PV, moment, and image fitting. 
Session resume works fine. No regression from e2e tests.

markccchiang (approved):

confluence (approved):

YuHsuan-Hwang (approved):

loveluthien (approved):

PR #1323, opened 2023-10-21
Replace map with unordered map

### LINKED ISSUE #1295: Replace map with unordered_map
Code maintenance: ensure we use `unordered_map` instead of `map` for better performance.

> This should probably be `unordered_map`, but I see that it's `map` in the existing code. There are a few other places where we're using `map`; this should probably be raised in a separate PR (I don't think there are any cases where we actually need an ordered map -- ensuring that we use `unordered_map` may improve performance, but `map` uses less memory, so perhaps this warrants further discussion).
>
> _Originally posted by @confluence in https://github.com/CARTAvis/carta-backend/pull/1294#discussion_r1293206281_

markccchiang:
**Description**

Closes #1295. Being a pure lookup table, `unordered_map` is supposed to perform better than `map`. According to the discussions from the [stake overflow](https://stackoverflow.com/questions/2196995/is-there-any-advantage-of-using-map-over-unordered-map-in-case-of-trivial-keys). However, using `unordered_map` would need more memory space than that for `map`. The new tester `TestMapPerf` simply tests the performances of `map` and `unordered_map`. `unordered_map` is a little bit faster than `map` when inserting a new data element or finding & copying a specific data element. But it is significantly slower than `map` when erasing map elements or manipulating the map structure. Ref: [map vs. unordered_map in C++](https://www.geeksforgeeks.org/map-vs-unordered_map-c/).

Replace all `map` with `unordered_map`, except for the ordered array from `HttpServer`.

No.

It will not affect the original tests. @acdo2002 could you please check if there are any performance differences compared to the dev branch?

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+* Replace map with unordered_map ([#1295](https://github.com/CARTAvis/carta-backend/issues/1295)).
+
 ### Fixed
 * Fixed scripting interface and symlink directory issues ([#1283](https://github.com/CARTAvis/carta-frontend/issues/1283), [#1284](https://github.com/CARTAvis/carta-frontend/issues/1284), [#1314](https://github.com/CARTAvis/carta-frontend/issues/1314)).
 * Include casacore log messages in carta log ([#1169](https://github.com/CARTAvis/carta-backend/issues/1169)).
--- a/src/Cache/RequirementsCache.h
+++ b/src/Cache/RequirementsCache.h
@@ -174,10 +174,10 @@ struct RegionSpectralConfig {
 };
 
 struct SpectralCache {
-    std::map<CARTA::StatsType, std::vector<double>> profiles;
+    std::unordered_map<CARTA::StatsType, std::vector<double>> profiles;
 
     SpectralCache() {}
-    SpectralCache(std::map<CARTA::StatsType, std::vector<double>>& profiles_) : profiles(profiles_) {}
+    SpectralCache(std::unordered_map<CARTA::StatsType, std::vector<double>>& profiles_) : profiles(profiles_) {}
 
     bool GetProfile(CARTA::StatsType type_, std::vector<double>& profile_) {
         if (!profiles.empty() && profiles.count(type_)) {
@@ -200,14 +200,14 @@ struct RegionStatsConfig {
 };
 
 struct StatsCache {
-    std::map<CARTA::StatsType, double> stats;
+    std::unordered_map<CARTA::StatsType, double> stats;
 
     StatsCache() {}
-    StatsCache(std::map<CARTA::StatsType, double>& stats_) {
+    StatsCache(std::unordered_map<CARTA::StatsType, double>& stats_) {
         stats = stats_;
     }
 
-    bool GetStats(std::map<CARTA::StatsType, double>& stats_) {
+    bool GetStats(std::unordered_map<CARTA::StatsType, double>& stats_) {
         if (!stats.empty()) {
             stats_ = stats;
             return true;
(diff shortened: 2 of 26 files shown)


kswang1029 (approved):
I simply run e2e tests and no regression is spotted. If there are specific parts I need to pay attend to, please let me know. Thanks.

kswang1029:
@kswang1029 to do performance test via e2e tests. check test elapsed time!

kswang1029:
@acdo2002 please execute ICD test manually in a sequence (not in parallel) and compare elapsed time of each test to see if there is significant performance boost or degradation. I will do a similar one but from e2e perspective. @confluence these test results would provide a hint about performance impact hopefully.

kswang1029:
@markccchiang @confluence Not sure if I am seeing a thing here. It appears to me that dev (with map) performs better than unordered_map branch. Basically all tests with unordered_map branch require more time to execute. 

Screenshot 2023-11-02 at 09 29 58

confluence:
Hmm. In that case, I suggest that we move this to draft until I can dig further into the code and look at how we're using the maps in different components. It's possible that only some of the maps should be changed, none of them should be changed, or that some of our unordered maps should be ordered maps! Clearly it's not as simple as changing them across the board.

kswang1029:
> Hmm. In that case, I suggest that we move this to draft until I can dig further into the code and look at how we're using the maps in different components. It's possible that only some of the maps should be changed, none of them should be changed, or that some of our unordered maps should be ordered maps! Clearly it's not as simple as changing them across the board.

Any naive explanations why?

IMG_2237

confluence:
> Any naive explanations why?

My short answer is: the way we're accessing the maps is (on average!) triggering cases where `map` is better more than the cases where `map` is worse. Some guesses:

* unexpected hash collisions (could be a bug in a hash function we've defined -- that's happened before!)
* something else to do with key hashing (e.g. we have a complicated key type and the hashing and/or equality comparison is expensive)
* we're performing one of the slow operations more often than I thought

But these are just guesses, and this requires further investigation. It's possible that only one of these maps is the source of the slowdown, and all the other maps do in fact benefit from being converted (but the one slow map is disproportionately affecting the outcome because it's accessed a lot more frequently).

~(Edit: my best guess is that using the `StatsRegionId` struct as a key, with custom hashing and comparison functions, is expensive. I have an idea for testing this theory, but I need to think about how to implement it correctly.)~ This is only used in a very limited way, so it's probably not that. String keys may be expensive to hash, though, and we use those in a few places in the metadata code.

confluence:
I've also noticed that some of the region code repeatedly creates and destroys local variables which are small maps, and this might be more expensive for unordered maps. But I also don't know if this is happening frequently enough to be an issue. Rather than continuing to guess, I will try to use a profiler. :)

kswang1029:
> I've also noticed that some of the region code repeatedly creates and destroys local variables which are small maps, and this might be more expensive for unordered maps. But I also don't know if this is happening frequently enough to be an issue. Rather than continuing to guess, I will try to use a profiler. :)

indeed, region e2e tests do have the performance degradation problem more significantly. All the e2e tests have a set of common step: load file list, load file info, and then load image or images.

markccchiang:
> Hmm. In that case, I suggest that we move this to draft until I can dig further into the code and look at how we're using the maps in different components. It's possible that only some of the maps should be changed, none of them should be changed, or that some of our unordered maps should be ordered maps! Clearly it's not as simple as changing them across the board.

Yes. Considering the current performance test results, we may switch between map and unordered_map one by one, and see which is better. This is trivial but time-consuming work -- so far there is no clear guide on which kind of map is better for us to use.

acdo2002:
I did a quick test, here is the result.
the current dev
current performance regression tests (not much region related, big 35 GB fits, casa and 70GB hdf5):102.882 s, 100.716 s, 100.641 s, 100.945 s, 100.977 s
ICD-RxJS (small image size) region related tests:3.341 s, 3.442 s, 3.49 s, 3.465 s, 3.473 s

the unordered map branch:
current performance regression tests (not much region related, big 35 GB fits, casa and 70GB hdf5):101.107 s, 101.149 s
101.007 s, 100.931 s, 100.984 s => **NO big difference**
ICD-RxJS (small image size) region related tests:3.535 s, 3.48 s, 3.467 s, 3.498 s, 3.527 s => **little bit slower, perhaps need the specific large region sizes test rather than bigger image size tests**

NOTE: there is a regression failure happened on SET_REGION_ANNOTATION_EXPORT_IMPORT_CASA_PIXEL.test.ts
The exported region does not match to the expect
Expected: 163
Received: 157.71336364746094

PR #1324, opened 2023-10-25
Improve the code style in HTTP server

### LINKED ISSUE #1260: Code style improvements in HTTP server code
As discussed in [this comment](https://github.com/CARTAvis/carta-backend/pull/1240#discussion_r1181267458) and [this comment](https://github.com/CARTAvis/carta-backend/pull/1240#discussion_r1181269702), the object types passed as function parameters should be enum values rather than string literals, and schema URLs should be looked up in an unordered map with object types as keys.

markccchiang:
**Description**

Closes #1260.

To avoid the magic hardcoded strings, I think replacing them with global const string variables is good enough. Applying new enum values, we may need an additional map that transfers the enum to string, which is inefficient. If my understanding is correct (?)

No.

All tests should not be affected.

**Checklist**

### DIFF
--- a/src/HttpServer/HttpServer.cc
+++ b/src/HttpServer/HttpServer.cc
@@ -72,16 +72,23 @@ void HttpServer::RegisterRoutes() {
         app.put("/api/database/preferences", [&](auto res, auto req) { HandleSetPreferences(res, req); });
         app.del("/api/database/preferences", [&](auto res, auto req) { HandleClearPreferences(res, req); });
 
-        std::vector<std::string> object_types = {LAYOUT, SNIPPET, WORKSPACE};
-        for (const auto& object_type : object_types) {
-            app.get(fmt::format("/api/database/list/{}s", object_type),
-                [&](auto res, auto req) { HandleGetObjectList(object_type, res, req); });
-            app.get(fmt::format("/api/database/{}s", object_type), [&](auto res, auto req) { HandleGetObjects(object_type, res, req); });
-            app.get(
-                fmt::format("/api/database/{}/:name", object_type), [&](auto res, auto req) { HandleGetObject(object_type, res, req); });
-            app.put(fmt::format("/api/database/{}", object_type), [&](auto res, auto req) { HandleSetObject(object_type, res, req); });
-            app.del(fmt::format("/api/database/{}", object_type), [&](auto res, auto req) { HandleClearObject(object_type, res, req); });
-        }
+        app.get(fmt::format("/api/database/list/{}s", LAYOUT), [&](auto res, auto req) { HandleGetObjectList(LAYOUT, res, req); });
+        app.get(fmt::format("/api/database/{}s", LAYOUT), [&](auto res, auto req) { HandleGetObjects(LAYOUT, res, req); });
+        app.get(fmt::format("/api/database/{}/:name", LAYOUT), [&](auto res, auto req) { HandleGetObject(LAYOUT, res, req); });
+        app.put(fmt::format("/api/database/{}", LAYOUT), [&](auto res, auto req) { HandleSetObject(LAYOUT, res, req); });
+        app.del(fmt::format("/api/database/{}", LAYOUT), [&](auto res, auto req) { HandleClearObject(LAYOUT, res, req); });
+
+        app.get(fmt::format("/api/database/list/{}s", SNIPPET), [&](auto res, auto req) { HandleGetObjectList(SNIPPET, res, req); });
+        app.get(fmt::format("/api/database/{}s", SNIPPET), [&](auto res, auto req) { HandleGetObjects(SNIPPET, res, req); });
+        app.get(fmt::format("/api/database/{}/:name", SNIPPET), [&](auto res, auto req) { HandleGetObject(SNIPPET, res, req); });
+        app.put(fmt::format("/api/database/{}", SNIPPET), [&](auto res, auto req) { HandleSetObject(SNIPPET, res, req); });
+        app.del(fmt::format("/api/database/{}", SNIPPET), [&](auto res, auto req) { HandleClearObject(SNIPPET, res, req); });
+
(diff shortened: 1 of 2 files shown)


confluence:
When I made the enum suggestion I didn't realise that C++ did not have an enum type with string values -- I'm happy to drop the enum idea and continue using strings. However, I have more specific remarks about how this code should be modified.

src/HttpServer/HttpServer.cc
```diff
@@ -24,7 +24,13 @@ using json = nlohmann::json;
 
 namespace carta {
 
-const std::string success_string = json({{"success", true}}).dump();
+const std::string SUCCESS_STRING = json({{"success", true}}).dump();
```
confluence, line 27: This is unnecessary and should be reverted. While this is repeated across multiple functions, we don't abstract away other kinds of responses, and it's more readable to keep this directly in the functions.

src/HttpServer/HttpServer.cc
```diff
@@ -54,39 +60,41 @@ HttpServer::HttpServer(std::shared_ptr<SessionManager> session_manager, fs::path
 void HttpServer::RegisterRoutes() {
     uWS::App& app = _session_manager->App();
 
+    const std::string scripting("/api/scripting/action");
```
confluence: This is unnecessary and should be reverted.

src/HttpServer/HttpServer.cc
```diff
@@ -54,39 +60,41 @@ HttpServer::HttpServer(std::shared_ptr<SessionManager> session_manager, fs::path
...
     if (_enable_scripting) {
-        app.post("/api/scripting/action", [&](auto res, auto req) { HandleScriptingAction(res, req); });
+        app.post(scripting, [&](auto res, auto req) { HandleScriptingAction(res, req); });
     } else {
-        app.post("/api/scripting/action", [&](auto res, auto req) { NotImplemented(res, req); });
+        app.post(scripting, [&](auto res, auto req) { NotImplemented(res, req); });
     }
 
+    const std::string database("/api/database/");
```
confluence: This is unnecessary and should be reverted. The repetition problem in this block of code is not the repetition of this string (and substituting concatenation with a variable is also repetitive -- the concatenation just makes it slightly harder to read).

The problem is that identical code is repeated for the three object types, and that is what should be eliminated. Please use a foreach loop over `LAYOUT`, `SNIPPET` and `WORKSPACE` to combine the three blocks of routes for objects into one, and please use `fmt::format` for the concatenation. So for example the first line would read `app.get(fmt::format("/api/database/list/{}s", object_type), [&](auto res, auto req) { HandleGetObjectList(object_type, res, req); });` (assuming you use `object` as your loop variable, like you did below).

confluence (changes requested):

src/HttpServer/HttpServer.cc
```diff
@@ -25,6 +25,12 @@ using json = nlohmann::json;
 namespace carta {
 
 const std::string success_string = json({{"success", true}}).dump();
```
confluence: My mistake; I didn't see that this was already a global variable and you just capitalised the name. In my original comment I suggested that this value should be used directly inside the functions -- so instead of `res->end(success_string)` we would use `res->end(json({{"success", true}}).dump())`. But now that I write that out, I think that it looks cleaner to keep this as a const variable, and you were correct to capitalise it (that's a more consistent style for constants). I apologise for the confusing feedback! Please restore the capitalisation.

confluence (approved):

confluence:
Looks good to me. Added @kswang1029 as a reviewer, for testing.

markccchiang:
Looks like we can not register the routes using a loop, otherwise, it will fail when executing the command `/carta_backend --frontend_folder /Users/path/to/carta-frontend/build --port 3002 --omp_threads 8 --debug_no_auth --no_browser --verbosity 5`. I am wondering if it is related to the life circle of the handler functions. @confluence do you have any ideas?

markccchiang:
> Looks like we can not register the routes using a loop, otherwise, it will fail when executing the command `/carta_backend --frontend_folder /Users/path/to/carta-frontend/build --port 3002 --omp_threads 8 --debug_no_auth --no_browser --verbosity 5`. I am wondering if it is related to the life circle of the handler functions. @confluence do you have any ideas?

I guess this may be due to the interpretation of the code during the compiling time. The argument variable in the handler function has to be global. Otherwise, the registered handler function will not be valid and fail to open the frontend. @kswang1029 could you please check again if the problem is solved? Thank you!

confluence:
@markccchiang what error did you get with the previous version of the code? I can't reproduce it here (and valgrind isn't reporting any issues either). This could be a difference between compilers.

I agree that it has to do with the lifetime of the `object_type` variable. The lambdas are executed after the `RegisterRoutes` function has exited, so any objects defined in the function's scope have been destroyed. Your original `object_type` was a reference to a value in a vector defined in the function scope. This variable doesn't have to be global, but it does have to outlive the execution of the function. I believe that you could also have fixed this by explicitly capturing the variable by copy in all of the lambdas, but I think that the current code is fine -- using the existing map keys instead of repeating the strings is cleaner.

kswang1029 (approved):
@markccchiang with the new solution, it works nicely. No regression from e2e tests.

markccchiang:
> @markccchiang what error did you get with the previous version of the code? I can't reproduce it here (and valgrind isn't reporting any issues either). This could be a difference between compilers.
> 

@confluence with my Mac compiler I got an error message when opening the frontend by the command `/carta_backend --frontend_folder /Users/path/to/carta-frontend/build ...` like this:

Screenshot 2023-10-31 at 13 39 44

This was first found by @kswang1029. It was also found on a specific computer when opening the CARTA v4.0 electron app (with the dev branch). Not sure if it is related to the same root cause of the problem.

PR #1326, opened 2023-10-27
Normalize the style of guard name from headers

### LINKED ISSUE #1023: Include guards are not consistently named
The `#ifndef` guards in our headers and template files are inconsistently named -- some still reflect old file names or locations, some never used the same convention in the first place, etc.. It would be simple to extend our `style.py` script to check and fix these automatically.

[Here's a discussion](https://stackoverflow.com/questions/4867559/naming-include-guards) of various possible naming conventions. My suggestion would be something like `CARTA_PATH_TO_FILE_H_`, which is close to what we have already. Maybe with the top-level directory included, so that we can distinguish between `CARTA_SRC...` and `CARTA_TEST...` ?

markccchiang:
**Description**

Closes #1023.

A new bash script, `scripts/check_guard_name_style.sh`, is added to normalize the style of guard names from `*.h` and `*.tcc` files under the `src` and `test` folders automatically. The usage is simply by executing this bash script under the `scripts` folder.

No.

All tests are not affected.

**Checklist**


kswang1029 (approved):
I just run e2e tests to see if there are regressions. None is observed. 👍

confluence:
There are a few issues with this approach:

* The script should be callable from the main directory, not from the scripts subdirectory.
* The sed implementation test should not rely on a platform check, because the user could have GNU sed installed on MacOS. This check should also be done in one place and not repeated.
* This script only does replacements, and doesn't have an option just to check for incorrect guards.
* It doesn't automatically fix missing guards.
* We need this check to be performed automatically on PRs, just like the other style checks (and the automated check should only report, not fix).

I think that the cleanest way to implement this (which is also portable, and sidesteps the sed compatibility issue) is to extend the `style.py` Python script (as suggested in the issue). I have [committed what I think is a working implementation to a branch](https://github.com/CARTAvis/carta-backend/tree/confluence/guard-check-in-style-script) -- please have a look and confirm that it works (it should give the same result as your bash implementation). In addition to detecting incorrect guards and missing guards, it separately reports when the closing `endif` comment doesn't match the guard name.

To perform a check, run `./scripts/style.py headerguards check`, and to perform a fix, run `./scripts/style.py headerguards check`. This additional check type is also included in `all`, so it should be picked up automatically by the CI. In my branch I have just committed the script change, not any modified files.

If you're happy with that approach, please feel free to merge that branch into your PR (or to make a new PR; whatever is simpler).

markccchiang:
> There are a few issues with this approach:
> 
> * The script should be callable from the main directory, not from the scripts subdirectory.
> * The sed implementation test should not rely on a platform check, because the user could have GNU sed installed on MacOS. This check should also be done in one place and not repeated.
> * This script only does replacements, and doesn't have an option just to check for incorrect guards.
> * It doesn't automatically fix missing guards.
> * We need this check to be performed automatically on PRs, just like the other style checks (and the automated check should only report, not fix).
> 
> I think that the cleanest way to implement this (which is also portable, and sidesteps the sed compatibility issue) is to extend the `style.py` Python script (as suggested in the issue). I have [committed what I think is a working implementation to a branch](https://github.com/CARTAvis/carta-backend/tree/confluence/guard-check-in-style-script) -- please have a look and confirm that it works (it should give the same result as your bash implementation). In addition to detecting incorrect guards and missing guards, it separately reports when the closing `endif` comment doesn't match the guard name.
> 
> To perform a check, run `./scripts/style.py headerguards check`, and to perform a fix, run `./scripts/style.py headerguards check`. This additional check type is also included in `all`, so it should be picked up automatically by the CI. In my branch I have just committed the script change, not any modified files.
> 
> If you're happy with that approach, please feel free to merge that branch into your PR (or to make a new PR; whatever is simpler).

@confluence Thank you for providing the new edition of `./scripts/style.py`! I confirm it works well and is a better way to implement this guard name style check. So I removed the original bash script `./scripts/check_guard_name_style.sh`.

confluence (approved):
Looks good to me! I'll hold off merging until I can check the other PRs, since this will touch every header.

PR #1330, opened 2023-11-07
Fix the hanging problem of region spectral profiles when a region is deleted

### LINKED ISSUE #1328: backend hangs if we delete a region when its region spectral profile is updating
**Describe the bug**
The backend seems to stop responding and hangs if we delete a region when its region spectral profile is updating. Further region create request cannot be made (seeing a dimmed region rendered). Cannot request cursor spectral profile either.

**To Reproduce**
1. Load a larger cube (eg 2000x2000x2000)
2. enable spectral profiler to see cursor profile
3. freeze the cursor
4. make a region to see region spectral profile 
5. move the region to see another profile update
6. while it is updating, delete the region.
7. unfreeze the cursor to see profile update but we cannot see the update
8. create a region but we see a dimmed region forever (ie no ack from the backend)

**Expected behavior**
backend should be still responsive in this case

 - OS [e.g. macOS Monterey]: macOS Ventura / ubuntu 22.04
 - Browser [e.g. chrome, safari, electron app]: electron app / firefox
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v4-stable / v4-stable
 - Frontend branch [e.g. dev, v3b2 release]: v4-stable / v4-stable

markccchiang:
**Description**

Closes #1328.

This problem is due to the deadlock of two mutexes, `_active_task_mutex` and `_region_mutex`, in the RegionHandler methods. When a region spectral profile process starts, `_active_task_mutex` is taken in the method [GetRegionSpectralData](https://github.com/CARTAvis/carta-backend/blob/dev/src/Region/RegionHandler.cc#L1847). Then the `_region_mutex` will be frequently taken inside this method [RegionFileIdsValid](https://github.com/CARTAvis/carta-backend/blob/dev/src/Region/RegionHandler.cc#L2080) to check whether the region object exists or it is disconnected. If we delete the region during this process, the `RemoveRegion` method is called, and the `_region_mutex` is taken [here](https://github.com/CARTAvis/carta-backend/blob/dev/src/Region/RegionHandler.cc#L113) inside this function. Then it waits to take `_active_task_mutex` in the [WaitForTaskCancellation](https://github.com/CARTAvis/carta-backend/blob/dev/src/Region/RegionHandler.cc#L116). Two methods `GetRegionSpectralData` and `RemoveRegion` are waiting for the same mutexes `_active_task_mutex` and `_region_mutex`. The solution is to remove the `_region_mutex` from the `RegionSet` method. Since it only reads the status of region objects and does not modify the `_regions` map. Removing this mutex allows the [RegionFileIdsValid](https://github.com/CARTAvis/carta-backend/blob/dev/src/Region/RegionHandler.cc#L2080) to get rid of `_region_mutex` (since it can be taken by `RemoveRegion` at the same time), so it can pass the disconnected signal and stop the region spectral profile process. Finally, the region object can be deleted.

No.

The hanging problem of region spectral profiles should be solved.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Include casacore log messages in carta log ([#1169](https://github.com/CARTAvis/carta-backend/issues/1169)).
 * Fixed the problem of opening old IRAM fits images ([#1312](https://github.com/CARTAvis/carta-backend/issues/1312)).
 * Fixed incorrect std calculation when fitting images with nan values ([#1318](https://github.com/CARTAvis/carta-backend/issues/1318)).
+* Fixed the hanging problem when deleting a region during the spectral profile process ([#1328](https://github.com/CARTAvis/carta-backend/issues/1328)).
 
 ## [4.0.0]
 
--- a/src/Region/RegionHandler.cc
+++ b/src/Region/RegionHandler.cc
@@ -137,7 +137,6 @@ std::shared_ptr<Region> RegionHandler::GetRegion(int region_id) {
 bool RegionHandler::RegionSet(int region_id, bool check_annotation) {
     // Check whether a particular region is set or any regions are set
     bool region_set(false);
-    std::lock_guard<std::mutex> region_guard(_region_mutex);
     if (region_id == ALL_REGIONS) {
         region_set = _regions.size() > 0;
     } else {
--- a/src/Region/RegionHandler.cc
+++ b/src/Region/RegionHandler.cc
@@ -110,14 +110,20 @@ void RegionHandler::RemoveRegion(int region_id) {
         return;
     }
 
-    std::unique_lock<std::mutex> region_lock(_region_mutex);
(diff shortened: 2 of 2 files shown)


pford (changes requested):
The purpose of the region mutex is to prevent one thread accessing a region while another is deleting it, and the purpose of the active task mutex is to prevent the region from being destroyed mid-task.  So I think we need them both.

We could move the region mutex lock in RemoveRegion around the regions map: WaitForTaskCancellation then lock-erase-unlock.  For all regions, use an iterator to do this one region at a time.  (1) RemoveRegion disconnects the region and waits for the active task mutex, (2) GetRegionSpectralData uses RegionSet (with a region mutex) to detect that the region exists but is disconnected and exits to release the active task mutex, (3) RemoveRegion safely removes (with a region mutex) the region from the map.

pford (approved):

kswang1029 (approved):
Tested. no regression from e2e tests.

confluence (approved):

Issue #1309, opened 2023-09-04
Feature request: custom URL


appolloford:
Following the discussion in PR #1307, we agree to some extent that it is good for CARTA to support custom URL. This function is particularly helpful for setting a reverse proxy (e.g. via Open OnDemand) for CARTA and is also implemented in Jupyter `(ServerApp.base_url)` and Matlab `(MWI_BASE_URL)`. Similar to the two softwares, we propose to have CARTA support adding URL prefix behind the `host:port` for proxy to parse.

kswang1029:
@confluence do you think this is implemented with https://github.com/CARTAvis/carta-backend/pull/1338?

confluence:
@kswang1029 I believe so.

PR #1338, opened 2023-11-28
Add custom prefix to all HTTP server URLs

confluence:
**Description**

This is a re-creation of #1307 with a copy of the branch in our main repo, and implements #1309.

To be done before this can move out of draft:

- [X] Resolve merge conflict
- [X] Replace low-level string manipulation with regular expressions
- [X] Convert prefix env variable to a normal setting (wait for #1321 to be merged)

**Checklist**


confluence:
@appolloford and @Micket, does this solution still work for you with the prefix set using a configuration option rather than an environment variable? This option can be passed on the commandline (`carta_backend --http_url_prefix /foo/bar ...`), or set globally or per-user in one of the config files.

appolloford:
Having a configuration option in commandline should work well, but I am wondering whether the priority is higher than users' config file. We need the node name and port number information in the URL prefix so that we can have OnDemand find the computing node. There may be some trouble if users overwrite it with some arbitrary value.

confluence:
@appolloford yes, the commandline parameters have the highest priority, and override the user config file, which overrides the global config file. You can also use commandline parameters to disable the user and/or global config file entirely.

kswang1029:
@confluence do we consider the following use cases?

```
url_without_token = 'http://localhost:3003'


session = Session.start_and_create(Chrome(headless=True, options=custom_chrome_options), "/Users/kswang/carta_build/carta-backend/build/carta_backend", params=("/Users/kswang/set_QA_e2e_v2", "--frontend_folder", "/Users/kswang/carta_build/carta-frontend/build", "--port", "3003", "--debug_no_auth", "--http_url_prefix", "/foo/bar"))
```

If I do so, I see errors as 

```
Traceback (most recent call last):
  File "/Users/kswang/carta_python_e2e_test/test.py", line 12, in <module>
    session = Session.start_and_create(Chrome(headless=True, options=custom_chrome_options), "/Users/kswang/carta_build/carta-backend/build/carta_backend", params=("/Users/kswang/set_QA_e2e_v2", "--frontend_folder", "/Users/kswang/carta_build/carta-frontend/build", "--port", "3003", "--debug_no_auth", "--http_url_prefix", "/foo/bar"))
  File "/Users/kswang/anaconda3/lib/python3.10/site-packages/carta/session.py", line 224, in start_and_create
    return browser.new_session_with_backend(executable_path, remote_host, params, timeout, token, frontend_url_timeout)
  File "/Users/kswang/anaconda3/lib/python3.10/site-packages/carta/browser.py", line 144, in new_session_with_backend
    return self.new_session_from_url(backend.frontend_url, backend.token, backend=backend, timeout=timeout, debug_no_auth=backend.debug_no_auth)
  File "/Users/kswang/anaconda3/lib/python3.10/site-packages/carta/browser.py", line 98, in new_session_from_url
    self.exit(f"Could not parse session ID from frontend. Last error: {last_error}")
  File "/Users/kswang/anaconda3/lib/python3.10/site-packages/carta/browser.py", line 149, in exit
    raise CartaBadSession(msg)
carta.util.CartaBadSession: Could not parse session ID from frontend. Last error: Message: no such element: Unable to locate element: {"method":"css selector","selector":"[id="info-session-id"]"}
  (Session info: chrome-headless-shell=125.0.6422.141)
...
```

The core question is how do we support the scripting interface when the `--http_url_prefix` flag is set? Or this is not a problem as that flag is used and set by remote server manager, so that we (the client) should just set the URL as seen with the custom prefix directly?

confluence:
@kswang1029 that's a wrapper issue which should be discussed [here](https://github.com/CARTAvis/carta-python/issues/152).

I thought that the wrapper didn't require changes, but I hadn't fully tested it yet.

confluence:
@kswang1029 actually, this is an issue with `debug_no_auth` here in the backend code -- I can reproduce it without the wrapper. (I'm also cleaning up some URL parsing code in the wrapper, but it's not related to this issue). I will investigate further.

confluence:
@kswang1029 I have pushed a change which I think fixes this issue. With this fix, I believe that your use case should work with the current development wrapper. Additionally, [wrapper PR #154](https://github.com/CARTAvis/carta-python/pull/154) fixes a parsing issue which breaks starting a backend from the wrapper with a prefix *and* auth.

kswang1029 (approved):
this works fine as far as I can tell. No other regression from e2e tests.
I would prefer to have other reviewers to have another look and tests just in case I missed something.

kswang1029:
just wondering, what happens if some weird string (ie containing special char etc) is set as the custom prefix? Would that cause problems or security issues? Maybe the backend needs to validate the prefix first before applying it and throw errors to stop the backend process.

confluence:
> just wondering, what happens if some weird string (ie containing special char etc) is set as the custom prefix? Would that cause problems or security issues? Maybe the backend needs to validate the prefix first before applying it and throw errors to stop the backend process.

The prefix is limited to a subset of characters, otherwise it is ignored (I kept the selection of characters from the original PR).

daikema:
So testing this in a local k8s environment set to run CARTA at a subpath, I'm able to replicate the error [reported here](https://github.com/CARTAvis/carta-frontend/issues/2336).

When testing the code in the branch here where I have the `--http_url_prefix /carta-a/` added (with or without a final slash after CARTA) I see a page saying:
```
File Not Found
uWebSockets/20 Server
```
rather than the expected CARTA session.

daikema (approved):
So I managed to figure out what was causing the problem I reported earlier in conversation with @confluence.  I was having the ingress rewriting requests such that an inbound request to `/carta-a/` would be send to the container with backend as a request for `/`.  After removing the prefix-stripping code from the ingress configuration it seems that the app is now functioning as expected with the custom prefix.

veggiesaurus (approved):

PR #1348, opened 2023-12-19
ajm/GitHub actions icd tests

ajm-ska:
**Description**

This one took a lot of trial and error. It effectively moves the daily ICD tests from Jenkins to Github Actions.

The 3 macOS and 5 Linux platforms are combined in one workflow file called `ICD Tests`. They use local runners hosted at ASIAA. The macOS platforms run natively on the Mac hardware. The 5 Linux platforms run via Apptainer (Singularity) on three different Linux servers. It is a long story, but I found Apptainer easier to use instead of Docker and the workflow file can be a lot shorter and simpler.

To reduce the length of the workflow file, we use two different actions; `run-macos/action.yml` and `run-apptainer/action/yml` that start a carta-backend and run the ICD tests for each "job" (ICD test stage) on macOS and Linux, respectively. I don't think the workflow file can be shortened much more. Each "job" requires that long matrix block. Unless we consolidate stages. But I did not do that as I was trying to keep the layout similar to the previous Jenkins ICD tests where we can visually see each test stage.

I have put a list of the actual ICD tests per stage in the ICD-RxJS repo in ICD-RxJS/ICD_test_stages. The workflow file reads that list of tests. This means we can easily add or remove tests to existing stages without needing to make any commits to the carta-backend repo. However, we would need to make a commit to icd_tests.yml on the carta-backend repo if we were to add or remove test stages.

Just like on Jenkins, the ICD test stages run sequentially. e.g. Animator only runs after File-Browser. This was not necessary for Linux/Apptainer, but for some strange reason it is necessary for the macOS platforms. I was unable to figure out why. But is not really important. It still works fine running sequentially and would not be any faster because one runner only runs one job at a time.

The status of all the jobs and steps and the log files can be seen, for example, here: https://github.com/CARTAvis/carta-backend/actions/runs/7259363238

The final part that I have not been able to test is the schedule:
> on:
>   schedule:
>     - cron: '0 0 * * *' # UTC time

This _should_ allow it to run daily at 08:00 Taiwan time. I believe 'schedule' only runs on the default branch, which is 'dev' in our case. So we can not actually test this until it is merged.

I guess this is an unusual PR. It does not affect the carta-backend code base so doesn't require much review or testing.

**Checklist**

### DIFF
--- a/.github/actions/run-apptainer/action.yml
+++ b/.github/actions/run-apptainer/action.yml
@@ -1,0 +1,52 @@
+name: 'Run ICD tests via Apptainer'
+description: 'Start carta_backend and run ICD tests'
+inputs:
+  os_version:
+    description: 'Platfrom'
+    required: true
+  image:
+    description: 'Apptainer image'
+    required: true
+  port:
+    description: 'Port number for carta_backend'
+    required: true
+  test_stage_name:
+    description: 'ICD test stage'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - run: |
+        SRC_DIR=$GITHUB_WORKSPACE/source
+        BUILD_DIR=$GITHUB_WORKSPACE/build-${{ inputs.os_version }}
+        TEST_STAGE="$BUILD_DIR/ICD-RxJS/ICD_test_stages/${{ inputs.test_stage_name }}.tests"
+        LOG_FILE="/tmp/carta_icd_${{ inputs.os_version }}_${{ inputs.test_stage_name }}.log"
+        apptainer exec \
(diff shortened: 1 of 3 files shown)


kswang1029:
@ajm-asiaa any suggestions on testing this? or it is not possible?

ajm-ska:
@kswang1029 I think the only way to test it is to merge it and see if it runs automatically tomorrow at 0800. That is the only part I'm not sure of at the moment, but it is only supposed to work when it is in the main branch ('dev' in our case).  If there is a problem, I could try to issue a direct fix. It will not affect the carta-backend codebase at all.
Besides that, you can see that all the ICD tests on each platform are currently capable of passing.

PR #1352, opened 2024-01-23
Refactor Region code to RegionState and RegionConverter

pford:
**Description**

* What is implemented or fixed? Mention the linked issue(s), if any.
Implements #1347 
* How does this PR solve the issue? Give a brief summary.

  - **Code maintenance**: the Region class needed to be streamlined.
    - RegionState struct was moved to a separate file, with some Region functions now in the struct.
    - RegionConverter class contains code (formerly in Region) to convert the region to a non-reference image, either as the same region type or as a polygon approximation, and return it as casacore::LCRegion or casacore::Record.
    - Region is now only for the region in the reference image.  It creates a RegionConverter when needed for matched images.
    - Other cleanup: unused methods and includes removed.
  - **Testing**:  additional backend tests were created for reference and matched regions, to validate the refactor and for better code coverage.
  - **Performance**:  to retrieve the Region as LCRegion or a Record, for all images (1) the control points were converted from pixel to world coordinates and used to create a casacore::WCRegion, (2) WCRegion::toLCRegion converts back to pixel coordinates, (3) (if requested) LCRegion::toRecord creates a Record.  For the reference image, however, the conversion to world coordinates and back to pixel coordinates is unnecessary and adds time.  This especially affects the performance of line spatial profiles and PV image generation where numerous regions are created along the line and their LCRegions are used for the statistics.  Instead, for the reference file ID a Record is created from the control points, which can be returned as a Record for export or as LCRegion using LCRegion::fromRecord for analytics.

* Are there any companion PRs (frontend, protobuf)?
No, backend change only

* Is there anything else that testers should know (e.g. exactly how to reproduce the issue)?
The behavior should be unchanged by this refactor, except in small underlying details (e.g. LCRegion::toRecord returns the coordinates 1-based but creating the Record from control points is 0-based, as specified in the Record's "oneRel" field).

**Checklist**

- [x] changelog updated / ~no changelog update needed~
- [ ] e2e test passing / corresponding fix added / new e2e test created
- [ ] ICD test passing / corresponding fix added / new ICD test created
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] ~protobuf version bumped~ / protobuf version not bumped
- [x] added reviewers and assignee
- [x] added ZenHub estimate, milestone, and release

PR #1353, opened 2024-01-23
Improve region spatial profile and PV performance

### LINKED ISSUE #1339: performance issue in spatial profile generation along a line or a polyline
as of v4-stable, the spatial profiler generation process along a line or a polyline is not that performant. The impact to users is they will need to wait a while before seeing the results. If performance cannot be improved significantly, we will need to implement "progressive update + cancellation" similar to the region spectral profiler (https://github.com/CARTAvis/carta-frontend/issues/1954 + https://github.com/CARTAvis/carta-backend/issues/1179).

pford:
**Description**

Fixes #1339
  This PR improves spatial profile and PV performance.  Both features involve creating box regions along a line/cut with a width, then calculating stats for each box.  Although the performance improvement may be small for a short line, it will have greater impact when a long line is drawn on a "pancake"-type image (as described at the F2F).

  In addition to the Region changes in #1352 (submitted separately as a large refactor), this PR improves performance in these methods:
    - Frame::GetRegionData applies a casacore::Slicer to the loaded image plane data when possible (current channel/stokes for spatial profile), rather than creating a casacore::SubImage to get region data.  A SubImage is a casacore::ImageInterface which includes coordinate system information, unneeded for data only.
    - Frame::GetSlicerData applies a casacore::Slicer to the loaded image plane data when possible (current channel/stokes for spatial profile), rather than using the FileLoader to apply the Slicer to the Image to create a SubImage.
    - RegionHandler::GetLineProfiles iterates through the box regions created along the line to get stats for each; this for loop is parallelized with OMP (spatial profile and PV).  Note: cannot `break` out of omp parallel loop when cancelled, so set a cancel flag and just `continue` instead.
    - RegionHandler::GetTemporaryRegionProfile calculates the mean and number of pixels for the region data rather than using BasicStats to calculate all stats.

No, backend change only

The behavior and results should be unchanged by this refactor. Should improve performance.

**Checklist**

- [x] ICD test passing / corresponding fix added / new ICD test created


src/Region/RegionHandler.cc
```diff
@@ -2606,19 +2615,27 @@ casacore::Vector<float> RegionHandler::GetTemporaryRegionProfile(int region_idx,
...
             std::vector<float> region_data;
             if (_frames.at(file_id)->GetRegionData(stokes_region, region_data, false)) {
-                // Get BasicStats
-                BasicStats<float> basic_stats;
-                CalcBasicStats(basic_stats, region_data.data(), region_data.size());
-                num_pixels = basic_stats.num_pixels;
-                profile[0] = basic_stats.mean;
+                // Very small region, just calc needed stats here
+                num_pixels = 0;
+                double sum(0.0);
```
markccchiang, line 2627: I am just wondering does the performance can be further improved by adding `#pragma omp parallel for reduction(+ : sum, num_pixels)` before this for loop (?)

kswang1029:
@pford I did a quick (somewhat extreme) test and see a significant improvement on performance in spatial profile generation (eg, v4.1 17s-ish vs thisPR 4s-ish) 🚀🚀🚀🚀
I will continue with a more detailed test after the region refactoring PR is in. 

The e2e test looks good except a pv cancellation test (due to performance improvement which is good).

kswang1029:
@pford just to double check: does the code change in this PR include the changes in https://github.com/CARTAvis/carta-backend/pull/1352 as well?

confluence (approved):

pford:
> @pford just to double check: does the code change in this PR include the changes in #1352 as well?

@kswang1029 I thought the Region refactor should be a separate branch and PR because of the large code change.  I combined the branches for both PRs into `pam/combine_1347_1339_branches` if you want to check it out.

src/Region/RegionHandler.cc
pford: Each box region along the line is very small, 1 x line_width px (~max 10~: this max was for rendering not computation).  So I thought the overhead would not be worth it, but I could try it.

pford: I added `pragma omp parallel for` and tested a 30000-px line in the "pancake" FITS image with line widths of 3, 10, and 20.  The averages after 5 runs each (parallel vs not) showed parallel runs were 30 ms slower (width 3), 9 ms faster (width 10), and 47 ms slower (width 20).  Then I ran 5 more runs and averaged all 10; parallel runs were 5 ms faster (width 3), 42 ms slower (width 10), and 124 ms slower (width 20).  So it looks like overall the losses outweigh the gains with omp parallelization.

pford: Default width is 3

veggiesaurus: yeah, I'd really expect it to be slower due to thread overhead. This is a very tight loop that is probably memory bandwidth bound more than anything else.

veggiesaurus:
> @pford I did a quick (somewhat extreme) test and see a significant improvement on performance in spatial profile generation (eg, v4.1 17s-ish vs thisPR 4s-ish) 🚀🚀🚀🚀 I will continue with a more detailed test after the region refactoring PR is in.
> 
> The e2e test looks good except a pv cancellation test (due to performance improvement which is good).

@kswang1029 can we benchmark these improvements against any other tools? 🤔

kswang1029:
@pford I found that the pv generation progress bar displayed at frontend is not that accurate compared to dev. Now the progress bar goes quickly close to the right end and then hangs there until the process is done. Is that because now the progressive update logic has been changed?

kswang1029:
> > @pford I did a quick (somewhat extreme) test and see a significant improvement on performance in spatial profile generation (eg, v4.1 17s-ish vs thisPR 4s-ish) 🚀🚀🚀🚀 I will continue with a more detailed test after the region refactoring PR is in.
> > The e2e test looks good except a pv cancellation test (due to performance improvement which is good).
> 
> @kswang1029 can we benchmark these improvements against any other tools? 🤔

Yes, I tried psrecord for spatial profile. Roughly it is 4x improvement if we request the profile from the reference image (ie no spatial matching in action). @pford are the CPU footprints expected in the new approach? 

dev: 18.83 s
dev_spatial_profile

pr: 4.81s
pr_spatial_profile


However, if we request a spatial profile from the spatially-matched image, the improvement is much less. (dev: 124.6s vs pr: 110.6s). @pford is this also expected? Any chance to improve this as well?

markccchiang (approved):

pford:
Fixed the PV progress bar.  Also updated the combined branch `pam/combine_1347_1339_branches` with this change and the changes in the 1347 branch.

Spatial profiles for matched regions take each box region along the line in parallel and convert pixel control points -> world points -> WCRegion -> matched LCRegion then apply the box region to the matched image (= SubImage, with an exclusive lock due to casacore) for region data. A region in the reference image takes each box in parallel to create an LCRegion from pixel control points and apply it to the image cache stored in memory (with a read lock) for region data, which is faster.  But I will look for improvements for matched regions and investigate the psrecord CPU spike.

kswang1029:
> Fixed the PV progress bar. Also updated the combined branch `pam/combine_1347_1339_branches` with this change and the changes in the 1347 branch.

Confirmed that the pv generation progress bar now behaves normally and as expected. 👍

pford:
@kswang1029 it was difficult to reproduce the cpu spike mid-profile, but I suspect it was a cursor update.  I moved my mouse in the image while it was calculating the profile and I finally did see a spike like yours.

kswang1029:
> @kswang1029 it was difficult to reproduce the cpu spike mid-profile, but I suspect it was a cursor update. I moved my mouse in the image while it was calculating the profile and I finally did see a spike like yours.

I see. Thanks for investigating the spikes. As for the performance of the matched images, if you think there is not much we can do about it due to mutex, then I will approve the PR.

loveluthien (approved):

pford:
@kswang1029 I found two issues with the matched region analytics implementation, when the region is a matched rotbox (rotated rectangle region for stats/spectral profiles, or box region along a line for spatial profiles/PV image):

1. When the converted region is not distorted, the rotbox was still being approximated as a polygon with (constant) 1000 points.  I changed this to convert the 4 corners of the box to the matched image and create a casacore::LCPolygon from the converted points.  Much faster.
2. When the converted region is distorted, and the rotbox is approximated as a polygon with (constant) 1000 points, if the region is very small (by definition each box along the line is small 3 x line_width pixels, default 3 max 20) each segment length between points can be as low as 0.01 pixels.  This seemed like overkill to me, so I set a minimum length of 0.1 pixel so there are fewer polygon points to convert to the matched image. (Perhaps still too low?  Any suggestions for the appropriate interval?)

kswang1029:
> @kswang1029 I found two issues with the matched region analytics implementation, when the region is a matched rotbox (rotated rectangle region for stats/spectral profiles, or box region along a line for spatial profiles/PV image):
> 
> 
> 
> 1. When the converted region is not distorted, the rotbox was still being approximated as a polygon with (constant) 1000 points.  I changed this to convert the 4 corners of the box to the matched image and create a casacore::LCPolygon from the converted points.  Much faster.
> 
> 2. When the converted region is distorted, and the rotbox is approximated as a polygon with (constant) 1000 points, if the region is very small (by definition each box along the line is small 3 x line_width pixels, default 3 max 20) each segment length between points can be as low as 0.01 pixels.  This seemed like overkill to me, so I set a minimum length of 0.1 pixel so there are fewer polygon points to convert to the matched image. (Perhaps still too low?  Any suggestions for the appropriate interval?)

1/10th pixel size as interval sounds sensible. 👍 I will retest the new implementation.

kswang1029:
@pford ok I have re-tested the latest commit 
8d7bb1d with a flat-field image (1920x1920) and a wide-field image (16384x16384). 
flat
S255_IR_sci spw29 cube I pbcor fits-image-2024-03-05-14-47-15
wide
cluster_16384 fits-image-2024-03-05-14-47-39

I made two copies of each and loaded them with one as the reference image and the other as the matched image.
The performance is indeed improved SIGNIFICANTLY. 
Screenshot 2024-03-05 at 14 39 13
One thing I spotted is for the flat case, with this PR, on matched image the performance is better than than on the reference image? Is this expected? 

I also checked the profile values between the reference image and the matched image using this PR. In the 2nd and 3rd tab of https://docs.google.com/spreadsheets/d/1JmHmKlDOXWoTDSnW8NAbyOwUbYtwjmCyUDPzt3vG80c/edit?usp=sharing I computed the relative errors using the reference image as the reference. For the wide field case, we see in general the relative error is less than 5%. We do see errors ~ 10% or more. I suspect these errors happened at the kinks of the polyline region. Not sure if applying 1/20th or 1/100th sampling could improve this error? On the flat field image, the error is zero as it should be.

acdo2002 (commented):
@pford please check my performance tests report.
[Here](https://docs.google.com/spreadsheets/d/11Ail7emcHY4Saf5jtkjwa2XH06x_GsIRGeae3wKZkWM/edit?usp=sharing) is the performance tests report.

I compare your PR branch and current dev branch and found:
1. "PV generator"
If looking at the protobuf response time, there is no significant difference in the PV generator.
2. "Set Spatial Profiler"
There is a significant improvement in 
SET_SPATIAL_REQUIREMENTS ==>
<== SPATIAL_PROFILE_DATA
Set Spatial Profiler has very significant improvement in pancake-like file (i.e. x and y size is much larger than z), 3-4 times better in your PR branch.
3. "PV preview"
Even the ICD-RxJS tests of `PV_PREVIEW.test.ts` has improvement.
(Step 4) PV request and response is almost the same, but
(Step 6) SET_SPATIAL_REQUIREMENTS and the following (Step 7) moving the region and update PV_PREVIEW DATA & SPATIAL_PROFILE_DATA have improved.

Please check whether  this performance is consistent to your expect? If yes then let me know I will approve this PR. I also confirmed there is no regression effect in daily ICD-RxJS or performance tests.

pford:
@kswang1029 I improved the rotbox handling for the reference image (create the LCPolygon directly from the box corners computed from the control points and rotation), so its performance is better than for the matched image as expected.  However, I cannot reproduce the errors in the matched widefield spatial profiles you described (after my code change...).  Could you try again, and if you still see the large errors, can you post the widefield image and the region you used in the test?  Thanks!

kswang1029:
> @kswang1029 I improved the rotbox handling for the reference image (create the LCPolygon directly from the box corners computed from the control points and rotation), so its performance is better than for the matched image as expected.  However, I cannot reproduce the errors in the matched widefield spatial profiles you described (after my code change...).  Could you try again, and if you still see the large errors, can you post the widefield image and the region you used in the test?  Thanks!

Will re-test the new commit👍

kswang1029:
@pford the e2e tests detected a backend crash when requesting pv preview. The macOS crash log is attached.

> Thread 2 Crashed:
> 0   carta_backend                 	       0x10468e74c casacore::ArrayLattice<bool>::getAt(casacore::IPosition const&) const + 16 (ArrayLattice.tcc:167)
> 1   carta_backend                 	       0x10476e540 carta::PvPreviewCube::GetRegionProfile(casacore::Slicer const&, casacore::ArrayLattice<bool> const&, std::__1::function<void (float)>, std::__1::vector<float, std::__1::allocator<float>>&, double&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&) + 468 (PvPreviewCube.cc:177)
> 2   carta_backend                 	       0x1047e31c8 carta::RegionHandler::CalculatePvPreviewImage(int, int, bool, std::__1::shared_ptr<carta::PvPreviewCut>, std::__1::shared_ptr<carta::PvPreviewCube>, std::__1::function<void (float)>, CARTA::PvResponse&, carta::GeneratedImage&) + 2204 (RegionHandler.cc:1281)
> 3   carta_backend                 	       0x1047e0c4c carta::RegionHandler::CalculatePvPreviewImage(int, int, int, AxisRange&, bool, std::__1::shared_ptr<carta::Frame>&, CARTA::PvPreviewSettings const&, std::__1::function<void (float)>, CARTA::PvResponse&, carta::GeneratedImage&) + 3028 (RegionHandler.cc:1179)
> 4   carta_backend                 	       0x1047dfee4 carta::RegionHandler::CalculatePvImage(CARTA::PvRequest const&, std::__1::shared_ptr<carta::Frame>&, std::__1::function<void (float)>, CARTA::PvResponse&, carta::GeneratedImage&) + 1324 (RegionHandler.cc:1019)
> 5   carta_backend                 	       0x104816b60 carta::Session::OnPvRequest(CARTA::PvRequest const&, unsigned int) + 972 (Session.cc:1356)
> 6   carta_backend                 	       0x10482c38c carta::GeneralMessageTask<CARTA::PvRequest>::execute() + 28 (OnMessageTask.tcc:34)
> 7   carta_backend                 	       0x1048595d4 carta::ThreadManager::StartEventHandlingThreads(int)::$_0::operator()() const + 96 (ThreadingManager.cc:54) [inlined]
> 8   carta_backend                 	       0x1048595d4 decltype(static_cast<carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(fp)()) std::__1::__invoke<carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(carta::ThreadManager::StartEventHandlingThreads(int)::$_0&&) + 96 (type_traits:3918) [inlined]
> 9   carta_backend                 	       0x1048595d4 void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>&, std::__1::__tuple_indices<>) + 96 (thread:287) [inlined]
> 10  carta_backend                 	       0x1048595d4 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, carta::ThreadManager::StartEventHandlingThreads(int)::$_0>>(void*) + 192 (thread:298)
> 11  libsystem_pthread.dylib       	       0x19481bfa8 _pthread_start + 148
> 12  libsystem_pthread.dylib       	       0x194816da0 thread_start + 8

pford:
@kswang1029 I fixed the bugs introduced by the faster rotbox implementation.  Thanks for catching this!

Side note: these changes will be removed in the Region -> RegionConverter implementation (where LCRegion for the reference image is from casacore::Record for all regions, rather than directly from the control points and rotation for rotbox only).  Should we combine these PRs, since the same tests (verification and performance) need to be run for both?  With the original PRs, the changes in this PR were *outside* the Region and the changes in PR #1352 were mainly *inside* the Region, moving its code to some new files and fixing includes, so there was basically no overlap.  However, the latest changes here are in the Region class.

acdo2002 (approved):

kswang1029:
@pford with the latest commit ([ee4ce1b](https://github.com/CARTAvis/carta-backend/pull/1353/commits/ee4ce1bd90c51de27ddf905d25f7d26eefc963cb)) the pv preview crashing issue is fixed. 👍 

I tried again to see the difference of the derived spatial profiles from the reference image and the matched image. Still I see errors in the case of matched wide field image. I compared the values in the x axis and in the y axis, respectively. The x axis is identical in both flat and wide cases. The y axis is identical in the flat case. Only in the wide case we see errors in y axis.

Figure_1

 I attached the test images and the region file here: https://drive.google.com/file/d/1MvKlVjysO4_iUCrPanv-Z86DnfUzp7Ce/view?usp=sharing

I think this is something to do with the polygon approximation in the wide-field (with projection distortion) case. Are we using 1/10th sampling approach now? Maybe using a finer sampling factor can improve the error with cost of performance. Could you point me where I can adjust that factor so that I can perform experiments here directly? I would like to see if we can have errors less than 5% (there is no right answer on this though).

src/Region/Region.cc
```diff
@@ -638,36 +680,47 @@ std::vector<std::vector<CARTA::Point>> Region::GetApproximatePolygonPoints(int n
...
-        auto first_x(first_segment_point.x());
-        auto first_y(first_segment_point.y());
-
-        for (size_t j = 1; j < target_nsegment; ++j) {
-            auto length_from_first = j * target_length;
-            auto x_offset = dir_x * length_from_first;
-            auto y_offset = dir_y * length_from_first;
-            segment_points.push_back(Message::Point(first_x + x_offset, first_y + y_offset));
-        }
+    double min_length = 0.1;
```
pford: @kswang1029 here is where the minimum segment length between polygon points is set to 0.1 pixels.

pford: I commented out this length and the maximum error was still 10.25%.  Will look into implementation details particularly at the box corners.

pford:
@kswang1029 I do not see a correlation between errors and the pivot points of the polyline.  The polyline segments correspond to these profile indexes:

- segment 0-1 = profile 0 - 8702
- segment 1-2 = profile 8703 - 18915
- segment 2-3 = profile 18916 - 27377
- segment 3-4 = profile 27378 - 36486

kswang1029:
> @kswang1029 I do not see a correlation between errors and the pivot points of the polyline. The polyline segments correspond to these profile indexes:
> 
> * segment 0-1 = profile 0 - 8702
> * segment 1-2 = profile 8703 - 18915
> * segment 2-3 = profile 18916 - 27377
> * segment 3-4 = profile 27378 - 36486

With a horizontal line, with 0.1 sampling factor, I see a perfect fit of the two profiles. However, once a line segment has an angle, we start seeing errors. I also tried 0.001, and 0.00001 sample factors but still errors persist and do not become smaller.

kswang1029:
@pford I also see similar in v4.1-stable. So if you like, we can file a separate issue for this error issue?

pford:
@kswang1029 the issue was in the polygon approximation, which removes points on a nearly horizontal line but favors those near an integral pixel.  After fixing this (ignoring nearest pixel), the maximum error I got was ~3.35% with the minimum length 0.1.  Without minimum length, the error goes to 0 but time increases 6-7x for the matched region profile (5.1s to 34.1s for me).  Also was ~100ms faster in both cases (reference and matched) without the parallel loop for the profiles, so I took that out.

kswang1029:
> @kswang1029 the issue was in the polygon approximation, which removes points on a nearly horizontal line but favors those near an integral pixel. After fixing this (ignoring nearest pixel), the maximum error I got was ~3.35% with the minimum length 0.1. Without minimum length, the error goes to 0 but time increases 6-7x for the matched region profile (5.1s to 34.1s for me). Also was ~100ms faster in both cases (reference and matched) without the parallel loop for the profiles, so I took that out.

@pford Nice! With the minimum length 0.1, having an error of a few percent should be fine. We may consider to have a configurable parameter for users if higher precision (with cost of performance) is needed. But this is out of the current scope. 👍

kswang1029 (approved):

PR #1361, opened 2024-03-01
Remove copyright end date

confluence:
I have noticed that we haven't bumped the copyright years since 2022. This draft PR just updates the end date to 2024, but [recent guidance](https://hynek.me/til/copyright-years/) appears to be that this is legally unnecessary, and several large projects have switched to either removing the end date or not including any dates at all. So we could take this opportunity to do the same.


confluence:
We agreed in the meeting to remove the end date.

kswang1029 (approved):
looks good and no regression from e2e tests as expected.

pford (approved):

PR #1371, opened 2024-04-17
Fix hdf5 tile data after animation stops

### LINKED ISSUE #1368: odd small pixel tiles seen in hdf5 image after animation stops
**Describe the bug**
odd small pixel tiles seen in hdf5 image after animation stops

https://user-images.githubusercontent.com/20819712/167797705-a5d4049f-8406-4c49-b6ce-87d8421b79d4.mov

**To Reproduce**
1. load an hdf5 cube
2. start and stop animation playback repeatedly 
3. observe the image quality, especially near the NaN edges.

**Expected behavior**
No such image defect after animation playback stopped.

 - Browser version [e.g. 22]: 101
 - Backend branch [e.g. dev, v3b2 release]: v3b3 test electron app 
 - Frontend branch [e.g. dev, v3b2 release]: v3b3 test electron app

also seen in the v3b3 AppImage packages. 
also seen in dev
also seen in v3b2 macOS electron app

pford:
**Description**

Fixes #1368 
By default, image quality is decreased during animation for better speed.  For each channel, tile data for HDF5 images is requested from the tile cache, encoded so that NaN values may be restored, and then compressed.  However, the encoding step **changes** the cached tile data (passed with a shared_ptr) by replacing a NaN values in a block with the average of finite values in that block.  When the cached tile data is requested again after animation stops, for image data with a higher compression quality, the image data is distorted since some values are no longer NaN.  To fix this, the tile data is assigned to a new vector for encoding and compression, leaving the cached data unchanged.

No
Zoom in an HDF5 image to an area with NaN values, then start and stop animation.  There should be no defect in the image after the animation is stopped.

**Checklist**

### DIFF
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -501,9 +501,7 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
             return true;
         } else if (compression_type == CARTA::CompressionType::ZFP) {
             // GetNanEncodingsBlock changes tile data (cached in TileCache)!
-            std::vector<float> tile_data_to_compress;
-            tile_data_to_compress.assign(tile_data_ptr->begin(), tile_data_ptr->end());
-            auto nan_encodings = GetNanEncodingsBlock(tile_data_to_compress, 0, tile_width, tile_height);
+            auto nan_encodings = GetNanEncodingsBlock(*tile_data_ptr, 0, tile_width, tile_height);
             tile_ptr->set_nan_encodings(nan_encodings.data(), sizeof(int32_t) * nan_encodings.size());
 
             if (ZStokesChanged(z, stokes)) {
@@ -516,7 +514,7 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
             std::vector<char> compression_buffer;
             size_t compressed_size;
             int precision = lround(compression_quality);
-            Compress(tile_data_to_compress, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
+            Compress(*tile_data_ptr, 0, compression_buffer, compressed_size, tile_width, tile_height, precision);
             float compression_ratio = (float)tile_image_data_size / (float)compressed_size;
             bool use_high_precision(false);
 
@@ -524,8 +522,7 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
                 // re-compress the data with a higher precision
                 std::vector<char> compression_buffer_hq;
                 size_t compressed_size_hq;
-                Compress(
-                    tile_data_to_compress, 0, compression_buffer_hq, compressed_size_hq, tile_width, tile_height, HIGH_COMPRESSION_QUALITY);
+                Compress(*tile_data_ptr, 0, compression_buffer_hq, compressed_size_hq, tile_width, tile_height, HIGH_COMPRESSION_QUALITY);
                 float compression_ratio_hq = (float)tile_image_data_size / (float)compressed_size_hq;
 
                 if (compression_ratio_hq > 10) {
@@ -584,7 +581,7 @@ bool Frame::GetRasterTileData(std::shared_ptr<std::vector<float>>& tile_data_ptr
         loaded_data = _loader->GetDownsampledRasterData(tile_data, _z_index, _stokes_index, bounds, mip, _image_mutex);
     } else if (!_image_cache_valid && _loader->UseTileCache()) {
         // Load a tile from the tile cache only if this is supported *and* the full image cache isn't populated
-        tile_data_ptr = _tile_cache.Get(TileCache::Key(bounds.x_min(), bounds.y_min()), _loader, _image_mutex);
+        tile_data_ptr = _tile_cache.GetCopy(TileCache::Key(bounds.x_min(), bounds.y_min()), _loader, _image_mutex);
         if (tile_data_ptr) {
             return true;
         }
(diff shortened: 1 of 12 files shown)


pford:
Not sure if this should be code reviewed or tested first since it involves copying data, which we try to avoid.

confluence:
In this case I think copying is unavoidable -- the compression function should not be modifying data in the tile cache!

I do remember adding a tile pool to the tile cache to avoid constantly creating and destroying vector objects, since this was causing quite a significant slowdown. I wonder if a tile pool would be appropriate here as well (I think that we should test the performance impact of this change first).

If yes, then we could either create a new tile pool to be used by this function, or add a `GetCopy` function to the tile cache which performs the copy internally and uses the tile cache's own tile pool (we may then need to adjust the tile cache pool size when we start and stop the animation, but maybe the tile requirements aren't going to exceed the default size, which IIRC is enough to fit one row and one column of 2x2 tile chunks).

confluence:
Actually, now I see that the current change is in `FillRasterTileData`, which means that the copy is being performed regardless of how the data was obtained in `GetRasterTileData` (which means that it will also affect cases where the data does not need to be copied). It's `GetRasterTileData` which should be modified -- there's already a new vector which is created there (for the cases which don't use the tile cache), which could be used to copy the cached data.

Perhaps adding a new tile pool in `GetRasterTileData`, and using it in all cases instead of creating a new tile vector, would be a good idea.

pford:
@confluence I added TileCache::GetCopy.  If I added a TilePool to Frame, wouldn't it have the same problem as the TileCache pool?

confluence:
> @confluence I added TileCache::GetCopy.

Currently `GetCopy` is creating a new vector. I was originally suggesting using the tile cache's pool to recycle a vector object to use for the copy (i.e. using `_pool->Pull()` to get a vector pointer).

> If I added a TilePool to Frame, wouldn't it have the same problem as the TileCache pool?

I'm not sure what problem you mean here.

I think I've confused the issue by conflating multiple suggestions -- for clarity, I'm talking about two separate things:

1. In `GetRasterTileData` we already create a new vector to read data into, in the cases that don't use the tile cache. I would suggest that we use this vector as well in the tile cache case, for consistency, instead of creating a vector in a different place. Perhaps the `GetCopy` function should be called `CopyInto` and take that vector as an input/output parameter -- but maybe this function isn't necessary at all, and the copy can just be inlined in `GetRasterTileData`. (I originally suggested the tile cache function to enable the use of the tile cache's pool, before I realised that the tile cache case could be integrated more generically with the other cases.)

2. Additionally, as a performance improvement, I think it would be a good idea to use a tile pool for that new vector in `GetRasterTileData`, instead of repeatedly creating and destroying it. My updated suggestion for this is to add a separate tile pool object to the frame, to be used here in all the cases, not just the tile cache case. I haven't checked if there are other places in the frame where a new tile vector is created (which could make use of the same pool).

confluence:
@pford I was suggesting to use only a tile _pool_ in the frame for all the tiles, not the tile _cache_ -- we only added the full-resolution tile cache to HDF5 files because in the HDF5 schema the main dataset is chunked, which makes it efficient to read a chunk (which is 2x2 tiles) at a time. For FITS files and other formats, reading a chunk at a time is not efficient, which is why we still use the full channel cache for those formats. I don't think it makes sense to copy tiles read from the full channel cache into the tile cache (this is currently bypassing the tile cache's tile pool, and I believe that downsampled tiles are also being saved -- this cache is only written to handle full-resolution tiles).

I am suggesting only that the frame gets its own tile pool object (separate to the tile cache's tile pool), and uses that pool to recycle the tile pointers (instead of creating new pointers with newly allocated data). This wouldn't be caching any tile data; just keeping a buffer of reusable objects allocated in memory. They would be overwritten with new data every time they are used (so it would be fine to use them for all tiles in `GetRasterTileData`, whether they're downsampled or not).

So the frame would have a `std::shared_ptr<TilePool> _pool`, with some reasonable capacity (how many tiles are likely to be allocated at the same time during an animation?), and in `GetRasterTileData` instead of creating a new `tile_data` vector and later a new pointer with `make_shared`, we'd get a pointer from the pool with `_pool->Pull()`, and write to that pointer's data. The pool class handles the recycling internally (with a custom deleter).

So it would be just this small change, plus the copying in the tile cache case.

pford:
I added a TilePool to Frame with the capacity set to the number of omp threads, since Session fills the tiles in parallel using omp.  The TilePool is created the first time tiles are requested, since tiles are not used when Frame holds the downsampled image for PV preview.

src/Frame/Frame.cc
```diff
@@ -480,6 +483,11 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
         raster_tile_data.clear_tiles();
     }
 
+    if (!_tile_pool) {
+        _tile_pool = std::make_shared<TilePool>();
+        _tile_pool->Grow(num_threads);
+    }
```
confluence: This isn't threadsafe, surely -- multiple threads may try to create the pool at once.

`Grow` doesn't allocate any memory -- it just sets the capacity, which is the upper bound of tiles to keep. `Pull` creates tiles on demand if there are no recycled tiles available in the pool. So it should be fine to create the pool and set its capacity in the constructor.

confluence (commented):
Other than the place the pool is created, I think this is good now. Using the number of threads to calculate the pool capacity makes sense. Perhaps it should be the number of threads plus one or two, to avoid race conditions in which a new tile is needed before any tiles have been returned to the pool?

confluence (approved):

confluence:
@kswang1029 in addition to testing the bug fix, please check the performance impact for animation (of both HDF5 files and other file types). I expect this to slightly decrease performance for HDF5 files but also slightly increase performance across the board, and I'd like to find out what the overall impact is (and to make sure that something unexpected isn't happening).

kswang1029 (approved):
@pford this looks good. @confluence As for performance test, please refer to the following table. I test the performance with a custom e2e test which has a 30s animation playback and then I count the stopped channel index and hence an average fps. I do not see obvious performance improvement/degradation. 

Screenshot 2024-05-20 at 14 26 08

confluence:
@kswang1029 @pford in that case I'm happy to merge this.

kswang1029:
> @kswang1029 @pford in that case I'm happy to merge this.

the effect (if any) _may_ be more prominent using "large" cube but I do not have the resources to test with, unfortunately.

PR #1378, opened 2024-05-22
Fix save image critical bug

### LINKED ISSUE #1377: [CRITICAL BUG] backend removes the target directory if we save an image without a filename
**Describe the bug**
If we save an image without giving a filename in the file browser dialog and click the save button, there will be an error toast showing the target saving directory not found. But then the original target directory is actually gone. This is a critical bug as user data will be lost.

**To Reproduce**

TEST WITH CAUTIONS!

1. create a test directory call "test_save_image" somewhere in the directory tree (eg `~/carta_image_pool/test_save_image`)
2. load an image
3. use the save image dialog to save it _with_ a filename in `~/carta_image_pool/test_save_image`. This should be successful and we do see a new file is created in the target directory.
4. now use the save image dialog to save it once again but with giving a filename (ie empty filename). Then we will see an error toast and if we check the existence of `~/carta_image_pool/test_save_image`, it is gone!

**Expected behavior**
We should apply to fixes:
1. at the frontend, if the target saving filename is gone, disable the save button
2. at the backend, if the target saving filename is an empty string, throw error messages and do nothing saving.

 - OS [e.g. macOS Monterey]: macOS Ventura
 - Browser [e.g. chrome, safari, electron app]: chrome, electron app
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev, v4.1-stable
 - Frontend branch [e.g. dev, v3b2 release]: dev, v4.1-stable

pford:
**Description**

#1377 

Saving the image fails if:
  - SaveFile filename is empty (always fails)
  - the resolved path is the same as the source image (always fails)
  - the resolved path is a non-image directory, to prevent deletion/overwrite of existing directory (always fails)
  - the resolved path is a non-image file and overwrite is false
  - the resolved path is a symlink and overwrite is false (else overwrites the link itself, not its target)
  - the resolved path is an image file or directory and overwrite is false

  Exporting regions fails if:
    - the resolved path is a directory
    - the resolved path is an existing file and overwrite is false

This should be backported to carta 4.2.0.

CARTAvis/carta-frontend#2389, CARTAvis/carta-protobuf#97

See issue description.  As stated, use caution!!! Copy test image to new directory.  Create a region and export to an existing file: should fail unless frontend dialog confirms overwrite.

**Checklist**

- [x] protobuf version bumped / ~protobuf version not bumped~
- [ ] Backport to 4.2

### DIFF
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1956,13 +1956,6 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
         }
     }
 
-    if (output_filename.string() == in_file) {
-        message = "The source file can not be overwritten!";
-        save_file_ack.set_success(success);
-        save_file_ack.set_message(message);
-        return;
-    }
-
     double rest_freq(save_file_msg.rest_freq());
     bool change_rest_freq = !std::isnan(rest_freq);
 
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1923,10 +1923,25 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
     bool success(false);
     casacore::String message;
 
+    if (output_filename.empty()) {
+        message = "Cannot save image with no filename.";
+        save_file_ack.set_success(success);
+        save_file_ack.set_message(message);
+        return;
+    }
+
     // Get the full resolved name of the output image
     fs::path temp_path = fs::path(root_folder) / directory;
     fs::path abs_path = fs::absolute(temp_path);
     output_filename = abs_path / output_filename;
+
+    if (fs::exists(output_filename) && fs::is_directory(output_filename) &&
+        (CasacoreImageType(output_filename.string()) == casacore::ImageOpener::UNKNOWN)) {
+        message = "Cannot overwrite existing directory.";
+        save_file_ack.set_success(success);
+        save_file_ack.set_message(message);
+        return;
+    }
 
     if (output_filename.string() == in_file) {
         message = "The source file can not be overwritten!";
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1939,9 +1939,9 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
     if (fs::exists(output_filename)) {
         if (output_filename.string() == in_file) {
             message = "Cannot overwrite the source image.";
-	} else if (CasacoreImageType(output_filename.string()) == casacore::ImageOpener::UNKNOWN) {
+        } else if (CasacoreImageType(output_filename.string()) == casacore::ImageOpener::UNKNOWN) {
             // Not an image
-	    if (fs::is_directory(output_filename)) {
+            if (fs::is_directory(output_filename)) {
                 message = "Cannot overwrite existing directory.";
             }
         } else if (!overwrite) {
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1949,10 +1949,12 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
             } else if (!overwrite) {
                 // Only overwrite file or symlink with confirmation
                 message = "Cannot overwrite existing file or symlink.";
+                save_file_ack.set_overwrite_confirmation_required(true);
             }
         } else if (!overwrite) {
             // Only overwrite image with confirmation
             message = "Cannot overwrite existing image.";
+            save_file_ack.set_overwrite_confirmation_required(true);
         }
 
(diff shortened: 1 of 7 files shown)


kswang1029:
@pford I tested this briefly with https://github.com/CARTAvis/carta-frontend/pull/2389. Please feel free to change this PR as a formal one when you feel ready.

src/Frame/Frame.cc
```diff
@@ -1916,18 +1916,46 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
...
+        }
+
+        if (!message.empty()) {
+            save_file_ack.set_success(success);
+            save_file_ack.set_message(message);
+            return;
+        }
+    }
+
     if (output_filename.string() == in_file) {
```
confluence: Is this part redundant now? Are there any circumstances under which this could ever be triggered if the file *doesn't* already exist? I guess someone could delete the file on disk while it is open in CARTA. But in that case we wouldn't be overwriting anything, right?

confluence: (This is referring to the `if (output_filename.string() == in_file) {` block below the added code.)

pford: Good catch, this block was mistakenly left in.  I removed the redundant check.

confluence:
I wonder if we should also require confirmation for overwriting a non-image ordinary file -- at the moment it seems a little unexpected / inconsistent that we'll silently overwrite it. I would suggest:

Any image OR an unrecognised ordinary file -> requires confirmation
Unrecognised directory -> never overwritten

This would mean that either we would need to issue the same error for an existing image or an existing file, or (this is probably better) the frontend would need to check for an additional error message (so that we could use two different ones, for clarity).

pford:
Ok, I understood your issue comment "I think non-image files are OK to overwrite" as _without_ confirmation, but you meant _with_ confirmation.  I agree with your suggestion and will make this change.  I just realized I did not handle symlinks, which may resolve to a directory, file, image, or outside the root folder.  Besides needing write permissions, where are we allowed to write?

We do not have any checks on exporting region files except checking that an existing file is writable, or can create a new file.  I wonder if these issues exist there too? Uh oh -- I will check.

confluence:
Yes, I was thinking of casa images being a special case of a directory being treated like a file (and otherwise consistently overwriting files only with confirmation, and not overwriting directories).

A symlink should be overwriteable (with confirmation), since overwriting a symlink won't touch whatever it was pointing to. But maybe that warrants a *third* message.

Good point about the regions. I haven't been able to reproduce the same directory overwriting issue: I can't save a region file with a blank name; and if I type in the name of an existing directory, I don't get an error, and briefly see a success message, but nothing seems to happen (the directory is not deleted, and the file is not saved anywhere). I *can*, however, overwrite an ordinary non-region file without confirmation, so that is a problem. I'd say that we need to replicate the same logic there (including the additional flag for that request).

That's the only other place where we save files to the server, right? All the other save / export operations save to the client, which is the browser's problem.

confluence:
@pford we still need to add an overwrite flag to the export region request, and handle that in the same way as we do when saving images, right?

src/Region/RegionHandler.cc
```diff
@@ -221,31 +221,43 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
...
                 return;
             }
+
+            if (export_file.isRegular(false)) {
+                if (!export_file.isWritable()) {
+                    message = "Export region failed: cannot overwrite read-only file.";
+                }
+            } else {
+                message = "Export region failed: path is not a file.";
+            }
```
confluence: I'm not sure about the logic here. Shouldn't this be `isRegular(true)`? But also we need different responses for directories and non-directories. This conditional can't be identical to the save file code because we're using casacore file functions and not filesystem paths (and also the image detection is not applicable), so maybe something like:

* if file exists
  * if it's a directory -> error about directory (never overwritten)
  * if it's a regular file or symlink
    * if no overwrite flag -> error about file or symlink (need confirmation)
    * else if not writeable -> error
  * otherwise -> error about it not being a file
* else if it can't be created... (as before)

pford: I was trying to simplify the logic here, since it always fails if overwrite is false (no need to check the type) or if the target is not a file (only need to check type is file).  Currently, the frontend only shows the error and does not respond with a confirmation dialog regardless of the error.  But I will change it if you think additional information would be helpful.

confluence (approved):
I think that the code is fine. Now we need to test integration with the frontend, which should retry both saving a file and exporting regions if it receives a specific error message.

kswang1029 (approved):
looks good.

confluence (approved):

kswang1029 (approved):
looks good. no regression from e2e.

PR #1381, opened 2024-06-14
Optimisation of directory type checking in file list

confluence:
This is an attempt to improve the performance of generating a file list for directories with lots of subdirectories. The main performance bottlenecks in the current code are:

1. Checking whether a directory is a known image format, and should be treated as a file,
2. Counting the children of a subdirectory.

There are three options for displaying file lists: filtering by content, filtering by file extension, and no filtering. Filtering by content additionally slows down the file list generation because it reads the signature of each ordinary file to determine the type. Guessing the type from the file extension speeds up the processing of ordinary files, but not subdirectories, which are always checked with casacore's `ImageOpener`. Turning off filtering entirely has no additional performance impact on the image file list (because detection of directory-based images is still required), but speeds up the region file list (because if no filtering is required, all directories can be shown without additional checks).

This PR is an attempt to add a less expensive heuristic for directory-based images, to be used when the option to filter by content is not selected. It performs almost the same checks as `ImageOpener`, but only for directories, by looking for files inside the directory with `fs::exists`, and without distinguishing between different CASA image subtypes.

The existing code assumes that there are directory image formats which we do not support, and handles them differently. However, it's clear from the `ImageOpener` code that the GIPSY format is a pair of files, not a directory (so `ImageOpener` would never return that type for a directory), and the `CAIPS` and `NEWSTAR` types are obsolete and never returned by `ImageOpener` (at all). So I have removed this option from the code, and not implemented it in the alternative code.

The result: the alternative code appears to be slightly faster, but I don't know if it's faster enough for it to make sense to add it as an alternative to the casacore code. If this implementation is sufficient for our purposes (e.g. we don't need to read the `table.info` file because we don't need do distinguish between CASA sub-types here), then perhaps we should *replace* the casacore check with this (for a modest speed improvement in all cases).

Other optimizations we discussed:

* Not counting subdirectory children
* Disabling directory type-checking entirely (if not filtering by content?), and changing the frontend UI to fetch the type information when a directory is clicked on
* Progressively requesting info for batches of directories as the user scrolls (this would require a more substantial redesign, with ICD changes)

I think we're planning to use the last option as our long-term solution, and I would suggest applying that strategy to all files: instead of loading file information up-front when generating the file list, we could initially return *just* a bare list of files and directories, and then return information for lists of files *and* directories as the frontend requests them.

**Checklist**

### DIFF
--- a/src/Util/File.cc
+++ b/src/Util/File.cc
@@ -9,6 +9,7 @@
 #include <spdlog/fmt/fmt.h>
 #include <fstream>
 
+#include "Casacore.h"
 #include "String.h"
 
 uint32_t GetMagicNumber(const std::string& filename) {
@@ -95,6 +96,38 @@ CARTA::FileType GuessImageType(const std::string& path_string, bool check_conten
             return CARTA::FITS;
         } else if (HasSuffix(filename, ".hdf5")) {
             return CARTA::HDF5;
+        }
+    }
+
+    return CARTA::UNKNOWN;
+}
+
+CARTA::FileType GuessImageDirectoryType(const std::string& path_string, bool check_content) {
+    if (check_content) {
+        // Guess file type using casacore::ImageOpener
+        auto image_type = CasacoreImageType(path_string);
+
+        switch (image_type) {
+            case casacore::ImageOpener::AIPSPP:
+            case casacore::ImageOpener::IMAGECONCAT:
+            case casacore::ImageOpener::IMAGEEXPR:
+            case casacore::ImageOpener::COMPLISTIMAGE:
+                return CARTA::CASA;
+            case casacore::ImageOpener::MIRIAD:
+                return CARTA::MIRIAD;
+            default:
+                return CARTA::UNKNOWN;
+        }
+    } else {
+        // Guess file type by checking for files inside directory
+        fs::path dir(path_string);
+
+        if (fs::exists(dir / "table.info") || fs::exists(dir / "imageconcat.json") || fs::exists(dir / "imageexpr.json")) {
+            return CARTA::CASA;
+        }
+
+        if (fs::exists(dir / "image") && fs::exists(dir / "header")) {
+            return CARTA::MIRIAD;
         }
     }
 
--- a/src/Util/File.cc
+++ b/src/Util/File.cc
@@ -10,7 +10,6 @@
 #include <fstream>
 #include <regex>
 
-#include "Casacore.h"
 #include "String.h"
 
 uint32_t GetMagicNumber(const std::string& filename) {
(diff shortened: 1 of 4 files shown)


pford (approved):

kswang1029:
I did a test with a directory with 20000 fits image files, and a directory with 20000 casa image directories. The text computer is a local build with a fast SSD. I did purge disk cache for each test case. 

The test results (in seconds) are summarized in the following table:
Screenshot 2024-07-01 at 11 06 30

We see improvements of ~2x for the casa image with the "filter by extension" and the "all files" modes. Is this expected in this improvement and consistent with what you tested and observed?

The psrecord plots are attached below as a record. The first plateau is casa image and the second is fits image.

old - content
old_content

new - content
new_content

old - extension
old_extension

new - extension
new_extension

old - all
old_all

new - all
new_all

It appears to me that once we access the context of a file, the CPU usage won't be 100% in my test case.

pford (changes requested):
I just found a place in ProgramSettings.cc:338 which uses casacore::ImageOpener to check whether the positional folder argument is a CASA image to be opened.  This should also use GuessImageDirectoryType.

src/Util/File.cc
```diff
@@ -10,6 +10,7 @@
 #include <fstream>
 #include <regex>
 
+#include "Casacore.h"
```
pford: If this include necessary?

pford (changes requested):

src/Util/File.cc
pford: Asking because I am looking at a file list bug and might include File.h in Casacore.h.

confluence:
Update: I'm currently incorporating the code feedback, and testing the performance on Ilifu to confirm that this provides a satisfactory speedup when the file *metadata* is stored in fast storage even when the file *content* is stored in slower storage (because it is no longer necessary to access the file content). My current intention is to *replace* the previous heuristic with this one, assuming that this doesn't cause any regressions.

confluence:
I have tested this on Ilifu (on a partition where the data is in slow storage but the metadata is in fast storage), and unfortunately it kind of makes things worse! In the release version, the first time I load a directory of 1000 subdirectories with 1000 empty subdirectories each, the first access is extremely slow (as expected), but subsequent accesses are instantaneous (I believe because the directory data loaded by Casacore is being cached by the OS). In this branch, the first access is still quite slow but noticeably faster (maybe by an order of magnitude), but every subsequent access takes the same amount of time (because there's no caching benefit). So it looks like this change would cause a regression in performance for our users on Ilifu, and should *not* be merged.

I'm going to leave this draft open while I investigate further, in case I've missed something, but we should probably proceed with our longer-term plan to load directory information on demand in a more scalable way.

confluence:
I'm closing this; I don't think this is going to be a useful intermediate step.

PR #1390, opened 2024-11-01
Minor refactoring of polarization values and defaults in the slicer function 

confluence:
**Description**

This is a set of refactoring changes:

* Removing redundant lookups of integers with same integer enums
* Using correct protobuf function to cast integers to integer enums
* General cleanup of polarization values (descriptions; CASA and FITS conversion...)
* Replacing unnecessarily complex and inefficient ordering of small unordered maps with polarization keys (these can just be ordered maps)
* Precalculating fixed ranges for full X, Y and Z and using the X and Y ranges as defaults to GetImageSlicer instead of using a placeholder default which has to be checked and possibly replaced whenever the function is called
* Reimplementing functions for mapping polarization index to polarization value and vice versa

**Checklist**

- [X] ~changelog updated~ / no changelog update needed

### DIFF
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -2320,8 +2320,8 @@ casacore::Slicer Frame::GetExportRegionSlicer(const CARTA::SaveFile& save_file_m
 bool Frame::GetStokesTypeIndex(const string& coordinate, int& stokes_index) {
     // Coordinate could be profile (x, y, z), stokes string (I, Q, U), or combination (Ix, Qy)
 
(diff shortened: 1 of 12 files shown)


pford (approved):
I did not realize the protobuf enums had these built-in functions.  We should check if this can be used in other refactoring!

kswang1029 (approved):
all e2e tests are passing. regular Stokes cubes are fine (including hypercubes). rotated cubes are fine in switching IQUV (will crash if switching computed components).

PR #1392, opened 2024-11-08
Aggregate information about image axes and dimensions into structs

confluence:
**Description**

This is a refactoring PR which aggregates information about image axes and dimensions into two structs. This simplifies passing the information around and storing it as properties: we can use the structs instead of inconsistent subsets of individual axes and dimensions.

Major changes:

* `FileLoader::::FindCoordinateAxes` signature has changed; it now *only* sets `_axes` and `_dims` on the loader and does not also modify return parameters (but it's still a `bool` and may set the message parameter).
* Individual axes properties in `FileLoader` and `Frame` replaced with structs. Individual getters have been kept; getters for the structs have been added.
* Calculation of dims from axes and shape moved to dims struct constructor to enable cleaner reuse from multiple places where axes can be calculated.
* Frame, HDF5 loader, polarization calculator, Stokes file connector, extended file info loader updated to use axes and/or dims structs.

There are two non-refactoring changes:

1. While I was making this change I noticed that previously the spectral axis of a compressed FITS file was being set to the depth axis, and have fixed this. At the moment compressed FITS files with rotated axes can't be viewed at all for unrelated reasons; I will see how easy that would be to fix, and possibly factor these changes out into a separate PR.
2. `FileLoader::::FindCoordinateAxes` now explicitly bails out with an error message if there are fewer than 2 render axes (e.g. in a weird case like a 2D image with one Stokes axis). Previously I believe that this would have crashed. Maybe we should give images like this a second render axis with size 1 (like we handle images with no depth or Stokes), but that would be a much more complicated change (and it may not be something we need).

Possible outstanding issue

* There are hardcoded axes in Frame's image export functions. Should these be updated to support rotated axes?

**Checklist**

### DIFF
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -713,8 +713,6 @@ void FileExtInfoLoader::AddInitialComputedEntries(const std::string& hdu, CARTA:
 
     if (compressed_fits) {
         compressed_fits->SetShape(shape);
-        compressed_fits->SetSpectralAxis(axes.spectral); // TODO: this was set to depth axis previously -- which is correct???
-        compressed_fits->SetStokesAxis(axes.stokes);
     }
 }
 
@@ -1201,6 +1199,8 @@ void FileExtInfoLoader::AddComputedEntriesFromHeaders(
         }
         entry->set_entry_type(CARTA::EntryType::STRING);
     }
+
+    std::cout << "+++++ Before  if (compressed_fits); projection is " << projection << std::endl;
 
     if (compressed_fits) {
         casacore::CoordinateSystem coordsys;
--- a/src/FileList/FileExtInfoLoader.cc
+++ b/src/FileList/FileExtInfoLoader.cc
@@ -204,16 +204,17 @@ bool FileExtInfoLoader::FillFileInfoFromImage(CARTA::FileInfoExtended& extended_
 
                 AddDataTypeEntry(extended_info, data_type, equivalent_type);
 
-                std::vector<int> spatial_axes, render_axes;
-                int spectral_axis, stokes_axis, depth_axis;
-                if (_loader->FindCoordinateAxes(image_shape, spatial_axes, spectral_axis, stokes_axis, render_axes, depth_axis, message)) {
+                if (_loader->FindCoordinateAxes(message)) {
+                    auto image_shape = _loader->GetShape();
+                    auto axes = _loader->GetAxes();
+
                     casacore::Vector<casacore::String> axes_names;
-                    AddShapeEntries(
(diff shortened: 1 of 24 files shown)


src/FileList/FileExtInfoLoader.cc
```diff
@@ -712,12 +709,12 @@ void FileExtInfoLoader::AddInitialComputedEntries(const std::string& hdu, CARTA:
...
     AddDataTypeEntry(extended_info, data_type, equivalent_type);
 
-    AddShapeEntries(extended_info, shape, spatial_axes, spectral_axis, stokes_axis, render_axes, depth_axis, axes_names);
+    AddShapeEntries(extended_info, shape, axes, axes_names);
 
     if (compressed_fits) {
         compressed_fits->SetShape(shape);
-        compressed_fits->SetSpectralAxis(depth_axis);
-        compressed_fits->SetStokesAxis(stokes_axis);
+        compressed_fits->SetSpectralAxis(axes.spectral); // TODO: this was set to depth axis previously -- which is correct???
```
pford: There is no need to set spectral and stokes axis in CompressedFits, these values are not currently used.  I commented out the lines here as well as the set/get methods in CompressedFits and it still worked. The shape, however, is used later for extended file info computed entries.  For performance reasons, all file browser info is from the headers only (using cfitsio), rather than opening the image which decompresses the file into memory.

pford (changes requested):
Remove spectral and stokes axis from CompressedFits

pford:
I could not find anything that was missed.
I can fix the Save Image bug for issue #1178 (additional support for rotated cubes).

src/FileList/FileExtInfoLoader.cc
pford: So at some point these axes needed to be saved but is now unnecessary.

pford (approved):

kswang1029 (changes requested):
Loading individual Stokes images as a hypercube does not work. This is picked up by e2e. Requesting @pford a re-review once addressed.
selenium-screenshot-5

The test images in the screenshot all have `Shape = [256, 256, 480, 1] (RA, DEC, FREQ, STOKES)`. 

Other tests images with `Shape = [256, 256, 480] (RA, DEC, FREQ)` or `Shape = [6280, 3140] (GLON, GLAT)` work however.

confluence:
I think I found the issue. I used the same test files here and it seems to be fixed -- @kswang1029 can you confirm?

~CI is broken for unrelated reasons; merging #1396 should fix it.~

pford (approved):

kswang1029 (approved):
looks fine now. 👍

PR #1398, opened 2025-01-27
Prevent segfault in casacore log sink destructor

### LINKED ISSUE #1382: Backend segfaults if the starting dir isn't contained in the top level folder
**Describe the bug**
Unexpected segfault instead of just an error message when the starting dir isn't contained in the top level folder.

**To Reproduce**
```
ubuntu@carta-dave:~$ carta_backend --version
4.1.0
ubuntu@carta-dave:~$ carta_backend --top_level_folder /mnt /home
[2024-06-14 10:27:36.154Z] [CARTA] [info] Writing to the log file: /home/ubuntu/.carta-beta/log/carta.log
[2024-06-14 10:27:36.155Z] [CARTA] [info] /usr/bin/carta_backend: Version 4.1.0
[2024-06-14 10:27:36.155Z] [CARTA] [critical] Starting /home must be a subdirectory of top level /mnt. Exiting carta.
Segmentation fault (core dumped)
```
**Expected behavior**
An error message, but no segfault.

 - OS [e.g. macOS Monterey]: Ubuntu 22.04
 - Browser [e.g. chrome, safari, electron app]:  N/A
 - Browser version [e.g. 22]:  N/A
 - Backend branch [e.g. dev, v3b2 release]: 4.1.0
 - Frontend branch [e.g. dev, v3b2 release]: N/A

**Additional context**
Picked dirs here at random just to duplicate on a more "normal" box an issue originally detected on a minimalist Ubuntu 22 pod.

confluence:
**Description**

This fixes #1382. Testing is required to check if this is still doing the correct thing with casacore logs.

The segfault happens in the shared pointer destructor called from the `casacore::LogSink` destructor. I believe that this is because the shared pointer's underlying pointer is zeroed by `casacore::LogSink::globalSink` when it is used to set the global sink. This PR bypasses this issue by not reusing the same raw pointer both to set the global sink and to create a `LogSink` object which the `LogIO` object is then attached to. Instead, it uses the default `LogIO` constructor, which should attach the object to the global sink, which should be set to the correct sink.

I tested this locally and it appears that the casacore messages are still being formatted correctly in the output, but I'm not 100% familiar with the way that this logging model is supposed to work.

If there is no regression, I think that this PR is low-risk enough to be included in the freeze, but I'm creating it as a draft first just in case.

**Checklist**

### DIFF
--- a/src/Main/Main.cc
+++ b/src/Main/Main.cc
@@ -74,9 +74,8 @@ int main(int argc, char* argv[]) {
             default:
                 carta_log_sink = new CartaLogSink(casacore::LogMessage::NORMAL);
         }
-        casacore::LogSink log_sink(carta_log_sink->filter(), std::shared_ptr<casacore::LogSinkInterface>(carta_log_sink));
         casacore::LogSink::globalSink(carta_log_sink);
-        casacore::LogIO casacore_log(log_sink);
+        casacore::LogIO casacore_log;
 
         if (settings.wait_time >= 0) {
             Session::SetExitTimeout(settings.wait_time);
--- a/src/Main/Main.cc
+++ b/src/Main/Main.cc
@@ -75,7 +75,6 @@ int main(int argc, char* argv[]) {
                 carta_log_sink = new CartaLogSink(casacore::LogMessage::NORMAL);
         }
         casacore::LogSink::globalSink(carta_log_sink);
-        casacore::LogIO casacore_log;
 
         if (settings.wait_time >= 0) {
             Session::SetExitTimeout(settings.wait_time);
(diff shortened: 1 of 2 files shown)


pford (approved):
As stated in the description, these changes solve the reported bug and still send casacore log messages to the carta log for consistent formatting.

src/Main/Main.cc
```diff
@@ -74,9 +74,8 @@ int main(int argc, char* argv[]) {
             default:
                 carta_log_sink = new CartaLogSink(casacore::LogMessage::NORMAL);
         }
-        casacore::LogSink log_sink(carta_log_sink->filter(), std::shared_ptr<casacore::LogSinkInterface>(carta_log_sink));
         casacore::LogSink::globalSink(carta_log_sink);
-        casacore::LogIO casacore_log(log_sink);
+        casacore::LogIO casacore_log;
```
pford: LogIO not needed?

confluence: Good point; this object is actually never used.

kswang1029 (approved):

PR #1399, opened 2025-01-28
Clip rounded profile start to end after clipping rounded end to width / height

### LINKED ISSUE #1395: Backend crash due to 'n' exceeds maximum supported size after spatial matching of two images
**Describe the bug**
Using the two attached images, the following backend error caused the backend to crash after spatial matching.
```
libc++abi: terminating due to uncaught exception of type std::length_error: allocator<T>::allocate(size_t n) 'n' exceeds maximum supported size
zsh: abort      ./carta_backend --port 3002 --omp_threads 8 --debug_no_auth --no_browser  5
```

**To Reproduce**
1. Open VLASS3.1.ql.T27t14.J205320+683000.10.2048.v1.I.iter1.image.pbcor.tt0.subim.0__1862_1864_1860_1862.fits
2. Open NGC4321_MUSE_subcube_with_restfreq.fits
3. Turn on spatial matching
4. Scroll and maybe click to zoom until the backend crashes

**Screenshots or videos**
If applicable, add screenshots/videos to help explain your problem.

 - OS: macOS Sequoia
 - Browser: chrome
 - Backend branch [e.g. dev, v3b2 release]:
 - Frontend branch [e.g. dev, v3b2 release]:

confluence:
**Description**

This should fix #1395.

I believe that this bug is triggered by the spatially matched images being non-overlapping, which leads to a profile sometimes being requested from the large image with both endpoints being the largest y value (i.e. 999 to 999). When the image is zoomed out and a large mip is used, the endpoints are rounded for decimation (from 999 to 1000). The _end_ is then restricted to the height of the image (999), but we did not consider the possibility that the _start_ value could be equal to the end and thus _also_ need to be adjusted in the same way here.

This PR adds a line to restrict the start so that it does not exceed the end.

In my local test (on Ubuntu), the crash was happening in `profile.reserve(end - start);`. I'm not 100% sure if @crocka's crash on the Mac (a value being too large to fit in a `size_t`) was happening in exactly the same place, but it's likely to have had the same root cause (a negative number being incorrectly converted to an unsigned type).

If you cannot reproduce the crash in `dev`, try reducing the area available for the image viewer component (e.g. by opening the browser console in the window). It's only triggered when the mip is high enough (and I believe when your cursor is in the appropriate area of the image).

If there is no regression, I think this is a low-risk addition.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fix save image/export regions bug which could cause directory overwrite or deletion ([#1377](https://github.com/CARTAvis/carta-backend/issues/1377)).
 * Fix bug in cache slicer transformation which affects some images with rotated axes ([#1389](https://github.com/CARTAvis/carta-backend/pull/1389)).
 * Fix bug accessing top (root) folder of file browser ([#2354](https://github.com/CARTAvis/carta-frontend/issues/2354)).
+* Fix crash when spatially matching two images ([#1395](https://github.com/CARTAvis/carta-backend/issues/1395)).
 
 ### Changed
 * Move the loader cache to separate files ([#1021](https://github.com/CARTAvis/carta-backend/issues/1021)).
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1194,6 +1194,7 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
                     start = decimated_start * mip;
                     end = decimated_end * mip;
                     end = config.coordinate().back() == 'x' ? std::min(end, _width) : std::min(end, _height);
+                    start = std::min(start, end);
                 }
 
                 if (is_current_stokes) {


pford:
I tested on a Mac and got almost the same error in the dev branch: `libc++abi: terminating due to uncaught exception of type std::length_error: vector`.  This is fixed in the branch for this PR.

I find it odd that when the second image goes out of view due to spatial matching, the frontend does not update the cursor in the backend with the pixel values outside the image (x < 0, y > max_y), sends two spatial requirements for the matched image as shown below, then does not show the profiles that are sent with the image's last valid cursor (if no stored valid cursor, no profiles).  If the spectral profile is set to the active image, spectral requirements are requested and spectral data is sent for the last valid cursor (I assume, not in spectral data message) and shown.

1. Frontend sends spatial requirements {'x': start=0, end=0}  {'y': start=999, end=999} with mip=4 (image shape [1231, 999, 301]).  Backend resolves the x-range to full range and sends the x-profile for the last valid cursor, but no y-profile. (no error)
2. Frontend sends spatial requirements {'x': start=0, end=0}  {'y': start=0, end=0} with mip=1. Backend resolves both to the full range and sends x and y profiles for the last valid cursor.

pford (approved):

kswang1029 (approved):

PR #1409, opened 2025-03-05
Fix the coverage badge and dependencies in README.md

### LINKED ISSUE #1400: README issues
In https://github.com/CARTAvis/carta-backend/blob/dev/README.md the badge up top is broken - shows as "custom badge | resource not found"

And `libhdf5-cpp-100` seems to be slightly too version-specific a package name - on jammy I think `libhdf5-cpp-103-1` is the one that I want.  (Would honestly be nice to have the ability to copy/paste an apt install commands for all the installable dependencies, rather than having to copy/paste individual package names.

confluence:
**Description**

This PR fixes #1400.

1. The badge URL which we already have in the generated coverage report is parsed out of the report and downloaded, and the generated SVG file is stored in a dedicated `gh-storage` branch. The persistent path to this file can then be used in the README. This may be slightly overengineered for a single file, but we could potentially use this branch for persistent versioned storage of other assets.
2. It's unnecessary to mention `libhdf5-cpp-*` at all, because the `-dev` package will install it. But in any case we should not mention package names here, and link to the dockerfiles instead. I'm rewriting most of this README; a lot of it is incomplete or very outdated.

TODO:

- [X] Once the `gh-storage` branch exists, the fallback commands can be removed.
- [X] Badge update should only be triggered on pushes to dev, not on pull requests.
- [X] General README rewrite for accuracy and clarity

**Checklist**

- [X] ~changelog updated~ / no changelog update needed

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -107,15 +107,15 @@ jobs:
       - name: Store coverage badge in gh-storage branch
         shell: bash
         run: |
-          set -x
           git config --global user.email "backend.ci@github.action"
           git config --global user.name "Backend CI"
           git add badge-coverage.svg
           git stash
           git fetch origin gh-storage
           git checkout gh-storage
+          git pull
           git checkout stash@{0} -- badge-coverage.svg
-          git diff --staged --quiet || { git commit -m "Updated coverage" && git pull && git push; }
+          git diff --staged --quiet || { git commit -m "Updated coverage" && git push; }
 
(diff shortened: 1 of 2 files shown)


confluence:
@kswang1029 I added you just for a review of the README text (at the top I tried to make it clearer that these are docs for the backend only, and to direct people elsewhere if they're looking for desktop packages).

daikema (approved):

kswang1029 (approved):

PR #1413, opened 2025-03-17
Unit tests and documentation for Image, String, File, Casacore files in the Util subfolder

### LINKED ISSUE #1420: Add unit tests and documentation for Image, String, File, Casacore files in the Util subfolder
Add test coverage and documentation to the Image, String, File, and Casacore files in the Util sub-folder.
Closed with PR #1413

Michaelazzz:
**Description**

Unit test and documentation for the Image, String, File, and Casacore files in the Utils subfolder.
Adds test and documentation coverage to the codebase.
No.
No.

**Checklist**

- [x] new e2e test created
- [x] ICD test passing


confluence:
Just FYI, `AxisRange` was originally only used for the Z axis, which is why the default constructor is from 0 to `ALL_Z`, which doesn't really make sense in the current code. In future I think that the default range constructor should be from 0 to 0, and we should explicitly use one of the other constructors when we want something else. It may be possible to do this now, separately from the other refactoring related to the cache and Stokes calculations.

In the broader refactoring I am also planning to remove the `ALL_X` and `ALL_Y` constants. Unlike `ALL_Z`, they don't correspond to any external placeholder in a request from the protobuf API -- and internally we should never be using placeholders like this, otherwise it's very difficult to decipher when a range is referring to concrete values and when it contains placeholders that still have to be replaced, so there are repeated checks for this at multiple levels. They're also both set to the same placeholder value as `ALL_Z`, which makes it impossible to replace them without additional context. The only place where they currently can't easily be removed is the default constructor of the `StokesSource` object, which I'm also planning to remove together with the `StokesRegion` and `StokesSlicer` wrappers (this is one of the reasons why, apart from the general problem with constructing those wrappers too early, and the difficulty of detecting what the wrappers actually represent as they're being passed around). So I would skip writing tests for those.

General comments on this PR so far:

1. Run `./scripts/style.py all fix` to fix the formatting and pass the CI format check[^1]. If you're already doing that, your `clang-format` version may be too new -- you can install multiple versions in parallel and pass an explicit executable path to `style.py` with `--clang-format`. At the moment it looks like 14 is the latest version that is compatible using our current settings, but I'm going to have a look at fixing that (the 14 - 15 change is minor, and the modification it suggests is a good idea). The failing format check is blocking the CI that runs the tests on different systems.
2. I would suggest organising the test code in subdirectories matching the current `src` subdirectory structure. Putting `Util` in the test names is probably still a good idea, though.

Other than that, this is looking good -- this is exactly the kind of unit test coverage we need. 👍

[^1]: Stuff like this should be recorded in the [developer docs](https://cartavis.org/carta-backend/md_docs_2static_2faq.html). Please feel free to expand these if you find something that they don't address. Any markdown files you put in [docs/static](docs/static) will automatically be picked up. The docs are built from the dev branch; to build them locally, run `doxygen` in the root of the repo and open `docs/html/index.html`. IIRC the light/dark theme doesn't work unless you use a local webserver.

confluence:
Oh, also, these classes and functions should have doxygen comments added at the same time -- have a look at `TileCache` to see what format we're using. The comments only go in the header files (they should document the API, not the implementation).

pford:
We have two PRs #1424 and #1429 which make changes to Util files.  It would be nice to have these test files merged so we can update them in the other PRs.

confluence (changes requested):
I would like to address the commented-out tests before merging this. What is the current behaviour of the failing tests? Are all the commented-out tests failing?

Michaelazzz:
> I would like to address the commented-out tests before merging this. What is the current behaviour of the failing tests? Are all the commented-out tests failing?

Yes, all the commented out tests were failing and it's mainly to do with casacore and how it resolves symlink paths.
Also, the failing tests are not for one of the functions I removed casacore from. I can a add refactored GetResolvedFilename to PR #1424 with the additional tests, but it would be good to understand what behavior is expected before attempting to implement a refactor.

Michaelazzz:
> I would like to address the commented-out tests before merging this. What is the current behaviour of the failing tests? Are all the commented-out tests failing?

Yes all the commented-out tests were failing, but all the failing tests are related to issues with how casacore resolves the path to folders with it expandedPath function.

confluence:
@pford we've gone back and forth on these failing tests a bit -- we established that they're actually not related to the reimplementation in #1424, so they've been moved back here, and also per the discussion in #1430 the current behaviour of the function is correct and the tests should be modified (and the function documentation updated accordingly). After these changes this PR should be ready for final review.

test/Util/TestUtilCasacore.cc
```diff
@@ -233,15 +233,21 @@ TEST_F(GetResolvedFilenameTest, ResolvesNormalFile) {
...
-*
-* path.resolvedName() -> it is a known casacore issue that it fails to resolve symlinks
-* path.expandedName() -> if resolvedName() fails it tries expandedName() which returns the expanded name but neccessarily the resolved path
-*/
+TEST_F(GetResolvedFilenameTest, ResolvesToSameDirectory) {
+    auto pwd = TestRoot();
+	fs::path dir1 = temp_dir / "subdir";
+	fs::create_directory(dir1);
+    fs::path dir2 = temp_dir / "" / "subdir";
+    fs::create_directory(dir2);
```
confluence: This is not necessary -- this just depends on fs::path behaviour, which we aren't testing.

test/Util/TestUtilCasacore.cc
```diff
@@ -233,15 +233,21 @@ TEST_F(GetResolvedFilenameTest, ResolvesNormalFile) {
...
+	std::string message;
+	std::string result1 = GetResolvedFilename(temp_dir.string(), "subdir", "", message);
+    std::string result2 = GetResolvedFilename(temp_dir.string(), "subdir", "", message);
```
confluence: Different parameters should be passed here -- `"subdir", ""` and `"", "subdir"`

confluence (changes requested):

pford:
For GetResolvedFilename tests, std::filesystem::temp_directory_path uses `the one specified in the environment variables TMPDIR, TMP, TEMP, TEMPDIR, and, if none of them are specified, the path "/tmp" is returned.` On macOS, the default temporary directory is a symlink (see below), so the resolved path (/private/var/folders/...) is not the input path to the tests (/var/folders/...) regardless of whatever is added after temp_dir.

```
% env | grep TMP
TMPDIR=/var/folders/vr/v9yknqkx5wq3w_6rlz20x3480000gp/T/
% ls -l / | grep var
lrwxr-xr-x@  1 root  wheel    11 Jul 18  2024 var -> private/var
```

confluence:
@Michaelazzz that `AxesInfo` test is failing the address sanitization check because you've found a genuine bug in my lazy constructor code. :) ~~The constructors which accept vector parameters should check the vector length before trying to access the elements, and raise an exception (`std::invalid_argument` is probably appropriate) if the size is not 2 (empty vectors should also fail).~~ The simplest thing to do is to use `.at` instead of `[]` in the initializer list -- this will already raise an appropriate exception.

There's also a bit of repetition in the constructors that it would be best to eliminate at the same time: ~~add these checks only to the body of~~ modify only the 3-parameter constructor, and have the 5-parameter constructor use that constructor and the two extra fields in its initializer list, instead of accessing all of the fields directly. Then the logic will only be in one place.

confluence:
Something that we don't have is a test in the CI that tries to build the docs and makes sure there are no obvious errors -- I added something like that for Sphinx to `carta-python`. I will see if I can find or write an equivalent for Doxygen and add it to your branch. It's going to be important as we add more docs.

src/Util/App.cc
```diff
@@ -83,6 +107,19 @@ std::string GetReleaseInformation() {
...
 }
 
+/**
+ * @details This function opens a pipe using `popen()` to execute the given command,
+ * reads its output, and returns it as a `std::string`.
+ *
+ * @note
+ * - The function trims trailing newline characters from the output.
+ * - The caller must ensure the command is safe to execute (avoid shell injection).
+ * - Works only on Unix-like systems; **not compatible with Windows** (consider `_popen()` for Windows support).
```
confluence: We don't officially support Windows, and have no plans to do so (since CARTA works in WSL), so it's probably not necessary to go into this much detail about the lack of Windows compatibility, here and in other places (e.g. I think just the first part of the note, `Works only on Unix-like systems`, is fine here).

src/Util/App.h
```diff
@@ -12,8 +12,31 @@
 // version
```
confluence, line 12: General note: docstrings should also be added for:
* each file at the top level ([using the `@file` command](https://www.doxygen.nl/manual/docblocks.html)) -- this isn't as important in files that correspond to a single class, but maybe we should do it there too, for consistency
* each simple preprocessor macro (like `VERSION_ID` here). I checked, and our current doxygen config is detecting and including them, so they should be documented just like global variables. I think we want the [`@def` command](https://www.doxygen.nl/manual/commands.html) for macros.

src/Util/Casacore.cc
```diff
@@ -12,6 +12,18 @@
...
+ * The function verifies that both directories exist and are accessible, and ensures that
+ * the starting directory is a valid subdirectory of the top-level directory.
+ *
+ * @note If `starting_string` is invalid, it is replaced with `top_level_string`.
+ * @note If `starting_string` is not a subdirectory of `top_level_string`, the function logs a critical error and returns `false`.
+ *
+ * @warning If both `top_level_string` and `starting_string` are set to their default placeholders ("base" and "root"),
+ *          the function logs a critical error and returns `false`.
+ */
 bool CheckFolderPaths(string& top_level_string, string& starting_string) {
```
confluence: Doxygen isn't correctly matching and merging the docs for this function because it's using bare `string` as a type here (it's `std::string` in the header). Years ago we decided to standardise and use the `std` namespace explicitly everywhere, so this should be fixed -- please add the `std::` to all the bare type names in this file (I had a look and I think it's just `string` in a bunch of places).

src/Util/Message.tcc
```diff
@@ -7,6 +7,10 @@
 #ifndef CARTA_SRC_UTIL_MESSAGE_TCC_
 #define CARTA_SRC_UTIL_MESSAGE_TCC_
 
+/**
+ * @details This function extracts the payload from a given message buffer, ignoring the event header,
+ * and deserializes it into an object of template type `T` using `ParseFromArray`.
+ */
```
confluence, line 13: This docstring isn't being picked up (at least by my older version of Doxygen) because the templated definition in the `.tcc` file isn't being handled in the same way as a normal definition in a `.cc` file. I'm going to see if there's a way to work around this (or if it's a bug fixed in a newer version).

confluence (commented):
This review isn't complete, but I have a few comments about the docstrings so far.

src/Util/App.h
Michaelazzz: Should I add these before we merge this PR?

src/Util/Image.h
```diff
@@ -128,32 +224,102 @@ struct PointXy {
...
+ * @struct AxesInfo
+ * @brief Stores information about image axes, including spatial and spectral components.
+ *
+ * This struct holds indices for different axes used in image rendering, spatial mapping,
+ * spectral representation, and stokes parameters.
+ */
 struct AxesInfo {
-    int x, y, spatial_x, spatial_y, spectral, z, stokes;
+    int x;         ///< Index of the x-axis for rendering.
+    int y;         ///< Index of the spatial x-axis.
```
confluence: should be "y axis for rendering"

src/Util/Image.h
```diff
@@ -128,32 +224,102 @@ struct PointXy {
...
+    int spatial_x; ///< Index of the spatial x-axis.
+    int spatial_y; ///< Index of the spectral y-axis.
```
confluence: should be "spatial y axis"

src/Util/Image.h
```diff
@@ -128,32 +224,102 @@ struct PointXy {
     }
 };
 
+/**
+ * @struct AxesInfo
+ * @brief Stores information about image axes, including spatial and spectral components.
+ *
+ * This struct holds indices for different axes used in image rendering, spatial mapping,
+ * spectral representation, and stokes parameters.
```
confluence, line 232: Minor issue: Stokes is a proper noun, so it should be capitalised in the docs. Technically in a lot of places we should be saying "polarization" rather than "Stokes", but that's a bigger issue for another time.

src/Util/Image.h
```diff
@@ -128,32 +224,102 @@ struct PointXy {
...
+ */
 struct AxesInfo {
-    int x, y, spatial_x, spatial_y, spectral, z, stokes;
+    int x;         ///< Index of the x-axis for rendering.
+    int y;         ///< Index of the spatial x-axis.
+    int spatial_x; ///< Index of the spatial x-axis.
+    int spatial_y; ///< Index of the spectral y-axis.
+    int spectral;  ///< Index of the spectral axis.
```
confluence: This should also be "if applicable" -- we always pass in an explicit value for this, but it may be -1 if there is no spectral axis.

src/Util/Stokes.h
```diff
@@ -38,28 +95,75 @@ class Stokes {
...
+ * specification and utility functions for comparison and checking if the data
+ * represents an original image.
+ */
 struct StokesSource {
-    int stokes;
-    AxisRange z_range;
-    AxisRange x_range;
-    AxisRange y_range;
+    int stokes;        ///< The Stokes parameter identifier.
+    AxisRange z_range; ///< The range along the Z-axis.
```
confluence, line 111: We should standardise on either e.g. "Z-axis" or "z-axis" in all the docs -- I don't really mind which one as long as it's consistent. I have a slight preference for the capitalised version.

confluence (commented):
A few more comments about the Image docs.

src/Util/Stokes.h
pford: @confluence if Z-axis then also X-axis and Y-axis?

src/Util/App.h
confluence: I don't want to hold up this PR unnecessarily, and it's not related to the tests, so I think it can wait.

Michaelazzz: I'll remember for any docs that follow

confluence:
> Something that we don't have is a test in the CI that tries to build the docs and makes sure there are no obvious errors -- I added something like that for Sphinx to `carta-python`. I will see if I can find or write an equivalent for Doxygen and add it to your branch. It's going to be important as we add more docs.

I'm going to make a separate issue for this -- currently Doxygen already issues some warnings about unrelated parts of the code, so I think it would be best to handle this in a separate PR.

src/Util/Stokes.h
confluence: > @confluence if Z-axis then also X-axis and Y-axis?

Yes. Only in these docstrings; it doesn't matter what we do in internal comments.

src/Util/Message.tcc
confluence: I checked, and the issue is still present in the latest Doxygen version built from source. I will look for a solution; in the meantime I suggest leaving this as-is (it's in the most sensible place, and it's not that critical for now if it's not appearing in the docs).

src/Util/Image.h
Michaelazzz: Noted

src/Util/Message.tcc
confluence: @Michaelazzz I figured it out! It was a config issue -- the `.tcc` files were being *read*, but they weren't being recognised as C++ files. I had to add a mapping. I have pushed a fix to the branch.

Michaelazzz: Amazing!!!

src/Util/File.cc
```diff
@@ -35,16 +51,40 @@ bool IsCompressedFits(const std::string& filename) {
     return false;
 }
 
+/**
+ * @details This function uses a regular expression (Regex) to determine if the provided filename
```
confluence: It's not necessary to include the abbreviation in brackets here.

src/Util/File.cc
```diff
@@ -60,6 +100,17 @@ int GetNumItems(const std::string& path) {
...
 // quick alternative to bp::search_path that allows us to remove
 // boost:filesystem dependency
+/**
+ * @details This function retrieves the `PATH` environment variable, splits it into individual
+ * directory paths, and searches for the specified file within those directories.
+ * If the file is found, its full path is returned. If not found or if an error occurs,
+ * an empty `fs::path` is returned.
+ *
+ * @note This function assumes that path entries in `PATH` are separated by colons (`:`),
+ *       which is standard on UNIX-like systems. It may need modification for Windows.
```
confluence: The windows comment here is unnecessary (we don't support Windows).

src/Util/File.cc
```diff
@@ -79,6 +130,18 @@ fs::path SearchPath(std::string filename) {
...
 }
 
+/**
+ * @details This function attempts to identify the type of an image file by either checking
+ * its magic number (file signature) or examining its file extension. If `check_content`
+ * is set to `true`, the function inspects the file's magic number to classify it as
+ * FITS or HDF5. If `check_content` is `false`, it relies on common file extensions.
+ *
+ * @note When `check_content` is enabled, compressed FITS files (`.fits.gz`) are identified
+ *       by checking their decompressed filename extension.
```
confluence: Dangling participle -- I suggest just `identified by their decompressed filename extension`.

src/Util/File.cc
```diff
@@ -79,6 +130,18 @@ fs::path SearchPath(std::string filename) {
     return fs::path();
 }
 
+/**
+ * @details This function attempts to identify the type of an image file by either checking
```
confluence: Minor nitpick: I would suggest "either by checking... or by..." instead of "by either..." (and again in the next docstring).

src/Util/File.cc
```diff
@@ -141,6 +217,17 @@ CARTA::FileType GuessRegionType(const std::string& path_string, bool check_conte
     return CARTA::UNKNOWN;
 }
 
+/**
+ * @details This function attempts to classify the type of a catalog table file (e.g., FITS table or VOTable)
+ * by first checking its magic number (if `check_content` is `true`) or, if content checking is
```
confluence: Here I would also suggest "either by..." instead of "by first..." (I would say "first by...", but later in the sentence we have "or" rather than "then").

src/Util/File.cc
```diff
@@ -141,6 +217,17 @@ CARTA::FileType GuessRegionType(const std::string& path_string, bool check_conte
...
+ * disabled, by inspecting the file extension.
+ *
+ * @note If `check_content` is enabled, the function may attempt to read the file's magic number.
+ *       Ensure the file is accessible to avoid potential I/O errors.
+ *
+ * @warning This function does not validate file integrity, it only determines type based on
```
confluence: `,` should be `;` here

src/Util/Image.h
```diff
@@ -61,63 +62,158 @@
...
         to = to_;
     }
+
+    /**
+     * @brief Checks if two AxisRange objects are equal.
+     *
+     * Two AxisRange objects are considered equal if both their `from`
+     * and `to` values match.
+     *
+     * @param rhs The other AxisRange to compare with.
```
confluence: I suggest just `the other AxisRange` or `the AxisRange to compare` or `the reference AxisRange`

src/Util/Image.h
```diff
@@ -61,63 +62,158 @@
...
         return false;
     }
+
+    /**
+     * @brief Checks if two AxisRange objects are not equal.
+     *
+     * Two AxisRange objects are considered not equal if either their
+     * `from` or `to` values differ.
+     *
+     * @param rhs The other AxisRange to compare with.
```
confluence: As above.

src/Util/Image.h
```diff
@@ -61,63 +62,158 @@
...
         if ((x != rhs.x) || (y != rhs.y)) {
             return false;
         }
         return true;
     }
+
+    /**
+     * @brief Converts floating-point coordinates to integer indices.
+     * @param x_index Reference to store the x-coordinate index.
+     * @param y_index Reference to store the y-coordinate index.
```
confluence: The capitalisation here should match what is used for X-axis, etc..

src/Util/Image.h
confluence: Looks like this was added to the wrong line (the spatial Y-axis)

src/Util/Image.h
```diff
@@ -128,32 +224,102 @@ struct PointXy {
...
 
+    /**
+     * @brief Default constructor initialising all dimensions to 1.
+     */
     DimsInfo() : width(1), height(1), depth(1), num_channels(1), num_stokes(1) {}
+
+    /**
+     * @brief Constructor that initialises dimensions based on given axes and shape.
+     * @param axes The AxesInfo struct containing axis indices.
+     * @param shape The shape of the data in terms of dimensions.
```
confluence: just `the shape of the image` or `the image dimensions` is fine

src/Util/String.h
```diff
@@ -10,21 +10,61 @@
...
+ */
 std::string SafeStringUnescape(const std::string& input);
 
-// split input string into a vector of strings by delimiter
+/**
+ * @brief Splits a string into parts based on a specified delimiter.
+ *
+ * @param[in] input The string to be split.
+ * @param[in] delim The character delimiter used to separate the input string.
+ * @param[out] parts A vector to store the resulting substrings. It is cleared before adding new values.
```
confluence: "it is cleared before new values are added"

src/Util/Token.cc
```diff
@@ -11,6 +11,10 @@
 
 #include <uuid/uuid.h>
 
+/**
+ * @details This function generates a new random UUID (Universally Unique Identifier)
```
confluence: We don't need to define UUID here.

confluence (changes requested):
I think this completes my docstring review; I will start a separate review for the unit tests.

test/Util/TestUtilImage.cc
```diff
@@ -0,0 +1,296 @@
...
+}
+
+TEST_F(ImageUtilTest, NegativeAxisRange) {
+    AxisRange range(-10, -5);
+    EXPECT_TRUE(range.is_in_range(-7));
+    EXPECT_TRUE(range.is_in_range(-10));
+    EXPECT_TRUE(range.is_in_range(-5));
+    EXPECT_FALSE(range.is_in_range(-11));
+    EXPECT_FALSE(range.is_in_range(-4));
+}
```
confluence, line 69: We don't expect to use AxisRange like this -- real values should always be positive; we only use negative values for placeholders (like ALL_Z or CURRENT_Z).

test/Util/TestUtilImage.cc
```diff
@@ -0,0 +1,296 @@
...
+
+TEST_F(ImageUtilTest, ReversedAxisRange) {
+    AxisRange range(8, 3);              // This might indicate an invalid range
+    EXPECT_FALSE(range.is_in_range(5)); // Should be out-of-range in a valid case
+}
```
confluence, line 74: We should perhaps add better placeholder logic to these objects, but we would have to consider all the edge cases carefully.

I'm in favour of redesigning the code so that AxisRange *never* contains placeholders and is an unsigned integer (once I have removed ALL_X and ALL_Y) -- but this is also a larger issue for another time.

confluence (approved):

confluence:
@pford are you happy with the Casacore-related docstrings and tests?

If so, then I think this is ready to be merged. @kswang1029 there are some minor changes to the code -- they're low-risk, but perhaps an e2e test is a good idea just in case.

pford (changes requested):
The docstrings and tests look great, but I think the PR title and description need to be updated to indicate the tests and docstrings are for all of the Util functions not just Image (for future reference and searches).

pford:
This PR is linked to #1420 and will close the issue when it is merged.  However, the Util code coverage increased from 38% to 45%.  @confluence is this coverage level enough to close the issue?

kswang1029 (commented):
@confluence e2e tests are fine. Any specific thing I should check manually? @Michaelazzz

confluence:
@pford good idea -- @Michaelazzz I think that both the title of this PR and the title of issue #1420 (which was created for this PR) should be modified to be more specific (e.g. "... related to files and images" ?).

confluence:
> @confluence e2e tests are fine. Any specific thing I should check manually? @Michaelazzz

I think as long as the e2e tests are fine, there's no need -- these are changes to low-level objects and functions that are used in a lot of places, rather than a specific feature.

kswang1029 (approved):

pford (approved):

Michaelazzz:
Closes issues #1430  #1431

PR #1414, opened 2025-03-17
Fix memory leak in closed PV and other images

### LINKED ISSUE #1289: [performance] potential memory leakage when we generate pv images repeatedly
**Describe the bug**
The follow psrecord plot is a test of generating pv images three times in a row. We can see the blue curve (RAM usage) increases.

pv_cpu2

If we request pv generation, let it run, and cancel it repeatedly, we see similar pattern.

pv_cpu3

**To Reproduce**
1. load a cube
2. enable psrecord to trace carta_backend
3. generate a pv image three times (without keeping the previous pv)
4. kill psrecord and see the plot

**Expected behavior**
Memory usage in such test should not accumulate.

 - OS [e.g. macOS Monterey]: macOS ventura
 - Browser version [e.g. 22]: 104

pford:
**Description**

Fixes #1289 .

When an image is closed, the Frame holding it is deleted.  I used a debugger in the Frame destructor and noticed that for a unique_ptr it deleted the managed pointer (`delete __ptr;`).  But the _image_cache managed pointer is to an array of a size designated at construction with make_unique.  unique_ptr should delete this array when going out of scope, but deleting it manually (`delete[] image_cache_ptr;`) seems to have fixed the issue on macOS.  Since this fix is in Frame, it will be applied to all images opened in CARTA.

This is the psrecord plot for requesting 4 PV images on macOS 14 Sonoma (M1, clang 15.0.0):
pv_mem

No

Run psrecord while requesting a PV image multiple times.

**Checklist**

- [ ] changelog updated / ~no changelog update needed~
- [ ] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [ ] ~protobuf version bumped~ / protobuf version not bumped

### DIFF
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -38,6 +38,7 @@ Frame::Frame(uint32_t session_id, std::shared_ptr<FileLoader> loader, const std:
       _tile_cache(0),
       _z_index(default_z),
       _stokes_index(DEFAULT_STOKES),
+      _image_cache_size(0),
       _image_cache_valid(false),
       _tile_pool(std::make_shared<TilePool>()),
       _use_tile_cache(false),
@@ -388,8 +389,13 @@ bool Frame::FillImageCache() {
 
     Timer t;
     StokesSlicer stokes_slicer = GetImageSlicer(AxisRange(_z_index), _stokes_index);
-    _image_cache_size = stokes_slicer.slicer.length().product();
-    _image_cache = std::make_unique<float[]>(_image_cache_size);
+    auto frame_size = stokes_slicer.slicer.length().product();
+    if (frame_size != _image_cache_size) {
+        _image_cache.reset();
+        _image_cache_size = frame_size;
+        _image_cache = std::make_unique<float[]>(_image_cache_size);
+    }
+
     if (!GetSlicerData(stokes_slicer, _image_cache.get())) {
         spdlog::error("Session {}: {}", _session_id, "Loading image cache failed.");
         return false;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Fix empty response to HTTP request with underscores in header names ([#1406](https://github.com/CARTAvis/carta-backend/issues/1406))
+* Fix memory leak in Frame for PV and other images ([#1289](https://github.com/CARTAvis/carta-backend/issues/1289))
 
 ## [5.0.0-beta.1]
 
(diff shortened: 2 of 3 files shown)


confluence:
When I started looking at the cache code recently, I noticed that the unique pointer is recreated (and the underlying array reallocated??) whenever the cache is filled, not just once when the frame is created, and it isn't clear to me why. Not to prevent the data from being overwritten while it's being read, surely, because it's protected by a mutex.

The change was introduced [here](https://github.com/CARTAvis/carta-backend/commit/d6eef0743a3cf27342c5ead9a68f06f1d43750ef#diff-6252db2d5198896acf737639ce0f0d22d82fd4dc6f6474bb2c58f15dac2b9201L347), when the cache was initially converted from a float vector to a shared pointer to an array (the shared pointer was subsequently [changed to a unique pointer](https://github.com/CARTAvis/carta-backend/commit/7c84cb4f98b354818b49b2c65e5288746e528041)). In the original code, the vector was [created once](https://github.com/CARTAvis/carta-backend/commit/d6eef0743a3cf27342c5ead9a68f06f1d43750ef#diff-af1f6f5542140b007c59f4f060ea7d20b894702f512ce3641ec555705aa88e29L280) and then [resized in place the first time it was filled](https://github.com/CARTAvis/carta-backend/commit/d6eef0743a3cf27342c5ead9a68f06f1d43750ef#diff-6252db2d5198896acf737639ce0f0d22d82fd4dc6f6474bb2c58f15dac2b9201L347). After that, the data would have been written to the same vector.

I can't say for sure that this is related to the memory leak (I didn't look into this too deeply at the time because I'm intending to refactor the cache into its own object very soon, which will completely change this part of the code), but maybe it would be a good idea to fix this here. Unless I've missed something, we're reallocating an array the size of a channel every time we change the channel, and that can't be good for performance.

kswang1029 (changes requested):
@pford using my dev environment (macOS), I still see memory increases when repeating pv generation (just keep hitting the generate button with "keep previous pv images" toggle off).

pv

In your psrecord plot, I see memory usage is ~700MB. What image did you use?

In addition, I also tested moment generator and saw same issue.
moment

src/Frame/Frame.cc
```diff
@@ -388,8 +389,13 @@ bool Frame::FillImageCache() {
 
     Timer t;
     StokesSlicer stokes_slicer = GetImageSlicer(AxisRange(_z_index), _stokes_index);
-    _image_cache_size = stokes_slicer.slicer.length().product();
-    _image_cache = std::make_unique<float[]>(_image_cache_size);
+    auto frame_size = stokes_slicer.slicer.length().product();
```
confluence, line 392: Will the frame size ever change after the creation of the Frame object?

pford:
Closing PR until solution is found.

PR #1417, opened 2025-03-19
Allow scripting to find resumed sessions

### LINKED ISSUE #1416: Scripting can't find session ID after session resumed
When a session is resumed by the backend, the ID of the session is changed internally in the `Session` object, but the `SessionManager` object uses the original ID of the new session as the key in its session map (since this is the ID that is used in the socket data). However, the scripting code (which uses the internal session ID) is using the session map to find sessions -- so after a session has resumed, it can no longer be found.

We need to add another mapping in the session manager from the internal ID to the external ID, and use it to look up the correct key from the scripting code.

confluence:
**Description**

This PR maps the internal session IDs used by scripting to the real session IDs used in the sessions map. These are different for resumed sessions, which makes it impossible for resumed sessions to be used with scripting.

Fixes #1416.

How to test with a local backend:

1. Start the backend with a constant token and a short idle timeout:
```bash
CARTA_AUTH_TOKEN=banana ./carta_backend --frontend_folder /path/to/frontend --idle_timeout 60 /path/to/test/image
```
2. Copy the session ID from the running CARTA session
3. Connect to the session and try to do something
```python
from carta.session import Session
from carta.constants import Colormap
session = Session.interact("URL_WITH_TOKEN_HERE", SESSION_ID_HERE)
img = session.image_list()[0]
img.raster.set_colormap(Colormap.VIRIDIS)
```
5. Wait for the idle timeout
6. Restart the backend (with the same token)
7. Reconnect from the browser
8. Verify that the session ID in the browser is unchanged
9. Try to do something again with the existing Python objects:
```python
session.image_list()
```

In `dev`, this should result in an error that the session ID cannot be found. In this branch, the session ID should continue to work.

- [X] Test this on Ilifu

**Checklist**

- [X] changelog updated


confluence:
@Jordatious This should fix that session bug. ~~The specific fix needs to be tested with a controller; please let me know if you have access to that kind of test environment and would like to be added as a reviewer.~~ Actually, I've realised that it's possible to test it with a local CARTA backend -- I'll add instructions to the description.

Jordatious (approved):
It's been a while since I've programmed in C++! But generally looks good, and I quite enjoy the name `_real_session_id`!

kswang1029 (approved):
looks good.

pford (approved):

PR #1424, opened 2025-03-31
Remove casacore file handling and replaced with filesystem in the CheckFolderPaths and IsSubdirectory functions

Michaelazzz:
**Description**

Casacore file handling replaced with filesystem functions.
A step towards the larger goal of phasing out casacore from the codebase, except for function directly pertaining to handling the cascaore file format.
No.
N/A

**Checklist**

- [x] no regression from e2e tests
- [ ] ICD test passing


Michaelazzz:
@confluence there's a test failing but its something nowhere close to where I was working in the codebase. It's something to do with the RoundTrip test in the TileEncoding test file where a tile is encoded and decoded. Should I be worried about it?

pford:
I looked at using filesystem at one point, but I could not find a way to check user permissions (readable / executable).  It looks like casacore includes <unistd.h> and calls access() in casacore/casa/OS/File.cc.

confluence:
> @confluence there's a test failing but its something nowhere close to where I was working in the codebase. It's something to do with the RoundTrip test in the TileEncoding test file where a tile is encoded and decoded. Should I be worried about it?

This seems to be an unrelated bug in the test (not easily reproducible because the test generates random values). I'm looking into it separately. I re-ran the failing action in this PR, and it's fine now.

Michaelazzz:
> I looked at using filesystem at one point, but I could not find a way to check user permissions (readable / executable). It looks like casacore includes <unistd.h> and calls access() in casacore/casa/OS/File.cc.

@pford there were some notes in the codebase that mentioned some specific places where casacore can be phased out, I haven't gotten to this dilemma you mentioned but I will make a note of it. Thank you!

src/Util/Casacore.cc
```diff
@@ -13,88 +13,79 @@
...
-            top_level_string = top_level_folder.path().absoluteName();
-        } catch (casacore::AipsError& err) {
-            spdlog::error(err.getMesg());
-        }
-        if (top_level_string.empty())
-            top_level_string = "/";
-    }
-    // check starting folder
-    casacore::File starting_folder(starting_string);
-    if (!(starting_folder.exists() && starting_folder.isDirectory(true) && starting_folder.isReadable() &&
```
pford, line 65: @Michaelazzz this is where the folder permissions were checked using casacore::File

pford:
These Util files were named `Casacore` because they use casacore functions, so I think the functions should now be moved to `File`.  And ParseHistoryBeamHeader should live in `String` since it uses std::string not casacore::String, and I am the one who put it here!

confluence:
@pford @Michaelazzz we're going to move some of the additional tests from the testing PR to here, so that we can investigate why they're failing and discuss what the intended behaviour of this function is.

kswang1029 (approved):
this appears fine.

src/Util/String.h
```diff
@@ -67,4 +67,17 @@ bool ConstantTimeStringCompare(const std::string& a, const std::string& b);
  */
 bool StringToInt(const std::string& input, int& i);
 
+/**
```
pford: To me, ParseHistoryBeamHeader feels out of place in String utility functions (equally out of place in its former Casacore location).  It is used only for FITS images by ImageData/FitsLoader (is a FileLoader) and ImageData/CompressedFits (not a FileLoader) when the support_aips_beam preference is set.  Maybe create FitsUtil.h in ImageData so it is co-located where it is used?

pford (changes requested):

src/Util/String.h
Michaelazzz: You are absolutely right, I'll make the changes

pford (approved):

pford:
Sorry @Michaelazzz , I asked you to move ParseHistoryBeam to String then complained it was there!  Will try to do more careful reviews in the future.

Michaelazzz:
> Sorry @Michaelazzz , I asked you to move ParseHistoryBeam to String then complained it was there! Will try to do more careful reviews in the future.

No worries Pam, it was a completely valid request.

kswang1029 (approved):
looks good.

confluence (approved):

Issue #1439, opened 2025-04-29
Replacing the Session Manager OnMessage function switch statement


Michaelazzz:
There a very long switch statement in the `OnMessage` function within the `SessionManager` class, which is not great and has an O(n)  look-up time for messages and scales linearly with each added message.

I suggest that all the switch statement cases be put into handler functions within an unordered map.
This would improve the O(n) look-up time to O(1), which would remain consistent even after adding more messages.
(Closed in PR #1440 )

PR #1440, opened 2025-04-29
Changing the SessionManager OnMessage Switch statement to handler functions

Michaelazzz:
**Description**

It adds new handler functions to the session manager for incoming messages.
 issue #1439 
This PR replaces the switch statement handling incoming messages with handler functions to improve the lookup speed for incoming messages
No
No

**Checklist**

- [x] e2e test passing
- [x] ICD test passing


src/Session/SessionManager.cc
```diff
@@ -645,4 +269,508 @@ std::string SessionManager::IPAsText(std::string_view binary) {
...
 
+void SessionManager::InitMessageHandlers() {
+    _message_handlers[CARTA::EventType::REGISTER_VIEWER] = [](Session* session, const char* event_buffer, int event_length,
+                                                               const EventHeader& head) {
+        CARTA::RegisterViewer message;
+        if (!message.ParseFromArray(event_buffer, event_length)) {
+            spdlog::error("Error parsing message for event type {} in session {}: Failed to parse message.", session->GetId(),
+                CARTA::EventType_Name(CARTA::EventType::REGISTER_VIEWER));
+            return;
+        }
```
confluence: This is adding a lot of verbose repetition, and should be factored out. I was going to suggest adding a templated utility function to `Util/Message` for this -- but it already exists (`Message::DecodeMessage`), and isn't being used! You can modify to integrate it into this code.

The log message should be inside the function, and the unpacking of the message string should also be inside.

If you modify the function to take an in-out decoded message parameter and return a boolean, you can restore the original `if (message is decoded) { do stuff }` design.

However, you could potentially keep the current design (returning the decoded message) and raise an exception if `ParseFromArray` fails, and then catch the exception in the code that calls the handler (to avoid reintroducing verbose repetition into the handlers) -- that would make the handlers even cleaner (one-liner to decode the message, then just the handler-specific code).

src/Session/SessionManager.cc
```diff
@@ -139,406 +141,28 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
-            session->UpdateLastMessageTimestamp();
-
-            EventHeader head = *reinterpret_cast<const EventHeader*>(sv_message.data());
-            const char* event_buf = sv_message.data() + sizeof(EventHeader);
-            int event_length = sv_message.length() - sizeof(EventHeader);
+    if (op_code == uWS::OpCode::BINARY && sv_message.length() >= sizeof(EventHeader)) {
+        session->UpdateLastMessageTimestamp();
 
-            CARTA::EventType event_type = static_cast<CARTA::EventType>(head.type);
-            logger::LogReceivedEventType(event_type);
```
confluence, line 151: This is missing from the refactored code; we should keep it.

src/Session/SessionManager.cc
```diff
@@ -139,406 +141,28 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
-                spdlog::warn("Bad {} message!", event_type_name);
-            }
-
-            if (tsk) {
-                ThreadManager::QueueTask(tsk);
+        if (handler != _message_handlers.end()) {
+            auto start_time = std::chrono::high_resolution_clock::now();
+            try {
+                handler->second(session, event_buf, event_length, head);
+            } catch (const std::exception& e) {
```
confluence: This should only be catching the message parsing exception, for consistency with the old code (other exceptions should not be handled here; there are generic exception handlers at a higher level).

src/Util/Message.tcc
```diff
@@ -12,11 +12,20 @@
...
+ */
+T Message::DecodeMessage(uint32_t session_id, const char* event_buffer, int event_length, const carta::EventHeader& head) {
+    static_assert(std::is_member_function_pointer<decltype(&T::ParseFromArray)>::value,
+        "T must have a member function ParseFromArray(const void*, int)");
     T decoded_message;
-    char* event_buf = message.data() + sizeof(carta::EventHeader);
-    int event_length = message.size() - sizeof(carta::EventHeader);
-    decoded_message.ParseFromArray(event_buf, event_length);
+    if (!decoded_message.ParseFromArray(event_buffer, event_length)) {
+        throw std::runtime_error(
```
confluence: This should throw your custom exception type instead.

I think it's appropriate for it to inherit from `std::runtime_error` because this is external input and not an internal function call.

Since C++11 you can do this very simply, like this:
```
struct my_exception: std::runtime_error {
    using std::runtime_error::runtime_error;
}
```

(Previously you had to use a more verbose syntax to define a constructor explicitly.)

src/Session/SessionManager.cc
```diff
@@ -645,4 +268,298 @@ std::string SessionManager::IPAsText(std::string_view binary) {
     return result;
 }
 
+std::unordered_map<CARTA::EventType, SessionManager::MessageHandler> SessionManager::_message_handlers = {
```
confluence: I suggest moving this above the function definitions.

src/Session/SessionManager.cc
```diff
@@ -139,406 +139,29 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
+    if (op_code == uWS::OpCode::BINARY && sv_message.length() >= sizeof(EventHeader)) {
+        session->UpdateLastMessageTimestamp();
+
+        EventHeader head = *reinterpret_cast<const EventHeader*>(sv_message.data());
+        const char* event_buf = sv_message.data() + sizeof(EventHeader);
+        int event_length = sv_message.length() - sizeof(EventHeader);
+
+        CARTA::EventType event_type = static_cast<CARTA::EventType>(head.type);
+        logger::LogReceivedEventType(event_type);
+        auto handler = _message_handlers[event_type];
```
confluence: You shouldn't use the `[]` operator here because it will insert a default value if the key is not found.

You can either go back to the previous method (use `find` and compare to `end`) or use `at` and catch `std::out_of_range` (put `try` block around the handler lookup only, not the whole handler call), depending on what is cleaner (probably `at`, because it's more idiomatic).

src/Session/SessionManager.cc
```diff
@@ -139,406 +139,29 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
-            if (!message_parsed) {
-                spdlog::warn("Bad {} message!", event_type_name);
-            }
-
-            if (tsk) {
-                ThreadManager::QueueTask(tsk);
+    if (op_code == uWS::OpCode::BINARY && sv_message.length() >= sizeof(EventHeader)) {
+        session->UpdateLastMessageTimestamp();
+
+        EventHeader head = *reinterpret_cast<const EventHeader*>(sv_message.data());
```
confluence: You should unpack the buffer and length inside the `DecodeMessage` function, not out here (because it isn't used anywhere except inside that function). Then you can simplify the signatures of the handlers and the decode function (pass the whole `sv_message` instead  -- you should use a `std::string_view`, though, not a `std::vector<char>&` as in the original `DecodeMessage` signature).

You need the head out here, and in the handlers, so it's fine to keep that separate. You don't need the head inside `DecodeMessage`, just the header size, so there's no repetition.

confluence: Related to this:

1. `Util/Message` has a static `EventType` function which gets the type from an encoded message, but I believe it's also currently unused. Since we actually need the whole header here and not just the type, I suggest that you replace it with a static `GetEventHeader` function which returns the `EventHeader`, and add a `GetType` function to the `EventHeader` struct which returns the `CARTA::EventType`.
2. For consistency, it would be good to add an `EncodeMessage` function to `Message`, and move the encoding code from `Session::SendEvent` inside (not the compression decision; that can stay in `SendEvent` -- construct `msg_vs_compress` at the end using the final message size).

confluence (changes requested):
This is looking much better! I found a few other places where we can simplify things.

src/Session/SessionManager.cc
confluence: (I believe that the default value will always be zero, so the `if (handler)` will correctly prevent a crash, but the insertion is an undesirable side effect.)

src/Util/Message.cc
```diff
@@ -447,9 +448,27 @@ CARTA::ChannelMapFlowControl Message::ChannelMapFlowControl(int32_t file_id, int
...
 }
 
-CARTA::EventType Message::EventType(std::vector<char>& message) {
-    carta::EventHeader head = *reinterpret_cast<const carta::EventHeader*>(message.data());
-    return static_cast<CARTA::EventType>(head.type);
+carta::EventHeader Message::GetEventHeader(std::string_view message) {
+    return *reinterpret_cast<const carta::EventHeader*>(message.data());
+}
+
+std::pair<std::vector<char>, bool> Message::EncodeMessage(
```
confluence: This function should *only* create and return the encoded message (the char vector). The pair should be constructed afterwards in the calling function.

src/Session/Session.cc
```diff
@@ -2117,18 +2117,13 @@ void Session::SendEvent(CARTA::EventType event_type, uint32_t event_id, const go
...
-    head->type = event_type;
-    head->icd_version = ICD_VERSION;
-    head->request_id = event_id;
-    message.SerializeToArray(msg.data() + sizeof(EventHeader), message_length);
+    size_t required_size = sizeof(carta::EventHeader) + message_length;
+
+    std::pair<std::vector<char>, bool> msg_vs_compress = Message::EncodeMessage(event_type, event_id, message);
+
     // Skip compression on files smaller than 1 kB
     msg_vs_compress.second = compress && required_size > 1024;
```
confluence: The EncodeMessage function should return only the encoded message, not the pair. `required_size` should not be needed out here; only inside the function (for resizing the char vector).

This function should:
1. Call the encode function to get just the message
2. Create just the boolean (using the size of the returned vector; this should be equivalent to `required_size` here)
3. use `std::make_pair` to create the pair at the end

confluence (changes requested):
The handler signature simplification looks good. The encoding functions need some changes.

(Also, it looks like the doxygen test is failing.)

Michaelazzz:
The doxygen comments need some cleaning up. I wanted to wait until there was no more feedback or changes before updating the doc comments.

src/Util/Message.cc
```diff
@@ -447,9 +448,26 @@ CARTA::ChannelMapFlowControl Message::ChannelMapFlowControl(int32_t file_id, int
...
+    return *reinterpret_cast<const carta::EventHeader*>(message.data());
+}
+
+std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t event_id, const google::protobuf::MessageLite& message) {
+    size_t message_length = message.ByteSizeLong();
+    size_t required_size = sizeof(carta::EventHeader) + message_length;
+
+    std::vector<char> msg;
+    msg.resize(required_size, 0);
+    carta::EventHeader* head = (carta::EventHeader*)msg.data();
```
confluence: This should be removed.

confluence (commented):
The signatures look good now.

src/Util/Message.cc
Michaelazzz: Remove all of the mentioned lines?

confluence: No, I mean the creation of the duplicate header variable.

Michaelazzz: Great, that is what I thought

src/Util/Message.cc
```diff
@@ -447,9 +448,27 @@ CARTA::ChannelMapFlowControl Message::ChannelMapFlowControl(int32_t file_id, int
...
+
+/**
+ * @note This function creates a binary buffer containing a CARTA::EventHeader followed by the serialized protobuf message.
+ * The header includes the event type, protocol version, and event/request ID.
+ */
+std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t event_id, const google::protobuf::MessageLite& message) {
+    size_t message_length = message.ByteSizeLong();
+    size_t required_size = sizeof(carta::EventHeader) + message_length;
+
+    std::vector<char> msg(required_size, 0);
```
confluence: I think the `0` parameter is unnecessary here -- I believe that this is the default value for initializing a vector with chars.

confluence (approved):

Michaelazzz:
@kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?

kswang1029:
> @kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?

I can try to check the total elapsed time of all e2e test runs as a start as the aggregated message counts should be a lot and might be sufficient to see performance improvements.

kswang1029:
> > @kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?
> 
> I can try to check the total elapsed time of all e2e test runs as a start as the aggregated message counts should be a lot and might be sufficient to see performance improvements.

Which message was in the last if-else-if chain?

Michaelazzz:
@kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?

> > > @kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?
> > 
> > 
> > I can try to check the total elapsed time of all e2e test runs as a start as the aggregated message counts should be a lot and might be sufficient to see performance improvements.
> 
> Which message was in the last if-else-if chain?

The last message in the chain was `CHANNEL_MAP_FLOW_CONTROL`

kswang1029:
> @kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?
> 
> > > > @kswang1029 are there any end-to-end performance tests that you run that can determine if this PR provides a speedup for handling incoming requests?
> > > 
> > > 
> > > I can try to check the total elapsed time of all e2e test runs as a start as the aggregated message counts should be a lot and might be sufficient to see performance improvements.
> > 
> > 
> > Which message was in the last if-else-if chain?
> 
> The last message in the chain was `CHANNEL_MAP_FLOW_CONTROL`

Ok thanks. That’s ideal one for testing performance improvements. I will write a simple e2e test to see how it improves.

veggiesaurus:
I can't imagine there would be much of a performance difference here... we're not sending tens of thousands of messages per second 😆

Michaelazzz:
> I can't imagine there would be much of a performance difference here... we're not sending tens of thousands of messages per second 😆

Just making sure that there isn't a catastrophic slowdown 😳

kswang1029 (approved):
this looks fine. I do not see significant performance improvements nor degradation as expected since we do not have million messages in the if-else if chain.

Michaelazzz:
> this looks fine. I do not see significant performance improvements nor degradation as expected since we do not have million messages in the if-else if chain.

Great, thank you for looking into that for me.

PR #1443, opened 2025-05-13
Fix region import from top-level directory

### LINKED ISSUE #1403: Fail to Load Region List at the Top Level Folder
**Describe the bug**
Not sure if this is a backend issue, but the top level folder region list is not loaded when a image file is loaded from the top level folder. `RegionListRequest` with `directory = ""` might be returning `undefined` in some circumstances.

Note that this issue can be more easily triggered with the cartavis/carta docker image.

**To Reproduce**
1. Launch Carta with backend top level folder set to the directory of an image.
2. Open the image at the top level folder.
3. Open import regions and see the error.
4. Load an image in a subdirectory.
5. Open import regions and move to the top level folder.
6. The region list is shown without error.

**Expected behavior**
The region list is shown without error at step 3.

**Screenshots or videos**

Image

 - OS [e.g. macOS Monterey]: macOS
 - Browser version [e.g. 22]: 132
 - Backend branch [e.g. dev, v3b2 release]: dev, v5b1
 - Frontend branch [e.g. dev, v3b2 release]: dev, v5b1

**Additional context**

confluence:
**Description**

Fixes #1403 by initialising the value of the region list directory which is currently being processed, so that it is not initialised to the default empty string (which is also the valid path to the top-level directory).

Other changes:
* Changed the dummy value from `nofolder`, which could in theory be a valid path value, to a string which contains the null character (not valid in file names and therefore in paths).
* Defined a constant for the dummy value to eliminate hardcoded string
* Fixed `TestFileListSubdirectories` so that it doesn't assume that there will never be an image file present in the test environment's root directory (`FileListTest.SetTopLevelFolder` was failing on my system because I put an image file there). I believe that the intention of the test is fulfilled if it just checks for the presence of subdirectories.

To test:

1. Run the backend
2. Navigate to the top-level directory
3. Open an image in that directory
4. Try to import regions.

In `dev` this will result in an error and a blank directory will be displayed. In this branch there should be no error and the directory should be displayed correctly.

**Checklist**

- [X] changelog updated


pford (approved):

kswang1029:
@confluence I tested with the case having the initial folder same as the top_level_folder. After loading an image, I can see a region list successfully. But if I try to load an image, it gives me an error "Import region failed: cannot import by filename or contents."

confluence:
@kswang1029 this was an unrelated error: the check determining whether the region should be loaded from a file or from contents required both the filename and the directory to be non-empty (it's been like that since the first implementation!). So it wasn't possible to load a region file from the top-level directory (regardless of the starting directory, etc.). This should be fixed now.

kswang1029 (changes requested):
I am testing with the following setup

```
./carta_backend /Users/kswang/carta_image_pool/set_HD163296 --frontend_folder /Users/kswang/carta_build/carta-frontend/build --port 3002 --omp_threads 8 --debug_no_auth --no_browser --verbosity 5 --top_level_folder /Users/kswang/carta_image_pool/set_HD163296
```
where `set_HD163296` directroy contains images and region files. I found that I cannot save regions as files. Could you have a look?

confluence:
Ah, the same directory check is also in the export function. I'll fix it.

confluence:
~~The same kind of check may also be causing an issue with loading LEL expressions from the top-level directory. I will test this.~~ that's actually called with the resolved "real" path, so it shouldn't have the same problem.

confluence:
@kswang1029 I removed the extra check from the backend export code, but currently this is still broken because of [an additional check in the frontend](https://github.com/CARTAvis/carta-frontend/blob/7bdbdc15f7dd71b89b19b3ed51826dbb80121036/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx#L233C22-L233C22). I will file an issue.

kswang1029 (approved):
Looks good and passing e2e tests.

pford (approved):

PR #1445, opened 2025-05-21
Update OS in the test file: carta-backend/.github/workflows/icd_tests.yml

pshnghng0318:
**Description**

* Update os_version, with ICD tests passed: 
macOS-13, macOS-15
ubuntu-22.04, ubuntu-24.04
rhel-8, rhel-9
(macOS-14-Intel has an issue on fitting, update later)

* The runner with a label macOS-13 can't build backend, which try to access an non-exist folder: /Users/runner/, replaced by [macOS-13, ICD]

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -1,6 +1,6 @@
 
 name: ci
-on: [push, pull_request]
+on: [push]
 jobs:
 
 
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -312,7 +312,7 @@ jobs:
       - name: Run unit tests
         uses: nick-fields/retry@master
         with:
-          timeout_minutes: 5
+          timeout_minutes: 10
           retry_wait_seconds: 60
           max_attempts: 3
           retry_on_exit_code: 1
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -1,6 +1,6 @@
 
 name: ci
-on: [push, pull_request]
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):

confluence:
Can we also update the Docker containers in the CI workflow here, or would that be more appropriate to do in a separate PR?

.github/workflows/continuous_integration.yml
```diff
@@ -1,5 +1,6 @@
+
 name: ci
-on: [push, pull_request]
+on: [push]
```
confluence: I think that this will prevent checks from running when a new pull request is created.

But I agree that it would be a good idea to refine the criteria here so that we aren't duplicating checks 90% of the time.

Have a look at the discussion [here](https://github.com/orgs/community/discussions/26940). I think that the solution proposed [here](https://github.com/orgs/community/discussions/26940#discussioncomment-5686753) may suit our needs (with `dev` and `release/*` as our `push` branches). This would run the checks on any pull request change, and also on merge to `dev` or a release branch.

That wouldn't run them on pushes to any other branches, but I don't think that's necessarily a problem, since work-in-progress feature branches aren't intended to be test-ready. If someone wants to test their branch before it's ready, they can make a draft PR to trigger the checks.

confluence (changes requested):
We should *remove* the older platform versions, and have only one version of Ubuntu, Alma Linux, and macOS.

(In future it would be good to run the checks on each supported platform, but for now that would make CI prohibitively slow -- we should revisit this after investigating ways to speed it up.)

confluence (approved):
This looks good to me now.

PR #1458, opened 2025-05-29
Protobuf header message creation cleanup and helper functions

Michaelazzz:
**Description**

Streamline construction of FileInfoExtended message, including helper function for sub-messages.

**Checklist**

- [ ] changelog updated
- [x] e2e test passing
- [ ] ICD test passing


confluence:
@Michaelazzz from the diffs, it looks like you may be running your local formatting pass with a version of `clang-format` which is newer than 14 (which is what we're currently using for the format check).

At some point we should decide whether to switch to a newer version (and either change the options to be backwards compatible or reformat everything -- I'll make an issue for this), but for now please install `clang-format-14` and either reconfigure your default `clang-format` to point to it (by changing the symlink `/usr/bin/clang-format`; you may want to change `clang-format-diff` and `git-clang-format` to match to avoid later confusion) or explicitly select it with the `--clang-format` flag when you run `style.py`.

src/Util/Message.cc
```diff
@@ -717,6 +665,28 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    entry->set_name(name);
+    entry->set_value(value);
+    entry->set_entry_type(CARTA::EntryType::STRING);
+}
+
+void Message::AddComputedEntry(CARTA::FileInfoExtended& response, std::string name, const std::string &value, double numeric_value) {
+    auto entry = response.add_computed_entries();
+    entry->set_name(name);
+    entry->set_value(value);
+    entry->set_entry_type(CARTA::EntryType::INT);
```
confluence: We can't use `INT` for all numeric values -- we need another parameter for setting an explicit type here (either `CARTA::EntryType::INT` or `CARTA::EntryType::FLOAT`).

As discussed previously we can't easily deduce the type from the numeric value parameter (and also there may be cases where we pass in an integer but actually want the entry to be a float). So it's best to be explicit.

We should also have corresponding `AddHeaderEntry` functions for adding header entries that are not computed.

confluence (changes requested):
The wrapper for adding numeric entries needs an explicit type parameter. Other than that, this is looking good so far.

src/Util/Message.cc
```diff
@@ -717,6 +665,53 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+        CARTA::RegionInfo region_info;
+        region_info.set_region_type(region_state.type);
+        *region_info.mutable_control_points() = {region_state.control_points.begin(), region_state.control_points.end()};
+        region_info.set_rotation(region_state.rotation);
+
+        // Add info and style to import_ack; increment region id for next region
+        (*region_info_map)[region_id] = region_info;
+        (*region_style_map)[region_id++] = region_style;
+    }
+}
```
confluence: This doesn't compile because the function declaration / definition is incorrect (copy/paste error?).

But also parts of this function are manipulating RegionHandler internals, and those parts shouldn't be here. This function should *only* be modifying the import region ack message, adding the region information to the region info map and the region style map.

Most of this function should go back in RegionHandler, but the
```
auto region_info_map = import_ack.mutable_regions();
auto region_style_map = import_ack.mutable_region_styles();
```
should move inside the function.

I suggest that you make the function take the ack message as a parameter, as well as the region id and region parameters. Increment the ID outside. Get the info and style maps inside the function.

What form the parameters should be is tied to where various structs are defined -- e.g. if you pass in a RegionProperties struct, it (and RegionState) should move into Util (maybe not a bad idea) so that Util stuff doesn't depend on stuff inside Region. Or you could pass in most of the parameters as individual values, but that would be a bit verbose.

confluence (changes requested):

src/Util/Message.cc
```diff
@@ -717,6 +665,43 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    auto entry = response.add_computed_entries();
+    entry->set_name(name);
+    entry->set_value(value);
+    entry->set_entry_type(type);
+    entry->set_numeric_value(numeric_value);
+}
+
+CARTA::ImportRegionAck& Message::AddImportedRegion(CARTA::ImportRegionAck& import_ack, int region_id, CARTA::RegionType region_type, std::vector<CARTA::Point> control_points, float region_rotation, CARTA::RegionStyle region_style) {
+    auto region_info_map = import_ack.mutable_regions();
+    auto region_style_map = import_ack.mutable_region_styles();
```
confluence: These variables are unnecessary; you can use `(*import_ack.mutable_whatever())[...]` directly below.

You should also remove these lines where they appear in the RegionHandler code.

src/Util/Message.cc
```diff
@@ -717,6 +665,43 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+
+    // Set CARTA::RegionInfo
+    CARTA::RegionInfo region_info;
+    region_info.set_region_type(region_type);
+    *region_info.mutable_control_points() = {control_points.begin(), control_points.end()};
+    region_info.set_rotation(region_rotation);
+
+    // Add info and style to import_ack; increment region id for next region
+    (*region_info_map)[region_id] = region_info;
+    (*region_style_map)[region_id++] = region_style;
```
confluence: This increment of `region_id` has no effect. You should just use `region_id` as the index here, and increment `region_id` separately outside the function in RegionHandler.

confluence (changes requested):

confluence:
@Michaelazzz there are still lots of whitespace-only changes in this PR. If you previously reformatted it with a newer version of clang-format, it's possible that subsequent passes with the older version won't revert those changes, so you may have to clean those up manually. If you revert all the files that have whitespace-only changes (checking out the previous versions) that will leave only files with other changes, where you should be able to follow a diff and selectively fix just those parts.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -91,10 +92,8 @@ bool FileExtInfoLoader::FillFileExtInfo(
     // Set name from filename
     fs::path filepath(filename);
     std::string filename_nopath = filepath.filename().string();
-    auto entry = extended_info.add_computed_entries();
-    entry->set_name("Name");
-    entry->set_value(filename_nopath);
-    entry->set_entry_type(CARTA::EntryType::STRING);
+
+    Message::AddComputedEntry(extended_info, "Has Name", filename_nopath);
```
confluence: Typo: `Has Name` -> "Name"

src/FileList/FileExtInfoLoader.cc
```diff
@@ -846,19 +809,13 @@ void FileExtInfoLoader::AddComputedEntries(CARTA::FileInfoExtended& extended_inf
...
         }
 
         if (!reference_pixels.empty()) {
-            auto entry = extended_info.add_computed_entries();
-            entry->set_name("Image reference pixels");
-            std::string ref_pix = fmt::format("[{}, {}]", reference_pixels(axes.x) + 1.0, reference_pixels(axes.y) + 1.0);
-            entry->set_value(ref_pix);
-            entry->set_entry_type(CARTA::EntryType::STRING);
+            Message::AddComputedEntry(extended_info, "Image reference pixels",
+                fmt::format("[{}, {}]", reference_pixels(axes.x) + 1.0, reference_pixels(axes.y) + 1.0));
```
confluence: Maybe keep the intermediate variable here for legibility.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -895,18 +846,12 @@ void FileExtInfoLoader::AddComputedEntries(CARTA::FileInfoExtended& extended_inf
...
+            Message::AddComputedEntry(extended_info, "Pixel increment", pixel_inc);
         }
 
         casacore::String brightness_unit(image->units().getName());
         if (!brightness_unit.empty()) {
-            auto entry = extended_info.add_computed_entries();
-            entry->set_name("Pixel unit");
-            entry->set_value(brightness_unit);
-            entry->set_entry_type(CARTA::EntryType::STRING);
+            Message::AddComputedEntry(extended_info, "Brightness unit", brightness_unit);
```
confluence: String name has changed (should be "Pixel unit")

src/FileList/FileExtInfoLoader.cc
```diff
@@ -1581,10 +1484,8 @@ void FileExtInfoLoader::AddCoordRanges(
                 } else if (axis_names(spatial_axes[0]) == "Longitude") {
                     axis_names(spatial_axes[0]) = "LON";
                 }
-                auto* x_entry = extended_info.add_computed_entries();
-                x_entry->set_name(fmt::format("{} range", axis_names(spatial_axes[0])));
-                x_entry->set_value(fmt::format("[{}, {}]", x_start, x_end));
-                x_entry->set_entry_type(CARTA::EntryType::STRING);
+                Message::AddComputedEntry(
+                    extended_info, fmt::format("{} range", axis_names(spatial_axes[0])), fmt::format("[{}, {}]", x_start, x_end));
```
confluence: This would also look cleaner with intermediate variables for the name and value.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -1593,10 +1494,8 @@ void FileExtInfoLoader::AddCoordRanges(
                 } else if (axis_names(spatial_axes[1]) == "Latitude") {
                     axis_names(spatial_axes[1]) = "LAT";
                 }
-                auto* y_entry = extended_info.add_computed_entries();
-                y_entry->set_name(fmt::format("{} range", axis_names(spatial_axes[1])));
-                y_entry->set_value(fmt::format("[{}, {}]", y_start, y_end));
-                y_entry->set_entry_type(CARTA::EntryType::STRING);
+                Message::AddComputedEntry(
+                    extended_info, fmt::format("{} range", axis_names(spatial_axes[1])), fmt::format("[{}, {}]", y_start, y_end));
```
confluence: As above.

src/FileList/FileListHandler.cc
```diff
@@ -182,7 +183,7 @@ void FileListHandler::GetFileList(CARTA::FileListResponse& file_list_response, c
 
             if (cc_file.isReadable() && cc_file.exists() && name.firstchar() != '.') { // ignore hidden files/folders
                 casacore::String full_path(cc_file.path().absoluteName());
-                std::string name_only = cc_file.path().baseName();
+                casacore::String name_only = cc_file.path().baseName();
```
confluence: There's no reason to switch to `casacore::String` here; in general we should avoid using these types if we can use `std` types instead.

src/ImageStats/BasicStatsCalculator.tcc
```diff
@@ -56,7 +56,8 @@ BasicStatsCalculator<T>::BasicStatsCalculator(const T* data, size_t data_size)
 template <typename T>
 void BasicStatsCalculator<T>::reduce() {
     size_t i;
-#pragma omp parallel for private(i) shared(_data) reduction(min: _min_val) reduction(max:_max_val) reduction(+:_num_pixels) reduction(+:_sum) reduction(+:_sum_squares)
+#pragma omp parallel for private(i) shared(_data) reduction(min : _min_val) reduction(max : _max_val) reduction(+ : _num_pixels) \
```
confluence: This should be reverted (was it previously changed by a newer clang-format?).

src/Region/RegionHandler.cc
```diff
@@ -1584,7 +1584,7 @@ bool RegionHandler::FillSpectralProfileData(
                 profile_ok = GetRegionSpectralData(config_region_id, config_file_id, z_range, coordinate, stokes_index, required_stats,
                     report_error, [&](std::map<CARTA::StatsType, std::vector<double>> results, float progress) {
                         auto profile_message = Message::SpectralProfileData(
-                            config_file_id, config_region_id, stokes_index, progress, coordinate, required_stats, results);
+                            stokes_index, progress, config_file_id, config_region_id, coordinate, required_stats, results);
```
confluence: It's unclear why these parameters were reordered: the original ordering makes more sense. This should be reverted.

src/Region/RegionHandler.cc
```diff
@@ -2016,15 +2016,9 @@ bool RegionHandler::FillLineSpatialProfileData(int file_id, int region_id, std::
...
-                int end = profile_size - 1;
-                float crpix = profile_size / 2;
-                float cdelt = increment.getValue();
-                float crval = (axis_type == CARTA::ProfileAxisType::Offset ? 0.0 : crpix * cdelt);
-                std::string unit = increment.getUnit();
-
-                auto profile_message = Message::SpatialProfileData(file_id, region_id, x, y, channel, stokes_index, value, start, end,
-                    profile, coordinate, mip, axis_type, crpix, crval, cdelt, unit);
+                auto profile_message = Message::SpatialProfileData(x, y, channel, stokes_index, value);
+                Message::AddProfile(profile_message, file_id, region_id, start, profile, coordinate, mip, axis_type, increment);
```
confluence: These calculations are at the wrong level to be performed inside the message wrapper. The wrapper should *only* set message fields. The calculations should be performed outside the function, as before.

src/Region/RegionHandler.cc
```diff
@@ -2016,15 +2016,9 @@ bool RegionHandler::FillLineSpatialProfileData(int file_id, int region_id, std::
...
-                auto profile_size = profile.size();
-                int end = profile_size - 1;
-                float crpix = profile_size / 2;
-                float cdelt = increment.getValue();
-                float crval = (axis_type == CARTA::ProfileAxisType::Offset ? 0.0 : crpix * cdelt);
-                std::string unit = increment.getUnit();
-
-                auto profile_message = Message::SpatialProfileData(file_id, region_id, x, y, channel, stokes_index, value, start, end,
-                    profile, coordinate, mip, axis_type, crpix, crval, cdelt, unit);
+                auto profile_message = Message::SpatialProfileData(x, y, channel, stokes_index, value);
```
confluence: file_id and region_id are fields on the main message, not on the profile submessage, and should be set in the main message wrapper, not the wrapper for adding a profile.

src/Util/Message.cc
```diff
@@ -133,39 +133,12 @@ CARTA::Point Message::Point(const std::vector<double>& input, int x_index, int y
...
-    stats_configs->add_stats_types(CARTA::StatsType::FluxDensity);
-    stats_configs->add_stats_types(CARTA::StatsType::Mean);
-    stats_configs->add_stats_types(CARTA::StatsType::RMS);
-    stats_configs->add_stats_types(CARTA::StatsType::Sigma);
-    stats_configs->add_stats_types(CARTA::StatsType::SumSq);
-    stats_configs->add_stats_types(CARTA::StatsType::Min);
-    stats_configs->add_stats_types(CARTA::StatsType::Max);
-    stats_configs->add_stats_types(CARTA::StatsType::Extrema);
-    return set_stats_requirements;
+CARTA::SetRegion Message::SetRegion(int32_t file_id, int32_t region_id, const CARTA::RegionInfo& region_info) {
```
confluence: Removing the other SetRegion wrapper is fine (it's never used), but there should be more refactoring of related messages (e.g. adding the region info objects that will be passed here to the ImportRegionAck message).

src/Util/Message.cc
```diff
@@ -475,9 +440,9 @@ std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t e
     return msg;
 }
 
-CARTA::SpectralProfileData Message::SpectralProfileData(int32_t file_id, int32_t region_id, int32_t stokes, float progress,
-    std::string& coordinate, std::vector<CARTA::StatsType>& required_stats,
-    std::map<CARTA::StatsType, std::vector<double>>& spectral_data) {
+CARTA::SpectralProfileData Message::SpectralProfileData(int32_t stokes, float progress, int32_t file_id, int32_t region_id,
+    const std::string& coordinate, const std::vector<CARTA::StatsType>& required_stats,
+    const std::map<CARTA::StatsType, std::vector<double>>& spectral_data) {
```
confluence: See above re: reordering.

src/Util/Message.cc
```diff
@@ -494,19 +459,12 @@ CARTA::SpectralProfileData Message::SpectralProfileData(int32_t file_id, int32_t
             double nan_value = DOUBLE_NAN;
             new_profile->set_raw_values_fp64(&nan_value, sizeof(double));
         } else {
-            new_profile->set_raw_values_fp64(spectral_data[stats_type].data(), spectral_data[stats_type].size() * sizeof(double));
+            new_profile->set_raw_values_fp64(spectral_data.at(stats_type).data(), spectral_data.at(stats_type).size() * sizeof(double));
```
confluence: It shouldn't be necessary to make this change; the key is guaranteed to be in the map. Alternatively, the conditional could be refactored to attempt `at` first, catch an exception, and fall back to the default.

src/Util/Message.cc
```diff
@@ -740,6 +698,85 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    CARTA::FileInfoExtended& response, std::string name, const std::string& value, CARTA::EntryType type, double numeric_value) {
+    auto entry = response.add_computed_entries();
+    entry->set_name(name);
+    entry->set_value(value);
+    entry->set_entry_type(type);
+    entry->set_numeric_value(numeric_value);
+    return response;
+}
+
+CARTA::ImportRegionAck Message::AddImportedRegion(CARTA::ImportRegionAck& import_ack, int region_id, CARTA::RegionType region_type,
```
confluence: This doesn't appear to be used in the region code.

src/Util/Message.cc
```diff
@@ -740,6 +698,85 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    std::vector<CARTA::Point> control_points, float region_rotation, CARTA::RegionStyle region_style) {
+    // Set CARTA::RegionInfo
+    CARTA::RegionInfo region_info;
+    region_info.set_region_type(region_type);
+    *region_info.mutable_control_points() = {control_points.begin(), control_points.end()};
+    region_info.set_rotation(region_rotation);
+
+    // Add info and style to import_ack; increment region id for next region
```
confluence: Comment out of date? Increment should be happening outside; this function just gets the id passed in.

src/Util/Message.cc
```diff
@@ -740,6 +698,85 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    return message;
+}
+
+CARTA::SpatialProfileData Message::AddProfile(CARTA::SpatialProfileData& message, int32_t file_id, int32_t region_id, int32_t start,
+    std::vector<float>& profile, std::string& coordinate, int32_t mip, CARTA::ProfileAxisType axis_type, casacore::Quantity& increment) {
+    auto profile_size = profile.size();
+    int end = profile_size - 1;
+    float crpix = profile_size / 2;
+    float cdelt = increment.getValue();
+    float crval = (axis_type == CARTA::ProfileAxisType::Offset ? 0.0 : crpix * cdelt);
```
confluence: See above: all these calculations should happen *outside* the function; the function should just set the fields.

src/Util/Message.cc
```diff
@@ -740,6 +698,85 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    std::string unit = increment.getUnit();
+    message.set_file_id(file_id);
+    message.set_region_id(region_id);
```
confluence: This should happen in the function that creates the parent message.

src/Util/Message.cc
```diff
@@ -740,6 +698,85 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+    spatial_profile->set_end(end);
+    spatial_profile->set_raw_values_fp32(profile.data(), profile.size() * sizeof(float));
+    spatial_profile->set_coordinate(coordinate);
+    spatial_profile->set_mip(mip);
+    auto* profile_axis = spatial_profile->mutable_line_axis();
+    profile_axis->set_axis_type(axis_type);
+    profile_axis->set_crpix(crpix);
+    profile_axis->set_crval(crval);
+    profile_axis->set_cdelt(cdelt);
+    profile_axis->set_unit(unit);
```
confluence: There should be a separate wrapper for adding a profile axis (sub-)submessage to the spatial profile submessage. Then this alternative `AddProfile` can be eliminated. See general remarks re: return values -- all these wrappers that add submessages should return the *submessage*, not the parent message. Then you can chain operations (e.g. adding a submessage to a submessage, or setting another field on it).

src/Util/Message.cc
```diff
@@ -740,6 +698,85 @@ CARTA::RemoteFileRequest Message::RemoteFileRequest(int32_t file_id, const strin
...
+
+CARTA::FileInfoExtended Message::AddComputedEntry(
+    CARTA::FileInfoExtended& response, std::string name, const std::string& value, CARTA::EntryType type, double numeric_value) {
+    auto entry = response.add_computed_entries();
+    entry->set_name(name);
+    entry->set_value(value);
+    entry->set_entry_type(type);
+    entry->set_numeric_value(numeric_value);
+    return response;
+}
```
confluence: I know that we went back and forth on this interface, but I think it makes sense to rework this as follows:

1. You can combine both these functions into one by providing default parameters for `type` (`CARTA::EntryType::STRING`) and `numeric_value` (`0`) to make them optional. Inside the function, you don't even need an explicit check before setting the numeric value: setting it to zero should have the same result as not setting it at all (because that's the default value of a numeric field, and [protobuf won't serialize default values](https://protobuf.dev/programming-guides/field_presence/)).
2. You should add a corresponding `AddHeaderEntry` wrapper for adding header entries rather than computed entries. This is done in a few places (in `ImageData/CompressedFits.cc` and `FileList/FileExtInfoLoader.cc`). You could factor out the common code for *creating* the entry into a separate `HeaderEntry` wrapper, but this is such a simple object that it's not really necessary.
3. The code that adds normal header entries sets comments on them, but I'd rather not add a parameter for this (because then we'd either need to have two functions for each type of header or to specify unnecessary empty optional parameters in a subset of calls). Instead we can set the comment separately on the submessage (see my comments in other places that all these submessage-adding wrappers should return the submessage, not the main message). So you can do `Message::AddHeaderEntry(main_message, ....).set_comment(...)`, or do the set_comment later in a separate step, depending on how the code is arranged.
4. In at least one place in `FileExtInfoLoader.cc`, we set *only* the name in a header entry (when we're skipping over `COMMENT`, etc.). Rather than making the value optional in the wrapper, we should pass in the empty string here -- empty string is the default value for string fields, so again this will have the same effect as not setting the field.

confluence (changes requested):
General remarks:

1. I suggest splitting this by topic into different PRs. This PR can be for header entries (almost complete except for minor changes). Other message wrappers, which need a bit more work, should move to separate PRs: a) file / directory entries in the file list response; b) region import ack; c) spatial and spectral profiles.
2. All wrappers that add a submessage to a message should return the *submessage*, not the main message. This provides a reference to the calling code that can be used to chain operations. For example, you may want to set additional fields on the submessage, or to add a sub-submessage to it.

src/FileList/FileExtInfoLoader.cc
```diff
@@ -1369,35 +1235,22 @@ void FileExtInfoLoader::AddBeamEntry(CARTA::FileInfoExtended& extended_info, con
...
+                Message::AddHeaderEntry(extended_info, "BMIN", fmt::format("{:E}", minor_deg), CARTA::EntryType::FLOAT, minor_deg);
             header_entry->set_comment("extracted from HISTORY");
 
-            header_entry = extended_info.add_header_entries();
-            header_entry->set_name("BPA");
-            header_entry->set_entry_type(CARTA::EntryType::FLOAT);
-            header_entry->set_value(fmt::format("{:E}", pa));
-            header_entry->set_numeric_value(pa);
+            header_entry = Message::AddHeaderEntry(extended_info, "BPA", fmt::format("{:E}", pa), CARTA::EntryType::FLOAT, pa);
             header_entry->set_comment("extracted from HISTORY");
```
confluence: This is where I suggest using a map and a loop to avoid copying this verbose and identical formatting code three times:
```
std::map<std::string, double> beam_entries = {
    {"BMIN", major.get("deg").getValue()},
    // (...)
};

for (auto const& [name, value] : beam_entries) {
    // write the code once
}
```

confluence (changes requested):

confluence (approved):

kswang1029 (changes requested):
@Michaelazzz @confluence  loading a .fits.gz file results in a backend crash.

> Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   carta_backend                 	       0x104671d88 google::protobuf::internal::InternalMetadata::HasUnknownFieldsTag() const + 0 (metadata_lite.h:202) [inlined]
1   carta_backend                 	       0x104671d88 google::protobuf::internal::InternalMetadata::have_unknown_fields() const + 0 (metadata_lite.h:133) [inlined]
2   carta_backend                 	       0x104671d88 google::protobuf::internal::InternalMetadata::arena() const + 0 (metadata_lite.h:125) [inlined]
3   carta_backend                 	       0x104671d88 google::protobuf::MessageLite::GetArenaForAllocation() const + 0 (message_lite.h:440) [inlined]
4   carta_backend                 	       0x104671d88 void CARTA::HeaderEntry::set_comment<casacore::String&>(casacore::String&) + 4 (defs.pb.h:7538) [inlined]
5   carta_backend                 	       0x104671d88 carta::CompressedFits::AddHeaderEntry(casacore::String&, casacore::String&, casacore::String&, CARTA::FileInfoExtended&) + 348 (CompressedFits.cc:422)
6   carta_backend                 	       0x10466ec8c carta::CompressedFits::GetFitsHeaderInfo(std::__1::map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, CARTA::FileInfoExtended, std::__1::less<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const, CARTA::FileInfoExtended>>>&) + 1840 (CompressedFits.cc:113)
7   carta_backend                 	       0x1045dff44 carta::FileExtInfoLoader::FillFitsFileInfoMap(std::__1::map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, CARTA::FileInfoExtended, std::__1::less<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const, CARTA::FileInfoExtended>>>&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&) + 100 (FileExtInfoLoader.cc:44)
8   carta_backend                 	       0x1047bae4c carta::Session::FillExtendedFileInfo(std::__1::map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>, CARTA::FileInfoExtended, std::__1::less<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const, CARTA::FileInfoExtended>>>&, CARTA::FileInfo&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, bool, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>&) + 484 (Session.cc:215)
9   carta_backend                 	       0x1047bde48 carta::Session::OnFileInfoRequest(CARTA::FileInfoRequest const&, unsigned int) + 160 (Session.cc:429)
10  carta_backend                 	       0x1047d76b4 carta::SessionManager::FileInfoRequestHandler(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&) + 56 (SessionManager.cc:397)
11  carta_backend                 	       0x1047e1228 decltype(*std::declval<carta::SessionManager*>().*std::declval<void (carta::SessionManager::*&)(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&)>()(std::declval<carta::Session*&>(), std::declval<std::__1::basic_string_view<char, std::__1::char_traits<char>>&>(), std::declval<carta::EventHeader&>())) std::__1::__invoke[abi:ne180100]<void (carta::SessionManager::*&)(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&), carta::SessionManager*, carta::Session*&, std::__1::basic_string_view<char, std::__1::char_traits<char>>&, carta::EventHeader&, void>(void (carta::SessionManager::*&)(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&), carta::SessionManager*&&, carta::Session*&, std::__1::basic_string_view<char, std::__1::char_traits<char>>&, carta::EventHeader&) + 40 (invoke.h:312) [inlined]
12  carta_backend                 	       0x1047e1228 std::__1::invoke_result<void (carta::SessionManager::*&)(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&), carta::SessionManager*, carta::Session*&, std::__1::basic_string_view<char, std::__1::char_traits<char>>&, carta::EventHeader&>::type std::__1::invoke[abi:ne180100]<void (carta::SessionManager::*&)(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&), carta::SessionManager*, carta::Session*&, std::__1::basic_string_view<char, std::__1::char_traits<char>>&, carta::EventHeader&>(void (carta::SessionManager::*&)(carta::Session*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, carta::EventHeader const&), carta::SessionManager*&&, carta::Session*&, std::__1::basic_string_view<char, std::__1::char_traits<char>>&, carta::EventHeader&) + 40 (invoke.h:28) [inlined]
13  carta_backend                 	       0x1047e1228 carta::SessionManager::OnMessage(uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode) + 972 (SessionManager.cc:220)
14  carta_backend                 	       0x1047e521c ofats::any_detail::any_invocable_impl<void, false, uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode>::call(uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode) + 24 (MoveOnlyFunction.h:250) [inlined]
15  carta_backend                 	       0x1047e521c ofats::any_invocable<void (uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode)>::operator()(uWS::WebSocket<false, true, carta::PerSocketData>*, std::__1::basic_string_view<char, std::__1::char_traits<char>>, uWS::OpCode) + 24 (MoveOnlyFunction.h:357) [inlined]
16  carta_backend                 	       0x1047e521c uWS::WebSocketContext<false, true, carta::PerSocketData>::handleFragment(char*, unsigned long, unsigned int, int, bool, uWS::WebSocketState<true>*, void*) + 1248 (WebSocketContext.h:100)
17  carta_backend                 	       0x1047e3fb4 bool uWS::WebSocketProtocol<true, uWS::WebSocketContext<false, true, carta::PerSocketData>>::consumeMessage<6u, unsigned char>(unsigned char, char*&, unsigned int&, uWS::WebSocketState<true>*, void*) + 920 (WebSocketProtocol.h:366)
18  carta_backend                 	       0x1047e3ad8 uWS::WebSocketProtocol<true, uWS::WebSocketContext<false, true, carta::PerSocketData>>::consume(char*, unsigned int, uWS::WebSocketState<true>*, void*) + 176 (WebSocketProtocol.h:479)
19  carta_backend                 	       0x1047e39c4 auto uWS::WebSocketContext<false, true, carta::PerSocketData>::init()::'lambda'(auto*, char*, int)::operator()<us_socket_t>(auto*, char*, int) const + 120 (WebSocketContext.h:308)
20  carta_backend                 	       0x1047e3940 auto uWS::WebSocketContext<false, true, carta::PerSocketData>::init()::'lambda'(auto*, char*, int)::__invoke<us_socket_t>(auto*, char*, int) + 32 (WebSocketContext.h:286)
21  carta_backend                 	       0x1048972cc us_internal_dispatch_ready_poll + 1004 (loop.c:358)
22  carta_backend                 	       0x104897bcc us_loop_run + 204 (epoll_kqueue.c:147)
23  carta_backend                 	       0x1047e252c uWS::Loop::run() + 4 (Loop.h:216) [inlined]
24  carta_backend                 	       0x1047e252c uWS::run() + 12 (Loop.h:233) [inlined]
25  carta_backend                 	       0x1047e252c uWS::TemplatedApp<false>::run() + 12 (App.h:650) [inlined]
26  carta_backend                 	       0x1047e252c carta::SessionManager::RunApp() + 280 (SessionManager.cc:290)
27  carta_backend                 	       0x104743278 main + 3876 (Main.cc:236)
28  dyld                          	       0x18209ab98 start + 6076

src/FileList/FileExtInfoLoader.cc
```diff
@@ -1096,63 +992,42 @@ void FileExtInfoLoader::AddComputedEntriesFromHeaders(
...
         }
     }
 
     if (disp1_crpix != min_double && disp2_crpix != min_double) {
         std::string ref_pix = fmt::format("[{}, {}]", disp1_crpix, disp2_crpix);
-        auto comp_entry = extended_info.add_computed_entries();
-        comp_entry->set_name("Image reference pixels");
-        comp_entry->set_value(ref_pix);
-        comp_entry->set_entry_type(CARTA::EntryType::STRING);
+        Message::AddComputedEntry(extended_info, "Image reference pixels", ref_pix);
```
kswang1029, line 1002: @Michaelazzz Typo: Image reference pixels -> Image reference pixel

src/FileList/FileExtInfoLoader.cc
```diff
@@ -846,19 +768,14 @@ void FileExtInfoLoader::AddComputedEntries(CARTA::FileInfoExtended& extended_inf
...
         }
 
         if (!reference_pixels.empty()) {
-            auto entry = extended_info.add_computed_entries();
-            entry->set_name("Image reference pixels");
-            std::string ref_pix = fmt::format("[{}, {}]", reference_pixels(axes.x) + 1.0, reference_pixels(axes.y) + 1.0);
-            entry->set_value(ref_pix);
-            entry->set_entry_type(CARTA::EntryType::STRING);
+            auto name = fmt::format("Image reference pixel");
+            auto value = fmt::format("[{}, {}]", reference_pixels(axes.x) + 1.0, reference_pixels(axes.y) + 1.0);
```
kswang1029, line 776: @confluence shall we use "reference_pixels" or "reference_pixel"?

confluence: Good catch; I missed this. This PR shouldn't change any of the header values, and the entry is for a pair of pixels, so it should be plural.

src/FileList/FileExtInfoLoader.cc
confluence: The plural here is correct.

src/FileList/FileExtInfoLoader.cc
confluence: I will fix this now.

kswang1029 (approved):

PR #1464, opened 2025-06-06
Modified tile compression precision heuristic

### LINKED ISSUE #1447: zfp compression for moment 1 image in frequency
**Describe the bug**

user feedback via freshdesk

I am having a CARTA issue; I have a moment 1 map of a MeerKAT HI data cubelet generated with SoFiA. In CARTA we only see black where there is emission, despite the correct values in each pixel displaying when moused over (1st image). It doesn't help to change the displayed flux range (2nd image). However, everything displays fine in DS9 (3rd image). Also, the moment 1 map of the full SoFiA-generated HI cube (not the Sofia-generated cubelet at one source position) displays fine in Carta (4th image). Any ideas of what could be happening? I have multiple cubelet mom1 maps of this data which are having the same issue.

KSW: after investigating this deeper, it is something to do with the default tile compression quality factor we selected. If we bump it to the highest available (32), we can see the image correctly. 

**To Reproduce**
1. using the default zfp compression quality factor, load the attached image and see the issue like the following

Image

Get the test image: https://carta.freshdesk.com/helpdesk/attachments/69155254892?download=true

**Expected behavior**
The test image should be zfp compressed properly. 

 - OS [e.g. macOS Monterey]: macOS Sequoia 
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev, v5-beta
 - Frontend branch [e.g. dev, v3b2 release]: dev, v5-beta

confluence:
**Description**

Fixes #1447 and possibly other artifact problems.

The [previous heuristic for upgrading the precision](https://github.com/CARTAvis/carta-backend/pull/792) would always try the maximum, and discard the results if the new compression ratio was smaller than 10. I'm not sure what the rationale for this was, since typical initial ratios are often well below 10.

The new heuristic guesses higher precisions by performing a bisection and selecting the first guess with an appropriate ratio (or in the worst case eventually the maximum). It can handle the edge case where the ratio is a constant very high value regardless of the precision (e.g. in a tile with all NaNs). In my brief tests, one guess is sufficient, so for a starting precision of 11 it upgrades to 22. The theoretical worst case is 5 or 6 guesses, depending on the starting precision.

**Checklist**

- [X] changelog updated


src/Frame/Frame.cc
```diff
@@ -531,42 +529,40 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
-                    spdlog::debug("Using high compression quality. Previous compression ratio: {:.3f}", compression_ratio);
-                    compression_ratio = compression_ratio_hq;
-                    use_high_precision = true;
+
+            // originally requested precision
+            int requested_precision(lround(compression_quality));
+
+            auto find_precision = [&](const auto& self, int current, float previous_ratio) -> int {
+                Compress(*tile_data_ptr, 0, compression_buffer, compressed_size, tile_width, tile_height, current);
+                float compression_ratio = (float)tile_image_data_size / compressed_size;
```
kswang1029, line 541: shouldn't we compare the highest compression quality (32) with the desired one and then judge? IIRC that's the original idea in the previous heuristic.

confluence: @kswang1029 but it isn't always *necessary* to use the highest precision -- a lower precision, higher than the requested one, may be sufficient to make the ratio low enough, while not being as low as it would be at the maximum precision (which would be inefficient).

Previously only the highest precision would be tried, and then discarded anyway in favour of the original, if the ratio was lower than the second threshold (even though the ratio at the original precision is never tested against this threshold). So it was all or nothing -- the maximum precision with the lowest possible compression, if it wasn't lower than an arbitrary threshold, or the original (possibly bad) compressed data, with no other fallbacks.

This heuristic tries other values between the current precision and the maximum, so it is more likely to find a compromise value for which the ratio is not excessively high *or* unnecessarily low.

It isn't guaranteed to find the *lowest possible* precision, because that would require a more complex search in both directions, and more guesses. I initially considered that, but opted for this simpler solution until we can investigate the library options further.

I believe that the second threshold is unnecessary: if the initial precision has a ratio only slightly above the first threshold, it's likely that the first guess halfway to the maximum will be accepted, with a ratio that is unlikely to be very low. If the algorithm terminates at the maximum, the maximum is either still above the ratio threshold or not far below.

If a lower final value is desired, I think I could tune this to make the guesses smaller increments away from the current value (e.g. a third rather than a half), but I'd have to think about the rounding so that the algorithm can terminate. And there will be more guesses in the worst case. Or I could try implementing a more complex bidirectional search. But this level of complexity may not be necessary, depending on how the current algorithm performs.

kswang1029: do you mean if we request 11, then we compute 22, then compute the ratio to see if the threshold is met (and iterates)?

kswang1029 (approved):
Tested with some images that was causing compression artifacts (model image and moment 1 images) and they all look fine. 👍

src/Frame/Frame.cc
confluence: Yes, that's correct, but we only start iterating if 11 does not meet the threshold. The initial calculation isn't a separate step; it's been folded into the recursive function as the starting guess.

PR #1465, opened 2025-06-09
Map resumed session ID on viewer registration

confluence:
**Description**

This is a followup to #1417 which resolves a design conflict with #1440 and also fixes a small bug in the original PR.

Instead of adding the mapping between each newly created session ID and itself on connection, and also adding the mapping between a resumed session ID and the new session ID when a session is resumed, we can do this once when the viewer is registered -- this message triggers the change of the ID in the session object, so we can obtain both values from the session object (before and after), which is available in the handler function in #1440. Additionally, in the original PR, we were creating duplicate mappings for resumed sessions and never removing them (although they would never be accessed).

**Checklist**

- [X] no changelog update needed


Michaelazzz (approved):
looks good

confluence:
I have tested the change to confirm that scripting can still find both new and resumed sessions correctly by ID.

PR #1469, opened 2025-06-13
Yuhsuan/fitting init val

YuHsuan-Hwang:
**Description**

  Supports initial value auto generation for image fitting.

  * When there are NaN values in the request message `FittingRequest.initial_values`, `ImageFitter.SetInitialValues` returns false and initial values are regenerated using `ImageFitter.CalculateInitialValues`.
  * Added a new class `InitialValueCalculator` for generating initial values.
    * A KMeans++ clustering method is used to classify data points into different sources.
    * The method of moments is calculated twice to estimate Gaussian parameters. The first calculation obtains the source area and select a circular area so that the estimate will not be affect by the image boundaries. The second calculation obtains the Gaussian parameters based on the selected area.
    * The threshold is set to 4 * MAD by default. It is lowered when the result is not ideal, for example, when there are too few data points, KMeans++ produces fewer groups than needed, the method of moments obtains extremely large FWHM, or most part of a source is outside the image.
  * The result of the generated initial values are added in the response message `FittingResponse.log`.

  Frontend PR: https://github.com/CARTAvis/carta-frontend/pull/2546

ToDo: update documentation, update change log

**Checklist**

- [ ] ICD test passing / ~corresponding fix added~ / ~new ICD test created~
- [x] GitHub Project estimate added


src/ImageFitting/ImageFitter.cc
```diff
@@ -188,6 +226,15 @@ void ImageFitter::SetInitialValues(
         }
     }
     _fdf.p = p;
+
+    return true;
+}
+
+bool ImageFitter::CalculateInitialValues(std::vector<CARTA::GaussianComponent>& initial_values) {
+    InitialValueCalculator* calculator =
+        new InitialValueCalculator(_fit_data.data, _fit_data.width, _fit_data.n / _fit_data.width, _fit_data.offset_x, _fit_data.offset_y);
```
confluence: I suggest passing a single FitData struct parameter into this constructor and unpacking it in the constructor.

These helper structs should also be moved from this file to a separate file in this directory, e.g. `ImageFitting/Util.[h|cc]` so that both the fitter and the calculator can use them.

src/ImageFitting/InitialValueCalculator.cc
```diff
@@ -0,0 +1,328 @@
...
+#include "InitialValueCalculator.h"
+
+using namespace carta;
+
+InitialValueCalculator::InitialValueCalculator(float* image, size_t width, size_t height, size_t offset_x, size_t offset_y) {
+    _image = image;
+    _width = width;
+    _height = height;
+    _offset_x = offset_x;
+    _offset_y = offset_y;
```
confluence: Please move these into an initializer list (in addition to passing in one FitData parameter). This should also be done for the other constructors (and the structs themselves should have constructors), but that may be outside the scope of this PR.

src/ImageFitting/ImageFitter.cc
```diff
@@ -48,7 +48,35 @@ bool ImageFitter::FitImage(size_t width, size_t height, float* image, double bea
...
     CalculateNanNumAndStd();
-    SetInitialValues(initial_values, background_offset, fixed_params);
+    success = SetInitialValues(initial_values, background_offset, fixed_params);
+
+    std::string initial_value_log = "";
+    if (!success) {
+        spdlog::info("Generating initial values for fitting.");
+        success = CalculateInitialValues(initial_values);
+        if (success) {
+            initial_value_log = InitialValueCalculator::GetLog(initial_values, _unit);
```
confluence: I would eliminate `ImageFitter::CalculateInitialValues` entirely, and construct the calculator object instance directly inside this function. This would also allow streamlining the design of the calculator, which is currently instantiated with some parameters, has one instance function run with additional parameters, and then has the instance discarded and has a static function run with the same initial parameters.

I suggest instead passing `initial_values` and `_image_std` into the calculator constructor as well, calling `CalculateInitialValues` here (no additional parameters), making `GetLog` an instance function, and calling it here with only `_unit`. Inside the calculator, the functions can use the initial values and standard deviation properties set on the instance by the constructor.

confluence (changes requested):
I have some suggestions for the structure of the objects; other than that the code looks fine.

src/ImageFitting/InitialValueCalculator.cc
YuHsuan-Hwang: @confluence For `the other constructors`, do you suggest we apply initializer list to any fitting related constructors, or do you mean codebase not related to fitting?

src/ImageFitting/ImageFitter.cc
YuHsuan-Hwang: Changed the input parameter to FitData.

src/ImageFitting/InitialValueCalculator.cc
YuHsuan-Hwang: Applied initializer list to the calculator constructor.

src/ImageFitting/ImageFitter.cc
YuHsuan-Hwang: Removed `ImageFitter::CalculateInitialValues`, passed `initial_values` and `_image_std` into the calculator constructor, and adjusted `GetLog` function.

YuHsuan-Hwang:
@confluence The structure is adjusted. Thanks for the suggestions.

src/ImageFitting/InitialValueCalculator.cc
confluence: > @confluence For `the other constructors`, do you suggest we apply initializer list to any fitting related constructors, or do you mean codebase not related to fitting?

I meant the other constructors in these image fitting files -- since they aren't modified by this PR, it's not strictly related. I will re-review.

src/ImageFitting/InitialValueCalculator.cc
```diff
@@ -0,0 +1,329 @@
...
+
+using namespace carta;
+
+InitialValueCalculator::InitialValueCalculator(FitData* fit_data, std::vector<CARTA::GaussianComponent>& initial_values, float image_std)
+    : _image(fit_data->data),
+      _width(fit_data->width),
+      _height(fit_data->n / fit_data->width),
+      _offset_x(fit_data->offset_x),
+      _offset_y(fit_data->offset_y),
+      _initial_values(initial_values),
```
confluence: Sorry; I've realised that this is an in-out parameter, and setting it as a property is a bit awkward. I also think that the log construction can be merged into the public calculation function (since this is a log of the calculation process).

My revised suggestion:
* Don't pass `initial_values` to the constructor
* Do pass it into `CalculateInitialValues` as an in-out parameter
* Pass the image unit into the constructor instead
* Pass `initial_value_log` into `CalculateInitialValues` as another in-out parameter
* Build the log string inside `CalculateInitialValues`.

src/ImageFitting/ImageFitter.cc
confluence: I've realised that half of this refactoring request was a bit silly; I have included an updated suggestion in my review.

confluence (changes requested):
I have an updated suggestion for the function signatures; I think that's the last thing.

src/ImageFitting/InitialValueCalculator.cc
YuHsuan-Hwang: I see your point. I'll adjust it.

src/ImageFitting/ImageFitter.cc
```diff
@@ -48,7 +48,36 @@ bool ImageFitter::FitImage(size_t width, size_t height, float* image, double bea
     _progress_callback = progress_callback;
 
     CalculateNanNumAndStd();
-    SetInitialValues(initial_values, background_offset, fixed_params);
+    success = SetInitialValues(initial_values, background_offset, fixed_params);
+
+    std::string initial_value_log = "";
+    if (!success) {
+        spdlog::info("Generating initial values for fitting.");
+        InitialValueCalculator* calculator = new InitialValueCalculator(&_fit_data, initial_values, _image_std);
```
pford: Memory leak--`new` but no `delete`, use unique_ptr

src/ImageFitting/ImageFitter.h
```diff
@@ -220,13 +201,13 @@ class ImageFitter {
      * @param initial_values Initial fitting parameter
      * @param offset_x X-axis offset from the fitting region to the entire image
      * @param offset_y Y-axis offset from the fitting region to the entire image
-     * @return A tuple of Gaussian parameters
+     * @return A tuple of Gaussian parameters: center x, center y, amplitude, FWHM x, FWHM y, and position angle
```
pford: Add struct to Util.h instead of tuple, use in ImageFitter and InitialValueCalculator

pford (changes requested):

src/ImageFitting/ImageFitter.cc
confluence: Good point; I missed this. But why is this a pointer at all? It could just be a normal object variable.

pford: That's true, it does not get passed anywhere.

src/ImageFitting/InitialValueCalculator.cc
YuHsuan-Hwang: Revised based on the new suggestions.

src/ImageFitting/ImageFitter.cc
YuHsuan-Hwang: I see. Changed to a normal object.

src/ImageFitting/ImageFitter.h
YuHsuan-Hwang: Added a new struct `GaussianParams` in `Util.h`. Also moved `GetGaussianComponent` to `Util.h` and applied it to the calculator.

YuHsuan-Hwang:
@confluence @pford Thanks for the comments. Please check my new changes, thanks!

src/ImageFitting/Util.h
```diff
@@ -0,0 +1,82 @@
...
+ * @brief Create a Gaussian component sub-message from Gaussian parameters.
+ * @param params Gaussian parameters including center x, center y, amplitude, FWHM x, FWHM y, and position angle
+ * @return A Gaussian component sub-message
+ */
+inline CARTA::GaussianComponent GetGaussianComponent(GaussianParams params) {
+    auto center = Message::DoublePoint(params.center_x, params.center_y);
+    auto fwhm = Message::DoublePoint(params.fwhm_x, params.fwhm_y);
+    auto component = Message::GaussianComponent(center, params.amp, fwhm, params.pa);
+    return component;
+};
```
confluence: This should be an instance method of GaussianParams.

src/ImageFitting/ImageFitter.cc
```diff
@@ -492,8 +538,8 @@ void ImageFitter::ErrorHandler(const char* reason, const char* file, int line, i
     spdlog::error("gsl error: {} line{}: {}", file, line, reason);
 }
 
-std::tuple<double, double, double, double, double, double> ImageFitter::GetGaussianParams(const gsl_vector* value_vector, size_t index,
-    std::vector<int>& fit_values_indexes, std::vector<double>& initial_values, size_t offset_x, size_t offset_y) {
+GaussianParams ImageFitter::GetGaussianParams(const gsl_vector* value_vector, size_t index, std::vector<int>& fit_values_indexes,
```
confluence, line 538: This should be an additional constructor of `GaussianParams`.

src/ImageFitting/ImageFitter.cc
```diff
@@ -268,9 +307,11 @@ void ImageFitter::CalculateErrors() {
...
 
     for (size_t i = 0; i < _num_components; i++) {
-        double center_x, center_y, amp, fwhm_x, fwhm_y, pa;
-        std::tie(center_x, center_y, amp, fwhm_x, fwhm_y, pa) =
-            GetGaussianParams(_fit_values, i * 6, _fit_data.fit_values_indexes, _fit_data.initial_values, 0, 0);
+        GaussianParams params = GetGaussianParams(_fit_values, i * 6, _fit_data.fit_values_indexes, _fit_data.initial_values, 0, 0);
+        double amp = params.amp;
+        double fwhm_x = params.fwhm_x;
+        double fwhm_y = params.fwhm_y;
+        double pa = params.pa;
```
confluence: You can use a structured binding for this unpacking instead:
```
auto [center_x, center_y, amp, fwhm_x, fwhm_y, pa] = GetGaussianParams(...);
```

(But see also the previous comment about turning that function into a constructor.)

confluence (changes requested):

src/ImageFitting/ImageFitter.cc
YuHsuan-Hwang: @confluence I would prefer to remain this as a method in the fitter because it performs complex calculation related to fitting (checking `fit_values_indexes` and selecting fixed or fitted values). What do you think?

confluence: OK, that's fair -- I was on the fence about this. I think it's fine to leave it where it is.

src/ImageFitting/Util.h
YuHsuan-Hwang: Changed to method of GaussianParams.

src/ImageFitting/ImageFitter.cc
YuHsuan-Hwang: Applied structured binding.

YuHsuan-Hwang:
@confluence @kswang1029 I have updated the PR:
* Adjusted codebase according to reviews.
* Fixed incorrect guessed center with FOV and regions.
* Improved debug log.

src/ImageFitting/ImageFitter.cc
```diff
@@ -440,8 +478,13 @@ int ImageFitter::FuncF(const gsl_vector* fit_values, void* fit_data, gsl_vector*
...
 
-        double center_x, center_y, amp, fwhm_x, fwhm_y, pa;
-        std::tie(center_x, center_y, amp, fwhm_x, fwhm_y, pa) = GetGaussianParams(fit_values, k, d->fit_values_indexes, d->initial_values);
+        GaussianParams params = GetGaussianParams(fit_values, k, d->fit_values_indexes, d->initial_values);
+        double center_x = params.center_x;
+        double center_y = params.center_y;
+        double amp = params.amp;
+        double fwhm_x = params.fwhm_x;
+        double fwhm_y = params.fwhm_y;
+        double pa = params.pa;
```
confluence, line 487: The structured binding should also be used here.

confluence (changes requested):

src/ImageFitting/Util.h
```diff
@@ -0,0 +1,81 @@
...
+     * @brief Create a Gaussian component sub-message.
+     * @return A Gaussian component sub-message
+     */
+    CARTA::GaussianComponent GetGaussianComponent() {
+        auto center = Message::DoublePoint(center_x, center_y);
+        auto fwhm = Message::DoublePoint(fwhm_x, fwhm_y);
+        auto component = Message::GaussianComponent(center, amp, fwhm, pa);
+        return component;
+    };
+};
```
confluence, line 77: I have a much cleaner suggestion for this. The only reason that this function exists is to group the center and fwhm parameters into sub-messages -- but these are all message internals, and a more appropriate place for them is `Util::Message`.

`Message::GaussianComponent` is *only* used by this code, so there's no reason for it to require the parameters in this format -- you can make that function take these six separate parameters instead, and do this packing inside.

Everywhere that this function is called, you already have access to the separate parameters, or you can use structured binding to get them instead of the single struct. So if you change the signature of `Message::GaussianComponent`, you can call it directly instead.

(Michaela is currently working on extending `Message` to functions which add submessages to messages, to simplify how this is done, but I think that level of refactoring is outside the scope of this PR. We can handle that in a later refactoring pass.)

confluence (changes requested):

kswang1029 (approved):
this looks good to me. e2e tests are adjusted accordingly.

src/ImageFitting/Util.h
YuHsuan-Hwang: Ok. Let's work on this in future PRs.

src/ImageFitting/ImageFitter.cc
YuHsuan-Hwang: @confluence I get compile errors if applying here:
```shell
error: capturing a structured binding is not yet supported in OpenMP
  503 |                 double dx = i % d->width - center_x;
      |                                            ^
```
Or do you prefer we use `std::tie`?

confluence: Oh, I see. Let's leave it alone for now.

confluence (approved):

pford (approved):

PR #1471, opened 2025-06-13
Restore uWebSockets submodule, with uSockets managed with separate top-level submodule

### LINKED ISSUE #1411: Investigate options for submodules for third-party dependencies
1. Apparently the [`git` functionality for cloning submodules shallowly](https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules) has advanced a lot in the past few years. It is now possible to configure this per submodule. This would speed up cloning of the repository, both for end-users building from source and in our CI, and may make it unnecessary to e.g. use [this indirect release-tracking repository](https://github.com/astoeckel/json/), which hasn't updated in 5 years. The CI checkout action already checks out the main repo with depth 1.
2. The reason we switched away from using a submodule for uWebSockets is that the new version has a lot of submodules that we don't need, and which take a long time to clone. We could instead disable automatic updates for `uWebsockets` and separately update it with the only child submodule we need. Alternatively, while there doesn't seem to be a way to [whitelist a single sub-submodule](https://stackoverflow.com/questions/79284756/exclude-almost-all-nested-submodules-from-one-submodule) in the configuration, I think it's possible to [blacklist all the others explicitly](https://stackoverflow.com/questions/56745097/exclude-submodule-of-a-submodule/) (needs testing).

confluence:
**Description**

Fixes #1411.

This change mostly affects the build process. My rationale for this is that requiring cmake to download source code *during the build process* removes the clean separation between fetching a copy of the source and building the source, and makes it more difficult to build the backend in certain sandboxed environments that either disallow arbitrary external downloads or require additional configuration to allow them. To build the debian packages I currently have to abuse a flag that forces cmake to proceed with the build even though it thinks that the downloaded source is missing (because cmake did not download it).

This PR preserves separate top-level directories for `uWebSockets` and `uSockets` because of limitations in git's submodule management options (git *really* does not handle this use case well: selective recursive fetches are only possible if you manually blacklist each unwanted nested module on the commandline).

To avoid the download of unwanted nested submodules, we should stop using `--recursive` to update and stop using `--recurse-submodules` when cloning. We started updating recursively in the first place because of `uWebSockets`; none of the other submodules have nested submodules.

This PR also bumps uSockets to latest version, to match submodule version used by the version of uWebSockets we're using. It's a minor version update with a handful of commits, so this should not cause any issues.

**Checklist**

- [X] changelog updated


pford (approved):
Works for me on macOS 14.  Was able to build and run backend.

Issue #1475, opened 2025-06-18
Messages missing from handler


pford:
**Describe the bug**
Backend logged a "handler not found" error for a valid message type:

```
[2025-06-18 19:43:31.242Z] [CARTA] [error] Handler not found for event type: 17
Out of range error: unordered_map::at: key not found
```

This message is STOP_ANIMATION. I went through the message list and also found no handler for CLOSE_PV_PREVIEW.

**To Reproduce**
1. Open image
2. Start animation
3. Stop animation

**Expected behavior**
All valid messages handled.

Michaelazzz:
Thank you for catching this @pford 

A patch is on the way 🪄

Michaelazzz:
Closed by PR #1476

PR #1476, opened 2025-06-19
Addition of missing message handlers

Michaelazzz:
**Description**

The handlers for the STOP_ANIMATION and STOP_PV_PREVIEW were missing, highlighted in issue #1475 
The missing message handlers are added back into the codebase.
No
1. Open image
2. Start animation
3. Stop animation

If there is no error which resembles the following message, then the issue has been resolved.
`[2025-06-18 19:43:31.242Z] [CARTA] [error] Handler not found for event type: 17
Out of range error: unordered_map::at: key not found`

**Checklist**

- [ ] no changelog update needed
- [x] e2e test passing / corresponding fix added 
- [ ] ICD test passing / corresponding fix added

### DIFF
--- a/src/Session/SessionManager.cc
+++ b/src/Session/SessionManager.cc
@@ -210,7 +210,7 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
         try {
             std::invoke(handler, this, session, sv_message, head);
         } catch (const message_parsing_exception& e) {
-            std::string message = fmt::format("Error handling event: {}", CARTA::EventType_Name(event_type));
+            std::string message = fmt::format("Error parsing {} message", CARTA::EventType_Name(event_type));
             spdlog::error(message);
             session->SendLogEvent(message, {"event"}, CARTA::ErrorSeverity::ERROR);
         }
--- a/src/Session/SessionManager.cc
+++ b/src/Session/SessionManager.cc
@@ -212,7 +212,7 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
         } catch (const message_parsing_exception& e) {
             std::string message = fmt::format("Error handling event: {}", CARTA::EventType_Name(event_type));
             spdlog::error(message);
-            session->SendLogEvent(message);
+            session->SendLogEvent(message, {"event"}, CARTA::ErrorSeverity::ERROR);
         }
     } else if (op_code == uWS::OpCode::TEXT) {
         if (sv_message == "PING") {
--- a/src/Session/SessionManager.cc
+++ b/src/Session/SessionManager.cc
@@ -196,16 +196,17 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
 
         logger::LogReceivedEventType(event_type);
 
-        auto handler = _message_handlers.find(event_type);
-        if (handler == _message_handlers.end()) {
+        MessageHandler handler;
+        try {
+            handler = _message_handlers.at(event_type);
+        } catch (const std::out_of_range& e) {
             spdlog::error("Handler not found for event type: {}", CARTA::EventType_Name(event_type));
-            return;
         }
 
         try {
-            std::invoke(handler->second, this, session, sv_message, head);
+            std::invoke(handler, this, session, sv_message, head);
         } catch (const message_parsing_exception& e) {
-            spdlog::error("Error handling event {} in session {}: {}", event_type, session->GetId(), e.what());
+            spdlog::error("Error handling event {} in session {}: {}", CARTA::EventType_Name(event_type), session->GetId(), e.what());
         }
     } else if (op_code == uWS::OpCode::TEXT) {
         if (sv_message == "PING") {
--- a/src/Session/SessionManager.cc
+++ b/src/Session/SessionManager.cc
@@ -191,6 +191,7 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
 
         if (!CARTA::EventType_IsValid(event_type)) {
             spdlog::error("Bad event type: {}", event_type);
+            session->SendLogEvent("Bad event type: " + event_type, {"event"}, CARTA::ErrorSeverity::ERROR);
             return;
         }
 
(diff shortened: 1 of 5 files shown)


confluence (approved):
I checked the previous revisions, and I think there are no other missing handlers.

src/Session/SessionManager.cc
```diff
@@ -361,6 +363,11 @@ void SessionManager::StartAnimationHandler(Session* session, std::string_view sv
     ThreadManager::QueueTask(tsk);
 };
 
+void SessionManager::StopAnimationHandler(Session* session, std::string_view sv_message, const EventHeader& head) {
+    CARTA::StopAnimation message = Message::DecodeMessage<CARTA::StartAnimation>(sv_message);
```
confluence: You have a copy & paste error here (please check the CI output).

confluence (changes requested):

confluence (approved):

confluence:
I don't want to delay merging this (code freeze tomorrow!), but it would be good if the log message printed the human-readable name of the event type rather than the numeric value (it did this in the old code).

The old code used `CARTA::EventType_Name(CARTA::EventType(event_type))` to get the name. `event_type` is already a `CARTA::EventType` (even if the numeric value is not valid; we use proto3; see [here](https://protobuf.dev/reference/cpp/cpp-generated/#enum)), so it's not necessary to wrap it in a new object -- you should be able to call _Name on the bare variable.

However, in the block where you handle the missing handler, you should distinguish between the case we hit here (valid event type; missing handler) and what might happen in the case of a backend / frontend protobuf version mismatch (event type is not in the backend's enum, in which case you'll get a blank string for the name). You could use `CARTA::EventType_IsValid` to check, or check whether the name is blank. If it's a valid event type, log the missing handler error using the name. If it's an invalid event type, log a "bad event type" error instead, using the numeric value.

Also, please remove the printing of the "out of range" error: it's bypassing the logger by going to cerr, and it's unnecessary because it's exposing an implementation detail which is not relevant. This try/catch block is _replacing_ this internal error by interpreting it as a more specific error.

src/Session/SessionManager.cc
```diff
@@ -185,20 +187,23 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
+
+        auto handler = _message_handlers.find(event_type);
+        if (handler == _message_handlers.end()) {
+            spdlog::error("Handler not found for event type: {}", CARTA::EventType_Name(event_type));
             return;
         }
 
         try {
-            std::invoke(handler, this, session, sv_message, head);
+            std::invoke(handler->second, this, session, sv_message, head);
```
confluence: I would keep the original at + try/catch -- it's a bit cleaner, and lets you use the `handler` variable name for the actual handler object rather than the iterator. Just don't print the original exception when you catch it.

confluence (changes requested):

confluence (approved):

confluence (changes requested):
@Michaelazzz one more thing. It's currently difficult to test for missing handlers / bad event types / malformed messages with the E2E tests because the backend isn't sending any messages back to the frontend to inform it of these failures -- it only logs them to its own log.

There is an `ErrorData` message which can be used for this, and the `Session::SendLogEvent` function is used to send these messages. Currently it's only used by functions inside the session, so it's protected -- I suggest making it public. You can then use it inside `OnMessage` to send these error events in the same places as you are currently logging these three error cases (bad event; no handler; bad message). You have a session object at that point. I suggest doing both (logging locally *and* sending the event). You can use "event" as the tag for the log event.

Please also change the error message that you print for the message parsing failure. You are *only* catching the custom `message_parsing_exception` there, so you know it's a message parsing error, and the message can be more brief. The session ID and exception contents can be omitted (the message never reaches the session, and the contents are always the same).

I have also noticed that:

1. The documentation comments for `Message::DecodeMessage` say that it can throw a runtime error, but it specifically only throws `message_parsing_exception`, so this description should be updated.
2. in all the event handlers, it should be possible to use `auto message = Message::DecodeMessage<CARTA::SomeMessageType>(sv_message);`, because the output type should be unambiguous from the explicit function type. This would both be less verbose and less error-prone (removes repetition).

kswang1029 (approved):
I run e2e tests and investigated the resulting backend log file, and did not see errors about missing handlers.

src/Session/SessionManager.cc
```diff
@@ -185,22 +187,30 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
         session->UpdateLastMessageTimestamp();
 
         EventHeader head = Message::GetEventHeader(sv_message);
-
         CARTA::EventType event_type = head.GetType();
+
+        if (!CARTA::EventType_IsValid(event_type)) {
+            spdlog::error("Bad event type: {}", event_type);
+            session->SendLogEvent("Bad event type: " + event_type, {"event"}, CARTA::ErrorSeverity::ERROR);
```
confluence: Please create a string variable for the error message (using `fmt::format`) and reuse it for both the local log and the event.

Normally you would avoid doing this kind of interpolation before using the log function (because it interpolates on demand depending on the log level), but in this case avoiding repetition is more important, and typically the log level will include errors.

(This also applies to the other two places.)

src/Session/SessionManager.cc
```diff
@@ -185,22 +187,30 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
+                "Handler not found for event type: " + CARTA::EventType_Name(event_type), {"event"}, CARTA::ErrorSeverity::ERROR);
         }
 
         try {
             std::invoke(handler, this, session, sv_message, head);
-        } catch (const message_parsing_exception& e) {
-            spdlog::error("Error handling event {} in session {}: {}", event_type, session->GetId(), e.what());
+        } catch (const parsing_exception& e) {
+            spdlog::error("Error handling event: {}", CARTA::EventType_Name(event_type));
+            session->SendLogEvent("Error handling event: " + CARTA::EventType_Name(event_type), {"event"}, CARTA::ErrorSeverity::ERROR);
```
confluence: Please see my previous comment about rewriting this message.

src/Util/Message.tcc
```diff
@@ -9,16 +9,15 @@
 
 #include "Message.h"
 
-struct message_parsing_exception : std::runtime_error {
+struct parsing_exception : std::runtime_error {
```
confluence: I would keep the original name -- it's less ambiguous. This struct is available to any file that imports `Message.h`.

confluence (changes requested):

src/Session/SessionManager.cc
confluence: For clarity, what I meant was something like: "Error parsing {} message." (Because you know that this can only be a message parsing error.)

src/Session/SessionManager.cc
```diff
@@ -185,22 +187,32 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
         session->UpdateLastMessageTimestamp();
 
         EventHeader head = Message::GetEventHeader(sv_message);
-
         CARTA::EventType event_type = head.GetType();
+
+        if (!CARTA::EventType_IsValid(event_type)) {
+            std::string message = fmt::format("Bad event type: {}", event_type);
```
confluence: Minor request: please call these string variables `error`; this is a function which handles messages, so `message` is confusing.

confluence (changes requested):

confluence (approved):

PR #1479, opened 2025-06-24
Documentation: corrected beta and rc versions in release checklist examples

confluence:
We haven't been very consistent in our numbering of `beta` and `rc` versions. We usually start at `beta.1` but `rc.0`. The release checklist currently recommends `.1`. We have discussed this and decided to standardise on `.0` -- that's how we originally did this.


loveluthien (approved):
😎

PR #1482, opened 2025-06-25
Fix test executable linker error on Ubuntu Noble

### LINKED ISSUE #1481: Linker fails on test executable on Ubuntu Noble if build mode is not Debug
This was discovered by @Michaelazzz on a Noble desktop and reproduced by @confluence in a Noble docker container.

Two unit tests which use `DataColumn` don't import `DataColumn.tcc`, where the template is defined. This leads to a linker error in the test executable (since one of the recent changes).

I can't fully explain why this previously *did* work -- I assume that the objects were included by the linker because of a side effect which is eliminated by some combination of: our recent changes to the `cmake` config, updated versions of the tools, and optimisations outside `Debug` mode.

I didn't trigger it locally because I'm still using Jammy, and the Noble CI didn't trigger it because it builds in `Debug` mode. The Noble package shouldn't trigger it because it doesn't build tests (so this should not affect the 5.0 stable release).

I haven't tried `Release` mode; the default `RelWithDebInfo` is definitely affected. I also haven't attempted to narrow down the exact commit which introduces the behaviour. Regardless of the details, the root cause is clearly a mistake in the code and should be fixed.

**To Reproduce**
1. Launch a brand new Noble container
2. Follow the instructions from the Noble dockerfile to install dependencies
3. Check out the repo and update / init submodules
4. Attempt to build with tests enabled but with the default `RelWithDebInfo` build mode
5. The build should compile correctly, but fail during linking of the test executable.

**Expected behavior**
The build should succeed.

**Platform info**
 - OS [e.g. macOS Monterey]: Ubuntu Noble
 - Backend branch [e.g. dev, v3b2 release]: latest dev / 5.0.0-rc.0

**Additional context**
I have a draft fix which requires more refactoring.

confluence:
Fixes #1481.

This PR:
* Includes the missing `DataColumn.tcc` file in the two tests that need it
* Removes an unnecessary include of `OnMessageTask.h` from `ThreadingManager.h`, which was pulling *a lot* of nested includes from `Session`, `SessionManager`, etc., into low-level components which had no business including them (and only needed access to the parallel sort or static `ThreadManager` functions).
* Fixes errors in code which was previously relying on side effects of those includes (including a `using namespace std` somewhere) -- mostly explicitly specifying the `std` namespace and including a few headers directly.

Still to do before PR is ready for review:
- [X] Refactor to include `DataColumn.tcc` only in `Columns.h` -- this approach was previously impossible because of a circular dependency, but removing the unnecessary include may have fixed this.
- [ ] ~Possibly clean up some other namespace issues in tests: a lot of unnecessary or inconsistent use of the `carta` namespace (we should use `using`).~ We have inconsistent namespace usage in multiple places; I would prefer to handle this in an unrelated PR.
- [X] Possible refactoring: move `parallel_sort` into `Concurrency`; move `Concurrency` into `Util`
- [X] Possible refactoring: decouple generic / low-level functionality of `ThreadManager` from `OnMessageTask` dependency entirely by creating a generic `Task` interface to be a base class for `OnMessageTask`.
- [X] Rename files to match class names

To test: follow the steps to reproduce #1481 -- this branch should build correctly.

**Checklist**

- [X] changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+Fix linker error during unit test build on some platforms ([#1481](https://github.com/CARTAvis/carta-backend/issues/1481)).
+
 ## [5.0.0]
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [5.0.2]
+
+### Fixed
+* Fixed bug in region spectral profile with flux density statistic ([#1493](https://github.com/CARTAvis/carta-backend/issues/1493)).
+
+## [5.0.1]
+
+### Fixed
+* Fixed bugs in database validation error handling which caused preferences to be wiped, and corrected preferences schema ([#1485](https://github.com/CARTAvis/carta-backend/issues/1485)).
 
 ### Fixed
 * Fix linker error during unit test build on some platforms ([#1481](https://github.com/CARTAvis/carta-backend/issues/1481)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+* Fixed linker error during unit test build on some platforms ([#1481](https://github.com/CARTAvis/carta-backend/issues/1481)).
+
 ## [5.0.2]
 
 ### Fixed
@@ -13,9 +18,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Fixed bugs in database validation error handling which caused preferences to be wiped, and corrected preferences schema ([#1485](https://github.com/CARTAvis/carta-backend/issues/1485)).
-
-### Fixed
-* Fix linker error during unit test build on some platforms ([#1481](https://github.com/CARTAvis/carta-backend/issues/1481)).
 
 ## [5.0.0]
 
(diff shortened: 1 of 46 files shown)


Michaelazzz (approved):
I tested the compiling on my problem machine and everything is working 👍️

pford (commented):
Very bizarre that execute() was returning OnMessageTask* which was always nullptr!  Probably a refactoring mistake that was missed.

I wonder if ThreadManager (and its directory) should have a less generic name, since it is specifically for handling OnMessageTask events.  And therefore should it be co-located with OnMessageTask?

confluence:
> Very bizarre that execute() was returning OnMessageTask* which was always nullptr! Probably a refactoring mistake that was missed.

I think so -- it was a holdover from TBB. I looked all the way back in the history, and we only returned `nullptr` even when we were using TBB. I guess it was an optional feature for scheduling a new task when a task is completed.

> I wonder if ThreadManager (and its directory) should have a less generic name, since it is specifically for handling OnMessageTask events. And therefore should it be co-located with OnMessageTask?

Perhaps it should be called `TaskManager`? I'd have no objection to moving the files into `Session`, since currently we use this only with `OnMessageTask`. I think we should still keep the `Task` parent class to maintain a clean separation between the generic functionality of the manager and the specifics of the `OnMessageTask` implementation.

pford:
> Perhaps it should be called `TaskManager`? I'd have no objection to moving the files into `Session`, since currently we use this only with `OnMessageTask`. I think we should still keep the `Task` parent class to maintain a clean separation between the generic functionality of the manager and the specifics of the `OnMessageTask` implementation.

So... I searched the code for ThreadManager (should have done this first) and it is used in numerous places to apply thread limits--Contouring, Smoothing, Table, Main, and Histogram.  Maybe keep ThreadManager for basic thread handling and add a TaskManager for the Task queue handling which is derived from or uses ThreadManager? (or just leave it as is and not complicate things)

confluence:
> So... I searched the code for ThreadManager (should have done this first) and it is used in numerous places to apply thread limits--Contouring, Smoothing, Table, Main, and Histogram. Maybe keep ThreadManager for basic thread handling and add a TaskManager for the Task queue handling which is derived from or uses ThreadManager? (or just leave it as is and not complicate things)

This was the reason that I added the generic `Task` superclass (really just an interface with an `execute` function, which is the only thing that the manager needs to care about). This allows the manager not to depend on `OnMessageTask` at all, which allows other files that need access to the thread limit functions to import it without polluting their namespaces / creating circular imports.

Technically I think I mostly fixed that in the first pass by forward-defining `OnMessageTask` in the header and only including it from the cpp file, but there's really no reason for the manager to know about the `OnMessageTask` implementation.

But maybe the thread limiting functions mean that it should still be called `ThreadManager` and stay where it is -- it's a low-level component like the logger, which is also in its own subdirectory.

Subclassing the manager may be tricky because we're currently using it as a singleton (everything is static).

pford (approved):
Let's not change the ThreadManager name and location.

kswang1029 (approved):
Using a native Ubuntu 24.04, all e2e tests are running fine.

PR #1486, opened 2025-07-07
Handle invalid database objects correctly

### LINKED ISSUE #1485: setting preference will reset preferences.json
**Describe the bug**
This is a case to reproduce the issue. Not sure if there are others.

If we set WCS format to degrees, the preferences.json will be set to default so that the original custom preferences will be gone.

Image

**To Reproduce**
1. Go to Preferences, Render configuration, set colormap to jet
2. load an image to see it does what it supposed to do 
3. Check preferences.json to see the colormap setup
4. Go to Preferences, WCS and Image Overlay, and set WCS format as degree.
5. load an image to see it does what it supposed to do 
6. but now preferences.json has been reset so we do not see the colormap preference.
7. close CARTA and launch CARTA again
8. see prompt for telemetry. This is also an indication that the preferences.json has been reset.
 
 
**Expected behavior**
Being able to write preferences without altering existing ones.

 - OS [e.g. macOS Monterey]: Sequoia 
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev, v5
 - Frontend branch [e.g. dev, v3b2 release]: dev, v5

confluence:
**Description**

Fixes #1485.

The functions which save database objects to disk open the file for writing before performing validation. This means that if validation fails, the file is left blank. This PR moves the opening of the file to just before it is needed for writing, after the validation step. The invalid update is just ignored (which is what currently happens in the controller).

The actual invalid preferences which triggered this issue need separate fixes in these companion PRs:

1. In the [schema definition](https://github.com/CARTAvis/schemas/pull/10) (which is incorrect)
2. In the [frontend](https://github.com/CARTAvis/carta-frontend/pull/2563) (which needs a shim in the upgrade function to correct existing preferences which previous versions of the backend would have allowed to be saved).
3. In the [controller](https://github.com/CARTAvis/carta-controller/pull/239) (which needs the schema update)

This also needs to be merged into the 5.0 branch.

Still to be done:

- [X] The validator library documentation suggests that the default exception thrown by the validator includes a detailed description of the validation error(s), but it's actually just reported as a generic exception. This can be fixed with a custom validator (we already have one that we can reuse with a custom exception).
- [ ] ~The controller sends data in failure responses which contains the error status and a specific error message. The backend does not do this. For parity with the controller, it should send back the same data with the same message under the same circumstances.~ This is not critical, and I would like to adjust the controller's messages. I will make separate issues for this.
- [X] This PR should also include the schemas fix.
- [X] Updated prefs should be validated *before*, not *after* merge with existing prefs before write to file. Otherwise an existing invalid preference will block writes forever.
- [X] Assorted other small issues found and fixed.

**Checklist**

- [X] changelog updated

### DIFF
--- a/src/HttpServer/HttpServer.cc
+++ b/src/HttpServer/HttpServer.cc
@@ -261,14 +261,19 @@ bool HttpServer::WritePreferencesFile(nlohmann::json& obj) {
     auto preferences_path = _config_folder / "preferences.json";
 
     try {
-        fs::create_directories(preferences_path.parent_path().string());
-        std::ofstream file(preferences_path.string());
         // Ensure correct schema and version values are written
         obj["$schema"] = Json::Schema("preferences")["$id"];
         obj["version"] = 2;
+
+        // Validate the preferences
         Json::Validator("preferences").validate(obj);
+
+        // Write out the preferences to file
         auto json_string = obj.dump(4);
+        fs::create_directories(preferences_path.parent_path().string());
+        std::ofstream file(preferences_path.string());
         file << json_string;
+
         return true;
     } catch (json::type_error e) {
         spdlog::warn(e.what());
@@ -624,8 +629,6 @@ bool HttpServer::WriteObjectFile(const std::string& object_type, const std::stri
     auto object_path = _config_folder / (object_type + "s") / (object_name + ".json");
 
     try {
-        fs::create_directories(object_path.parent_path());
-        std::ofstream file(object_path.string());
         // Ensure correct schema value is written
         if (OBJECT_TYPES.count(object_type)) {
             obj["$schema"] = Json::Schema(object_type)["$id"];
@@ -634,10 +637,15 @@ bool HttpServer::WriteObjectFile(const std::string& object_type, const std::stri
             return false;
         }
 
+        // Validate the object
         Json::Validator(object_type).validate(obj);
 
+        // Write out the object to file
         auto json_string = obj.dump(4);
+        fs::create_directories(object_path.parent_path());
+        std::ofstream file(object_path.string());
         file << json_string;
+
         return true;
     } catch (json::type_error e) {
         spdlog::warn(e.what());
--- a/src/HttpServer/HttpServer.cc
+++ b/src/HttpServer/HttpServer.cc
@@ -223,12 +223,44 @@ void HttpServer::AddCorsHeaders(Res* res) {
     res->writeHeader("Access-Control-Max-Age", "3600");
 }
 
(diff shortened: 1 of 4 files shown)


loveluthien (approved):
Looks good. 😎

veggiesaurus (approved):

kswang1029:
@confluence if I manually add 
`"colormap": x`
in the preferences.json

once I launch carta I will be asked for telemetry permission and it seems my preferences.json is being reset. Could you confirm?

In addition, if we add an undefined key:value pair like
`"sponge":"bob",`
it will stay forever in the json file. Should we remove it plus a warning to user? @loveluthien

loveluthien:
Follow the previous comment, any key with a value that is not a number or a string will cause the preferences.json file to be reset.

confluence:
> @confluence if I manually add `"colormap": x` in the preferences.json
> 
> once I launch carta I will be asked for telemetry permission and it seems my preferences.json is being reset. Could you confirm?

I will check -- I may have missed incorrect handling of malformed JSON.

> In addition, if we add an undefined key:value pair like `"sponge":"bob",` it will stay forever in the json file. Should we remove it plus a warning to user? @loveluthien

We don't currently do anything about unrecognised properties -- all the schemas allow extra properties. It's up to you whether the frontend should handle them -- I agree that it may be helpful if the user makes a key typo, but perhaps this should wait for future refactoring.

src/HttpServer/HttpServer.cc
```diff
@@ -255,6 +255,19 @@ bool HttpServer::ValidateObject(const std::string& object_type, nlohmann::json&
...
+    auto preferences_path = _config_folder / "preferences.json";
+    auto backup_path = _config_folder / "preferences.json.bak";
+    std::error_code error_code;
+
+    fs::copy_file(preferences_path, backup_path, fs::copy_options::update_existing, error_code);
+    if (error_code) {
+        spdlog::warn("Could not back up preferences: {}", error_code.message());
+    } else {
+        spdlog::info("Backed up preferences to {}.", backup_path.string());
+    }
```
kswang1029, line 268: would be good to add another log like

"Resetting factory defaults."

kswang1029 (approved):
this looks good after a final tweak in the log message when a malformed preference is detected.

src/HttpServer/HttpServer.cc
confluence: I will add an explanatory message to the handler for setting preferences (which is where this is triggered -- the existing preferences will only be reset when the frontend tries to update them).

Issue #1427, opened 2025-04-01
Intermittent failure of TileEncodingTest.RoundTrip test caused by std::uniform_real_distribution bug


confluence:
`TileEncodingTest.RoundTrip` sometimes fails (see e.g. [this unrelated PR](https://github.com/CARTAvis/carta-backend/pull/1424#issuecomment-2769363158)). This failure is hard to reproduce, because the test generates random parameters. This is the error output:
```
  [ RUN      ] TileEncodingTest.RoundTrip
  /Users/acdc/actions-runner/_work/carta-backend/carta-backend/test/TestTileEncoding.cc:48: Failure
  Expected equality of these values:
    tile.x
      Which is: 4095
    x
      Which is: 64
  
  [  FAILED  ] TileEncodingTest.RoundTrip (2 ms)
```

I'm pretty sure that this is the cause:

1. [The test](https://github.com/CARTAvis/carta-backend/blob/576e5d9b54fef084cc162767ca99d8ae02694516/test/TestTileEncoding.cc#L34) creates a random integer distribution for generating integers in the interval [0, 12] and a random real distribution for generating floats in the interval [0, 1)
2. Inside a loop, the test generates a random layer using the integer distribution, calculates the layer width (`2**layer`), and generates a random x and y within the layer width by multiplying random floats generated with the real distribution by the layer width and taking the floor. This avoids redefining an integer distribution based on the width inside the loop.
3. _However_, [per the documentation](https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution), a bug in most implementations occasionally causes `std::uniform_real_distribution` to return the upper bound. So for example if the layer width is 64, x can sometimes be 64.
5. If these incorrect parameters are passed to [Tile::Encode](https://github.com/CARTAvis/carta-backend/blob/576e5d9b54fef084cc162767ca99d8ae02694516/src/DataStream/Tile.h#L21), it returns -1. If -1 is passed to [Tile::Decode](https://github.com/CARTAvis/carta-backend/blob/576e5d9b54fef084cc162767ca99d8ae02694516/src/DataStream/Tile.h#L30), the resulting layer, x, and y are 27, 4095, and 4095, respectively, which explains the error output.

I'm not sure that we're gaining a lot of value by generating all these random test cases -- perhaps we should instead be checking hardcoded edge cases for each permitted layer.

Michaelazzz:
[...]

PR #1488, opened 2025-07-15
Fix and expand Tile tests, add documentation, remove unnecessary MipToLayer function

Michaelazzz:
**Description**
Issue #1427  was caused by std::uniform_real_distribution bug.
It is an issue in the tests, not the code.
A fix was implemented by introducing more targeted edge cases and reducing the size of the round trip test.
Documentation for the Tile class was added as well.
There are no companion PR's.
The issues is intermittent and hard to reproduce but the new tests should account account for the edge case that produced the failure.

**Checklist**

- [x] e2e test passing / corresponding fix added
- [x] ICD test passing / new ICD test created


src/DataStream/Tile.h
```diff
@@ -12,28 +12,60 @@
...
+     */
     static int32_t Encode(int32_t x, int32_t y, int32_t layer) {
         int32_t layer_width = 1 << layer;
         if (x < 0 || y < 0 || layer < 0 || layer > 12 || x >= layer_width || y >= layer_width) {
             return -1;
         }
 
-        return ((layer << 24) | (y << 12) | x);
+        // Layout: [layer:8 bits][y:12 bits][x:12 bits]
+        return ((layer & 0xFF) << 24) | ((y & 0xFFF) << 12) | (x & 0xFFF);
```
confluence: As far as I can see, the bounds check above ensures that all these ands are no-ops.

src/DataStream/Tile.h
```diff
@@ -12,28 +12,60 @@
...
+    int32_t x;     /**< The X coordinate of the tile. */
+    int32_t y;     /**< The Y coordinate of the tile. */
+    int32_t layer; /**< The resolution layer of the tile. */
 
+    /**
+     * @brief Encodes tile coordinates and layer into a 32-bit integer.
+     *
+     * @param x The X coordinate (must be within bounds of the layer).
+     * @param y The Y coordinate (must be within bounds of the layer).
+     * @param layer The layer index (must be between 0 and 12).
```
confluence: It should perhaps be clearer that `within bounds of the layer` is referring to the layer width, and how that is calculated from the layer index, and which bounds are inclusive and exclusive.

src/DataStream/Tile.h
```diff
@@ -42,6 +74,16 @@ struct Tile {
...
 
+    /**
+     * @brief Converts a mipmap size to the corresponding layer index.
+     *
+     * @param mip The mipmap size to convert.
+     * @param image_width The width of the full-resolution image.
+     * @param image_height The height of the full-resolution image.
+     * @param tile_width The width of a tile in pixels.
+     * @param tile_height The height of a tile in pixels.
+     * @return The corresponding layer index, or -1 if invalid.
```
confluence: This is not correct: some invalid mip values will return a different negative number, and some will trigger an error. E.g. 0 will cause a division by zero, and a negative value will trigger some kind of mathematical error (because there's no valid log2 of a negative number). This is probably why the test is failing.

There's also no validation that the mip value is a power of 2.

`LayerToMip` doesn't have any consistent error handling either -- layer values that exceed the maximum layer produce fractional values (which I guess are coerced to zero), and negative values produce large integers which exceed the maximum mip.

We should perhaps add this error checking, but we'll have to make sure it doesn't break the calling code.

Each of the functions in this file is only used in one place (except `Encode`, which is currently unused).

test/TestTileEncoding.cc
```diff
@@ -37,17 +45,199 @@ TEST(TileEncodingTest, RoundTrip) {
...
+
+TEST(TileEncodingTest, MipToLayerInvalidInput) {
+    int img_width = 256;
+    int img_height = 256;
+    int tile_width = 256;
+    int tile_height = 256;
+
+    EXPECT_EQ(Tile::MipToLayer(0, img_width, img_height, tile_width, tile_height), -1);
+    EXPECT_EQ(Tile::MipToLayer(-5, img_width, img_height, tile_width, tile_height), -1);
+    EXPECT_EQ(Tile::MipToLayer(1000, img_width, img_height, tile_width, tile_height), -1);
```
confluence: None of these return -1, for three different reasons!

src/DataStream/Tile.h
Michaelazzz: Aplogies, this change was from when I was exploring a bit-masking approach compared to the current bit-shifting approach. I've reverted back to the original code.

src/DataStream/Tile.h
Michaelazzz: You're right there isn't any explicit validation or error checking in the functions using LayerToMip and MipToLayer.

Michaelazzz: Should I include some validation and error checking in this PR?

test/TestTileEncoding.cc
Michaelazzz: Since they all return different values for different reasons, I put -1 because we don't necessary want these tests to pass. There isn't an expected behaviour for all these cases because of the lack of validation and error checking in the MipTo Layer function. The current expected behaviour is that they return invalid values.

confluence: Right, but then the test should check for that (an exception, or that the value is negative). We should not have failing tests.

src/DataStream/Tile.h
confluence: Maybe that's a good idea. I don't think adding appropriate bounds checks and returning -1 will break anything, since currently invalid parameters would either cause a negative value to be returned or throw an exception.

We may not want to require the mip parameter to be a power of 2, despite its name, if this function is also used for other values (it does return a valid result, effectively rounding to an appropriate mip).

Adding error handling to the calling code may be overly complicated, because we would have to decide on an appropriate response in what may be many layers of function calls. It would be good, however, to trace all the parameters passed to these functions back to their source (which may be the frontend) and confirm whether they are (currently) always valid. For example, `MipToLayer` is currently used by the vector field calculation code; the origin of the mip value is the smoothing factor set with the `SetVectorOverlayParameters` message -- how is that value calculated in the frontend? Is it always a power of 2, or are intermediate values also possible? 

This should be done for each of the parameters -- since each function is only actually called in one place, it should be doable.

test/TestTileEncoding.cc
Michaelazzz: I will keep that in mind for future tests.

src/DataStream/Tile.h
```diff
@@ -12,41 +12,104 @@
...
+     *
+     * @param layer The layer index to convert.
+     * @param image_width The width of the full-resolution image.
+     * @param image_height The height of the full-resolution image.
+     * @param tile_width The width of a tile in pixels.
+     * @param tile_height The height of a tile in pixels.
+     * @param error A reference to a boolean that will be set to true if an error occurs.
+     * @return The mip size corresponding to the given layer.
+     */
+    static int32_t LayerToMip(
```
confluence: I think we had our wires crossed about this change: I suggest returning -1 here for invalid inputs (no in-out error parameter at this level). The in-out parameter should be passed from the `Session::OnAddRequiredTiles` to `Frame::FillRasterTileData` and from there to `Frame::GetRasterTileData`, which should set it if this function returns -1. The session function should then send an error message only if the error is set.

confluence (changes requested):

src/Frame/Frame.cc
```diff
@@ -504,7 +504,8 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
     std::shared_ptr<std::vector<float>> tile_data_ptr;
     int tile_width;
     int tile_height;
-    if (GetRasterTileData(z, tile_data_ptr, tile, tile_width, tile_height)) {
+    bool tile_error(false);
```
confluence: This intermediate variable is unnecessary; you can pass in `error` directly.

src/Frame/Frame.cc
```diff
@@ -577,15 +578,16 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
 }
 
-bool Frame::GetRasterTileData(int z, std::shared_ptr<std::vector<float>>& tile_data_ptr, const Tile& tile, int& width, int& height) {
-    bool error = false;
+bool Frame::GetRasterTileData(int z, std::shared_ptr<std::vector<float>>& tile_data_ptr, const Tile& tile, int& width, int& height, bool& error) {
     int mip = Tile::LayerToMip(tile.layer, _dims.width, _dims.height, TILE_SIZE, TILE_SIZE);
     if (mip == -1) {
         spdlog::error("Invalid tile layer {} for image size {}x{}.", tile.layer, _dims.width, _dims.height);
-        return false;
+        error = true;
```
confluence, line 588: You should set the error *and* return `false` here.

src/Session/Session.cc
```diff
@@ -722,9 +722,14 @@ void Session::OnAddRequiredTiles(const CARTA::AddRequiredTiles& message, int z,
...
-
+                bool tile_error(false);
                 if (_frames.count(file_id) && _frames.at(file_id)->FillRasterTileData(raster_tile_data, tile, requested_z, stokes,
-                                                  compression_type, compression_quality, is_current_z)) {
+                                                  compression_type, compression_quality, is_current_z, tile_error)) {
+                    if (tile_error) {
+                        SendLogEvent(
+                            fmt::format("Invalid mip calculation: channel={}, x={}, y={}, layer={}", requested_z, tile.x, tile.y, tile.layer),
+                            {"invalid mip calculation"}, CARTA::ErrorSeverity::WARNING);
+                    }
```
confluence: Because `FillRasterTileData` should also return `false` if the mip is invalid, you should move this into the `else` block.

The current `stale tile request` message should only be logged if there is no error.

You'll have to define the error variable outside the conditional.

confluence (changes requested):

src/Session/Session.cc
```diff
@@ -722,15 +722,21 @@ void Session::OnAddRequiredTiles(const CARTA::AddRequiredTiles& message, int z,
...
                     // Only use deflate on outgoing message if the raster image compression type is NONE
                     SendFileEvent(
                         file_id, CARTA::EventType::RASTER_TILE_DATA, 0, raster_tile_data, compression_type == CARTA::CompressionType::NONE);
                 } else {
-                    spdlog::warn(
-                        "Discarding stale tile request for channel={}, x={}, y={}, layer={}", requested_z, tile.x, tile.y, tile.layer);
+                    if (tile_error) {
+                        SendLogEvent(fmt::format("Invalid mip calculation: channel={}, x={}, y={}, layer={}", requested_z, tile.x, tile.y,
+                                         tile.layer),
+                            {"invalid mip calculation"}, CARTA::ErrorSeverity::WARNING);
```
confluence: Minor nitpick: this should be something more general, like `animation`. We haven't been consistent about this elsewhere.

confluence (approved):

kswang1029 (changes requested):
@Michaelazzz @confluence  Vector field rendering is broken in this PR, resulting six e2e test failures. Please check. 
(the schema submodule seems requiring an update, seeing mismatched commit)

confluence:
@kswang1029 is there any failing test output that we could look at? This PR does add some explicit error case handling which may change the response to certain kinds of input.

kswang1029:
The backend throws

> [2025-10-14 01:22:16.472Z] [CARTA] [error] Invalid tile layer -1 for image size 512x512.
[2025-10-14 01:22:16.472Z] [CARTA] [error] Failed to get tiles for vector field calculation.
[2025-10-14 01:22:16.472Z] [CARTA] [debug] Session 894761960: Error processing vector field image

without data returned to frontend for rendering. (will share the test image via slack)

src/DataStream/VectorField.h
```diff
@@ -136,7 +136,18 @@ class VectorField {
...
+/**
+ * @brief Computes the tile grid for a given mip level of an image and populates the provided tile vector.
+ *
+ * @param image_width The width of the image in pixels.
+ * @param image_height The height of the image in pixels.
+ * @param mip The mip level to generate tiles for.
+ * @param tiles Reference to a vector of `Tile` objects that will be filled with the computed tile data.
+ *              The vector is resized to fit all required tiles.
+ *
+ * @return `true` if the tiles were successfully computed and the layer was valid, `false` otherwise.
```
confluence: This comment should be updated -- the reference to the layer should be removed.

confluence (changes requested):
I was going to suggest correcting the `GetTiles` docstring in `VectorField`, and renaming `mip` to `smoothing_factor` throughout the `VectorField` code -- but I think that all of this is out of scope for this PR. The only reason that you added a docstring to this function is that it previously called the removed function -- it is now unrelated to the rest of the PR.

Instead, please remove the `GetTiles` docstrings for now. The `VectorField` code is a good candidate for refactoring, documentation and unit tests -- it's small and self-contained. But this should happen in a separate PR (I will create an issue with a list of tasks).

src/DataStream/VectorField.cc
```diff
@@ -189,11 +189,17 @@ void VectorField::FillTileData(CARTA::TileData* tile, int32_t x, int32_t y, int3
     }
 }
 
-void GetTiles(int image_width, int image_height, int mip, std::vector<Tile>& tiles) {
+/**
```
confluence: The docstring for this function should be removed, and the void -> bool change should be reverted. The only change in these files should be the removal of the `MipToLayer` call.

confluence (changes requested):

src/DataStream/VectorField.h
```diff
@@ -136,6 +136,15 @@ class VectorField {
...
 
+/**
+ * @brief Computes the tile grid for a given mip level of an image and populates the provided tile vector.
+ *
+ * @param image_width The width of the image in pixels.
+ * @param image_height The height of the image in pixels.
+ * @param mip The mip level to generate tiles for.
+ * @param tiles Reference to a vector of `Tile` objects that will be filled with the computed tile data.
+ *              The vector is resized to fit all required tiles.
+ */
```
confluence: This entire docstring needs to be removed.

confluence (changes requested):
This is almost ready; there's just one unnecessary docstring (in the vector field, which is being refactored separately).

confluence (approved):

confluence:
@kswang1029 could you please re-test?

kswang1029 (approved):
looks fine.

PR #1489, opened 2025-07-15
Remove test file generation from tests

Michaelazzz:
**Description**

This PR removes test file generation from the tests.
Addressed in issue #1496 
Instead the tests now use existing tests file that do need to be generated at the test runtime.
There are no accompanying PR's.
Hopefully this change speed up the time it takes for the tests to run.

**Checklist**

- [X] changelog updated
- [ ] e2e test passing
- [ ] ICD test passing / corresponding fix added


Michaelazzz:
@confluence do we also want to include the integration of the [git-lfs repo](https://github.com/CARTAvis/carta-backend-test-data) with this PR?

confluence:
> @confluence do we also want to include the integration of the [git-lfs repo](https://github.com/CARTAvis/carta-backend-test-data) with this PR?

@Michaelazzz yes, this PR should include the addition of that as a submodule.

carta-backend-test-data
confluence, line 1: This submodule should replace the `test/data` directory (the external name of the data repository is fine; only the reference in this repository should be moved and renamed). You should be able to use `git mv` for this (you don't have to remove it and re-add the submodule), but you probably need to `git rm` the old data directory first to avoid confusing git.

About the submodule implementation: instead of one script to regenerate all the files at once, I suggest putting scripts (or makefiles, which are more idiomatically suited to this kind of task) in each subdirectory for regenerating the files in that directory (e.g. for a makefile add a target per file and a default target that depends on all of them). This will make it easier to extend the repository with future collections of files that are generated in different ways with different tools.

It would be good to reorganise the existing files by topic rather than file type (but that can wait until you have a better idea of how they're used). If you use a makefile, then for files which aren't generated (but share a directory with generated files) you could add dummy targets that just print a message (for now, until they're moved somewhere -- in future, files like this should go in separate directories with documentation explaining where they came from).

.gitmodules
```diff
@@ -45,5 +45,5 @@
 	url = https://github.com/uNetworking/uSockets.git
 	shallow = true
 [submodule "carta-backend-test-data"]
-	path = carta-backend-test-data
+	path = test/carta-backend-test-data
```
confluence: The submodule directory should just be called `data`, like before. It doesn't make sense to use the full repository name here.

You can use `git mv` again to rename it.

confluence (changes requested):

Michaelazzz:
List of things that need to be done for this PR:
- [ ] Make a Python script to generate test data from the static test files
- [x] Remove function involved with the file generation in CommonTestUtilities
- [ ] Catch vertices data from existing tests
- [x] Make concentric squiggle test images for testing contours
- [x] remove GetGeneratedFilePath function from TestImageFitting

Note: `Make a Python script to generate test data from the static test files` and `Catch vertices data from existing tests` list items will be moved to another PR.

confluence:
@kswang1029 FYI we decided to limit this PR to the initial removal of the file generation -- there will be outstanding code that can be simplified and refactored, but it makes more sense to leave this for subsequent PRs so that this initial work can be merged in and future work can potentially be split up.

kswang1029 (approved):
looks fine.

test/CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ set(TEST_SOURCES
         TestProgramSettings.cc
         TestPvGenerator.cc
         TestRegion.cc
+        TestRegionImportExport.cc
```
confluence: Not sure where this came from; it should be removed -- it's a duplicate of line 61. Bad merge?

test/TestBlockSmooth.cc
```diff
@@ -166,25 +166,25 @@ TEST_F(BlockSmoothingTest, TestControl) {
...
+//                 auto max_error = nanmax(abs_diff);
+//                 EXPECT_EQ(MatchingNANs(smoothed_scalar, smoothed_sse), true);
+//                 if (std::isfinite(sum_error)) {
+//                     EXPECT_LE(sum_error, MAX_SUM_ERROR);
+//                     EXPECT_LE(max_error, MAX_ABS_ERROR);
+//                 }
+//             }
+//         }
+//     }
+// }
```
confluence: Why is this commented out?

test/TestContour.cc
```diff
@@ -11,27 +11,23 @@
...
-        int width, int height, std::string image_opts, const CARTA::FileType& file_type, const CARTA::SmoothingMode& smoothing_mode) {
-        std::string image_shape = std::to_string(width) + " " + std::to_string(height);
+    void GenerateContour(std::string filename, const CARTA::FileType& file_type, const CARTA::SmoothingMode& smoothing_mode) {
         std::string file_path;
         if (file_type == CARTA::FileType::HDF5) {
-            file_path = ImageGenerator::GeneratedHdf5ImagePath(image_shape, image_opts);
+            file_path = (TestRoot() / "data" / "images" / "hdf5" / filename);
         } else {
-            file_path = ImageGenerator::GeneratedFitsImagePath(image_shape, image_opts);
+            file_path = (TestRoot() / "data" / "images" / "fits" / filename);
```
confluence: We have `FitsImagePath`, `Hdf5ImagePath`, etc., for this in the common test utilities. However, those should also be removed. I suggest:

* Removing FileFinder entirely, and replacing the 5 functions in it with 5 standalone functions (like `TestRoot` and `UserDirectory`) that just return `fs::path`s to the 5 data subdirectories: `FitsImages`, `Hdf5Images`, `FitsTables`, etc.. Don't convert them to strings; leave them as path objects.
* In this specific function, pass in an `fs::path file_path` parameter instead of the filename and file type. Construct the path directly in the calling functions using the new directory functions, e.g. `GenerateContour(FitsImages() / "500_500_image_opts.fits", CARTA::SmoothingMode::NoSmoothing);`
* Replace all your `TestRoot() / "data" / "images" / whatever`s in the other test files with the new functions
* Replace all other calls to the FileFinder functions, e.g. `std::string image_path0 = FileFinder::FitsImagePath("noise_10px_10px.fits");` -> `std::string image_path0 = FitsImages() / "noise_10px_10px.fits";` ([`fs::path`](https://github.com/CARTAvis/carta-backend/blob/dev/src/Util/FileSystem.h) = [`std::filesystem::path`](https://en.cppreference.com/w/cpp/filesystem/path.html); they're implicitly convertible to strings, so I think that this should just work, and you shouldn't need to perform an explicit conversion -- it certainly seems to be working below in some of the other test files, where you're passing `fs::path`s directly into functions that take string parameters).

While in general we should leave further refactoring for later, this is a simple change which would make the code much more readable.

test/TestCursorSpatialProfiles.cc
```diff
@@ -100,7 +98,7 @@ class CursorSpatialProfileTest : public ::testing::Test, public ImageGenerator {
 };
 
 TEST_F(CursorSpatialProfileTest, SmallFitsProfile) {
-    auto path_string = GeneratedFitsImagePath("10 10", IMAGE_OPTS);
+    auto path_string = (TestRoot() / "data" / "images" / "fits" / "noise_10px_10px.fits");
```
confluence: 1. See comment above re: streamlining this with common functions that return the base subdirectory paths.
2. This is not a string; it's an `fs::path`. So the variable should be renamed to `path`. (This code works even though the type has changed because paths can be converted to strings implicitly.)

test/TestFitsTable.cc
```diff
@@ -12,26 +12,25 @@ using namespace carta;
 class FitsTableTest : public ::testing::Test, public FileFinder {};
 
 TEST_F(FitsTableTest, ParseIvoaExampleHeaderOnly) {
-    std::cout << FitsTablePath("test.fits") << std::endl;
-    Table table(FitsTablePath("ivoa_example.fits"), true);
+    Table table(TestRoot() / "data" / "tables" / "fits" / "ivoa_example.fits", true);
```
confluence: These table paths should also be simplified.

test/TestProgramSettings.cc
```diff
@@ -317,7 +317,7 @@ TEST_F(ProgramSettingsTest, ExpectedValuesLongJSON) {
 }
 
 TEST_F(ProgramSettingsTest, ValidateJSONFromFileWithGoodFields) {
-    const std::string input = DataPath("settings-good-fields.json");
+    const std::string input = TestRoot() / "data" / "settings" / "settings-good-fields.json";
```
confluence: I just noticed that you removed `DataPath`. It was only used to access the settings directory, so instead of restoring it I suggest adding a `SettingsPath` (in the same form as the other 5 path functions) to return the settings directory instead. So e.g. this would become `SettingsPath() / "settings-good-fields.json"`.

Technically, each of the table paths and the settings path is only used in one test file, so they could just be defined in those test files. But I think it's fine to leave them in the common file for now.

confluence (changes requested):
This is generally fine, but I would like to clean up the path code before merging.

test/CMakeLists.txt
Michaelazzz: Probably

test/TestBlockSmooth.cc
Michaelazzz: I'm not sure. I think I was debugging something.

test/CommonTestUtilities.cc
```diff
@@ -31,59 +31,28 @@ fs::path UserDirectory() {
...
-std::string FileFinder::FitsTablePath(const std::string& filename) {
-    return (TestRoot() / "data" / "tables" / "fits" / filename).string();
+fs::path Hdf5Images() {
+    return TestRoot() / "data" / "images" / "hdf5";
 }
 
-std::string FileFinder::XmlTablePath(const std::string& filename) {
-    return (TestRoot() / "data" / "tables" / "xml" / filename).string();
+fs::path MixImages() {
+    return TestRoot() / "data" / "images" / "mix";
```
confluence: You can use `ImagesPath` in here.

test/CommonTestUtilities.cc
```diff
@@ -31,59 +31,28 @@ fs::path UserDirectory() {
...
 
-std::string FileFinder::FitsTablePath(const std::string& filename) {
-    return (TestRoot() / "data" / "tables" / "fits" / filename).string();
+fs::path Hdf5Images() {
+    return TestRoot() / "data" / "images" / "hdf5";
 }
 
-std::string FileFinder::XmlTablePath(const std::string& filename) {
-    return (TestRoot() / "data" / "tables" / "xml" / filename).string();
+fs::path MixImages() {
```
confluence: This should probably be called `MixedImages`.

test/TestContour.cc
```diff
@@ -11,27 +11,23 @@
...
-        std::string image_shape = std::to_string(width) + " " + std::to_string(height);
+    void GenerateContour(std::string filename, const CARTA::FileType& file_type, const CARTA::SmoothingMode& smoothing_mode) {
         std::string file_path;
         if (file_type == CARTA::FileType::HDF5) {
-            file_path = ImageGenerator::GeneratedHdf5ImagePath(image_shape, image_opts);
+            file_path = (Hdf5Images() / filename);
         } else {
-            file_path = ImageGenerator::GeneratedFitsImagePath(image_shape, image_opts);
+            file_path = (FitsImages() / filename);
         }
```
confluence: This conditional should be removed entirely: you should construct the paths outside this function, and pass in the constructed path, not the filename and type.

test/TestContour.cc
```diff
@@ -127,43 +123,43 @@ class ContourTest : public ::testing::Test {
 };
 
 TEST_F(ContourTest, NoSmoothingFitsFile) {
-    GenerateContour(500, 500, IMAGE_OPTS, CARTA::FileType::FITS, CARTA::SmoothingMode::NoSmoothing);
+    GenerateContour("500_500_image_opts.fits", CARTA::FileType::FITS, CARTA::SmoothingMode::NoSmoothing);
```
confluence: In all these tests you should construct the path directly and pass it into the function. E.g.
```
GenerateContour(FitsImages() / "500_500_image_opts.fits", CARTA::SmoothingMode::NoSmoothing);
```

test/TestContour.cc
confluence: Lower down, where you need to know the type to create the reader, you can use [`parent_path`](https://en.cppreference.com/w/cpp/filesystem/path/parent_path.html) on the path and check that instead.

(Make sure that you make the function parameter a path object, not a string, so that you can do this.)

test/TestExprImage.cc
```diff
@@ -18,9 +18,9 @@ class ImageExprTest : public ::testing::Test {
     void GenerateImageExprTimesTwo(const std::string& file_name, const std::string& hdu, CARTA::FileType file_type, bool invalid = false) {
         std::string file_path;
         if (file_type == CARTA::FileType::FITS) {
-            file_path = FileFinder::FitsImagePath(file_name);
+            file_path = FitsImages() / file_name;
         } else if (file_type == CARTA::FileType::HDF5) {
-            file_path = FileFinder::Hdf5ImagePath(file_name);
+            file_path = Hdf5Images() / file_name;
```
confluence: As above -- you should pass in the path and construct it outside.

test/TestExprImage.cc
```diff
@@ -85,9 +85,9 @@ class ImageExprTest : public ::testing::Test {
     void SaveImageExpr(const std::string& file_name, const std::string& hdu, CARTA::FileType file_type) {
         std::string file_path;
         if (file_type == CARTA::FileType::FITS) {
-            file_path = FileFinder::FitsImagePath(file_name);
+            file_path = FitsImages() / file_name;
         } else if (file_type == CARTA::FileType::HDF5) {
-            file_path = FileFinder::Hdf5ImagePath(file_name);
+            file_path = Hdf5Images() / file_name;
```
confluence: As above -- you should pass in the path and construct it outside.

test/TestImageFitting.cc
```diff
@@ -56,8 +56,7 @@ class ImageFittingTest : public ::testing::Test {
         _fov_info.set_rotation(rotation);
     }
 
-    void FitImage(std::vector<float> gaussian_model, std::string failed_message = "") {
-        std::string file_path = GetGeneratedFilePath(gaussian_model);
+    void FitImage(std::string file_path, std::string failed_message = "") {
```
confluence: For consistency, the path parameter should still be a path object here (there's no need to convert it to a string at this stage).

test/TestImageFitting.cc
```diff
@@ -85,8 +84,8 @@ class ImageFittingTest : public ::testing::Test {
     }
 
     void FitImageWithFov(std::vector<float> gaussian_model, int region_id, std::string failed_message = "") {
-        std::string file_path = GetGeneratedFilePath(gaussian_model);
-        std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(file_path));
+        auto path = FitsImages() / "128_128_gaussian_model_one_component.fits";
```
confluence: This path should be constructed outside, and passed in (as in the function above).

test/TestImageFitting.cc
```diff
@@ -206,45 +193,49 @@ class ImageFittingTest : public ::testing::Test {
...
+    auto path = FitsImages() / "128_128_gaussian_model_one_component.fits";
     std::vector<float> gaussian_model = {1, 64, 64, 20, 20, 10, 135};
     std::vector<bool> fixed_params = {false, false, false, false, false, false, false};
     SetInitialValues(gaussian_model);
     SetFixedParams(fixed_params);
-    FitImage(gaussian_model);
+    FitImage(path);
 }
 
 TEST_F(ImageFittingTest, FittingWithFov) {
```
confluence, line 240: These functions should all pass the path into FitImageWithFov, not the model.

test/TestMoment.cc
```diff
@@ -128,7 +128,7 @@ class MomentTest : public ::testing::Test, public FileFinder {
 };
 
 TEST_F(MomentTest, CheckConsistency) {
-    std::string file_path = FitsImagePath("M17_SWex_unittest.fits");
+    std::string file_path = FitsImages() / "M17_SWex_unittest.fits";
```
confluence: These should all be `auto` (path objects), not strings.

test/TestProgramSettings.cc
```diff
@@ -181,34 +181,32 @@ TEST_F(ProgramSettingsTest, OverrideDeprecatedRoot) {
 TEST_F(ProgramSettingsTest, OverrideDeprecatedBase) {
     auto settings = SettingsFromVector({"carta_backend", "--base", "/tmp2", "/tmp"});
     EXPECT_EQ(settings.starting_folder, "/tmp");
-    auto image_dir = TestRoot() / "data/images";
-    settings = SettingsFromVector({"carta_backend", "--base", "/tmp2", image_dir.string()});
-    EXPECT_EQ(settings.starting_folder, image_dir.string());
+    settings = SettingsFromVector({"carta_backend", "--base", "/tmp2", ImagesPath().string()});
```
confluence: You shouldn't need to use `.string()` on path objects explicitly.

test/TestProgramSettings.cc
```diff
@@ -181,34 +181,32 @@ TEST_F(ProgramSettingsTest, OverrideDeprecatedRoot) {
...
+    EXPECT_EQ(settings.starting_folder, ImagesPath());
 }
 
 TEST_F(ProgramSettingsTest, StartingFolderFromPositional) {
-    auto image_dir = TestRoot() / "data/images";
-    auto settings = SettingsFromVector({"carta_backend", image_dir.string()});
-    EXPECT_EQ(settings.starting_folder, image_dir.string());
+    auto settings = SettingsFromVector({"carta_backend", ImagesPath().string()});
+    EXPECT_EQ(settings.starting_folder, ImagesPath().string());
```
confluence: `.string()` shouldn't be required here either.

test/TestProgramSettings.cc
```diff
@@ -181,34 +181,32 @@ TEST_F(ProgramSettingsTest, OverrideDeprecatedRoot) {
...
-    auto fits_image_path = FitsImagePath("invalid.fits");
+    auto fits_image_path = FitsImages() / "invalid.fits";
     auto settings = SettingsFromVector({"carta_backend", fits_image_path});
     EXPECT_EQ(settings.starting_folder, default_settings.starting_folder);
     EXPECT_TRUE(settings.files.empty());
 }
 
 TEST_F(ProgramSettingsTest, FileImageFromPositional) {
-    auto fits_image_path = FitsImagePath("noise_10px_10px.fits");
+    auto fits_image_path = (FitsImages() / "noise_10px_10px.fits").string();
```
confluence: `.string()` shouldn't be required for any of these.

test/TestProgramSettings.cc
```diff
@@ -317,7 +315,7 @@ TEST_F(ProgramSettingsTest, ExpectedValuesLongJSON) {
 }
 
 TEST_F(ProgramSettingsTest, ValidateJSONFromFileWithGoodFields) {
-    const std::string input = DataPath("settings-good-fields.json");
+    const std::string input = SettingsPath() / "settings-good-fields.json";
```
confluence: It should also be possible to use `auto` for all of these.

test/TestProgramSettings.cc
```diff
@@ -392,15 +390,15 @@ TEST_F(ProgramSettingsTest, TestFileQueryStringEmptyFiles) {
 }
 
 TEST_F(ProgramSettingsTest, TestFileQueryStringSingleFile) {
-    auto image_root = TestRoot() / "data" / "images";
+    auto image_root = ImagesPath();
     std::vector<std::string> files;
     files.push_back(image_root / "fits" / "noise_3d.fits");
     auto url_string = carta::HttpServer::GetFileUrlString(files);
     EXPECT_EQ(url_string, fmt::format("file={}{}", SafeStringEscape(fmt::format("{}/fits/", image_root.string())), "noise_3d.fits"));
```
confluence: The way this test concatenates the paths is unnecessarily convoluted. I think that this should work:
```
auto file_path = FitsImages() / "noise_3d.fits";
std::vector<std::string> files;
files.push_back(file_path);
auto url_string = carta::HttpServer::GetFileUrlString(files);
EXPECT_EQ(url_string, fmt::format("file={}", SafeStringEscape(file_path));
```

test/TestProgramSettings.cc
```diff
@@ -392,15 +390,15 @@ TEST_F(ProgramSettingsTest, TestFileQueryStringEmptyFiles) {
...
 }
 
 TEST_F(ProgramSettingsTest, TestFileQueryStringTwoFilesSameFolder) {
-    auto image_root = TestRoot() / "data" / "images";
+    auto image_root = ImagesPath();
```
confluence: First part can be simplified to:
```
std::vector<std::string> files;
files.push_back(FitsImages() / "noise_3d.fits");
files.push_back(FitsImages() / "noise_4d.fits");
auto folder = SafeStringEscape(FitsImages());
```

The rest is fine.

test/TestProgramSettings.cc
```diff
@@ -411,7 +409,7 @@ TEST_F(ProgramSettingsTest, TestFileQueryStringTwoFilesSameFolder) {
 }
 
 TEST_F(ProgramSettingsTest, TestFileQueryStringTwoFilesDifferentFolder) {
-    auto image_root = TestRoot() / "data" / "images";
+    auto image_root = ImagesPath();
```
confluence: As above, but for the two different directories.

test/Util/TestUtilFile.cc
```diff
@@ -84,7 +84,7 @@ TEST(IsSubdirectoryTest, SubdirectoryRel) {
     EXPECT_TRUE(IsSubdirectory("./data/images", "./"));
     EXPECT_FALSE(IsSubdirectory("./", "./data/images"));
     EXPECT_TRUE(IsSubdirectory("./data/images", "./data"));
-    EXPECT_FALSE(IsSubdirectory("./data", "./data/images"));
+    EXPECT_FALSE(IsSubdirectory("./data", "./data/data"));
```
confluence: This should be reverted.

test/CommonTestUtilities.cc
```diff
@@ -31,59 +31,28 @@ fs::path UserDirectory() {
...
-
-std::string FileFinder::DataPath(const std::string& filename) {
-    return (TestRoot() / "data" / filename).string();
+fs::path ImagesPath() {
+    return TestRoot() / "data" / "images";
 }
 
-std::string FileFinder::FitsImagePath(const std::string& filename) {
-    return (TestRoot() / "data" / "images" / "fits" / filename).string();
+fs::path SettingsPath() {
```
confluence: This is only used in the settings test, and can be moved there.

confluence (changes requested):

test/CommonTestUtilities.cc
```diff
@@ -31,59 +31,24 @@ fs::path UserDirectory() {
...
-std::string FileFinder::FitsImagePath(const std::string& filename) {
-    return (TestRoot() / "data" / "images" / "fits" / filename).string();
+fs::path ImagesPath() {
+    return TestRoot() / "data" / "images";
 }
 
-std::string FileFinder::CasaImagePath(const std::string& filename) {
-    return (TestRoot() / "data" / "images" / "casa" / filename).string();
+fs::path FitsImages() {
+    return TestRoot() / "data" / "images" / "fits";
```
confluence: You should use `ImagesPath()` in here and in all the other image subpaths.

test/TestFitsImage.cc
```diff
@@ -72,41 +72,41 @@ TEST_F(FitsImageTest, CorrectShape3dImage) {
 }
 
 TEST_F(FitsImageTest, CorrectShapeDegenerate3dImages) {
-    auto path_string = GeneratedFitsImagePath("10 10 10 1");
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_4d.fits";
```
confluence: Wrong images used in this test -- should be `[noise_3d_degen.fits](https://github.com/CARTAvis/carta-backend-test-data/blob/main/images/fits/noise_3d_degen.fits)` and `[noise_3d_degen_casa.fits](https://github.com/CARTAvis/carta-backend-test-data/blob/main/images/fits/noise_3d_degen_casa.fits)`. The dimensions in the test should be reverted.

test/TestCursorSpatialProfiles.cc
```diff
@@ -100,10 +98,10 @@ class CursorSpatialProfileTest : public ::testing::Test, public ImageGenerator {
 };
 
 TEST_F(CursorSpatialProfileTest, SmallFitsProfile) {
-    auto path_string = GeneratedFitsImagePath("10 10", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_10px_10px.fits";
```
confluence: Wrong image used here -- the image in this test should have been generated with NaN options.

test/TestCursorSpatialProfiles.cc
```diff
@@ -141,10 +139,10 @@ TEST_F(CursorSpatialProfileTest, SmallFitsProfile) {
 }
 
 TEST_F(CursorSpatialProfileTest, SmallHdf5Profile) {
-    auto path_string = GeneratedHdf5ImagePath("10 10", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = Hdf5Images() / "10_10_1_10_row_column.hdf5";
```
confluence: Wrong image used here -- this should be a 2D 10x10 image (with NaN options).

test/TestCursorSpatialProfiles.cc
```diff
@@ -182,10 +180,10 @@ TEST_F(CursorSpatialProfileTest, SmallHdf5Profile) {
 }
 
 TEST_F(CursorSpatialProfileTest, LowResFitsProfile) {
-    auto path_string = GeneratedFitsImagePath("130 100", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_10px_10px.fits";
```
confluence: Incorrect image used here.

test/TestCursorSpatialProfiles.cc
```diff
@@ -326,10 +324,10 @@ TEST_F(CursorSpatialProfileTest, LowResHdf5ProfileNoMipAvailable) {
 }
 
 TEST_F(CursorSpatialProfileTest, FullResFitsStartEnd) {
-    auto path_string = GeneratedFitsImagePath("400 300", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_10px_10px.fits";
```
confluence: Wrong image used here, which has also revealed a bug in this test (and several other tests where the image is currently wrong). This test should be failing, because clearly the profiles generated from this image can't be what we expect -- but we aren't checking that we're getting the correct number of `SpatialProfileData` messages back in the vector! We're just iterating over the vector, which is empty (because we're asking for profile ranges that are outside the bounds of the image), so we aren't checking anything and the test passes by default.

Before the loop, we need to check that the size of the vector is 1. This change should be made in all of these tests.

test/TestCursorSpatialProfiles.cc
```diff
@@ -396,10 +394,10 @@ TEST_F(CursorSpatialProfileTest, FullResHdf5StartEnd) {
 }
 
 TEST_F(CursorSpatialProfileTest, LowResFitsStartEnd) {
-    auto path_string = GeneratedFitsImagePath("400 300", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_10px_10px.fits";
```
confluence: Wrong image used here.

test/TestCursorSpatialProfiles.cc
```diff
@@ -545,10 +543,10 @@ TEST_F(CursorSpatialProfileTest, Hdf5MultipleChunkFullResStartEnd) {
 }
 
 TEST_F(CursorSpatialProfileTest, FitsChannelChange) {
-    auto path_string = GeneratedFitsImagePath("10 10 2", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_3d.fits";
```
confluence: Wrong image used here.

test/TestCursorSpatialProfiles.cc
```diff
@@ -588,10 +586,10 @@ TEST_F(CursorSpatialProfileTest, FitsChannelChange) {
 }
 
 TEST_F(CursorSpatialProfileTest, FitsChannelStokesChange) {
-    auto path_string = GeneratedFitsImagePath("10 10 2 2", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "noise_10px_10px.fits";
```
confluence: Wrong image used here.

test/TestCursorSpatialProfiles.cc
```diff
@@ -680,10 +678,10 @@ TEST_F(CursorSpatialProfileTest, ContiguousHDF5ChannelChange) {
 }
 
 TEST_F(CursorSpatialProfileTest, ChunkedHDF5ChannelChange) {
-    auto path_string = GeneratedHdf5ImagePath("1000 1000 2", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = Hdf5Images() / "1000_1000_2_2_row_column.hdf5";
```
confluence: Wrong image used here.

test/TestFileInfo.cc
```diff
@@ -14,7 +14,7 @@
 
 using namespace carta;
 
-static const std::string SAMPLE_FILES_PATH = (TestRoot() / "data" / "images" / "mix").string();
+static const std::string SAMPLE_FILES_PATH = MixedImages();
```
confluence: There's still some old path concatenation code in here. This global variable doesn't need to exist -- it's used in three places, where you can just use `MixedImages()` directly.

In `CheckFileInfoLoader` and `CheckFileExtInfoLoader` you can use path operations instead of string concatenation to create the full file path, e.g.
```
auto fullname = MixedImages() / request_filename
```

We should reorganise these image subdirectories, but that's a problem for later (when we refactor this).

If you want to keep the global variable, make it a path instead.

test/TestHdf5Image.cc
```diff
@@ -22,29 +22,29 @@ class TestFrame : public Frame {
     FRIEND_TEST(Hdf5ImageTest, ExampleFriendTest);
 };
 
-class Hdf5ImageTest : public ::testing::Test, public ImageGenerator {};
+class Hdf5ImageTest : public ::testing::Test {};
 
 TEST_F(Hdf5ImageTest, BasicLoadingTest) {
-    auto path_string = GeneratedHdf5ImagePath("10 10");
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = Hdf5Images() / "10_10_row_column.hdf5";
```
confluence: The original images used in these tests don't have any NaNs (although it shouldn't make any difference).

test/TestCursorSpatialProfiles.cc
```diff
@@ -217,10 +215,10 @@ TEST_F(CursorSpatialProfileTest, LowResFitsProfile) {
 }
 
 TEST_F(CursorSpatialProfileTest, LowResHdf5ProfileExactMipAvailable) {
-    auto path_string = GeneratedHdf5ImagePath("130 100", IMAGE_OPTS);
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = Hdf5Images() / "130_100_row_column.hdf5";
```
confluence: These images were generated with the same options as the NaN images in `TestContour`, so they should use the same naming convention (e.g. `130_100_nans_row_column.hdf5`).

I don't think these options make sense in this test. Since the point of including the NaNs is to make sure that they are handled correctly by various profile algorithms, these tests should probably check various different NaN patterns (rows, columns, pixel, whole channel) and make sure that the selected profiles actually hit the NaN rows and columns. But that's going to require more refactoring, and parameterised tests, so it's outside the scope of this PR (but we should create a new issue for this).

test/TestHdf5Image.cc
```diff
@@ -73,8 +73,8 @@ TEST_F(Hdf5ImageTest, CorrectShape3dImage) {
 }
 
 TEST_F(Hdf5ImageTest, CorrectShapeDegenerate3dImages) {
-    auto path_string = GeneratedHdf5ImagePath("10 10 10 1");
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = Hdf5Images() / "10_10_10_1_row_column.hdf5";
```
confluence: Inconsistent naming between these and the corresponding fits files.

test/TestMoment.cc
```diff
@@ -128,25 +128,25 @@ class MomentTest : public ::testing::Test, public FileFinder {
...
-    std::string file_path = FitsImagePath("M17_SWex_unittest.fits");
+    auto file_path = FitsImages() / "M17_SWex_unittest.fits";
     std::shared_ptr<casacore::ImageInterface<float>> image;
     int moment_axis(2);
 
     if (OpenImage(image, file_path)) {
         GenerateMoments(image, moment_axis);
     } else {
-        spdlog::warn("Fail to open the file {}! Ignore the Moment test.", file_path);
+        spdlog::warn("Fail to open the file {}! Ignore the Moment test.", file_path.string());
```
confluence: I've just realised that this whole conditional should be removed (here and in the other test). If the image can't be opened, that's a bug in the test code, and the test should crash! This should just be:
```
OpenImage(image, file_path);
GenerateMoments(image, moment_axis);
```

test/TestProgramSettings.cc
```diff
@@ -181,61 +185,59 @@ TEST_F(ProgramSettingsTest, OverrideDeprecatedRoot) {
...
+    EXPECT_EQ(settings.starting_folder, ImagesPath());
 }
 
 TEST_F(ProgramSettingsTest, StartingFolderFromPositional) {
-    auto image_dir = TestRoot() / "data/images";
-    auto settings = SettingsFromVector({"carta_backend", image_dir.string()});
-    EXPECT_EQ(settings.starting_folder, image_dir.string());
+    auto file_path = ImagesPath();
+    auto settings = SettingsFromVector({"carta_backend", file_path});
+    EXPECT_EQ(settings.starting_folder, ImagesPath());
```
confluence: You should use `file_path` here.

test/TestProgramSettings.cc
```diff
@@ -181,61 +185,59 @@ TEST_F(ProgramSettingsTest, OverrideDeprecatedRoot) {
...
-    // substr to remove leading "/" from expected path
-    EXPECT_EQ(settings.files[0], fits_image_path.substr(1));
+    EXPECT_EQ(settings.files[0], fs::relative(fits_image_path, "/"));
 }
 
 TEST_F(ProgramSettingsTest, RelativeFileImageFromPositional) {
-    auto absolute_image_path = FitsImagePath("noise_10px_10px.fits");
+    auto absolute_image_path = FitsImages() / "noise_10px_10px.fits";
     fs::current_path(TestRoot());
     std::string relative_image_path = "data/images/fits/noise_10px_10px.fits";
```
confluence: It should be possible to use `fs::relative(absolute_image_path)` here to create `relative_image_path`.

test/TestProgramSettings.cc
```diff
@@ -392,32 +394,30 @@ TEST_F(ProgramSettingsTest, TestFileQueryStringEmptyFiles) {
...
-    auto image_root = TestRoot() / "data" / "images";
     std::vector<std::string> files;
-    files.push_back(image_root / "fits" / "noise_3d.fits");
-    files.push_back(image_root / "hdf5" / "noise_10px_10px.hdf5");
-    auto folder1 = SafeStringEscape(fmt::format("{}/fits/", image_root.string()));
-    auto folder2 = SafeStringEscape(fmt::format("{}/hdf5/", image_root.string()));
+    files.push_back(FitsImages() / "noise_3d.fits");
+    files.push_back(Hdf5Images() / "noise_10px_10px.hdf5");
+    auto fits_file = SafeStringEscape(fs::absolute(FitsImages() / "noise_3d.fits"));
+    auto hdf5_file = SafeStringEscape(fs::absolute(Hdf5Images() / "noise_10px_10px.hdf5"));
```
confluence, line 419: Are these not absolute paths to start with?

test/TestPvGenerator.cc
```diff
@@ -301,7 +301,7 @@ TEST_F(PvGeneratorTest, FitsPvImageVerticalCut) {
 }
 
 TEST_F(PvGeneratorTest, TestNoSpectralAxis) {
-    auto path_string = GeneratedHdf5ImagePath("10 10 10");
+    auto path_string = Hdf5Images() / "10_10_10_row_column.hdf5";
```
confluence: This also doesn't need to have NaNs.

test/TestRegion.cc
```diff
@@ -31,7 +31,7 @@ class RegionTest : public ::testing::Test {
 };
 
 TEST_F(RegionTest, TestSetUpdateRemoveRegion) {
-    std::string image_path = FileFinder::FitsImagePath("noise_3d.fits");
+    std::string image_path = FitsImages() / "noise_3d.fits";
```
confluence: All of these paths should have type `auto` (so that they are paths).

confluence: Here's how you can find all of them:
```
ag 'std::string .*path.*Images\(\)' test/
```

test/TestContour.cc
```diff
@@ -127,43 +116,44 @@ class ContourTest : public ::testing::Test {
 };
 
 TEST_F(ContourTest, NoSmoothingFitsFile) {
-    GenerateContour(500, 500, IMAGE_OPTS, CARTA::FileType::FITS, CARTA::SmoothingMode::NoSmoothing);
+    GenerateContour(FitsImages() / "500_500_image_opts.fits", CARTA::SmoothingMode::NoSmoothing);
```
confluence: It's unnecessary to put `image_opts` in these names. The images generated with `IMAGE_OPTS` here only set the random seed, so that they were always the same images. This was already the default option in the generation functions, so it's what all the tests that were generating images without an explicit options parameter were doing. Since we are now using static images, these should just be called `500_500.fits`, etc..

test/TestContour.cc
```diff
@@ -127,43 +116,44 @@ class ContourTest : public ::testing::Test {
...
 }
 TEST_F(ContourTest, NoSmoothingFitsFileNaN) {
-    GenerateContour(500, 500, IMAGE_OPTS_NAN, CARTA::FileType::FITS, CARTA::SmoothingMode::NoSmoothing);
+    GenerateContour(FitsImages() / "500_500_image_opts_nan.fits", CARTA::SmoothingMode::NoSmoothing);
```
confluence: These names should describe what the options actually are, so e.g. `500_500_nans_row_column.fits`.

confluence (changes requested):

test/TestCursorSpatialProfiles.cc
Michaelazzz: I'll generate an appropriate test file for it

test/TestProgramSettings.cc
Michaelazzz: They were not. I was fighting with them quite a bit today.

test/TestCursorSpatialProfiles.cc
Michaelazzz: Why do we need to check if the vector is of size 1?

src/Frame/Frame.cc
```diff
@@ -653,7 +653,6 @@ bool Frame::ContourImage(ContourCallback& partial_contour_callback, int channel)
         // Always use the full image cache (for now)
         FillImageCache();
     }
-
```
confluence, line 656: This is an irrelevant change (probably left over from previous clang-format issues). Please revert this (manually replacing the line is probably simplest).

test/TestContour.cc
```diff
@@ -127,43 +116,44 @@ class ContourTest : public ::testing::Test {
 };
 
 TEST_F(ContourTest, NoSmoothingFitsFile) {
-    GenerateContour(500, 500, IMAGE_OPTS, CARTA::FileType::FITS, CARTA::SmoothingMode::NoSmoothing);
+    GenerateContour(FitsImages() / "500x500.fits", CARTA::SmoothingMode::NoSmoothing);
 }
 TEST_F(ContourTest, NoSmoothingFitsFileNaN) {
-    GenerateContour(500, 500, IMAGE_OPTS_NAN, CARTA::FileType::FITS, CARTA::SmoothingMode::NoSmoothing);
+    GenerateContour(FitsImages() / "500x500_nan-row-column_nanpct-10.0_seed-0.fits", CARTA::SmoothingMode::NoSmoothing);
```
confluence: Please rename all these files to `something_nans.something`, for simplicity (remove the `-row-column_nanpct-10.0_seed-0`). We will probably pick different NaN settings later.

confluence (changes requested):

test/Util/TestUtilCasacore.cc
```diff
@@ -52,7 +52,8 @@ TEST_F(GetResolvedFilenameTest, SymlinkToNonExistentFile) {
 TEST_F(GetResolvedFilenameTest, FileExists) {
     auto pwd = TestRoot();
     std::string message;
-    std::string resolved = GetResolvedFilename(pwd.string(), "data/images/fits", "noise_4d.fits", message);
+    std::string resolved =
+        GetResolvedFilename(pwd.string(), "data/images/fits", "10x10x10x10_nan-row-column_nanpct-10.0_seed-0.fits", message);
```
confluence: `.string()` should not be needed here.

test/TestExprImage.cc
```diff
@@ -116,33 +102,31 @@ class ImageExprTest : public ::testing::Test {
...
 
 TEST_F(ImageExprTest, FitsImageExprSave) {
-    SaveImageExpr("noise_10px_10px.fits", "0", CARTA::FileType::FITS);
+    SaveImageExpr(FitsImages() / "10x10_nan-row-column_nanpct-10.0_seed-0.fits", "0");
 }
 
 TEST_F(ImageExprTest, ImageExprFails) {
     // Forms invalid expression
-    ASSERT_THROW(GenerateImageExprTimesTwo("noise_10px_10px.fits", "", CARTA::FileType::FITS, true), casacore::AipsError);
+    ASSERT_THROW(GenerateImageExprTimesTwo(FitsImages() / "10x10_nan-row-column_nanpct-10.0_seed-0.fits", "", true), casacore::AipsError);
```
confluence: All of these tests should still be using the original `noise_10px_10px` files.

test/TestFitsImage.cc
```diff
@@ -72,8 +72,8 @@ TEST_F(FitsImageTest, CorrectShape3dImage) {
 }
 
 TEST_F(FitsImageTest, CorrectShapeDegenerate3dImages) {
-    auto path_string = GeneratedFitsImagePath("10 10 10 1");
-    std::shared_ptr<carta::FileLoader> loader(carta::FileLoader::GetLoader(path_string));
+    auto path = FitsImages() / "10x10x10_degen.fits";
```
confluence: This file should be called `10x10x10x1.fits`.

test/TestFitsImage.cc
```diff
@@ -88,8 +88,8 @@ TEST_F(FitsImageTest, CorrectShapeDegenerate3dImages) {
     EXPECT_EQ(frame->StokesAxis(), 3);
 
     // CASA-generated images often have spectral and Stokes axes swapped
-    path_string = GeneratedFitsImagePath("10 10 1 10");
-    loader.reset(carta::FileLoader::GetLoader(path_string));
+    path = FitsImages() / "10x10x10_degen_casa.fits";
```
confluence: This file should be called `10x10x1x10.fits`.

test/TestMoment.cc
```diff
@@ -121,32 +124,32 @@ class MomentTest : public ::testing::Test, public FileFinder {
...
+    std::shared_ptr<casacore::ImageInterface<float>> image = std::make_shared<casacore::FITSImage>(file_path.string());
     int moment_axis(2);
 
-    if (OpenImage(image, file_path)) {
-        GenerateMoments(image, moment_axis);
-    } else {
-        spdlog::warn("Fail to open the file {}! Ignore the Moment test.", file_path);
-    }
+    GenerateMoments(image, moment_axis);
+    spdlog::warn("Fail to open the file {}! Ignore the Moment test.", file_path.string());
```
confluence: You should *just* call `GenerateMoments`. The error message should not be printed at all.

test/TestMoment.cc
```diff
@@ -121,32 +124,32 @@ class MomentTest : public ::testing::Test, public FileFinder {
...
             CompareImageData(casa_moment_image, carta_moment_image);
         }
     }
 };
 
 TEST_F(MomentTest, CheckConsistency) {
-    std::string file_path = FitsImagePath("M17_SWex_unittest.fits");
-    std::shared_ptr<casacore::ImageInterface<float>> image;
+    auto file_path = FitsImages() / "M17_SWex_unittest.fits";
+    std::shared_ptr<casacore::ImageInterface<float>> image = std::make_shared<casacore::FITSImage>(file_path.string());
```
confluence, line 138: `.string()` should not be needed here (unless casacore's string wrapper complains about the implicit conversion).

test/TestMoment.cc
```diff
@@ -121,32 +124,32 @@ class MomentTest : public ::testing::Test, public FileFinder {
...
+    auto file_path = FitsImages() / "small_perplanebeam.fits";
+    std::shared_ptr<casacore::ImageInterface<float>> image = std::make_shared<casacore::FITSImage>(file_path.string());
     int moment_axis(2);
 
     if (OpenImage(image, file_path)) {
         GenerateMoments(image, moment_axis);
     } else {
-        spdlog::warn("Fail to open the file {}! Ignore the Moment test.", file_path);
+        spdlog::warn("Fail to open the file {}! Ignore the Moment test.", file_path.string());
     }
```
confluence: This conditional should also be eliminated; you should just call `GenerateMoments`. The OpenImage check is unnecessary; it just repeats the shared pointer creation above, and if that fails the test should crash.

test/TestProgramSettings.cc
```diff
@@ -181,102 +185,100 @@ TEST_F(ProgramSettingsTest, OverrideDeprecatedRoot) {
...
-    auto fits_image_path = FitsImagePath("invalid.fits");
+    auto fits_image_path = FitsImages() / "invalid.fits";
     auto settings = SettingsFromVector({"carta_backend", fits_image_path});
     EXPECT_EQ(settings.starting_folder, default_settings.starting_folder);
     EXPECT_TRUE(settings.files.empty());
 }
 
 TEST_F(ProgramSettingsTest, FileImageFromPositional) {
-    auto fits_image_path = FitsImagePath("noise_10px_10px.fits");
+    auto fits_image_path = FitsImages() / "10x10_nan-row-column_nanpct-10.0_seed-0.fits";
```
confluence: All the tests here that originally used `noise_10px_10px` files should still use them.

test/TestProgramSettings.cc
```diff
@@ -392,32 +394,31 @@ TEST_F(ProgramSettingsTest, TestFileQueryStringEmptyFiles) {
 }
 
 TEST_F(ProgramSettingsTest, TestFileQueryStringSingleFile) {
-    auto image_root = TestRoot() / "data" / "images";
+    auto file_path = FitsImages() / "10x10x10_nan-row-column_nanpct-10.0_seed-0.fits";
```
confluence: The `noise_3d`, etc., images used in these tests were also originally existing test files on disk, not generated files. The original files should be used in these tests (we may eliminate similar files later, but we shouldn't make that change now).

test/TestRegion.cc
```diff
@@ -31,7 +31,7 @@ class RegionTest : public ::testing::Test {
 };
 
 TEST_F(RegionTest, TestSetUpdateRemoveRegion) {
-    std::string image_path = FileFinder::FitsImagePath("noise_3d.fits");
+    auto image_path = FitsImages() / "10x10x10_nan-row-column_nanpct-10.0_seed-0.fits";
```
confluence: These tests should also use the original files.

test/TestRegionMatched.cc
```diff
@@ -32,11 +32,11 @@ class RegionMatchedTest : public ::testing::Test {
 
 TEST_F(RegionMatchedTest, TestMatchedImageRectangleLCRegion) {
     // frame 0
-    std::string image_path0 = FileFinder::FitsImagePath("noise_10px_10px.fits");
+    auto image_path0 = FitsImages() / "10x10_nan-row-column_nanpct-10.0_seed-0.fits";
```
confluence: These tests should use the original files.

test/TestRegionSpatialProfiles.cc
```diff
@@ -119,7 +118,7 @@ class RegionSpatialProfileTest : public ::testing::Test {
 };
 
 TEST_F(RegionSpatialProfileTest, TestSpatialRequirements) {
-    std::string image_path = FileFinder::FitsImagePath("noise_3d.fits");
+    auto image_path = FitsImages() / "10x10x10.fits";
```
confluence: These tests should use the original files.

test/TestRegionSpectralProfiles.cc
```diff
@@ -85,7 +85,7 @@ class RegionSpectralProfileTest : public ::testing::Test {
 
 TEST_F(RegionSpectralProfileTest, TestPolygonSpectralProfile) {
     // Box described as 4-corner polygon
-    std::string image_path = FileFinder::FitsImagePath("noise_3d.fits");
+    auto image_path = FitsImages() / "10x10x10.fits";
```
confluence: Also original files here.

test/TestRegionStats.cc
```diff
@@ -62,7 +62,7 @@ class RegionStatsTest : public ::testing::Test {
 };
 
 TEST_F(RegionStatsTest, TestFitsRegionStats) {
-    std::string image_path = FileFinder::FitsImagePath("noise_3d.fits");
+    auto image_path = FitsImages() / "10x10x10_nan-row-column_nanpct-10.0_seed-0.fits";
```
confluence: Also here.

confluence (changes requested):
In addition to the specific comments, I noticed something about the data repository:
* The image generator in that repository should be a submodule, not a copy
* The image generator submodule in the *backend* should be removed in this PR
* The backend's dependency on the `fits2idia` executable should be removed: that means removing the `find_program` line from `test/CMakeLists.txt` and also the `fits2idia` package from the dependency installation command in the current dockerfiles. Everything else should be fine.

test/TestMoment.cc
confluence: Turns out that it does.

confluence (commented):
All the outstanding changes are fixed now; I'm going to remove some unused test images from the submodule before approval.

confluence (approved):

PR #1506, opened 2025-09-09
attempt to remove image cache from core dump

veggiesaurus:
**Description**

This PR reduces the size of core dumps by excluding the image channel cache data. On supported platforms, `madvise` is used to mark the memory address range appropriately. We recognise both the Linux flag and the FreeBSD flag -- we don't officially support FreeBSD, but this flexibility doesn't require much overhead. On macOS `madvise` does not implement this functionality.

`madvise` only acts on page-aligned address ranges. Page-aligning the image cache is a good idea for other reasons, so we do it regardless of whether the core dump exclusion is supported. The amount of memory wasted to pad the size to a multiple of the page size is minimal in comparison to the typical size of the cache.

Steps to test (on Linux):

1. Enable core dumps temporarily  (run `ulimit -c unlimited`; this applies per shell)
2. Run the backend
3. Open an image with large channels (ideally > 500M per channel)
4. Type `ctrl` + `\` in the shell to quit the backend and trigger a dump (you should see something like `^\Quit (core dumped)`)
5. Check the size of the core dump (should be in the current directory and called `core.PID_OF_BACKEND`).
6. If you compare sizes between this branch and `dev`, the difference should be exactly the size of one image channel.

On macOS, core dumps won't be affected, but the cache will still be page-aligned. There should be no adverse side-effects.

**Checklist**

- [X] changelog updated


confluence (commented):
The AI is technically correct about compatibility, but the advice is dubious. We should print an error message (with the errno value) if the call failed *or* a warning / info message if the behaviour is unsupported on the platform.

There's a `MADV_NOCORE` on FreeBSD, but nothing equivalent on macOS. I don't know if the backend is currently buildable in FreeBSD, but it would be simple to check for this as well.

I suggest an `ifdef` to check for behaviour availability and define a new behaviour variable (if we want to support both), and another `ifdef` to either call `madvise` or print a warning.

Semi-related: we still haven't fixed the line where we recreate a new float pointer for the cache every time we load the cache. Should we fix that first in a separate PR, and then move this to wherever the pointer is created (once)? If we fix that, we will only need to do this once as well, right?

veggiesaurus:
> Semi-related: we still haven't fixed the line where we recreate a new float pointer for the cache every time we load the cache. Should we fix that first in a separate PR, and then move this to wherever the pointer is created (once)? If we fix that, we will only need to do this once as well, right?

don't we only load the cache once? it's always the same size, surely we just update the values instead of allocating new memory?

confluence:
> > Semi-related: we still haven't fixed the line where we recreate a new float pointer for the cache every time we load the cache. Should we fix that first in a separate PR, and then move this to wherever the pointer is created (once)? If we fix that, we will only need to do this once as well, right?
> 
> don't we only load the cache once? it's always the same size, surely we just update the values instead of allocating new memory?

That's what we *should* do, but instead we do [this](https://github.com/CARTAvis/carta-backend/blob/58b4a0393067f374d6c436c1359625312222f593/src/Frame/Frame.cc#L391). This originally happened [here](https://github.com/CARTAvis/carta-backend/commit/d6eef0743a3cf27342c5ead9a68f06f1d43750ef). It was almost fixed [here](https://github.com/CARTAvis/carta-backend/pull/1414), but the PR was closed unmerged because it was unclear whether it was fixing the reported issue. The cache code is due to be refactored very soon, but I'm just going to make a quick fix PR now. This can't be good for performance. 🙈

confluence:
@veggiesaurus Problem: `madvise` can only be used on page-aligned memory. There are probably sufficiently portable ways to allocate the cache to be page-aligned, but I believe that this would require padding the array to a multiple of the page size (4kB on Linux, and on macOS we wouldn't need to do this at all). Maybe this won't be a problem if we define a padded size for the purposes of `madvise` but only read the cache up to the existing data size everywhere else. 🤔 We can discuss it tomorrow; if we continue with this, some of the refactoring I did to make this more generic is probably overkill, especially if the allocation has to depend on the availability of the advice flag as well.

confluence:
@veggiesaurus I updated the description. I can't add you as a reviewer on your own PR, but please comment. I'll try to add some basic unit tests.

I used psrecord to confirm that the memory is freed correctly.

kswang1029:
I tested with a single channel image (1.6GB). With this branch I see the core dump size is reduced by ~ the size of the image itself.

> (base) kswang@olaf:/var/lib/apport/coredump$ ls -lh
total 1.8G
-r-------- 1 kswang root 158M Nov  3 15:07 core._home_kswang_carta_build_carta-backend_build_carta_backend.1000.9a7a3ea8-e758-42c2-859a-c26dd87c59ca.51146.146686
-r-------- 1 kswang root 1.7G Nov  3 15:13 core._home_kswang_carta_build_carta-backend_build_carta_backend.1000.9a7a3ea8-e758-42c2-859a-c26dd87c59ca.52971.182987

In dev we have 1.7GB in size and in this branch it is 158MB.

(ps in my ubuntu box, the core dump file sits in `/var/lib/apport/coredump`)

kswang1029 (approved):

PR #1510, opened 2025-09-19
Refactor region histogram code

### LINKED ISSUE #1509: Refactor region histograms
Move implementation from RegionHandler

pford:
**Description**

Closes #1509 

  - Moved region histogram implementation from RegionHandler to new class RegionHistogram in new subdirectory src/Region/RegionAnalysis (will move region stats and profile code there too)
  - Fixed region histograms for non-current and computed Stokes
  - Simplified implementation to get data for histogram from ImageRegion->SubImage->data (with coordinate system overhead) to LCRegion->Slicer->data
  - Added RegionHistogramTest test for non-current Stokes and invalid Stokes (4D test image has I and Q, no U so cannot get computed Stokes)

Nope

All region histogram tests should pass.

**Checklist**

- [x] changelog updated


pford:
@confluence this is ready to be reviewed and tested, but should it wait for the Stokes refactoring? I changed Frame::GetStokesTypeIndex to check if the components for computed Stokes types exist, rather than simply returning the computed stokes type.  This should not cause a conflict since you are moving the Stokes index logic to FileLoader and will continue to fail if the Stokes cannot be computed.

This way, the requested histogram consistently fails for a missing image Stokes or an invalid computed Stokes due to a missing image Stokes, rather than later when the LCRegion fails.  Usually the LCRegion fails when the region is outside the image, and its histogram "succeeds" with a default histogram containing a zero value in one bin.

confluence:
> @confluence this is ready to be reviewed and tested, but should it wait for the Stokes refactoring?

I think we should go ahead with this in the meantime; I can adjust the Stokes refactoring if it's necessary.

kswang1029 (approved):
this looks fine. 👍

src/Region/RegionHandler.cc
```diff
@@ -665,15 +648,12 @@ void RegionHandler::RemoveFileRequirementsCache(int file_id) {
...
-        // Iterate through requirements and remove those for given file_id
-        for (auto it = _histogram_req.begin(); it != _histogram_req.end();) {
-            if ((*it).first.file_id == file_id) {
-                it = _histogram_req.erase(it);
-            } else {
-                ++it;
-            }
+        // Remove histogram configurations and cache for given file_id
+        for (auto& histogram : _region_histograms) {
+            histogram.second->ClearFileConfigsCache(file_id);
```
confluence: I suggest unpacking the key and value when iterating over maps -- even if the key is unused, it makes the loop variable name more idiomatic, since it binds only to the value: `for (auto& [key, histogram] : _region_histograms)...`

We could maybe use `_` for unused variables in situations like this. It's already a convention in some languages, and [in C++26 it will be formalised, with additional special properties](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2169r4.pdf).

confluence: But it would probably be best to leave this as-is here, and create a separate PR later for adjusting the loop style throughout the code (it's not very consistent).

confluence (approved):

src/Region/RegionHandler.cc
```diff
@@ -303,16 +302,12 @@ bool RegionHandler::SetHistogramRequirements(
     // Save frame pointer
     _frames[file_id] = frame;
 
-    // Make HistogramConfig vector of requirements
-    std::vector<HistogramConfig> input_configs;
-    for (const auto& config : configs) {
-        HistogramConfig hist_config(config);
-        input_configs.push_back(hist_config);
+    if (_region_histograms.find(region_id) == _region_histograms.end()) {
+        _region_histograms[region_id] = std::unique_ptr<RegionHistogram>(new RegionHistogram(region_id, file_id, configs));
```
confluence: It would be better to use `make_unique` here.

confluence (changes requested):

confluence (approved):

PR #1521, opened 2025-10-21
Use ImageRegion for histogram calculation

pford:
**Description**
Recent region histogram refactor used LCRegion bounding box data for histogram rather than ImageRegion data.

No issue

RegonHandler sends a casacore ImageRegion to RegionHistogram for the calculations. This also removes the need for the RegionAnalysisUtil region slicer function to get slicer data from the LCRegion, so this file has been deleted.

No

Create a closed region (rotated rectangle, polygon, ellipse) and a rectangle region which describes the first region's bounding box.  The histograms for the two regions should differ.

**Checklist**


confluence (approved):

kswang1029 (approved):
@pford thanks for realizing this regression in the previous PR that I missed (sorry all!). I confirmed the regression in dev and fixed in this PR. I have used v5 as the baseline to create a set of histogram e2e tests and now we have identical results in this PR and v5.

PR #1523, opened 2025-10-23
MipToLayer incorrect value

Michaelazzz:
**Description**

* A preference value set on the front-end application is causing the calculation of an invalid mip layer
* This PR explicitly sets the mip layer to -1, to indicate failure or obvious default value.
* No
* No

**Checklist**

### DIFF
--- a/src/DataStream/VectorField.cc
+++ b/src/DataStream/VectorField.cc
@@ -193,7 +193,7 @@ void GetTiles(int image_width, int image_height, int mip, std::vector<Tile>& til
     int tile_size_original = TILE_SIZE * mip;
     int num_tile_columns = ceil((double)image_width / tile_size_original);
     int num_tile_rows = ceil((double)image_height / tile_size_original);
-    int32_t tile_layer = Tile::MipToLayer(mip, image_width, image_height, TILE_SIZE, TILE_SIZE);
+    int32_t tile_layer = -1;
     tiles.resize(num_tile_rows * num_tile_columns);
 
     for (int j = 0; j < num_tile_rows; ++j) {


confluence (changes requested):
This issue isn't caused by a frontend preference, but by the way that we're repurposing tile objects and downsampling code to apply smoothing to vector field overlays. When we request tiles for the raster image, we use `LayerToMip` to convert a "layer", or magnification level used in the frontend, to the mip (power of 2 downsampling factor) that should be used for the requested data in the backend.

When we request tiles of polarization intensity and angle values to use to create the vector overlay, we provide the exact downsampling factor to use for smoothing. We call it a mip because we reuse the same TileData submessage, but it's not actually a mip (it doesn't have to be a power of two, and it doesn't have an upper bound defined by the size of the original image).

This means that the reverse calculation of a layer is not meaningful, and usually produces an invalid result of -1 (for example when a smoothing factor of 4 is used with an image that is too small to be downsampled by a factor of 4 in the raster image view). It's also unnecessary, because the vector overlay code in the frontend ignores this field. Previously the invalid layer had no effect, but adding explicit validation checks to the calling function is now causing it to fail.

That's why we should just set the layer property here to a constant -1 (although it's not used, this is preferable to leaving it at the default zero, because that would be a valid layer value). As we discussed, `Tile::MipToLayer` should also be removed entirely (together with its tests). It is only used here, and has no other purpose.

These changes should be implemented directly in #1488 rather than a separate PR, since that PR has become a broader set of changes to the Tile functions. That PR should be renamed to something more general, e.g. `Fix and expand Tile tests, add documentation, remove unnecessary function`.

Issue #1513, opened 2025-09-25
Refactor Stokes calculator and related slicer and region code


confluence:
This is a placeholder issue for the current Stokes refactoring work, which is a dependency for the refactoring of the image cache code.

Partial list of expected changes:

* Refactor calculator: simplify and consolidate functions; remove unnecessary custom slicing of Stokes cubes.
* Refactor file loader: stop recreating subimages with different offsets on demand; create one child calculator with a set of all available computed subimages to reuse. This should not require calculating any more data, and should make no difference to performance.
* Possibly subclass new calculator to implement support for computed Stokes for mipmaps and the rotated dataset in HDF5 files.
* Clean up Stokes type / index lookup functions in frame and file loader  
* Refactor calling code: remove `StokesSource`, `StokesSlicer` and `StokesRegion` wrappers. Pass a Stokes index parameter together with a normal slicer or region object instead.
* Stop applying XYZ offsets for computed Stokes. Only the Stokes index to be set to zero for computed Stokes: slicers and lattice-coordinate regions should contain "real" ranges relative to their corresponding image objects (the original image for basic Stokes or the full computed subimage cube for computed Stokes).
* If possible, eliminate the creation of slicers / lattice regions in the frame entirely, and defer it to the file loader, using base range and region object parameters instead in the frame. This would be helpful later, because (especially when the image cache is refactored and extended to implementations with different dimensions) requests for raster data may be fulfilled by the cache with a different slicer.
* Determine which of the functions that currently accept the Stokes wrapper object parameters actually support computed Stokes or any Stokes other than current, which of them *should* support either, and which of them are intrinsically limited to a subset of polarizations. Adjust the parameters accordingly.
* Unit tests for the polarization calculator.
* Docstrings for modified functions

PR #1525, opened 2025-10-24
Refactor polarization calculator and a lot of related code

confluence:
**Description**

This PR implements #1513 to refactor the polarization calculator. This work is a precursor to factoring the image cache into its own component and implementing different cache types.

Done:

- [X] Calculator: simplify and consolidate functions; remove unnecessary slicing of source and result images.
- [X] File loader: stop recreating computed subimages for different slices on demand. Create one calculator property on loader creation and generate a set of full computed images to be reused.
- [X] Functions for mapping between Stokes type and index moved to loader. Deduced mappings populated in coordinate axes function. Frame now only has a wrapper for extracting the Stokes type from a coordinate string and looking it up on the loader.
- [X] Extensive refactoring of calling code: remove StokesSource, StokesSlicer and StokesRegion wrappers. Pass a Stokes index parameter together with a base slicer or region object or z / z_range instead, as required. The slicer must be a *real* slicer and the region a *real* region relative to the underlying image (the original image or the computed image, selected using the Stokes index). See later work for further refactoring idea.
- [X] Stop using raw pointers for loaders. Use shared pointers everywhere and create them with `make_shared`.
- [X] Removed unused range parameters from `GetImageData` function in tests. This function is likely to be removed as we rewrite the tests to stop generating random test images.

To do before moving out of draft:
- [X] Check compilation and tests with ASAN enabled
- [ ] Basic manual test
- [ ] `psrecord` test to check for unexpected memory allocations
- [X] Consider making the Stokes index parameter default to zero in some functions that are commonly called on the base image, especially if the Stokes index is sometimes arbitrary. Calling these functions with the default parameter is cleaner than passing an explicit zero that is meaningless or creating an overloaded function. A dummy value should not be used (it would require a check whenever the function is called). E.g. `GetCoordinateSystem` / `CoordinateSystem` should be refactored like this.
- [ ] Unit tests for the polarization calculator.
- [ ] Docstrings for the polarization calculator and some modified functions
- [ ] Consistently use `stokes_index` rather than `stokes` for every variable that refers to an index rather than a polarization type.
- [ ] Loader: merge `FindCoordinateAxes` into `OpenFile` to eliminate the need for it to be called explicitly (a problem in tests that create a bare loader without a frame). Also move other common loader metadata setup code into the parent loader to remove duplication across all the subclasses.

To check during code review:

- [ ] Confirm whether the casacore region types of the parameters are correct. Should we still be using `casacore::LattRegionHolder`, or are all these functions actually called with `casacore::ImageRegion`s?
- [ ] Confirm which of the functions that previously accepted the Stokes wrapper object parameters actually support: single z / z range / current z only / all z only, any Stokes / current Stokes only / particular Stokes only, xy ranges, etc., and that the new parameters correctly reflect this. E.g. we should not have a `stokes_index` parameter in a function that it only makes sense to call with the current Stokes, or a `z_range` parameter in a function that only uses a single z value or is only ever called with a single z value.

Related work to leave for later PRs:

* Subclass new calculator to implement support for computed Stokes for mipmaps and the rotated dataset in HDF5 files, and update calling code to ensure that these options are used. The tile cache should already be supported (tiles are read from the computed images for the normal dataset).
* Frame: investigate eliminating `GetImageSlicer` + `GetSlicerData` combination in favour of calling `GetSlicerData` directly with the original range parameters. The slicer is a low-level implementation detail specific to a particular data source; it should be created in the component that manages the source, e.g. the loader or (later) the image cache.
* Some `Frame` functions incorrectly treat the computed Stokes subimage as if contains in-memory data which must be copied and not referenced *or* as if creating it requires reading from disk and locking the image mutex. This should be checked and cleaned up. The image mutex should really be moved to the file loader, and all disk reads should be performed through loader functions.
* Regions: in the old implementation, coordinate regions for computed polarizations aren't cached, I believe because the computed subimages keep changing size. With this refactoring, the subimage dimensions are constant, so these regions could also be cached -- but in a separate cache, because the computed image dimensions may not match the original image (stokes dimension is always 1), except where the regions are always 2D and that dimension is irrelevant.
* Regions: the computed Stokes spectral profile code in `RegionHandler` is very convoluted and can be simplified greatly.
    * Would be simpler / more efficient to calculate the computed profiles from data read from the computed images than to derive them from component profiles?
    * Could be faster to use the components if they are already cached.
    * Reading from the computed image should be more or less equivalent to reading all the components and computing (assuming that the swizzled dataset is supported by the calculator, and that the current profile code is optimised).
    * In future, when we have the option of a full image cache, we should check the cache first (presumably we also won't cache any region data if a full cache is available). 

**Checklist**


src/Frame/Frame.cc
```diff
@@ -414,10 +382,10 @@ void Frame::InvalidateImageCache() {
 }
 
 void Frame::GetZSlice(std::vector<float>& z_slice, size_t z, size_t stokes) {
```
pford: Should rename `z_slice` parameter to indicate it is the return data, not a casacore::Slice.

src/Frame/Frame.cc
```diff
@@ -1531,9 +1500,9 @@ bool Frame::FillSpectralProfileData(std::function<void(CARTA::SpectralProfileDat
                     const auto N = slicer.length().product();
                     std::unique_ptr<float[]> buffer(new float[N]);
                     end_channel = start(_axes.z) + nz - 1;
```
pford, line 1489: Should call this end_z for consistency since the z axis may or may not be the spectral axis.

src/Frame/Frame.cc
```diff
@@ -1612,76 +1581,71 @@ bool Frame::HasSpectralConfig(const SpectralConfig& config) {
 // Region/Slicer Support (Frame manages image mutex)
 
 std::shared_ptr<casacore::LCRegion> Frame::GetImageRegion(
-    int file_id, std::shared_ptr<Region> region, const StokesSource& stokes_source, bool report_error) {
+    int file_id, std::shared_ptr<Region> region, int stokes_index, bool report_error) {
```
pford, line 1571: Could remove this method and call region->GetImageRegion directly in Frame::SaveFile, and do the same in RegionHandler::ApplyRegionToFile in this PR (this RH method is removed in the same way--calling the Region method directly--in the region spatial profile refactor PR)

src/Frame/Frame.cc
```diff
@@ -1696,11 +1660,11 @@ bool Frame::GetRegionData(const StokesRegion& stokes_region, std::vector<float>&
...
-            bool is_computed_stokes(!stokes_region.stokes_source.IsOriginalImage());
 
             // Get image data and mask, with image mutex locked
             std::unique_lock<std::mutex> ulock(_image_mutex);
-            if (_loader->IsGenerated() || is_computed_stokes) { // For the image in memory
+
+            if (_loader->IsGenerated() || Stokes::IsComputed(stokes_index)) { // For the image in memory
                 casacore::Array<float> tmpdata;
                 sub_image.doGetSlice(tmpdata, slicer);
                 data = tmpdata.tovector();
```
pford, line 1657: As discussed in Slack, get tmpdata slice using data pointer and shared storage

src/Frame/Frame.cc
```diff
@@ -1735,16 +1699,17 @@ bool Frame::GetRegionData(const StokesRegion& stokes_region, std::vector<float>&
...
+    if (_image_cache_valid && (Depth() == 1 || slicer.start()(_axes.z) == slicer.end()(_axes.z) == CurrentZ()) &&
+        stokes_index == CurrentStokes()) {
         // Slice image cache
         auto cache_shape = ImageShape();
-        auto slicer_start = stokes_slicer.slicer.start();
-        auto slicer_end = stokes_slicer.slicer.end();
+        auto slicer_start = slicer.start();
+        auto slicer_end = slicer.end();
 
         // Adjust cache shape and slicer for single channel and stokes
```
pford: comment = for single *z* and stokes

src/Frame/Frame.cc
```diff
@@ -1612,76 +1581,71 @@ bool Frame::HasSpectralConfig(const SpectralConfig& config) {
...
-    auto coord_sys = CoordinateSystem(stokes_region.stokes_source);
-    casacore::LatticeRegion lattice_region =
-        stokes_region.image_region.toLatticeRegion(*coord_sys.get(), ImageShape(stokes_region.stokes_source));
+    auto coord_sys = CoordinateSystem(stokes_index);
+    casacore::LatticeRegion lattice_region = region.toLatticeRegion(*coord_sys.get(), ImageShape(stokes_index));
     return lattice_region.shape();
 }
 
-bool Frame::GetRegionSubImage(const StokesRegion& stokes_region, casacore::SubImage<float>& sub_image) {
+bool Frame::GetRegionSubImage(const casacore::LattRegionHolder& region, int stokes_index, casacore::SubImage<float>& sub_image) {
```
pford: We should note that LattRegionHolder is the base class of ImageRegion, or just use ImageRegion (as in GetRegionData)

src/Frame/Frame.cc
```diff
@@ -1763,21 +1728,21 @@ bool Frame::GetSlicerData(const StokesSlicer& stokes_slicer, float* data) {
...
+        data_ok = _loader->GetSlice(tmp, slicer, stokes_index);
         _loader->CloseImageIfUpdated();
         ulock.unlock();
     }
     return data_ok;
 }
 
-bool Frame::GetRegionStats(const StokesRegion& stokes_region, const std::vector<CARTA::StatsType>& required_stats, bool per_z,
-    std::map<CARTA::StatsType, std::vector<double>>& stats_values) {
+bool Frame::GetRegionStats(const casacore::LattRegionHolder& region, int stokes_index, const std::vector<CARTA::StatsType>& required_stats,
```
pford: casacore::ImageRegion& ?

src/Frame/Frame.cc
```diff
@@ -1852,7 +1819,7 @@ void Frame::StopMomentCalc() {
 }
 
 bool Frame::FitImage(const CARTA::FittingRequest& fitting_request, CARTA::FittingResponse& fitting_response, GeneratedImage& model_image,
-    GeneratedImage& residual_image, GeneratorProgressCallback progress_callback, StokesRegion* stokes_region) {
+    GeneratedImage& residual_image, GeneratorProgressCallback progress_callback, casacore::ImageRegion* region) {
```
pford, line 1809: This is the only method with ImageRegion* instead of ImageRegion&.  Seems odd.

src/Frame/Frame.cc
```diff
@@ -2054,7 +2020,7 @@ void Frame::SaveFile(const std::string& root_folder, const CARTA::SaveFile& save
                 _loader->GetSubImage(slice_sub_image, latt_region_holder, sub_image);
             } else {
                 auto slice_sub_image = GetExportImageSlicer(save_file_msg, image_shape);
```
pford: sub_image_slicer would be clearer since it returns casacore::Slicer.  Also clearer to send channels and stokes from save_file_msg message rather than message itself?

src/Frame/Frame.cc
```diff
@@ -2495,8 +2424,8 @@ bool Frame::GetDownsampledRasterData(
         int y_max = bounds.y_max() - 1;
 
         auto tile_stokes_section = GetImageSlicer(AxisRange(x_min, x_max), AxisRange(y_min, y_max), AxisRange(z), stokes);
-        tile_data.resize(tile_stokes_section.slicer.length().product());
-        if (!GetSlicerData(tile_stokes_section, tile_data.data())) {
+        tile_data.resize(tile_stokes_section.length().product());
+        if (!GetSlicerData(tile_stokes_section, stokes, tile_data.data())) {
```
pford, line 2415: Should we consistently use `stokes_index` for int variable rather than `stokes`? (guilty of this myself)

src/ImageData/FileLoader.cc
```diff
@@ -914,50 +922,80 @@ double FileLoader::CalculateBeamArea() {
...
+            spdlog::warn("Could not get or deduce index for polarization {}.", Stokes::Name(stokes_type));
+            return false;
+        }
     }
 }
 
 bool FileLoader::GetStokesType(const int& stokes_index, CARTA::PolarizationType& stokes_type) {
+    if (Stokes::IsComputed(stokes_index)) {
+        // Computed type which is available for this image
+        if (_polarization_calculator->AvailablePolarizations().count((CARTA::PolarizationType)stokes_index)) {
```
pford, line 962: count -> find

src/ImageData/FileLoader.cc
```diff
@@ -914,50 +922,80 @@ double FileLoader::CalculateBeamArea() {
 }
 
 bool FileLoader::GetStokesTypeIndex(const CARTA::PolarizationType& stokes_type, int& stokes_index) {
+    // Computed type which is available for this image
+    if (_polarization_calculator->AvailablePolarizations().count(stokes_type)) {
```
pford, line 938: I think previously we preferred find() to count()

pford (changes requested):

src/Frame/Frame.cc
pford: In the spatial profile PR I renamed Region::GetImageRegion -> Region::GetLCRegion and moved RegionHandler::ApplyRegionToFile out to its own file as GetImageRegion.  I think these remain casacore::ImageRegion everywhere in the code until FileLoader::GetSubImage, where the casacore::SubImage constructor has a LattRegionHolder parameter.

confluence: To be resolved when I merge in dev (after merging the changes I'm moving out of this PR into dev.

src/Frame/Frame.cc
confluence: This is currently used to make the region optional. We can refactor the function to avoid doing that, but I'd rather not do it in this PR.

src/Frame/Frame.cc
confluence: I'm renaming the variable. I agree about the rest, but all these file-saving functions need a lot of other refactoring and to be moved out of Frame into a subcomponent, which should happen in a different PR.

src/Frame/Frame.cc
confluence: Agreed, but I will make a separate small PR to make the change more broadly first.

confluence: Now that I'm looking at this, I think it would make more sense to fix this *after* merging this PR (otherwise I'd be making conflicting changes to a lot of code that it touches, which would make the merge more difficult).

At the same time we can fix use of "channel" where we mean "z".

src/ImageData/FileLoader.cc
confluence: It's slightly more efficient, but in this case we'll have such a small number of items that it shouldn't make a significant difference. We use count in a lot of other places; this is also something that we should address in a new PR.

src/Frame/Frame.cc
confluence: Makes sense. I'll fix this here.

src/Frame/Frame.cc
confluence: Agreed, but we're still using "channel" in lots of other places -- I think we should fix this separately.

src/Frame/Frame.cc
confluence: What were we going to do here? 🤔 Should we be treating the computed Stokes in the same way as the normal (not generated) image read?

Issue #1522, opened 2025-10-23
cannot display polarization histogram other than Stokes I if there are matched images


kswang1029:
**Describe the bug**
When there is only one image, we can freely switch to other polarization components (eg Stokes U) to see its histogram in the histogram widget. However, if there are matched images, we cannot do so anymore. Only Stokes I is fine. Other polarization components will have a histogram of one bin returned from the backend. The protobuf message sent from the frontend seems okay which sets the correct file_id, region_id, and the selected polarization component (@loveluthien if you have time please help verifying the frontend behaves correctly).

**To Reproduce**
1. download test data from https://drive.google.com/drive/folders/1k7gAemQ9X2cib4KzvaMeHUz_1AacSY8h?usp=share_link
2. load HL_Tau_StokesIQUV_clean.pbcor.fits and HL_tau_Stokes_hypercube_IQU.pbcor.fits in order
3. match the two images spatially
4. create a region
5. enable the histogram widget
6. use the polarization dropdown menu to select Stokes I and we will see a histogram which is correct.
7. select Stokes Q so on we will see an empty histogram (actually it is a histogram with one bin only).

**Expected behavior**
With and without spatial matching, we should be able to see region histograms with different selections of polarization components.

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]: macOS Sequoia
 - Browser [e.g. chrome, safari, electron app]: chrome
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v5.0.3, dev
 - Frontend branch [e.g. dev, v3b2 release]: v5.0.3, dev

pford:
Just saw the backend messages on Slack, so we do not want to fix this now since it is fixed in the refactoring for v6b1?

kswang1029:
> Just saw the backend messages on Slack, so we do not want to fix this now since it is fixed in the refactoring for v6b1?

Will there be duplicated work? (fixing the issue using v5.0.3 as the base branch)

kswang1029:
> > Just saw the backend messages on Slack, so we do not want to fix this now since it is fixed in the refactoring for v6b1?
> 
> Will there be duplicated work? (fixing the issue using v5.0.3 as the base branch)

@pford after discussing with dev leads, the consensus now is to fix this in dev, not in the point release 5.0.4 to reduce work overhead. We can close this issue once your refactoring PR is merged. I will double-check to make sure this issue is gone in that PR.

[...]

PR #1550, opened 2025-12-12
Create casacore array correctly when fetching region data for other Stokes

confluence:
Previously the temporary array would be discarded and the data would be empty. This has been isolated for the backport from #1510.

The LCRegion from Region::GetImageRegion was also fixed to return a 2D LCRegion for the reference image for computed stokes, so that it could be used to get the region data as mentioned.

I will also update the changelog in this branch, and retroactively add the entry to the dev changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.1]
+
+### Fixed
+* Fixed fetching of region data for other Stokes ([#1551](https://github.com/CARTAvis/carta-backend/issues/1551)).
+
 ## [5.0.2]
 
 ### Fixed
--- a/src/Frame/Frame.cc
+++ b/src/Frame/Frame.cc
@@ -1694,13 +1694,13 @@ bool Frame::GetRegionData(const StokesRegion& stokes_region, std::vector<float>&
 
             // Get image data and mask, with image mutex locked
             std::unique_lock<std::mutex> ulock(_image_mutex);
-            casacore::Array<float> tmpdata;
             if (_loader->IsGenerated() || is_computed_stokes) { // For the image in memory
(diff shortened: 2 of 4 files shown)


pford (approved):

kswang1029 (approved):

kswang1029:
@pford thanks for addressing the missing computed components. Tested [a08415c](https://github.com/CARTAvis/carta-backend/pull/1550/commits/a08415c52ec0bfeaf1dddd4e235d39b7fa22978a) and it works fine. @confluence ok to proceed merging and backport.

kswang1029:
Spotted an issue in this `PR` and in `dev` that I think I missed before: using the test images described in https://github.com/CARTAvis/carta-backend/issues/1522 and doing the same tests but also check the histograms from the _reference_ image when selecting different polarization components (including computed ones). I see Stokes IQUV are missing but the computed ones can be displayed.

pford:
@kswang1029 I cannot see the issue in this PR branch.  I load HL_Tau_StokesIQUV_clean.pbcor.fits, append HL_tau_Stokes_hypercube_IQU.pbcor.fits and match XY, draw a rectangle, and request region histogram for all image Stokes and computed Stokes for reference image and matched image.  In all cases I get a histogram.

Update: Perhaps I misunderstood where the region is created. In my previous message I created the region in image 0.

I can reproduce the issue when the region is created in image 1 (spatially matched to image 0), then image 0 region histograms fail for  I, Q, U, V. Image 0 is the **reference** image for **spatial matching** but the **matched** image for the **region**.

pford:
Another update: today I cannot reproduce the issue in this branch or dev.  Works just fine. @confluence can you try? @kswang1029 does it fail reliably for you?

kswang1029:
@pford It seems it requires some special steps to reproduce the issue. Please check slack where I send you a screen recording.

pford:
@kswang1029 I found and fixed the matched image region histogram bug. Please test again.

kswang1029 (approved):

confluence:
@pford your changes look good to me. Is this ready for merge now?

pford (approved):

pford:
> @pford your changes look good to me. Is this ready for merge now?

Yes, please go ahead.

Issue #1354, opened 2024-01-29
Include full path in error messages about nonexistent files (e.g. when loading a workspace)


confluence:
At the moment if a workspace includes a file which is moved or deleted, when the workspace is loaded the error message about this file only contains the filename, not the full path to the file. This is unhelpful if the user has lots of files which have the same name but are in different subdirectories (this has actually come up recently, and seems like a common use case).

The error message should include the full path to the file. If possible (if it's not considered a security risk<sup>1</sup>) we should ensure that it's the correct absolute filesystem path in the case where the CARTA root is not the filesystem root. The user should be able to copy this path and use it directly for debugging.

1. My understanding is that we use the CARTA root mostly to avoid showing the user irrelevant system directories, and that we don't rely on it for security, so exposing the real filesystem path to the CARTA root should not be a problem.

confluence:
The relevant code is in `GetResolvedFilename` in `Util/Casacore`. This should be a simple fix: use the full path string instead of just the directory or file name in the messages.

PR #1555, opened 2026-01-07
Include full path in error messages about nonexistent files (e.g. when loading a workspace) issue #1354

marcinsokolowski:
**Description**
When error due to non-existing or non-readable file occurs the error message should contain both directory and filename.
This seems a very small on standalone change (Issue #1354). 

Done:
- [x] extended error message with the missing information (file or directory)

To do before moving out of draft:
- [ ] Adrianna to review and also verify if this satisfies any potential security concerns mentioned in the description.
- [ ] verify if this is what was required or anything else should be added
- [ ] what to show in the exception catch code (Casacore.cc : line ~47 } catch (const casacore::AipsError& err) {). Perhaps path.expandedName() should also be added to the error message (around like message = err.getMesg());

To check during code review:

Related work to leave for later PRs:

**Checklist**
- [x] changelog updated
- [x]  no protobuf update needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Moved region histogram implementation to analysis subdirectory ([#1509](https://github.com/CARTAvis/carta-backend/issues/1509)).
 * Moved region statistics implementation to analysis subdirectory ([#1512](https://github.com/CARTAvis/carta-backend/issues/1512)).
 * Excluded image cache data from core dumps on supported platforms ([#1506](https://github.com/CARTAvis/carta-backend/pull/1506)).
+* Moved region spatial profile implementation to analysis subdirectory ([#1515](https://github.com/CARTAvis/carta-backend/issues/1515)).
 * Full file path was added to the error message due to non-existing/non-readable file ([#1354](https://github.com/CARTAvis/carta-backend/issues/1354)).
 
 ## [5.0.2]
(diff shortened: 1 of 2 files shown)


confluence (commented):
1. In all these places you should only replace the file or directory variable with the whole `path` variable instead. That's what's missing -- the actual location of the error.
2. Please use more of the PR template in the description (specifically the checklist at the end, where we should track the update to the changelog, whether there are dependencies, etc.). You can copy and paste that from a new PR edit window (don't save the new PR).
3. Please include a link to the issue directly in the PR description (you can just type the number), so that GitHub links it automatically.
4. Re: security / what form of the path should be printed: ~~maybe it's better to print the path in its original form, since this is what is actually being tested (i.e. don't do anything else to it, and we'll test it and see if it works).~~ actually, now that I read the comments in the code, it looks like `exists` and `isReadable` use `.expandedName()`, so that's what we should put in the message(s).

confluence (commented):
This is looking good. I have explained my previous point 2 on Slack. I agree that we should also add the path to the fallback error -- maybe something like: `"Error resolving file [path here]: [error here]"`

Style suggestion: for consistency I think we should use `fmt::format` for this string interpolation instead of concatenating with `+`.

confluence (approved):
I think this is ready for testing. It should be added to the changelog.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Don't reallocate memory for image cache unnecessarily ([#1508](https://github.com/CARTAvis/carta-backend/pull/1508)).
 * Fixed crash in ICD channel map test with ASAN enabled ([#1518](https://github.com/CARTAvis/carta-backend/issues/1518)).
 * Fixed fetching of region data for other Stokes ([#1551](https://github.com/CARTAvis/carta-backend/issues/1551)).
+* Full file path was added to the error message due to non-existing/non-readable file ([#1354](https://github.com/CARTAvis/carta-backend/issues/1354)).
```
pford: I think this line should be moved to the "Changed" section.  The issue is labeled an enhancement not a bug.

pford (changes requested):

CHANGELOG.md
pford: Also note the merge conflict in CHANGELOG.md

kswang1029 (approved):

CHANGELOG.md
confluence: I believe that this is resolved.

pford (approved):

PR #1559, opened 2026-01-19
SpectralProfileData and SpatialProfileData protobuf cleanup

Michaelazzz:
**Description**

Streamlining of SpectralProfileData protobuf message and sub-messages.

**Checklist**


src/Util/Message.h
```diff
@@ -81,6 +81,38 @@ class Message {
...
+    static CARTA::SetSpectralRequirements_SpectralConfig SpectralConfig(const std::string& coordinate);
+    static CARTA::FileListRequest FileListRequest(
+        const std::string& directory, const CARTA::FileListFilterMode filter_mode = CARTA::FileListFilterMode::Content);
+    static CARTA::FileInfoRequest FileInfoRequest(const std::string& directory, const std::string& file, const std::string& hdu = "");
+    static CARTA::SetContourParameters SetContourParameters(uint32_t file_id, uint32_t ref_file_id, int32_t x_min, int32_t x_max,
+        int32_t y_min, int32_t y_max, const std::vector<double>& levels, CARTA::SmoothingMode smoothing_mode, int32_t smoothing_factor,
+        int32_t decimation_factor, int32_t compression_level, int32_t contour_chunk_size);
+    static CARTA::SetVectorOverlayParameters SetVectorOverlayParameters(uint32_t file_id, uint32_t mip, bool fractional, double threshold,
+        bool debiasing, double q_error, double u_error, int32_t stokes_intensity, int32_t stokes_angle,
+        const CARTA::CompressionType& compression_type, float compression_quality);
```
confluence: This is a merge error. These messages were removed in a recent PR.

src/Util/Message.cc
```diff
@@ -170,41 +170,50 @@ std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t e
...
-    for (auto stats_type : required_stats) {
+    return profile_message;
+}
+
+CARTA::SpectralProfileData Message::SpectralProfileData(int32_t stokes, float progress) {
+    CARTA::SpectralProfileData message;
+    message.set_stokes(stokes);
+    message.set_progress(progress);
+    return message;
+}
```
confluence: This doesn't need to be two separate functions. Put the IDs last as optional parameters defaulting to 0. As previously discussed, that is the protobuf default, so this has the identical result to not setting them.

src/Frame/Frame.cc
```diff
@@ -1465,10 +1461,8 @@ bool Frame::FillSpectralProfileData(std::function<void(CARTA::SpectralProfileDat
 
         // Create final profile message for callback
         auto profile_message = Message::SpectralProfileData(CurrentStokes(), 1.0);
-        auto spectral_profile = profile_message.add_profiles();
-        spectral_profile->set_coordinate(config.coordinate);
-        // point spectral profiles only have one stats type
-        spectral_profile->set_stats_type(config.all_stats[0]);
+        auto spectral_profile = Message::AddSpectralProfile(
+            profile_message, config.coordinate, config.all_stats[0]); // point spectral profiles only have one stats type
```
confluence: 1. This creation of the message and submessage should be removed from here and copied to the two places in the conditional below where these objects is used.
2. `AddSpectralProfile` should be refactored to (mostly) match `AddSpatialProfile`: it should take a data vector as a parameter and add one profile. You will probably need two versions for the two value data types. See other comments for more details.

src/Frame/Frame.cc
```diff
@@ -1566,9 +1560,7 @@ bool Frame::FillSpectralProfileData(std::function<void(CARTA::SpectralProfileDat
                         t_start_profile = t_end_slice;
 
                         auto partial_data = Message::SpectralProfileData(CurrentStokes(), progress);
-                        auto partial_profile = partial_data.add_profiles();
-                        partial_profile->set_stats_type(config.all_stats[0]);
-                        partial_profile->set_coordinate(config.coordinate);
+                        auto partial_profile = Message::AddSpectralProfile(partial_data, config.coordinate, config.all_stats[0]);
                         partial_profile->set_raw_values_fp32(spectral_data.data(), spectral_data.size() * sizeof(float));
```
confluence: This should happen inside the (refactored) `AddSpectralProfile` (you'd be calling the float version here).

You then shouldn't have to assign the return value to anything.

The same applies to the other two places where you'd be creating `profile_message`. I would also rename `partial_data` to `profile_message` here.

src/Region/RegionHandler.cc
```diff
@@ -1511,8 +1511,8 @@ bool RegionHandler::FillSpectralProfileData(
                 AxisRange z_range(0, _frames.at(config_file_id)->Depth() - 1); // all channels
                 profile_ok = GetRegionSpectralData(config_region_id, config_file_id, z_range, coordinate, stokes_index, required_stats,
                     report_error, [&](std::map<CARTA::StatsType, std::vector<double>> results, float progress) {
-                        auto profile_message = Message::SpectralProfileData(
-                            config_file_id, config_region_id, stokes_index, progress, coordinate, required_stats, results);
+                        auto profile_message = Message::SpectralProfileData(config_file_id, config_region_id, stokes_index, progress);
+                        Message::AddSpectralProfile(profile_message, coordinate, required_stats, results);
```
confluence: The loop logic inside this version of the function is too complex for a wrapper function, and should be moved outside into this calling code.

This calling code should iterate over the stats, and for each stat it should call (the refactored double version of) `AddSpectralProfile`, which (like the float version) should take one data vector as a parameter and set and return a profile (although you won't actually need the pointer, so you can keep calling the function without capturing the return value).

src/Util/Message.cc
```diff
@@ -170,41 +170,50 @@ std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t e
...
-    CARTA::SpectralProfileData message;
-    message.set_stokes(stokes);
-    message.set_progress(progress);
-    return message;
+CARTA::SpectralProfile* Message::AddSpectralProfile(
+    CARTA::SpectralProfileData& profile_data, std::string& coordinate, CARTA::StatsType& stats_type) {
+    auto* new_profile = profile_data.add_profiles();
+    new_profile->set_coordinate(coordinate);
+    new_profile->set_stats_type(stats_type);
+    return new_profile;
```
confluence: This should take a float vector parameter (the data) and set the fp32 values (like `AddSpatialProfile`).

src/Util/Message.cc
```diff
@@ -170,41 +170,50 @@ std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t e
...
         new_profile->set_coordinate(coordinate);
         new_profile->set_stats_type(stats_type);
 
-        if (spectral_data.find(stats_type) == spectral_data.end()) { // stat not provided
+        if (raw_values_fp64.find(stats_type) == raw_values_fp64.end()) { // stat not provided
             double nan_value = DOUBLE_NAN;
             new_profile->set_raw_values_fp64(&nan_value, sizeof(double));
         } else {
-            new_profile->set_raw_values_fp64(spectral_data[stats_type].data(), spectral_data[stats_type].size() * sizeof(double));
+            new_profile->set_raw_values_fp64(raw_values_fp64[stats_type].data(), raw_values_fp64[stats_type].size() * sizeof(double));
```
confluence: `for (auto stats_type : stats_type) {` may be causing a problem!

The loop and NaN logic is too complex for a wrapper function and should be moved outside (see previous comment about this).

This function should match the (refactored) float version, take a double vector as a parameter, and set the double values instead of the float values.

src/Region/RegionHandler.cc
confluence: The NaN decision should also be made here. To maintain a consistent parameter type, you should create a single-element double vector with the double NaN value and pass that in for each missing stat (you can create it outside the loop and reuse it instead of recreating it).

src/Util/Message.cc
confluence: You could in theory have a single function with two optional parameters, but then you'd have the problem of having to pass in a null value for the first one if you wanted to use the second one. So it's best to keep them split (especially since they'll be simple).

src/Util/Message.cc
```diff
@@ -238,6 +235,30 @@ CARTA::SpatialProfileData Message::SpatialProfileData(int32_t x, int32_t y, int3
     return message;
 }
 
+CARTA::SpatialProfile* Message::AddSpatialProfile(CARTA::SpatialProfileData& response, int32_t start, int32_t end,
+    std::vector<float>& raw_values_fp32, std::string coordinate, int32_t mip) {
```
confluence: `raw_values_fp32` should be called `values`. You know what the type is because it's in the function signature.

src/Util/Message.cc
```diff
@@ -170,41 +170,50 @@ std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t e
...
     }
-    return profile_message;
 }
 
-CARTA::SpectralProfileData Message::SpectralProfileData(int32_t stokes, float progress) {
-    CARTA::SpectralProfileData message;
-    message.set_stokes(stokes);
-    message.set_progress(progress);
-    return message;
+CARTA::SpectralProfile* Message::AddSpectralProfile(
```
confluence: I suggest renaming both of the `AddSpectralProfile` wrappers and `AddSpatialProfile` to `AddProfile`: they all have unambiguously different signatures because the parent messages have different types.

src/Util/Message.cc
confluence: The parent message parameter should be called `profile_message`, as in the `SpatialProfileData` wrapper below; `profile_data` is confusing (sounds like data values).

src/Util/Message.cc
confluence: `response` should be called `profile_message`, for consistency.

confluence: `spatial_profile` can just be called `profile`.

src/Util/Message.cc
```diff
@@ -238,6 +235,30 @@ CARTA::SpatialProfileData Message::SpatialProfileData(int32_t x, int32_t y, int3
...
+    spatial_profile->set_coordinate(coordinate);
+    spatial_profile->set_mip(mip);
+
+    return spatial_profile;
+}
+
+CARTA::LineProfileAxis* Message::AddLineProfileAxis(
+    CARTA::SpatialProfile* spatial_profile, CARTA::ProfileAxisType axis_type, float crpix, float crval, float cdelt, std::string unit) {
+    auto line_profile_axis = spatial_profile->mutable_line_axis();
+    line_profile_axis->set_axis_type(axis_type);
```
confluence: `line_profile_axis` can be called `axis` in here.

src/Util/Message.cc
confluence: `new_profile` can be called `profile`.

confluence (changes requested):

src/Util/Message.cc
confluence: I see that the existing `SpatialProfileData` wrappers have the same issue -- they should also be merged into one.

src/Frame/Frame.cc
```diff
@@ -1481,8 +1470,10 @@ bool Frame::FillSpectralProfileData(std::function<void(CARTA::SpectralProfileDat
             int xy_count(1);
             if (!Stokes::IsComputed(stokes) && _loader->GetCursorSpectralData(spectral_data, stokes, (start_cursor.x + 0.5), xy_count,
                                                    (start_cursor.y + 0.5), xy_count, _image_mutex)) {
-                // Use loader data
-                spectral_profile->set_raw_values_fp32(spectral_data.data(), spectral_data.size() * sizeof(float));
+                // Send final profile message with loaded data
```
confluence: This comment should say "loader data"

confluence (approved):

confluence:
@kswang1029 This is refactoring that should not change any behaviour.

kswang1029 (changes requested):
@Michaelazzz @confluence this PR breaks several spatial and spectral profiler tests. The symptom is empty plot.

https://github.com/user-attachments/assets/1e86cbe2-15c6-4833-89e9-c500266117b2

kswang1029:
still having issues with the spectral profiler. Cannot see a profile from a second region, nor having multiple profiles in one plot.

https://github.com/user-attachments/assets/749e3c4a-708b-42a4-ae10-4a517dd8d219

src/Frame/Frame.cc
```diff
@@ -1373,13 +1373,9 @@ bool Frame::FillSpatialProfileData(PointXy point, std::vector<CARTA::SetSpatialR
...
             if (have_profile) {
                 // add SpatialProfile to message
-                auto spatial_profile = spatial_profile_message.add_profiles();
-                spatial_profile->set_coordinate(coordinate);
                 // Should these be set to the rounded endpoints if the data is downsampled or decimated?
-                spatial_profile->set_start(requested_start);
-                spatial_profile->set_end(requested_end);
-                spatial_profile->set_raw_values_fp32(profile.data(), profile.size() * sizeof(float));
-                spatial_profile->set_mip(mip);
+                auto spatial_profile =
```
confluence: Minor point: we don't need to assign the return here.

src/Util/Message.cc
```diff
@@ -170,41 +170,39 @@ std::vector<char> Message::EncodeMessage(CARTA::EventType event_type, uint32_t e
...
 }
 
-CARTA::SpectralProfileData Message::SpectralProfileData(int32_t stokes, float progress) {
-    CARTA::SpectralProfileData message;
-    message.set_stokes(stokes);
-    message.set_progress(progress);
-    return message;
+CARTA::SpectralProfile* Message::AddProfile(
+    CARTA::SpectralProfileData& profile_data, std::string& coordinate, CARTA::StatsType& stats_type, std::vector<double>& values) {
+    // one SpectralProfile per stats type
```
confluence: This comment is obsolete and should be removed.

confluence (approved):
Two minor issues that should not hold up approval.

kswang1029:
@confluence @Michaelazzz the spectral profiler still has missing profile issue.

src/Region/RegionHandler.cc
```diff
@@ -1511,8 +1511,18 @@ bool RegionHandler::FillSpectralProfileData(
                 AxisRange z_range(0, _frames.at(config_file_id)->Depth() - 1); // all channels
                 profile_ok = GetRegionSpectralData(config_region_id, config_file_id, z_range, coordinate, stokes_index, required_stats,
                     report_error, [&](std::map<CARTA::StatsType, std::vector<double>> results, float progress) {
-                        auto profile_message = Message::SpectralProfileData(
-                            config_file_id, config_region_id, stokes_index, progress, coordinate, required_stats, results);
+                        auto profile_message = Message::SpectralProfileData(config_file_id, config_region_id, stokes_index, progress);
```
confluence: Sorry; I missed this: the parameters are still in the wrong order here.

confluence (changes requested):

src/Region/RegionHandler.cc
Michaelazzz: Sorry, I also missed it.

confluence (approved):

kswang1029 (approved):

PR #1560, opened 2026-01-19
File directories protobuf cleanup

Michaelazzz:
**Description**

Streamlining the construction of Directory related messages and sub messages.

**Checklist**


confluence (commented):
What is here so far is fine, but we need a few more related wrappers.

There are three types of file list responses (with corresponding requests): `FileListResponse`, `RegionListResponse`, and `CatalogListResponse`. The first two are identical except for the name; they use the same `FileInfo` and `DirectoryInfo` subobjects. `CatalogListResponse` uses `CatalogFileInfo` instead of `FileInfo`: while region and image file types are combined in the same `FileType` enum, `CatalogFileInfo` uses a different `CatalogFileType` enum.

The same `GetFileList` function is used for both region and image file lists, although a `FileListResponse` is always passed in. If the request is for a region list, the data is copied from the `FileListResponse` to a `RegionListResponse`. We never construct a `RegionListResponse` directly.

A `CatalogListResponse` is constructed in `TableController.cc`.

So we should definitely have an `AddDirectory` wrapper for a `CatalogListResponse` as well, and we need an `AddFile` wrapper for both `FileListResponse` and `CatalogListResponse`. The other fields can be set on these objects individually, as they are now.

It may be possible to simplify the `RegionListResponse` copy to some kind of cast (I need to check this); if not, we can add a `RegionListResponse` that takes a `FileListResponse` parameter and performs the copy inside.

src/Table/TableController.cc
```diff
@@ -237,14 +238,11 @@ void TableController::OnFileListRequest(
...
-                    directory_info->set_name(entry.path().filename().string());
-                    directory_info->set_item_count(GetNumItems(entry.path().string()));
-
-                    // Fill in file time
                     struct stat file_stats;
                     stat(entry.path().c_str(), &file_stats);
-                    directory_info->set_date(file_stats.st_mtim.tv_sec);
+
+                    auto directory_info = Message::AddDirectory(file_list_response, entry.path().filename().string(),
+                        static_cast<int64_t>(file_stats.st_mtim.tv_sec), static_cast<int32_t>(GetNumItems(entry.path().string())));
```
confluence: These explicit casts are not necessary.

In each of these wrapper functions, the parameter types should match the types of the values that are being passed in, not the internal types of the protobuf object fields, so that they are preserved when the logic for setting the fields is moved into the wrappers. Since setting the fields results in the appropriate cast being performed (by the protobuf library), this will continue to work inside the function.

For example, in this wrapper the type of `item_count` should be `int` (because `GetNumItems` returns an int) and I believe that the type of `date` should be `time_t` (the type of the nested value in the struct; this information is spread across the standard C library and the Linux system libraries).

src/Util/Message.cc
```diff
@@ -438,6 +438,47 @@ void FillStatistics(CARTA::RegionStatsData& stats_data, const std::vector<CARTA:
     }
 }
 
+CARTA::DirectoryInfo* Message::AddDirectory(CARTA::FileListResponse& response, std::string& name, int64_t date, int32_t item_count) {
```
confluence: All these counts should be `int` and all the dates should (probably) be `time_t`.

confluence (changes requested):

src/Table/TableController.cc
```diff
@@ -237,14 +238,11 @@ void TableController::OnFileListRequest(
...
-                    auto directory_info = file_list_response.add_subdirectories();
-                    directory_info->set_name(entry.path().filename().string());
-                    directory_info->set_item_count(GetNumItems(entry.path().string()));
-
-                    // Fill in file time
                     struct stat file_stats;
                     stat(entry.path().c_str(), &file_stats);
-                    directory_info->set_date(file_stats.st_mtim.tv_sec);
+
+                    auto directory_info = Message::AddDirectory(file_list_response, entry.path().filename().string(),
```
confluence: You don't need to assign this return value.

src/Table/TableController.cc
confluence: You should move this above the `AddFile` call, and include the date in the parameters.

src/Util/Message.cc
```diff
@@ -438,6 +438,47 @@ void FillStatistics(CARTA::RegionStatsData& stats_data, const std::vector<CARTA:
...
+    return directory_info;
+}
+
+CARTA::FileInfo* Message::AddFile(
+    CARTA::FileListResponse& response, std::string name, CARTA::FileType type, int64_t size, std::string hdu_list, int64_t date) {
+    auto* file_info = response.add_files();
+    file_info->set_name(name);
+    return file_info;
+}
+
```
confluence: You're ignoring all parameters except the name in here.

src/Util/Message.cc
```diff
@@ -438,6 +438,47 @@ void FillStatistics(CARTA::RegionStatsData& stats_data, const std::vector<CARTA:
...
+    directory_info->set_item_count(item_count);
+    return directory_info;
+}
+
+CARTA::CatalogFileInfo* Message::AddFile(
+    CARTA::CatalogListResponse& response, std::string name, CARTA::CatalogFileType type, int64_t size, std::string hdu_list, int64_t date) {
+    auto* file_info = response.add_files();
+    file_info->set_name(name);
+    return file_info;
+}
```
confluence: You're also ignoring all parameters except the name here.

confluence (changes requested):

src/Util/Message.cc
```diff
@@ -438,6 +438,55 @@ void FillStatistics(CARTA::RegionStatsData& stats_data, const std::vector<CARTA:
...
+CARTA::CatalogFileInfo* Message::AddFile(
+    CARTA::CatalogListResponse& response, std::string name, CARTA::CatalogFileType type, int64_t size, std::string hdu_list, int64_t date) {
+    auto* file_info = response.add_files();
+    file_info->set_name(name);
+    file_info->set_type(type);
+    file_info->set_file_size(size);
+    file_info->set_description(hdu_list);
+    file_info->set_date(date);
+    return file_info;
+}
```
confluence: The parameter names should be changed to match the field names (i.e. `file_size` and `description`).

confluence (changes requested):
I've realised that there are a few other messages where a file info is set, but I think that we should stick to the list responses here and handle those separately.

src/Table/TableController.cc
```diff
@@ -254,16 +252,14 @@ void TableController::OnFileListRequest(
...
-                file_info->set_file_size(fs::file_size(entry));
 
                 // Fill in file time
                 struct stat file_stats;
                 stat(entry.path().c_str(), &file_stats);
-                file_info->set_date(file_stats.st_mtim.tv_sec);
+
+                // Fill the file info
+                auto file_info = Message::AddFile(
+                    file_list_response, entry.path().filename().string(), file_type, fs::file_size(entry), "", file_stats.st_mtim.tv_sec);
```
confluence: Since description is not currently being set (and is the most optional of these properties), you should reorder the parameters to put it last, so that you don't have to include a dummy value.

confluence (changes requested):

src/Util/Message.h
```diff
@@ -123,6 +129,14 @@ class Message {
...
     static CARTA::ListProgress ListProgress(
         const CARTA::FileListType& file_list_type, int32_t total_count, int32_t checked_count, float percentage);
+    static CARTA::DirectoryInfo* AddDirectory(CARTA::FileListResponse& response, const std::string& name, time_t date, int item_count = 0);
+    static CARTA::FileInfo* AddFile(CARTA::FileListResponse& response, std::string name, CARTA::FileType type = CARTA::FileType::UNKNOWN,
+        int64_t size = 0, std::string hdu_list = "", int64_t date = 0);
+    static CARTA::DirectoryInfo* AddDirectory(
+        CARTA::CatalogListResponse& response, const std::string& name, time_t date, int item_count = 0);
+    static CARTA::CatalogFileInfo* AddFile(CARTA::CatalogListResponse& response, std::string name,
+        CARTA::CatalogFileType type = CARTA::CatalogFileType::Unknown, int64_t size = 0, int64_t date = 0, std::string description = "");
+    static CARTA::RegionListResponse RegionListResponse(CARTA::FileListResponse file_response);
```
confluence, line 146: Sorry; I missed this before:

For consistency (and so that the `AddFile` wrappers can be used with additional parameters in the future), `date` should be a `time_t` in both of them, and `hdu_list` should move to the end of the other `AddFile`'s parameter list.

confluence (changes requested):

confluence (approved):

confluence:
@kswang1029 this is a refactoring PR that should have no effect on behaviour.

kswang1029 (changes requested):
The filenames in the file list now contain extra string like HDU. 


Image

src/Util/Message.cc
```diff
@@ -435,6 +435,56 @@ void FillStatistics(CARTA::RegionStatsData& stats_data, const std::vector<CARTA:
...
+
+CARTA::FileInfo* Message::AddFile(CARTA::FileListResponse& response, std::string name, CARTA::FileType type, int64_t size, time_t date,
+    const std::vector<std::string>& hdu_list) {
+    auto* file_info = response.add_files();
+    file_info->set_name(name);
+    file_info->set_type(type);
+    file_info->set_size(size);
+    auto* list = file_info->mutable_hdu_list();
+    list->Add(hdu_list.begin(), hdu_list.end());
```
confluence: This is unnecessarily verbose: see how other repeated fields are set directly (e.g. in your region list wrapper).

confluence (changes requested):

confluence (approved):

kswang1029 (approved):

PR #1553, opened 2026-01-05
Refactoring: remove never-to-be-used wrappers from Util/Message #1533

marcinsokolowski:
**Description**

Moving wrapper functions which are only used in tests from from Message.cc to specific tests - as described in Issue [#1533](https://github.com/CARTAvis/carta-backend/issues/1533) 

**Done:**
- [x] the bulk of the wrappers were already moved from Message.cc to specific tests
- [x] fix style and make the CI/CD pass.

**To do before moving out of draft:**
- [X] consult/review by Adrianna
- [X] In the process, I've noticed that in Message.cc there are global functions ::FillHistogram and ::FillStatistics which seem to implement some common/general functionality used in a couple of places (e.g. Frame.cc). However, they do not seem to be message related. So, should they really be here, or rather moved elsewhere? To be discussed and decided. *addressed in comment*
- [X] Anything else I've missed?

**Checklist**

- [X] changelog updated
- [x] e2e test passing / corresponding fix added / new e2e test created
- [ ] ICD test passing / corresponding fix added / new ICD test created
- [x] no protobuf update needed
- [X] protobuf version not bumped
- [X] added reviewers and assignee
- [ ] GitHub Project estimate added

PR #1561, opened 2026-01-19
Region related protobuf cleanup

Michaelazzz:
**Description**

Streamline the construction Of Region related messages and sub-messages.

**Checklist**


src/Region/RegionImportExport/RegionExporter.cc
```diff
@@ -40,8 +41,7 @@ bool RegionExporter::CanExportToFile(const std::string& filename, bool overwrite
         }
     }
     if (!error.empty()) {
-        export_ack.set_success(false);
-        export_ack.set_message(error);
+        Message::ExportRegionAck(export_ack, false, error);
```
confluence: If we did use a wrapper for this, it should match the `ImportRegionAck` wrapper -- here you use an in-out parameter and there you assign the return value.

But we shouldn't use wrapper functions to construct ack messages like this -- the way that they're passed up and down nested functions means that there isn't a clean way to create them in one step. This message has other fields that are set somewhere else.

So this wrapper should be removed.

src/Region/RegionHandler.cc
```diff
@@ -155,23 +155,20 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
     try {
         importer = GetRegionImporter(region_file_type, csys, file_id, region_file, file_is_filename);
     } catch (const casacore::AipsError& err) {
-        import_ack.set_success(false);
-        import_ack.set_message("Region import failed: " + err.getMesg());
+        import_ack = Message::ImportRegionAck(false, "Region import failed: " + err.getMesg());
```
confluence: This wrapper should also be removed.

src/Util/Message.cc
```diff
@@ -46,6 +46,43 @@ CARTA::SetImageChannels Message::SetImageChannels(
...
+    CARTA::SetHistogramRequirements set_histogram_requirements;
+    set_histogram_requirements.set_file_id(file_id);
+    set_histogram_requirements.set_region_id(region_id);
+    auto* histograms = set_histogram_requirements.add_histograms();
+    histograms->set_coordinate(coordinate);
+    histograms->set_channel(channel);
+    histograms->set_num_bins(num_bins);
+    return set_histogram_requirements;
+}
+
```
confluence: This is another merge error (these messages were removed in #1553).

src/Util/Message.h
```diff
@@ -73,25 +73,55 @@ class Message {
         bool support_aips_beam, CARTA::RenderMode render_mode = CARTA::RenderMode::RASTER);
     static CARTA::SetImageChannels SetImageChannels(int32_t file_id, int32_t channel, int32_t stokes,
         CARTA::CompressionType compression_type = CARTA::CompressionType::NONE, float compression_quality = -1);
+    static CARTA::SetCursor SetCursor(int32_t file_id, float x, float y);
```
confluence: More merge errors here (messages that were removed in #1553 are being put back in).

src/Util/Message.cc
```diff
@@ -401,6 +435,21 @@ CARTA::ListProgress Message::ListProgress(
...
 
+CARTA::ImportRegionAck Message::AddImportedRegion(CARTA::ImportRegionAck& import_ack, int region_id, CARTA::RegionType region_type,
+    std::vector<CARTA::Point> control_points, float region_rotation, CARTA::RegionStyle region_style) {
+    // Set CARTA::RegionInfo
+    CARTA::RegionInfo region_info;
+    region_info.set_region_type(region_type);
+    *region_info.mutable_control_points() = {control_points.begin(), control_points.end()};
+    region_info.set_rotation(region_rotation);
+
+    // Add info and style to import_ack; increment region id for next region
```
confluence: This comment should be removed. You aren't incrementing the ID in here.

src/Util/Message.cc
```diff
@@ -401,6 +435,21 @@ CARTA::ListProgress Message::ListProgress(
     return message;
 }
 
+CARTA::ImportRegionAck Message::AddImportedRegion(CARTA::ImportRegionAck& import_ack, int region_id, CARTA::RegionType region_type,
```
confluence, line 388: `region_id` and `region_type` are fine, but it's not necessary to prepend `region` to the other names (use `rotation` and `style` instead).

src/Region/RegionHandler.cc
```diff
@@ -194,10 +191,8 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
             _regions[region_id] = std::move(region);
             region_lock.unlock();
 
-            CARTA::RegionInfo region_info;
-            region_info.set_region_type(region_state.type);
+            auto region_info = Message::SetRegionInfo(region_state.type, region_state.rotation);
             *region_info.mutable_control_points() = {region_state.control_points.begin(), region_state.control_points.end()};
-            region_info.set_rotation(region_state.rotation);
             (*region_info_map)[region_id] = region_info;
             (*region_style_map)[region_id++] = region_style;
```
confluence: You should be using your `AddImportedRegion` wrapper here to add the region to the `import_ack` message. `SetRegionInfo` should be removed.

* Don't forget to increment the region ID explicitly here! For clarity, do it separately on a new line after using the wrapper.
* Remove the map aliases created above outside the loop (they will no longer be needed here).
* I would also refactor the variable names in this loop for readability. Call the loop variable `props`; then you can eliminate the state and style aliases as well and refer to `props.state.type`, etc.. I think this is much less confusing than using multiple independent variables with similar names.

confluence (changes requested):

src/Region/RegionImportExport/RegionExporter.cc
```diff
@@ -82,8 +83,7 @@ void RegionExporter::ExportRegions(const std::string& filename, std::string& mes
         // Write to file
         success = ExportRegions(filename, message);
     }
-    export_ack.set_success(success);
-    export_ack.set_message(message);
+    Message::ExportRegionAck(export_ack, success, message);
```
confluence: Outstanding issue from previous comments: this wrapper should be removed.

src/Region/RegionHandler.cc
```diff
@@ -161,45 +161,37 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
     }
 
     if (!importer) {
-        import_ack.set_success(false);
-        import_ack.set_message("Region importer failed.");
+        import_ack = Message::ImportRegionAck(false, "Region importer failed.");
```
confluence: Outstanding issue from previous comments: this wrapper should be removed.

src/Region/RegionHandler.cc
```diff
@@ -161,45 +161,37 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
...
         return;
     }
 
     // Get regions and error message from importer
     std::string error;
     auto imported_regions = importer->GetRegions(error);
-    import_ack.set_message(error);
+    import_ack = Message::ImportRegionAck(false, error);
```
confluence: Outstanding issue from previous comments: this wrapper should be removed.

src/Region/RegionHandler.cc
```diff
@@ -161,45 +161,37 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
...
             region_lock.unlock();
 
-            CARTA::RegionInfo region_info;
-            region_info.set_region_type(region_state.type);
-            *region_info.mutable_control_points() = {region_state.control_points.begin(), region_state.control_points.end()};
-            region_info.set_rotation(region_state.rotation);
-            (*region_info_map)[region_id] = region_info;
-            (*region_style_map)[region_id++] = region_style;
+            Message::AddImportedRegion(import_ack, region_id, props.state.type,
+                {props.state.control_points.begin(), props.state.control_points.end()}, props.state.rotation, props.style);
```
confluence: You should *just* be passing `control_points` here. You are already making a copy inside the wrapper.

src/Region/RegionHandler.cc
```diff
@@ -212,8 +204,7 @@ void RegionHandler::ExportRegion(int file_id, std::shared_ptr<Frame> frame, CART
     // Export regions to given filename, or return export file contents in ack
     // Check if any regions to export
     if (region_styles.empty()) {
-        export_ack.set_success(false);
-        export_ack.set_message("Export region failed: no regions requested.");
+        Message::ExportRegionAck(export_ack, false, "Export region failed: no regions requested.");
```
confluence: This wrapper should be removed.

src/Region/RegionHandler.cc
```diff
@@ -224,8 +215,7 @@ void RegionHandler::ExportRegion(int file_id, std::shared_ptr<Frame> frame, CART
     bool export_pixel_coords(coord_type == CARTA::CoordinateType::PIXEL);
     if (!export_pixel_coords && !output_csys->hasDirectionCoordinate()) {
         // Export fails, cannot convert to world coordinates
-        export_ack.set_success(false);
-        export_ack.set_message("Cannot export regions in world coordinates for linear coordinate system.");
+        Message::ExportRegionAck(export_ack, false, "Cannot export regions in world coordinates without direction coordinate.");
```
confluence: This wrapper should be removed.

src/Session/Session.cc
```diff
@@ -1012,8 +1012,7 @@ void Session::OnExportRegion(const CARTA::ExportRegion& message, uint32_t reques
             string error = "Exporting region is not allowed in read-only mode";
             spdlog::error(error);
             SendLogEvent(error, {"Export region"}, CARTA::ErrorSeverity::ERROR);
-            export_ack.set_success(false);
-            export_ack.set_message(error);
+            Message::ExportRegionAck(export_ack, false, error);
```
confluence: This wrapper should be removed.

src/Util/Message.cc
```diff
@@ -401,6 +398,20 @@ CARTA::ListProgress Message::ListProgress(
...
+    std::vector<CARTA::Point> control_points, float rotation, CARTA::RegionStyle style) {
```
confluence: You need to pass the control points by reference here, not by value.

The only copy should be happening inside the function.

src/Util/Message.cc
```diff
@@ -372,6 +356,19 @@ CARTA::ImportRegionAck Message::ImportRegionAck(bool success, const std::string&
...
+    regionInfo.set_rotation(rotation);
+    return regionInfo;
+}
+
+CARTA::ExportRegionAck Message::ExportRegionAck(CARTA::ExportRegionAck& export_region_ack, bool success, const std::string& message) {
+    export_region_ack.set_success(success);
+    export_region_ack.set_message(message);
+    return export_region_ack;
+}
+
```
confluence: This should be removed.

src/Util/Message.h
```diff
@@ -73,25 +73,26 @@ class Message {
         bool support_aips_beam, CARTA::RenderMode render_mode = CARTA::RenderMode::RASTER);
     static CARTA::SetImageChannels SetImageChannels(int32_t file_id, int32_t channel, int32_t stokes,
         CARTA::CompressionType compression_type = CARTA::CompressionType::NONE, float compression_quality = -1);
+    static CARTA::SetStatsRequirements SetStatsRequirements(int32_t file_id, int32_t region_id);
+    static CARTA::SetHistogramRequirements SetHistogramRequirements(int32_t file_id, int32_t region_id, const std::string& coordinate = "z",
+        int32_t channel = CURRENT_Z, int32_t num_bins = AUTO_BIN_SIZE);
```
confluence: Outstanding merge error.

src/Util/Message.h
```diff
@@ -73,25 +73,26 @@ class Message {
...
         int32_t file_id, CARTA::CompressionType compression_type, float compression_quality, const std::vector<int32_t>& tiles);
     static CARTA::Point Point(float x, float y);
     static CARTA::Point Point(const casacore::Vector<casacore::Double>& input, int x_index = 0, int y_index = 1);
     static CARTA::Point Point(const std::vector<casacore::Quantity>& input, int x_index = 0, int y_index = 1);
     static CARTA::Point Point(const std::vector<double>& input, int x_index = 0, int y_index = 1);
-    static CARTA::SetRegion SetRegion(
-        int32_t file_id, int32_t region_id, CARTA::RegionType region_type, std::vector<CARTA::Point> control_points, float rotation);
-    static CARTA::ImageBounds ImageBounds(int32_t x_min, int32_t x_max, int32_t y_min, int32_t y_max);
     static CARTA::SetRegion SetRegion(int32_t file_id, int32_t region_id, const CARTA::RegionInfo& region_info);
+    static CARTA::ImageBounds ImageBounds(int32_t x_min, int32_t x_max, int32_t y_min, int32_t y_max);
```
confluence, line 82: Outstanding merge error.

src/Util/Message.h
```diff
@@ -73,25 +73,26 @@ class Message {
...
     static CARTA::ConcatStokesFiles ConcatStokesFiles(
         int32_t file_id, const google::protobuf::RepeatedPtrField<CARTA::StokesFile>& stokes_files);
     static CARTA::DoublePoint DoublePoint(double x, double y);
     static CARTA::GaussianComponent GaussianComponent(
         const CARTA::DoublePoint& center, double amp, const CARTA::DoublePoint& fwhm, double pa);
     static CARTA::ScriptingRequest ScriptingRequest(uint32_t scripting_request_id, const std::string& target, const std::string& action,
         const std::string& parameters, bool async, const std::string& return_path);
-
-    // Response messages
+    static CARTA::ChannelMapFlowControl ChannelMapFlowControl(int32_t file_id, int32_t received_channel);
```
confluence: Outstanding merge error.

src/Util/Message.cc
```diff
@@ -372,6 +356,19 @@ CARTA::ImportRegionAck Message::ImportRegionAck(bool success, const std::string&
     return import_region_ack;
 }
 
+CARTA::RegionInfo Message::SetRegionInfo(CARTA::RegionType& region_type, float rotation) {
+    CARTA::RegionInfo regionInfo;
+    regionInfo.set_region_type(region_type);
+    regionInfo.set_rotation(rotation);
+    return regionInfo;
+}
```
confluence: This is currently unused, and should be removed.

confluence (changes requested):

src/Util/Message.h
confluence: This should be moved above the `SetRegion` above it, so that the diff is clean: it should show only the removal of the other `SetRegion`, not this reordering.

confluence (changes requested):

src/Region/RegionHandler.cc
```diff
@@ -171,35 +171,28 @@ void RegionHandler::ImportRegion(int file_id, std::shared_ptr<Frame> frame, CART
     auto imported_regions = importer->GetRegions(error);
     import_ack.set_message(error);
     if (imported_regions.empty()) {
-        import_ack.set_success(false);
```
confluence, line 174: This should not have been removed. I think that false is the field default, but the value should still be set explicitly here (if the code is refactored and something else sets the field earlier, the intent needs to be clear).

confluence (changes requested):

confluence (approved):

confluence:
@kswang1029 this is a refactoring PR that should have no effect on behaviour.

kswang1029 (approved):

PR #1568, opened 2026-01-30
Retry failure tests once

pshnghng0318:
**Description**

* Added a retry mechanism for ICD tests on macOS.

* Each ICD test now runs once, and if it fails, it retries one additional time automatically.  
  This prevents intermittent failures caused by machine or network.

* No companion PR

### DIFF
--- a/.github/actions/run-macos/action.yml
+++ b/.github/actions/run-macos/action.yml
@@ -27,7 +27,7 @@ runs:
         cd $ICD_DIR
         for test_file in $(cat ICD_test_stages/${{ inputs.test_stage_name }}.tests); do
           if CI=true npm test "$test_file"; then
-            echo "$test_file 1st attempt"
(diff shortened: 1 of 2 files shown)


.github/actions/run-macos/action.yml
```diff
@@ -26,9 +26,16 @@ runs:
         ICD_DIR=$GITHUB_WORKSPACE/ICD-RxJS
         cd $ICD_DIR
         for test_file in $(cat ICD_test_stages/${{ inputs.test_stage_name }}.tests); do
-          CI=true npm test $test_file
+          if CI=true npm test "$test_file"; then
+            echo "$test_file 1st attempt"
```
markccchiang: Maybe changing the message to the following would make it clearer:
```echo "$test_file 1st attempt: PASS"```

markccchiang (approved):

PR #1588, opened 2026-03-24
Static generated data for contour tests

Michaelazzz:
**Description**
Static generated contour data added to the contour tests from the carta-backend-test-dat repo. 

* Generated test data was added to the carta-backend-test-data submodule
* It uses static test data instead of generating the comparison dataset within the test.
* Yes, CARTAvis/carta-backend-test-data#2
* No

**Checklist**


test/TestContour.cc
```diff
@@ -5,178 +5,163 @@
...
-TEST_F(ContourTest, BlockAverageHdf5FileNaN) {
-    GenerateContour(Hdf5Images() / "500x500_nans.hdf5", CARTA::SmoothingMode::BlockAverage);
+        spdlog::info("Level {}: {} vertices verified", level, gen_coords.size());
+    }
 }
+
+INSTANTIATE_TEST_SUITE_P(AllModesAndFormats, ContourTest,
+    ::testing::Values(ContourParams{FitsImages() / "500x500.fits", CARTA::SmoothingMode::NoSmoothing,
+                          {ContourData() / "500x500_contours/level_0.txt", ContourData() / "500x500_contours/level_-1.txt",
+                              ContourData() / "500x500_contours/level_1.txt"}},
```
confluence: There's no need to list the individual file paths here. You use the same three levels everywhere, so you can pass in the directory path here and hardcode iterating over the three numbers inside the test. Then instead of parsing the filename to get the level number you can construct the filename from the number you have.

If you wanted to use different compression levels in different tests, you could pass in a list of numbers as another parameter (but it's not necessary now).

confluence (changes requested):
I have also made a comment in the corresponding data PR.

Issue #1535, opened 2025-12-02
Basic information about user actions missing from logs


confluence:
Some very basic information that would be useful for debugging crashes is currently not logged at all. We rely on users self-reporting the context of errors they encounter when they actively file bugs. However, if an institution with an SDM installation wants to follow up on crashes that were never reported by users, there is currently very little information in the logs that can be used to reconstruct the circumstances that led to the crash. For example, we don't even log what files the user is opening! That seems important and infrequent enough to be logged at `info` level.

I think that (almost?) every handler in Session should log a message when it is triggered (although most of them can probably do it at `debug` level). For handlers that are called so frequently that this would cause excessive noise it should be possible to implement custom logger functionality to aggregate  / decimate multiple messages.

PR #1596, opened 2026-04-14
Logger verbosity improvements

Michaelazzz:
**Description**

Issue #1535 
1. Log major user actions on the `log`  (like opening a file) and log them in a meaningful way
Example: `[Session 3593344489] User is opening file: home/michaela/Desktop/carta/carta-backend-test-data/images/fits/128_128_gaussian_model_three_components.fits`
2. Log all handlers on the `debug` level
Example: `[debug] [protocol] <== SET_SPATIAL_REQUIREMENTS`
3. Handle noisey messages by only logging every X instance of the message.
No
Run with the `--verbosity 5 --log_protocol_messages`

**Checklist**


src/Session/SessionManager.cc
```diff
@@ -207,8 +207,20 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
+            CARTA::OpenFile open_msg;
+
+            auto payload_ptr = sv_message.data() + sizeof(EventHeader);
+            auto payload_len = sv_message.length() - sizeof(EventHeader);
+
+            if (open_msg.ParseFromArray(payload_ptr, payload_len)) {
+                spdlog::info("[Session {}] User is opening file: {}/{}", session_id, open_msg.directory(), open_msg.file());
+            }
+        }
+
```
confluence: This is unnecessary, and can be removed.

src/Session/SessionManager.cc
```diff
@@ -207,8 +207,20 @@ void SessionManager::OnMessage(WSType* ws, std::string_view sv_message, uWS::OpC
...
         MessageHandler handler;
         try {
+            // log message
```
confluence: This should be removed.

logging_rules.json
```diff
@@ -0,0 +1,20 @@
+{
```
confluence, line 1: These default settings shouldn't be stored here in a file; they should be in an object constructed at runtime (I'm describing this in a separate comment).

I was considering writing it out into the user's local directory if no configuration exists, but this would be complicated to integrate correctly with site configuration, so I think it's fine to skip that (as long as we document the default behaviour in the usage / help). I can provide example JSON configuration in the controller docs.

src/Logger/Logger.cc
```diff
@@ -8,14 +8,96 @@
...
+#include <nlohmann/json.hpp>
+#include <regex>
 #include <string>
 
 namespace carta {
 namespace logger {
 
 static bool log_protocol_messages(false);
 
+using LogAction = std::function<void(const std::string& name, uint64_t& count, const std::string& dir)>;
```
confluence: The event-logging functionality is now complex enough that it should be encapsulated in its own area of the code. That doesn't mean that it needs to have a completely separate logger (like the performance log); the messages can still go in the default logger, but all of the implementation should move into its own file:

* Create a new set of `ProtocolLogger` files and move all of the event-related code in there.
* Since you have a bunch of related state now (the registry and the event counts), it would be good to make this more object-oriented (make a `ProtocolLogger` class that groups these properties) and create it with a constructor.
* This should be a singleton instance accessible through static functions (kind of like the settings instance, but you won't need to retrieve the object -- just initialize it and use the functions to log events).
* *This* logger file should only call that class's initialization function in its own init.

I will elaborate on other design issues separately.

src/Logger/Logger.cc
```diff
@@ -8,14 +8,96 @@
...
+
+static std::unordered_map<CARTA::EventType, LogAction> registry;
+static std::mutex registry_mutex;
```
confluence: The registry doesn't need a mutex -- it's constructed once during initialization, and won't change.

What does need a mutex is the incrementation of the counts, so it should be renamed, and locked *only* around the incrementation.

src/Logger/Logger.cc
```diff
@@ -8,14 +8,96 @@
...
+static std::unordered_map<CARTA::EventType, uint64_t> inbound_counts;
+static std::unordered_map<CARTA::EventType, uint64_t> outbound_counts;
```
confluence: I believe that every event is unambiguously used either in one direction or the other direction, so you don't need two maps here -- just one.

src/Logger/Logger.cc
confluence: I will group all the boolean flag notes here:

The `log_protocol_messages` flag is set once at initialization and never changes, so instead of checking it repeatedly at runtime (like the current code does) you should set up your event logger class appropriately depending on what the flag is, and then never check it again.

Ideally what you want is:
* By default you don't initialise the registry, and the logging functions do nothing.
* If the flag is set, you *do* initialise the registry, and the logging functions do their thing.

My recommendation is to use inheritance:
* Base `ProtocolLogger` has a static `ProtocolLogger` instance property, implements a static `Init` member function, and defines virtual member functions for logging (not static; these will be accessed through the instance).
* `NullProtocolLogger` inherits from `ProtocolLogger`, has an empty constructor, and empty implementations of the logging functions.
* `ActiveProtocolLogger` (or maybe a better name) inherits from `ProtocolLogger`, does all the registry creation in the constructor, and has the "real" implementations of the logging functions. All the other properties and helper functions (registry, etc.) should go in this class.
* Base `ProtocolLogger`'s static `Init` function gets the settings instance, checks the flag, and sets its own instance by creating an instance of one of these subclasses.
* The base logger's init function calls that static init function. The base logger doesn't know anything else about event logging.
* The session calls the sent / received event logging functions on the retrieved protocol logger instance.

You could also eliminate the `NullProtocolLogger` subclass and just merge it into the base class, but it may be cleaner this way.

src/Logger/Logger.cc
```diff
@@ -8,14 +8,96 @@
...
+                spdlog::info("Successfully loaded custom config from: {}", settings.log_config_path);
+            } catch (const nlohmann::json::parse_error& e) {
+                spdlog::info("Error parsing JSON file: {}", e.what());
+            }
+        } else {
+            spdlog::debug("Could not open file: {}", settings.log_config_path);
+        }
+    }
+
+    if (config_data.contains("rules") && config_data["rules"].is_array()) {
```
confluence: 1. You need to build the registry whether there's any configuration or not -- if the config is empty, you don't want to be doing a linear search through the whole map before falling back to the default logging action every time you log an event. But I think you should have a default config that is used if no user-provided config exists (see below).
2. Reading this configuration should be done in `ProgramSettings`, not here.
3. The file should be checked against a (new) JSON schema (instead of these manual checks of the content).
4. There doesn't need to be a configurable settings path. `ProgramSettings` should look in two places (global `protocol_logging_rules.json` and user directory `protocol_logging_rules.json`, and resolve them in the same way as it currently resolves / merges `backend.json`).
5. You should create a `LoggingRule` struct (with fields for `match`, etc.). You could make a hierarchy for different action types, but it may be overkill for an intermediate helper object -- instead you could make the numeric properties default to -1 and only use them if it's a throttle action. The function in `ProgramSettings` should parse the rules JSON and construct a vector of these rules. The vector should be available as a property on the settings object. In the registry construction code, you should be retrieving this property and using the vector of structs instead of working with the raw JSON.
6. The default settings (to use if there's no custom config) should be stored in a static property on the `ActiveProtocolLogger` (also a vector of structs, not a JSON object). But you need to distinguish between *no* config and *empty* config -- the user must be able to provide a file with an empty rules array to clear the default config (i.e. reset to "normal" logging for every event). The place where this decision can be made is `ProgramSettings` -- the function that reads this config should get the default config from the logger class and set it on the settings object *only if* there is no config file at all. In the registry code you would only look at the settings object, and not have to care where the config came from.
 
All this is a bit convoluted, but I think that we need configure the default behaviour to be more sensible than "print every message", so we need those defaults. but I'm not 100% decided on whether the user settings should be applied *instead of* or *on top of* the defaults. Ideally I would like to provide a commandline option for printing out the JSON of the current configuration (something that I've also been wanting to do with the main backend config). For now I think it will be less magical if user rules are applied on top of a clean slate.

src/Logger/Logger.cc
```diff
@@ -8,14 +8,96 @@
...
+                    }
+
+                    matched = true;
+                    break;
+                }
+            }
+
+            if (!matched) {
+                registry[event_type] = createNormalLogger();
+            }
```
confluence: Two problems here:

1. You're only applying the first rule that matches an event type. Later rules have to be able to override earlier rules.
2. You don't handle the case where a matched rule is for normal logging.

As described above, in this part of the code you should be iterating over a vector of rule structs. For each event type, you should:
1. Create a rule struct variable, and set it to the default (the normal logger rule)
2. Iterate over the *whole* rule vector, and if a rule matches, update the variable to that rule
3. At the end, use whatever rule is in the variable (it will be the last rule to have matched, or the default if nothing matched).

src/Logger/Logger.cc
```diff
@@ -116,26 +198,27 @@ void InitLogger() {
...
-        auto event_name = CARTA::EventType_Name(event_type);
-        if (!event_name.empty()) {
-            spdlog::debug("[protocol] <== {}", event_name);
-        } else {
-            spdlog::debug("[protocol] <== unknown event type: {}!", event_type);
-        }
+void ExecuteLog(CARTA::EventType type, uint64_t& count, const std::string& arrow) {
+    std::lock_guard<std::mutex> lock(registry_mutex);
+
+    count++; // Increments the specific map passed in
```
confluence: This should be a mutex only for the counts, and it should only be locked for this increment.

src/Logger/Logger.cc
```diff
@@ -116,26 +198,27 @@ void InitLogger() {
...
+
+    auto it = registry.find(type);
+    if (it != registry.end()) {
+        it->second(CARTA::EventType_Name(type), count, arrow);
     }
```
confluence: Instead of this, try `.at` and catch the exception (print the "unknown event type" message if that happens). You've populated the registry with every known event, so this should only happen if there's some kind of protobuf version mismatch.

src/Logger/Logger.cc
confluence: You can't create an enum with string values, but you can create static string constants on the struct (`LoggingRule::NORMAL`, `LoggingRule::NONE`, `LoggingRule::THROTTLE`) and use those for comparisons instead of the raw strings.

(The correctness of the values in the JSON will be enforced by the new schema.)

src/Logger/Logger.cc
```diff
@@ -116,26 +198,27 @@ void InitLogger() {
...
 }
 
+void LogReceivedEventType(const CARTA::EventType& event_type) {
+    if (!log_protocol_messages)
+        return;
+    ExecuteLog(event_type, inbound_counts[event_type], "<==");
+}
+
```
confluence: As mentioned previously:
1. you shouldn't need to keep checking the flag (because this implementation should only be used if it was set up at the start).
2. you only need one map of counts

We can merge the received / sent functions entirely: I was originally thinking of detecting the arrow direction automatically from the event name, but that would be fragile. Instead, I suggest that you pass a `direction` string parameter directly from the calling code in `Session`, and put the arrow in there. Or rather make static constants on `ProtocolLogger` called `SENT` and `RECEIVED` and use those.

src/Logger/Logger.cc
```diff
@@ -8,14 +8,96 @@
...
+LogAction createThrottledLogger(uint32_t first_n, uint32_t every_m) {
+    return [first_n, every_m](const std::string& name, uint64_t& count, const std::string& dir) {
+        if (count <= first_n || count % every_m == 0) {
+            std::string suffix = (count > first_n) ? fmt::format(" (Total: {})", count) : "";
+            spdlog::debug("[protocol] {} {}{}", dir, name, suffix);
+        }
+    };
+}
```
confluence: Naming convention: these should be in UpperCamelCase.

You are currently creating many copies of potentially the same action for multiple events. This isn't a huge deal, since these aren't large objects, but it would be cleaner and simpler to rename these to `GetWhateverLogger` and return references to private action instances that you manage internally, creating them as needed (so that you only have one of each distinct action):

* A single "normal" action
* A single "no-op" action
* An unordered map of "throttle" actions, with (m,n) pairs as keys (create a `ThrottleActionKey` struct with a hash function; you can see how this is done for the `TileCacheKey`)

When a throttle action is requested, you can try to get it with `.at`, and if there's an exception, create it, put it in the map, and then return it.

confluence (changes requested):

PR #1613, opened 2026-07-22
channel map refactoring

### LINKED ISSUE #1612: channel map refactoring
To support channel map contour and vector overlay. See the [frontend issue](https://github.com/CARTAvis/carta-frontend/issues/2862) for details.

loveluthien:
**Description**

Closes #1612. Channel map flow refactoring.
Remove the channel map flow control from the backend and move it to the frontend. See the [frontend issue](https://github.com/CARTAvis/carta-frontend/issues/2862) for details. 
[frontend](https://github.com/CARTAvis/carta-frontend/pull/2863) and [protobuf](https://github.com/CARTAvis/carta-protobuf/pull/111)
no

**Checklist**

- [ ] ~ICD test passing~ / corresponding fix added / ~new ICD test created~
- [ ] protobuf updated to the latest dev commit / ~no protobuf update needed~
- [ ] protobuf version bumped / ~protobuf version not bumped~


kswang1029 (approved):
this looks good. @veggiesaurus @confluence please proceed code review at your earliest convenience as there are subsequent PRs that depend on this PR. Thanks.

kswang1029 (approved):
looks good. please proceed code review @confluence @veggiesaurus

src/Frame/Frame.cc
```diff
@@ -508,128 +510,73 @@ bool Frame::FillRasterTileData(CARTA::RasterTileData& raster_tile_data, const Ti
...
+            spdlog::debug("Upgraded precision to {} (originally requested precision: {}).", precision, requested_precision);
         }
-    }
 
-    return false;
-}
+        raster_tile_data.set_compression_quality((float)precision);
+        tile_ptr->set_image_data(compression_buffer.data(), compressed_size);
 
-bool Frame::GetRasterTileData(
```
confluence, line 590: I have no objection to removing this function and folding it into `FillRasterTileData`, since it's only called from there, but there's some important functionality in here that should not have been removed:

1: Tile cache

If the tile cache is being used and the full channel cache has not been filled (this happens in HDF5 files unless contours have been generated), the tile should be requested from the tile cache. Currently this is never triggered, so this cache is being bypassed.

2: Frame's tile pool:

The frame has a tile pool that should be used to obtain a shared pointer to a tile data vector whenever a new tile data vector would be created (the pointer should be passed around as a parameter, rather than a reference to the vector).

This prevents memory allocation and deallocation churn from vectors rapidly being created and destroyed -- that's why the tile cache has a tile pool; the frame's pool is separate.

The capacity of the frame's pool should be sufficient to hold all the tiles that will be allocated at one time -- if this is likely to increase while the channel map is enabled, there needs to be a way to notify the frame how many channels are in use. We will need this later anyway to adjust the cache(s).

It looks like you're currently removing the channel information from the protobuf message -- this should be kept in, even though it will no longer be used by the session to request the channels. Instead it should be passed directly to a function in the frame, which (for now) can use it to adjust the pool size and (later) to make adjustments to cache sizes.

confluence (changes requested):
This is not a full review, but I've found a few issues with the code that was removed.

src/Frame/Frame.cc
confluence: You can get an idea of what the tile pool size needs to be by profiling with psrecord and checking that you don't see a lot of rapid tile-sized memory allocations and deallocations: the pool should be the minimum size that avoids that. I'm not actually sure without a more thorough investigation whether this is affected by the channel map settings (it could be limited to the number of threads, plus a fuzz factor of one or two), but we should check.

loveluthien: I bring back the tile cache and frame's tile pool. I think we still cache the tiles of active channel (i.e., current channel in backend) of the current frame only. Scale up to 100 times of cache size with the channel number doesn't sound good to me. However, I am really not familiar with the cache staff, and I need a clear guide for it, so pool size is not modified yet.

confluence: You don't need to do anything to the tile cache (or to the full channel cache); that's something to investigate later after the cache refactoring.

For now we just need to check if we need to grow / shrink the frame's tile pool if the channel map is enabled / disabled -- possibly not (it depends on whether the number of tiles allocated in memory at once changes). If we do change it, this should not increase the actual memory usage; only prevent unnecessary allocations and deallocations.

Either way, I would like to keep the channel info in the message, because we will definitely need it to tune the cache(s) later.

loveluthien:
Forgot to reply to the comment... Since we will support the channel step in channel map mode in the next frontend PR, we replaced the channel range with the channel list in the protobuf.

veggiesaurus (approved):

src/Session/Session.cc
```diff
@@ -669,48 +667,140 @@ void Session::DeleteFrame(int file_id) {
     if (_region_handler) {
         _region_handler->RemoveFrame(file_id);
     }
-    if (_channel_map_settings) {
-        _channel_map_settings->RemoveFile(file_id);
-    }
 }
 
-void Session::OnAddRequiredTiles(const CARTA::AddRequiredTiles& message, int z, int animation_id, bool skip_data) {
+bool Session::OnAddRequiredTiles(
```
confluence, line 672: There are many changes here that seem to be outside the scope of the removal of the channel map logic, and I am concerned by the extent and direction of the refactoring of the tile-filling code. This is making the review more complex.

For example: previously, this session function had one (parallel) loop that called `Frame::FillRasterTileData`, and that function handled all conditionals (downsampled vs full resolution; downsampling computation vs mipmap; tile cache vs full channel cache vs slice bypassing cache) at the frame level by calling other frame functions. Here these frame functions have been modified, some decision logic has been moved into this session function, and some processing of data is now being performed in separate serial loops. I don't think that we should be doing this.

Could we try to pare down this PR to the minimal changes required to remove the channel map logic from the backend? Unless I'm overlooking something, these changes should mostly be in the session-level code.

confluence (changes requested):

src/Session/Session.cc
```diff
@@ -746,116 +838,62 @@ void Session::OnAddRequiredTiles(const CARTA::AddRequiredTiles& message, int z,
...
+        auto stokes_target = message.stokes();
+        bool channel_map_data_request =
+            message.channel_map_enabled() && message.has_required_tiles() && !message.required_tiles().tiles().empty();
+        if (channel_map_data_request) {
+            bool valid_z = z_target >= 0 && z_target < frame->Depth();
+            bool valid_stokes = (stokes_target >= 0 && stokes_target < frame->NumStokes()) || Stokes::IsComputed(stokes_target);
+            if (!valid_z || !valid_stokes) {
+                auto error = fmt::format("Channel {} or Stokes {} is invalid in image", z_target, stokes_target);
+                SendLogEvent(error, {"channels"}, CARTA::ErrorSeverity::ERROR);
+                return {CARTA::ChannelMapFlowControl::REJECTED, error};
```
confluence, line 864: Previously, `Session::OnSetImageChannels` would *either* set the channel and / or Stokes, *or* (in channel map mode) would request tile data for a range of channels and the current Stokes, without changing the channel or Stokes.

Is the intention now to be able to specify a different Stokes value when requesting channel map tiles?

